PR #177, opened 2014-09-02
BUG: Fix missing DLL for extensions that use OpenMP (requires CMake 3.1)

gregsharp:


gregsharp:
Plastimatch and therefore SlicerRT uses it.

jcfr:
Integrated in [r24125](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24125)

jcfr:

PR #189, opened 2014-10-07
BUG: Don't set empty osx version for exteranl non-cmake projects

blowekamp:
Building PCRE on OSX with the default configuration was failing during
configuration:

checking whether the C compiler works... no

configure: error: in
`/scratch/blowekamp/build/Slicer/PCRE-prefix/src/PCRE-build':
configure: error: C compiler cannot create executables
See `config.log' for more details.

This was due to the following flag being passed:
"-mmacosx-version-min=".

Now this flag is not set if CMAKE_OSX_DEPLOYMENT_TARGET does not have a
value.


hjmjohnson:
@jcfr @blowekamp 

Under yosemite, only 10.9 and 10.10 API's are supported.  Requiring 10.8 is no longer an option.

jcfr:
Closing this PR. After #221 will be integrated, these changes won't be needed.

PR #211, opened 2014-12-22
BUG: avoid false positives on geometry check

naucoin:
Add an epsilon value to the volumes logic and use it in the fuzzy compare instead
of the VTK default value. Add a precision calculation from the epsilon so that
any warning messages are printed out with the same precision as was used to
check the values.
Added a sanity check to see if the smallest volume spacing is within 3-10 orders
of magnitude of epsilon and print a warning if not.
Use the CTK significant decimals utility method to calculate the
precision with which to display differences in compared volumes.
Use the CTK order of magnitude utility method to compare the
spacing of the volumes with the current epsilon being used to
check them.
Added a python test to exercise the epsilon, precision and geometry check.

TBD: GUI for setting epsilon or just always use a volume spacing based value.

### DIFF
--- a/Modules/Loadable/Volumes/Logic/vtkSlicerVolumesLogic.cxx
+++ b/Modules/Loadable/Volumes/Logic/vtkSlicerVolumesLogic.cxx
@@ -53,8 +53,8 @@
 #include <vtkImageReslice.h>
 #include <vtkTransform.h>
 
-
-
+/// CTK includes
+#include <ctkUtils.h>
 
 //----------------------------------------------------------------------------
 namespace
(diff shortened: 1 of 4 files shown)


jcfr:
@naucoin 
* Should we wait for the GUI to be implemented before moving forward with this ? 
* Does it make sense to expose this into the GUI ? Will it be easy for the user to set this value ? Or should we always rely on a automatic approach ?

naucoin:
@pieper and I discussed it a little bit, I wanted to get some user feedback with the default settings to see if this stopped the false positives that had been happening. I had looked around the GUI to see if there was an obvious place to set it, it seems to be an application level setting rather than a Volumes module setting, but it's also dependent on the two volumes being compared. I think if we can get an automatic algorithm that works in most cases it would be better, it's a bit of an obscure and particular setting for a user to worry about, and it will change dramatically if volumes of different scales are used.

naucoin:
I'd like to discuss this at Winter Project Week and get the changes integrated.

naucoin:
@jcfr can we touch base on this today?

jcfr:
Sounds good. What about talking at 2.30pm ?

pieper:
I may have another meeting at that time but discuss without me if that
makes sense.

naucoin:
2:30 works for me, I think I got a sense of @piper's feedback when we chatted earlier.

naucoin:
Result of stand up meeting:
put the ctk methods in an anonymous name space in the logic .cxx file and use them from there (with notes/annotations/version from which it came/notes that it will move into vtk).
The user can hide the warning and not show it again so don't worry too much about popping it up often.

jcfr:
@naucoin Until an updated PR is submitted, I am closing this one.

PR #212, opened 2014-12-30
BUG: Fixed loading of DICOM image from files containing special characters

lassoan:
Got some DICOM files that contained a special character in the file names and in the patient name. Slicer failed to load them because lack of proper encoding before using them.

Solution:
* Encode strings to latin1 before using them in VTK set functions.
* Encode file names to a byte stream before sending them to hashlib.

### DIFF
--- a/Modules/Scripted/DICOMLib/DICOMPlugin.py
+++ b/Modules/Scripted/DICOMLib/DICOMPlugin.py
@@ -85,7 +85,8 @@ class DICOMPlugin(object):
       return None
     m = hashlib.md5()
     for f in files:
-      m.update(f)
+      # Unicode-objects must be encoded before hashing
+      m.update(f.encode('UTF-8', 'ignore'))
     return(m.digest())
 
   def getCachedLoadables(self,files):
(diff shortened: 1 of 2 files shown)


jcfr:
:+1:  Should we add function like `toLatin1` and `toUTF8` to `slicer.util` ?

lassoan:
It would be useful but I think a name such as toVTKString would be a better. I'm not sure if we need toUTF8.

jcfr:
Agreed. `toVTKString` should good. Let me know if you would like me to update `slicer.util` and integrate.

lassoan:
Looks good to me.

jcfr:
@lassoan  Could you review https://github.com/jcfr/Slicer/compare/dicom-loading-from-filenames-with-special-chars

More specifically, I wonder with [this particular changes](https://github.com/jcfr/Slicer/compare/dicom-loading-from-filenames-with-special-chars#diff-5c81585867008dffacf021ea0c9d23d4R297) is correct. (adapted from [here](https://github.com/Slicer/Slicer/pull/212/files#diff-5c81585867008dffacf021ea0c9d23d4R299)). In case, a filename as unicode characters ... I wonder how it can be located after converting to Ascii 8-bit and removing invalid characters.

lassoan:
This works if the file name only contains characters that are available in latin1 and the character is encoded the same way in the Windows system code page as in latin1.

http://msdn.microsoft.com/en-us/library/windows/desktop/dd317752(v=vs.85).aspx

When there are only pure ASCII (<127) characters in the file name then conversion is probably not needed because for these characters UTF8 and latin1 encoding is exactly the same.

What could be improved is to try to match the encoding with the Windows system code page when we pass file names to VTK  (different code page may be used by VTK text rendering components, so we would need to be careful when trying to display these file names). This would always work for FAT file systems. However, NTFS uses Unicode, so to have a 100% working solution VTK would need to use Unicode API and Slicer then could pass Unicode strings to VTK.

jcfr:
Thanks for clarifying. Makes sense.

In the mean time, I integrated the patches. See [r23867](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=23867) and [r23868](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=23868)

PR #214, opened 2015-01-06
COMP: Suppress quoted var warn under CMake 3.1.0

hjmjohnson:
Policy CMP0054 is not set:
Only interpret if() arguments as variables or keywords when unquoted.

Run "cmake --help-policy CMP0054" for policy details.

Quoted variables like "jqPlot_DIR" will no longer be dereferenced when the
policy is set to NEW.  Since the policy is not set the OLD behavior will be
used.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,11 @@ if(POLICY CMP0017)
   cmake_policy(SET CMP0017 OLD)
 endif()
 
+# Under CMake 3.1.0  quoted variables will no longer be dereferenced
+if(POLICY CMP0054)
+  cmake_policy(SET CMP0054 OLD)
+endif()
+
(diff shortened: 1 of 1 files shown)


hjmjohnson:
@jcfr  This avoids a warning with CMake 3.1.0.  

Perhaps there are changes that could be made to avoid the need for this.

Hans

jcfr:
Commit [r23894](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=23894) to [r23900](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=23900) fixes most of the CMP0054 warnings. Based on dashboard results, follow up commit will be done. 

The general idea is to set the policy to NEW after ensuring all warnings are all false positive.

PR #215, opened 2015-01-06
COMP: On OSX, CMAKE_OSX_SYSROOT must be set

hjmjohnson:
It is required that the SDK be set in order to
get the compiler flags correct for non-cmake builds.

This patch provides more robust failure mechanism
so that these variables are found and set prior
to building any code.  If they are re-set after
the project() command is used, then unknown behavior
may occur.

Building PCRE fails when CMAKE_OSX_SYSROOT, CMAKE_OSX_DEPLOYMENT_TARGET, CMAKE_OSX_ARCHITECTURES
are not set on OSX builds.


hjmjohnson:
Superceeded with solution.

PR #216, opened 2015-01-06
Allow mac osx1010 building

hjmjohnson:
@pieper @jcfr  This addresses isssues with building on 10.10 (Specifically Xcode 6).  Only 10.9 and 10.10 SDK's are available.  

The previous solutions were based on the now invalid assumption that QT would always be built agains libstd++.  As of QT 4.8.6 it is now possible to build against libc++.  The binaries from homebrew (for example) are defaulted to be built against libc++.  Additionally, if one builds QT from scratch on an Xcode 6 machine, it will default to be libc++ based build.

### DIFF
--- a/CMake/SlicerBlockSetCMakeOSXVariables.cmake
+++ b/CMake/SlicerBlockSetCMakeOSXVariables.cmake
@@ -47,12 +47,14 @@ if(APPLE)
   #  11.x == Mac OSX 10.7 (Lion)
   #  12.x == Mac OSX 10.8 (Mountain Lion)
   #  13.x == Mac OSX 10.9 (Mavericks)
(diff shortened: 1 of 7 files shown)


pieper:
+1

jcfr:
* Changes associated with commit b766114 are now captured by  #223
* Changes associated with commits fe7b236 and 33e79f9  are now captured by #221
* Part of the changes associated with commit 60ee98a are now captured by #224

Update of BRAINSTools would be done in an other commit.

jcfr:
Commit [r23858](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=23858) updates version of ITK, swig and SimpleITK

jcfr:
Closing this PR. Update of BRAINSTools will be done in a different commit.

PR #217, opened 2015-01-06
ENH: add ability to refocus 3d cameras on a fiducial

naucoin:
Added a right click context menu option to refocus the 3D camera to the
first selected fiducal. This can help localise the fiducial in a large scene.
Added logic method to get all camera nodes and set the focal point on them.
Added the call to a self test.

Issue #3683

### DIFF
--- a/Modules/Loadable/Markups/Logic/vtkSlicerMarkupsLogic.cxx
+++ b/Modules/Loadable/Markups/Logic/vtkSlicerMarkupsLogic.cxx
@@ -31,6 +31,7 @@
 #include "vtkMRMLAnnotationTextDisplayNode.h"
 
 // MRML includes
+#include "vtkMRMLCameraNode.h"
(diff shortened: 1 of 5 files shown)


jcfr:
Looks good. Would be nice to split the functions into:

* `FocusCamerasOnNthPointInMarkup(const char *id, int n)`

and 

* `FocusCameraOnNthPointInMarkup(const char* cameraNodeIf, const char *id, int n)`

naucoin:
Makes sense, will refactor and push another commit.

jcfr:
@naucoin Thanks for updating the request.

I fixed few issue with the latest patch and integrated all the patches as [r23869](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=23869) and [r23870](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=23870)

PR #220, opened 2015-01-07
BUG: This will fix issue 3881

mehrtash:
This will fix the bug [1] which hides the sliceview annotations scalarbar for window width values
less than 1.

[1] http://na-mic.org/Mantis/view.php?id=3881


jcfr:
Integrated in [r23860](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=23860)

PR #222, opened 2015-01-08
ENH: Deprecate ::Zero and ::One

hjmjohnson:
These were deprecated in 2012; just before the ITK 4.2 release.
ZeroValue() and OneValue() should be used instead.

-- This coforms to ITKv4 changes made in
    6c23e1daace315a40831cf255b2af57df5d3505c
    3f712cd09551e3c80d8beac8a7a097b8bfc73862

When building with clang on 10.9 where ITKv4 is a shared library,
a linking error occured for unfound symbols: NumericTraits<
OutputPixelType >::Zero but using the
NumericTraits< OutputPixelType >::ZeroValue() allows linkage to occur properly.


jcfr:

Integrated in [r23857](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=23857)

PR #229, opened 2015-01-19
ENH: Allow for absolute paths in SLICER_ADD_PYTHON_TEST macro

vovythevov:
The default is still CMAKE_CURRENT_SOURCE_DIR but it allows the user to
give a full path to the script instead of being forced to have their script
in the current source directory.


jcfr:
:+1:  Good for integration

jcfr:
Integrated in [r23882](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=23882)

PR #228, opened 2015-01-15
ENH: Use the latest vtkImageReslice filter instead of vtkImageResliceMask

lassoan:
Features:
* Performance optimization: about 2x speed increase in slice rendering speed when the volume is linearly transformed (e.g., transformed MRHead refresh rate 25fps -> 50fps)
* Thick slices, MIP, DRR, etc. display modes are available (no GUI yet, but already can be enabled using Python), see the code samples below

Implementation:
* Convert composite transform to simple linear filter before setting it in the reslice filter (because vtkImageResliceMask can do some additional optimizations if it gets a simple linear transform)
* Use the latest vtkImageReslice filter instead of vtkImageResliceMask:
** mask output of vtkImageReslice is a stencil, therefore slice processing pipeline is simplified: vtkImageCast filter is not needed anymore and the more efficient vtkImageStencil is used instead of vtkImageMathematics
** custom vtkImageResliceMask class is removed: no need to maintain it any longer, etc.

Thick slice:

sliceNode = slicer.mrmlScene.GetNodeByID('vtkMRMLSliceNodeRed')
appLogic = slicer.app.applicationLogic()
sliceLogic = appLogic.GetSliceLogic(sliceNode)
sliceLayerLogic = sliceLogic.GetBackgroundLayer()
reslice = sliceLayerLogic.GetReslice()
reslice.SetSlabModeToMean()
reslice.SetSlabNumberOfSlices(10)
reslice.SetSlabSliceSpacingFraction(0.3)
sliceNode.Modified()

Maximum intensity projection (MIP):

sliceNode = slicer.mrmlScene.GetNodeByID('vtkMRMLSliceNodeRed')
appLogic = slicer.app.applicationLogic()
sliceLogic = appLogic.GetSliceLogic(sliceNode)
sliceLayerLogic = sliceLogic.GetBackgroundLayer()
reslice = sliceLayerLogic.GetReslice()
reslice.SetSlabModeToMax()
reslice.SetSlabNumberOfSlices(600)
reslice.SetSlabSliceSpacingFraction(0.5)
sliceNode.Modified()

PR #230, opened 2015-01-22
BUG: avoid false positives on geometry check

naucoin:
Add an epsilon value to the volumes logic and use it in the fuzzy compare instead
of the VTK default value. Add a precision calculation from the epsilon so that
any warning messages are printed out with the same precision as was used to
check the values.
Added a sanity check to see if the smallest volume spacing is within 3-10 orders
of magnitude of epsilon and print a warning if not.
Use a copy of the CTK significant decimals utility method to calculate the
precision with which to display differences in compared volumes.
Use a copy of the CTK order of magnitude utility method to compare the
spacing of the volumes with the current epsilon being used to
check them.
To avoid CTK includes which pull in a dependency on Qt, rehome the CTK
core utility methods here in the anonymous namespace until they get ported
to VTK. Reimplemented significantDecimals and orderOfMagnitude using standard
strings and stringstreams rather than Qt strings.
Added a python test to exercise the epsilon, precision and geometry check.

TBD: GUI for setting epsilon or just always use a volume spacing based value.

### DIFF
--- a/Modules/Loadable/Volumes/Logic/vtkSlicerVolumesLogic.cxx
+++ b/Modules/Loadable/Volumes/Logic/vtkSlicerVolumesLogic.cxx
@@ -53,12 +53,128 @@
 #include <vtkImageReslice.h>
 #include <vtkTransform.h>
 
-
-
+/// CTK includes
+/// to avoid CTK includes which pull in a dependency on Qt, rehome some CTK
+/// core utility methods here in the anonymous namespace until they get ported
+/// to VTK
 
 //----------------------------------------------------------------------------
 namespace
 {
+/// Return a "smart" number of decimals needed to display (in a gui) a floating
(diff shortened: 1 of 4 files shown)


jcfr:
@naucoin Thanks for working on this. :+1: 

I tested on Ubuntu 13.10 and the test passed. 

As a side note, you should coordinate with @lassoan . After PR #228 will be integrated, this topic will have to be re-based to resolve merge conflicts.

lassoan:
Maybe the same files are modified but different parts, so if there are merge conflicts it should not be difficult to resolve.

naucoin:
Merged as svn 23934 and git hash 9fe70aaf70e7e24081c20e7eff7fba236eed5108
http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=23934

PR #233, opened 2015-02-05
BUG: update management of crop volume parameters node

naucoin:
The module was holding a pointer to the parameter node that could get out of
date on scene changes and cause crashes. Switch to using a mrml node combo
box to manage the parameters node.
Adjusted the UI file to allow the generic module widget test to pass.
Update the widget more fully on enter and scene changes. Select a newly added
input volume or ROI if none already selected. When picking a new input volume,
or switching ot a new parameter set one, set it to be active in the slice viewers.
Update the parameters node constructor, destructor, read/write xml, copy and print
for missing calls. Make sure that the GUI updates the parameters node (was missing
the input volume and ROI node IDs).
Added a scene close to the self test, and add the CMake command to make it show
up as a self test in the GUI.
Added error checking, took out debugging print outs.

Tested against issues: #3947 #1764 #3313

Issue #3947

### DIFF
--- a/Modules/Loadable/CropVolume/MRML/vtkMRMLCropVolumeParametersNode.cxx
+++ b/Modules/Loadable/CropVolume/MRML/vtkMRMLCropVolumeParametersNode.cxx
@@ -39,11 +39,10 @@ vtkMRMLCropVolumeParametersNode::vtkMRMLCropVolumeParametersNode()
   this->OutputVolumeNodeID = NULL;
   this->ROINodeID = NULL;
 
-  this->ROIVisibility = false;
-  this->InterpolationMode = 2;
-
+  this->ROIVisibility = true;
   this->VoxelBased = false;
-
+  this->InterpolationMode = 2;
+  this->IsotropicResampling = false;
   this->SpacingScalingConst = 1.;
 }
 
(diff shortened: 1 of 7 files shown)


fedorov:
@naucoin I checked out your branch, looks good to me and seems to work great. Interface changes are fine. Thanks a lot!

naucoin:
@fedorov - thanks for checking!

jcfr:
Thanks for working on this. 

Few questions: Is it normal that the Cropping technique is not stored in the parameter node ? Should we store the technique and associated properties in the parameter node ?

naucoin:
The cropping technique is stored as the VoxelBased flag.

jcfr:
In that case, it does not seem to be restored if i create two parameter
node with different technique and switch btw them in the ui.

--

naucoin:
I thought I'd double checked that before committing, I'll take another look.

naucoin:
Pushed another commit that should fix this, the radio button group wasn't acting as I expected (it wasn't set up to always have at least one button checked so setting the voxel based button to false wasn't forcing the interpolation button to true).

jcfr:
:+1:  Work great ! Feel to integrate.

And thanks for the descriptive commit message.



[LGTM](http://www.lgtm.in/i/wKLvOMVVs)

[:+1:](http://www.lgtm.in/u/wKLvOMVVs) [:-1:](http://www.lgtm.in/r/wKLvOMVVs)

naucoin:
Commited as
5d8e51a124220ce031750a584be3b24ea9c78c76 
ee84fba523d85be3539bf8c83c8eb73d4f59cfff

PR #234, opened 2015-02-06
BUG: improve loading of DICOM referenced datasets

fedorov:
See issue http://www.na-mic.org/Bug/view.php?id=3952

* try to improve the sizing policy for the popup (works as expected on Mac)
* make popup modal
* add checks for the referenced datasets that will prevent them from being
   loaded when the list of files for the referenced dataset matches the one
   user already selected in the DICOM widget GUI

@cpinter could you please review if it addresses the issue?

### DIFF
--- a/Modules/Scripted/DICOMLib/DICOMWidgets.py
+++ b/Modules/Scripted/DICOMLib/DICOMWidgets.py
@@ -1,4 +1,4 @@
-import os, glob, sys
+import os, glob, sys, copy
 from __main__ import qt
 from __main__ import vtk
 from __main__ import ctk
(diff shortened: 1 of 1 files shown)


fedorov:
Here is how popup shows up on my system

screenshot at feb 06 13-27-55

cpinter:
Looks nice, thank you! Maybe a bit more space between the message, the checkboxes and the button?

fedorov:
OK. Do you know is there a convention how these spacings should be set to
be consistent with the rest of the application? I will search, just
checking in case you know.

cpinter:
Good question.
I didn't find this in the guidelines: http://www.slicer.org/slicerWiki/index.php/Documentation/Nightly/Developers/Style_Guide#Layouts

But I know that most modules have spacing values of 9px

fedorov:
Thank you!

fedorov:
@jcfr do you have any concerns about this pull request? It's been over 2 weeks since it was submitted, so I wanted to check when you think it will be integrated.

jcfr:
If @pieper is on board. LGTM.

pieper:
Yes, this looks good to me.  @fedorov you can go ahead and commit this to svn and then close the PR.

I know that we aren't (yet) using the full utility of github when we have the git-svn bridge, but until we have fully migrated to git it's still nice to use PRs to organize our code and our comments.  But feel free to ping like this if you aren't getting the needed feedback.

fedorov:
Committed in the following:

http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24015
http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24016

PR #232, opened 2015-02-02
ENH: Fix 3943 and other enhancements for slice view annotaitons

mehrtash:
* Fixed #3943 [1] by replacing DICOM study date and study time with series date and series time.
* Moved slice view annotations settings gear from Data Probe small widget (main  screen) to the Data probe module in order to preserve more screen real estate for the module uis.
* Fixed problem with color scalarbar range labels.
* Added an option to change format of color scalarbar range label and save the user settings in application settings.
* Change ruler text font size and family according to other text annotations.

[1] http://na-mic.org/Mantis/view.php?id=3943

PR #235, opened 2015-02-17
ENH: Fix 3943 and other enhancements for slice view annotaitons

mehrtash:
* Fixed #3943 by replacing DICOM study date and study time with series date and
series time.
* Moved slice view annotations settings gear from Data Probe small widget (main
 screen) to the Data probe module in order to preserve more screen real
  estate for the module uis.
  * Fixed problem with color scalarbar range labels.
  * Added an option to change format of color scalarbar range label and save the user settings in application settings.
  * Change ruler text font size and family according to other text
  annotations.

### DIFF
--- a/Modules/Scripted/DataProbe/DataProbe.py
+++ b/Modules/Scripted/DataProbe/DataProbe.py
@@ -1,6 +1,6 @@
 import os
 import unittest
-import qt, vtk
+import qt, vtk, ctk
 from __main__ import slicer
 import DataProbeLib
 
@@ -24,7 +24,7 @@ See <a>http://www.slicer.org</a> for details.  Module implemented by Steve Piepe
     # TODO: need a DataProbe icon
     #parent.icon = qt.QIcon(':Icons/XLarge/SlicerDownloadMRHead.png')
     self.parent = parent
-    self.infoWidget = 0
+    self.infoWidget = None
 
     if slicer.mrmlScene.GetTagByClassName( "vtkMRMLScriptedModuleNode" ) != 'ScriptedModule':
       slicer.mrmlScene.RegisterNodeClass(vtkMRMLScriptedModuleNode())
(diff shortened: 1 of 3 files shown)


mehrtash:
@jcfr @fedorov

I could reproduce the bug that you've mentioned in [1]. It is fixed now.

[1] https://github.com/Slicer/Slicer/pull/232

jcfr:
I just tested this and I am not able to see the "Gear" annotation settings. See screenshot below.

3943-build-of-2015-02-23

mehrtash:
That is the correct behavior. Please open the Dataprobe module from the modules list. You can find the settings gear there. It is moved there since we got comments that the gear wasting screen real estate.

mehrtash:
Here is the attachment showing the dataprobe module and settings gear.

jcfr:
I see. Is there any advantage in keep the settings UI as a popup ?  I guess we could reduce the number of clicks by adding the slice view annotation control directly to the DataProbe module panel ?

mehrtash:
Good point @jcfr. I applied your suggestion to the code and squashed the commits. Please test it again. Thanks.

jcfr:
Looks good. 

A different issue that could probably be addressed by a follow-up commit: When disabling and then enabling the slicer view annotations, all settings are enabled independently of their previous value.

mehrtash:
@jcfr It should be fixed now. However this is for the main features including: corner annotation texts, ruler and color scalarbar.

jcfr:
Fixed in [r23981](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=23981)

jcfr:
If you happen to do some more work on slicer view annotation, there is now an enum associated with vtkCornerAnnotation that should avoid using the integer 0, 1, 2 and 3 to identify the text location.

It has been added to Slicer/VTK6 and backported to Slicer/VTK5.
See https://github.com/Slicer/VTK/tree/slicer-release-5.10_2014-04-03
and https://github.com/Slicer/VTK/tree/slicer-v6.2.0-d733280_2015-02-05

PR #236, opened 2015-02-22
ENH: Allow application name to be defined by argv0

pieper:
The qCoreApplication::applicationName controls the discovery
of QSettings files and some other parts of the internal
machinery of the application.  When making a custom version of
slicer, we want to keep the settings independent of the
original slicer, so that, for example, they can have different
module paths related to the extensions that are bundled.

See:

http://www.slicer.org/slicerWiki/index.php/Documentation/Labs/CustomSlicerGenerator

### DIFF
--- a/Applications/SlicerApp/Main.cxx
+++ b/Applications/SlicerApp/Main.cxx
@@ -105,7 +105,19 @@ int SlicerAppMain(int argc, char* argv[])
 #endif
 #endif
 
-  QCoreApplication::setApplicationName("Slicer");
+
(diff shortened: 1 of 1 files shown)


pieper:
@jcfr can you think of any downsides for this?  The wiki page has some links and more discussion.

jcfr:
Fixed in [r24004](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24004)

@vovythevov This may interest you

jcfr:
@pieper :+1:

PR #237, opened 2015-03-02
ENH: Do no disable the picking in 2D since the widget is never pickable

vovythevov:
This prevent the following error to appear when using 2D projection:

```
ERROR: In c:\d\p\slicer-440-package\vtkv6\interaction\widgets\vtkWidgetRepresentation.h, line 176
vtkPointHandleRepresentation2D (0000000010537080): Subclass should allow enable/disable picking
```


jcfr:
Nice catch. 

Could you submit a new PR with more details in the commit message. 

Topic name should be named `XXXX-name-of-the-topic`  where `XXXX` is the issue number. 

Instead of `ENH`, uses `BUG`.

PR #238, opened 2015-03-02
BUG: Do not disable the picking in 2D, widget is never pickable #3971

vovythevov:
@naucoin, @jcfr:

This prevent the following error to appear when using 2D projection:
ERROR: In c:\d\p\slicer-440-package\vtkv6\interaction\widgets\vtkWidgetRepresentation.h, line 176
vtkPointHandleRepresentation2D (0000000010537080): Subclass should allow enable/disable picking

The fix is to remove the DisablePicking call. It's not used since for the
2D projection, the widget is never enabled (i.e. it's visible but the user
cannot interact with it).


jcfr:
@naucoin Thanks for your feedback on this.

jcfr:
Fixed in [r24009](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24009)

PR #240, opened 2015-03-02
COMP: Fix redistributable system libraries in packaging

vovythevov:
When Slicer switched to using component for packaging (issue #2397), it
broke the packaging since CMake macro InstallRequiredSytemLibraries does
not support components (see http://www.cmake.org/Bug/view.php?id=15428).

Until the macro is fixed, we can install the system libs under the right
component. More information in issue #3972

### DIFF
--- a/CMake/SlicerCPack.cmake
+++ b/CMake/SlicerCPack.cmake
@@ -45,6 +45,24 @@ if(NOT APPLE)
     include(${Slicer_CMAKE_DIR}/SlicerBlockInstallLibArchive.cmake)
   endif()
   include(InstallRequiredSystemLibraries)
+
+  # XXX: Remove this once CMake minimum version has been updated.
+  #      See Slicer issue #3972 and CMake issue #15428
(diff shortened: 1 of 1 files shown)


jcfr:
Chaned commit message with  the following and commit. Thanks

```
COMP: Fix redistributable system libraries in packaging. Fixes #3972

This commit fixes a regression introduced by r23158 (Explicitly installs
the needed component and skips the development component if not explicitly
requested.). When Slicer switched to using component for packaging, the VS
redistributable libraries were not included anymore in the package. This
is explained by the fact the CMake module "InstallRequiredSystemLibraries"
does not support components. 

Waiting the issue reported on CMake bug tracker [1] is resolved and the
minimum required CMake version is updated, this commit duplicates the logic
installing the redistributable libraries specifying the "RuntimeLibraries"
component.

[1] http://www.cmake.org/Bug/view.php?id=15428)

```

jcfr:
Fixed in [r24010](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24010)

PR #248, opened 2015-03-16
ENH: Added option to control the PropagateVolumeSelection behavior

mehrtash:
Added the following features:
* Volume selection can propagate selectively to background, foreground and label layers.
* HelperBox and EditUtil were modified to use the new propagation mode. Default mode will propagate through all layers while "BackgroundAndLabel" mode only propagates to background and label and doesn't touch foreground. 

This feature is useful in cases where different foregrounds are set to different views and we don't want editor to reset foregrounds through propagate volume selection. A use case is the ongoing project PCampReview [1].

[1]  [PCampReview](https://github.com/fedorov/PCampReview )

### DIFF
--- a/Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
+++ b/Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
@@ -68,9 +68,17 @@ vtkStandardNewMacro(vtkMRMLApplicationLogic);
 class vtkMRMLApplicationLogic::vtkInternal
 {
 public:
-  vtkInternal();
+  vtkInternal(vtkMRMLApplicationLogic * external);
+  void PropagateVolumeSelection(int layer, int fit);
   ~vtkInternal();
 
+  enum{
+    Background=0,
+    Foreground,
+    Label,
+    };
+
+  vtkMRMLApplicationLogic*        External;
   vtkSmartPointer<vtkMRMLSelectionNode>    SelectionNode;
   vtkSmartPointer<vtkMRMLInteractionNode>  InteractionNode;
   vtkSmartPointer<vtkCollection> SliceLogics;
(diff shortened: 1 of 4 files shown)


fedorov:
+1

as a background, this change has been coordinated with @pieper

Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
```diff
@@ -312,6 +312,93 @@ void vtkMRMLApplicationLogic::PropagateVolumeSelection(int fit)
     }
 }
 
+//----------------------------------------------------------------------------
+void vtkMRMLApplicationLogic::PropagateBackgroundVolumeSelection(int fit)
```
pieper: There's a lot of duplicated code in these methods - can you factor it out?  Maybe make a new method that takes a bitfield with a boolean for propagating each of the layers.  Then the original method can call it with 0x7 and each of these methods can call it with 0x4, 0x2, 0x1 (with an enum).

fedorov: Steve, I thought when we talked, you gave strong preference to separate methods instead of using enum, which I initially suggested.... 

It's funny, Alireza actually implemented the enum approach first, but I asked him to switch to separate methods, because I thought that's what you wanted. A lot of confusion....

pieper: I want the separate methods in the API, but I want to avoid the duplication in the code.  So the only confusion is I didn't explain that well : )

jcfr: @mehrtash Earlier today, I provided comment with suggestion how to address Steve remarks.

pieper:
+1 to @jcfr's suggestions!

mehrtash:
Thanks @jcfr and @pieper. I will edit the pull request based on your suggestions.

mehrtash:
@jcfr @pieper I updated the pull request based on your suggestions.

jcfr:
Reviewed and integrated in [r24106](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24106)

jcfr:
@mehrtash  @fedorov @pieper Please, consider reviewing PR #253  where I made the layer enum public and generalized the `propagationMode` to be any combination of layer.

PR #251, opened 2015-03-20
CMake cmp0054 warning fix

aylward:
Quoted variables in if() statements will no longer be dereferenced when the policy is set to NEW.


jcfr:

Doing a clean configure & build using CMake 3.2.1,  I got only one warning related to `unknown argument` in `slicerFunctionAddPythonQtResources`. It has been fixed by Slicer/Slicer@bee786b7fdf8.

Which version of CMake were you using ? Did this happen for a clean build of Slicer or in a different context ?

PR #252, opened 2015-03-24
BUG: Fixed editor behaviour in restoring correct label.

mehrtash:
To reproduce the issue:
* Load a volume, go to editor and create a merge labelmap.
* With Paint effect paint something with label 1.
* While on Label 1 select eraser, erase something and go back to painteffect.
* Select another label e.g 2.
* Paint something with painteffect. It works fine (paint with label 2).
* Choose another effect (e.g draw or want effect) and it would go back to the label 1 as it only stores the backup label when eraser is selected.

Fix: Backup label should also be set when a new color (not zero) is chosen not only after switching to eraser.

### DIFF
--- a/Modules/Scripted/EditorLib/EditUtil.py
+++ b/Modules/Scripted/EditorLib/EditUtil.py
@@ -120,6 +120,8 @@ class EditUtil(object):
         return slicer.mrmlScene.GetNodeByID(labelID)
 
   def getColorNode(self):
+    if not self.isEraseEffectEnabled():
+      self.backupLabel()
     labelNode = self.getLabelVolume()
     if labelNode:
       dispNode = labelNode.GetDisplayNode()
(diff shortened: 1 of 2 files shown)


mehrtash:
@fedorov @pieper  This will fix label restoring issue.

mehrtash:
I also added another commit for making the structures expandable as the height of the structures list was small and fixed.

fedorov:
+1

pieper:
Sounds good to me - I won't have time for a detailed test right now, so go
ahead and merge if you both think it is okay.  Otherwise I can give
feedback next week.

-Steve

jcfr:
Commit 04a1e6b integrated in [r24108](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24108)

jcfr:
Regardrding  7f64491, each time the user click on the expandable corner .. the list view expand without having the possibility to reduce its size.

jcfr:
@mehrtash  Since 7f64491 is causing the issue desribed in https://github.com/Slicer/Slicer/pull/252#issuecomment-85928548, I am closing this PR.

PR #255, opened 2015-03-30
STYLE: Simplify main.cpp using "/ENTRY:mainCRTStartup" link flag if need...

jcfr:
...ed

Suggested-by: Julien Finet <julien.finet@kitware.com>

### DIFF
--- a/Applications/SlicerApp/Main.cxx
+++ b/Applications/SlicerApp/Main.cxx
@@ -56,13 +56,7 @@
 #include <itkFactoryRegistration.h>
 
 // VTK includes
-//#include <vtkObject.h>
 #include <vtksys/SystemTools.hxx>
-
-#if defined (_WIN32) && !defined (Slicer_BUILD_WIN32_CONSOLE)
-# include <windows.h>
-# include <vtksys/Encoding.hxx>
-#endif
 
 namespace
 {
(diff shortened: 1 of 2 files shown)


jcfr:
Ping @lassoan @finetjul @pieper

pieper:
According to this [1], you can just use a pragma - I think that would be cleaner (keeps things localized rather than changing the cmake file).


```
#ifdef _MSC_VER
#    pragma comment(linker, "/subsystem:windows /ENTRY:mainCRTStartup")
#endif
```
[1] http://stackoverflow.com/questions/11785157/replacing-winmain-with-main-function-in-win32-programs

jcfr:
Thanks for reviewing.

I also thought about this. 

That said, since we already specify the `MACOSX_BUNDLE` for MacOSx. Specifying everything in the CMakeLists.txt would then be consistent.

* `WIN32` and `/ENTRY:mainCRTStartup` on windows
* `MACOSX_BUNDLE` on MacOSx

lassoan:
We would lose potentially useful information. The hPrevInatance argument can be used to ensure that only a single instance of Slicer is running (if you start Slicer and it is running already you can bring the running instance to top instead of starting a new instance). This is useful for tablets that run Slicer in a "kiosk mode" (we are getting there soon). Startup mode can be useful, too (e.g., to be able to create a shortcut that starts Slicer minimized or maximized - although the same can be achieved using command-line arguments). Getting the command-line arguments in wide characters could be beneficial, too (although I don't know what encoding is used if simple main() is used, maybe that is just as good).

pieper:
Yes, that makes some sense.

jcfr:
Thanks for comments. I will close this PR then.

PR #256, opened 2015-04-01
ENH: Extension Wizard: Substitute CMake variable reference if possible.

jcfr:
In case the main `CMakeLists.txt` of an extension contain references
to CMake variable, these will be expanded if possible.

For example, if the CMakeLists.txt contains the following:

```
  set(EXTENSION_DESCRIPTION "The ${PROJECT_NAME} module provides Slicer user with any easy way to export models into a KiwiViewer scene file.")
```

the generated description file will have the following description:

```
  description The SlicerToKiwiExporter module provides Slicer user with any easy way to export models into a KiwiViewer scene file.
```

### DIFF
--- a/Utilities/Scripts/SlicerWizard/ExtensionProject.py
+++ b/Utilities/Scripts/SlicerWizard/ExtensionProject.py
@@ -33,6 +33,8 @@ class ExtensionProject(object):
 
   _moduleInsertPlaceholder = "# NEXT_MODULE"
 
+  _referencedVariables = re.compile(r"\$\{([\w_\/\.\+\-]+)\}")
+
   #---------------------------------------------------------------------------
   def __init__(self, path, encoding=None):
     """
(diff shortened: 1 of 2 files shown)


jcfr:
@mwoehlke-kitware  When you have a chance, let me know what you think. Thanks

Utilities/Scripts/SlicerWizard/ExtensionProject.py
```diff
@@ -33,6 +33,8 @@ class ExtensionProject(object):
 
   _moduleInsertPlaceholder = "# NEXT_MODULE"
 
+  _referencedVariables = re.compile("\$\{([\w\_]+)\}")
```
mwoehlke-kitware: This doesn't seem right; don't you mean to use a raw string here? (Then, the `_` should not be escaped, and the other escapes will be seen by `re` as intended.)

Also, CMake for better or worse allows a lot more than `[\w_]` in variable names... not sure if that's important here or not, however.

Utilities/Scripts/SlicerWizard/ExtensionProject.py
```diff
@@ -160,14 +196,19 @@ def getValue(self, name, default=None):
...
-    command sets the same ``name``, the result is the raw argument to the last
-    such command. If the value consists of more than one argument, only the
-    first is returned.
+    build script. By default, no substitution is performed (the result may
+    contain substitution placeholders like '``${var}``'). If more than one
+    ``set()`` command sets the same ``name``, the result is the raw argument
+    to the last such command. If the value consists of more than one argument,
+    only the first is returned.
+
+    If ``substitute`` is ``True``, each occurence '``${var}``' will be updated
```
mwoehlke-kitware: Missing 'of'.

Utilities/Scripts/SlicerWizard/ExtensionProject.py
```diff
@@ -145,12 +147,46 @@ def project(self, value):
...
+    if len(variableNames) == 0:
+      return text
+
+    prefinedVariables = {}
+    try:
+       prefinedVariables[u"PROJECT_NAME"] = self.project
+    except EOFError:
+      pass
+
+    for name in prefinedVariables.keys():
```
mwoehlke-kitware: I wonder if this shouldn't handle nested references... I would have instead written the loop to find one reference, replace it, then start over, until no references remain. (As an optimization, you can pre-fill a cache of known variables with supported built-ins, i.e. `PROJECT_NAME`, check for hits, and add to the dict with `getValue` as needed.)

You would need a little clever logic to deal with unknown substitutions, though, assuming you want to leave them unexpanded (or you could expand them to `name-NOTFOUND`).

jcfr:
@mwoehlke-kitware I updated the topic to considering your remarks. Nested variable references
can are now partially handled.

If we consider the following example:
```
set(foo "world")
set(hello_${foo} "sun")
set(world_hello "earth")

set(description "${hello_${foo}} and ${${foo}_hello} are different")
```
Getting the value of `description` will return `hello_world-NOTFOUND and earth are different`

jcfr:
@mwoehlke-kitware When you have a chance, let me  know if this recent changes looks good to you.

mwoehlke-kitware:
Can I bill something for this?

Matthew

mwoehlke-kitware:
> If we consider the following example: [...]

That seems fine; it's not a perfect parser and doesn't claim to be. (Anyway, a "correct" fix for that would require actually evaluating `set` commands, which we did not do previously, either. This isn't unlike the other documented issue we have where we might see the "wrong" value if something is set twice, as we don't have any concept of 'point of use'.)

jcfr:
Integrated in [r24196](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24196)

PR #257, opened 2015-04-01
Adding label type column in Label Statistics

rweiss42:
The idea is to add a column to the table in Label Statistics that contains the name of the label taken from the label node's color node by GetColorName(i).  It also adds this to the CSV output.

It may be better to add this information to the labelStats dictionary.  The reason I'm doing it outside of this is because I noticed that we are casting the data in labelStats to float when adding to the table.  So instead of checking in the loop whether we are currently adding the label name and not trying to cast, I am just doing it outside of the loop (and therefore outside of the labelStats dictionary).

Is "type" the most appropriate word to use?  Should it be "name" instead?

### DIFF
--- a/Modules/Scripted/LabelStatistics/LabelStatistics.py
+++ b/Modules/Scripted/LabelStatistics/LabelStatistics.py
@@ -209,6 +209,8 @@ class LabelStatisticsWidget:
     self.view.verticalHeader().visible = False
     row = 0
     for i in self.logic.labelStats["Labels"]:
+      col = 0
+      
       color = qt.QColor()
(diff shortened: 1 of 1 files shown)


jcfr:
@rweiss42  Thanks for your contribution :+1: 

@pieper Could you review ?

pieper:
I didn't test it but it looks good to me - thanks for the contribution!

jcfr:
Great. I will do a final review and integrate then.

jcfr:
Integrated in [r24143](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24143)

PR #258, opened 2015-04-02
Add module interactive property

jcfr:
This PR introduces a new module property named `interactive`, by default it is enabled for all module. If set to `False`, no widget representation is expected.

Ping @lassoan @pieper @fedorov

### DIFF
--- a/Base/QTCore/qSlicerAbstractCoreModule.cxx
+++ b/Base/QTCore/qSlicerAbstractCoreModule.cxx
@@ -192,7 +192,13 @@ CTK_GET_CPP(qSlicerAbstractCoreModule, vtkSlicerApplicationLogic*, appLogic, App
 //-----------------------------------------------------------------------------
 bool qSlicerAbstractCoreModule::isHidden()const
 {
-  return false;
+  return this->isInteractive() ? false : true;
+}
+
+//-----------------------------------------------------------------------------
+bool qSlicerAbstractCoreModule::isInteractive()const
+{
+  return true;
 }
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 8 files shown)


fedorov:
JC, I think this is a much needed functionality to be able to hide selected modules.

I have doubts however "interactive" is the right name. For example, many (most?) of the CLIs can be used in batch mode, and by definition are not interactive, but most of them will have a widget. Interactive has a different meaning, in my opinion.

Would it make more sense to have this property called "hidden"? This would make more sense to me, and also will be in line with what is done for MRML nodes that should not show up in selectors.

jcfr:
There is already a property named `hidden`:

```
/// This property holds whether the module is visible to the user.
/// If the module is hidden, it doesn't appear in the list of modules menu.
/// However, the module is programatically accessible.
/// By default, modules are visible (hidden == false).
/// \sa isHidden
```

The new property 'interactive` means that a module is `hidden` AND is not expected to have widget representation allowing the user to directly interact with the module. 

```
/// This property holds whether the user can interact with the module or not.
/// If the module is non-interactive, it will not have a widget representation.
/// By default, modules are interactive (interactive == true).
/// \sa isInteractive
```

fedorov:
This makes sense, sorry for not reviewing the code - I should have noticed this.

lassoan:
I think this name claims more/different than what it is. Some module may display a custom toolbar or interact with the user through other means (popups, sound, custom devices, etc).

Something more specific, such as hasWidget would avoid all possible confusions and also clarify the difference between hidden (it has widget but it is not shown by default).

jcfr:
Agreed. What about `hasWidgetRepresentation` ?

fedorov:
+1

lassoan:
sounds good!

pieper:
I was going to suggest hasWidget too - but hasWidgetRepresentation is also
good.  +1

jcfr:
Thanks for the reviews.

It has been integrated in [r24152](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24152).

Since the `qSlicerModuleWidget` is now smarter, it can check if a `Widget` class is available or not and set the property `widgetRepresentationCreationEnabled` accordingly. 

It means that if your scripted module doesn't have a widget class there are no need to explicitly set any property . See [24153](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24153) to look how the DICOM plugin have been updated.

pieper:
@fedorov, @lassoan for consistency it would be good for other slicer dicom plugins to make the same changes as @jcfr has made in the bundled slicer dicom plugins.

PR #259, opened 2015-04-03
3549 import scripted module as python module

jcfr:
ENH: Fixes #3549. Import scripted module as python module

This will allow each scripted module to defined top-level variables that
will not clobber each other

See associated commits for more details.

Ping @pieper @fedorov @lassoan @blowekamp @cpinter @mehrtash @naucoin

Associated mantis issue is [#3549](http://www.na-mic.org/Bug/view.php?id=3549)

All test passes.

Building on this, this could potentially allow scripted module to alternatively be implemented as regular python module with a `__init__.py` ( See issue [#1365](http://www.na-mic.org/Bug/view.php?id=1365)). It means scripted module could also be organized [1] like this:
```
/path/to/Editor
/path/to/Editor/__init__.py
/path/to/Editor/Editor.py
/path/to/Editor/Effects/WandEffect.py
[...]
```

[1] This is just an example. some more discussion would need to happen to agree on a layout.


pieper:
Sounds like a great improvement!  +1

jcfr:
Integrated in [r24155](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24155)

PR #261, opened 2015-04-15
BUG: No measurement frame option was not taking into account output imag...

fbudin69500:
...e orientation

When not using a measurement frame to specify the tensor orientation, but when instead using the image direction, the input image direction was taken into account, but the output image information was not taken into account. When resampling tensors in the same space then their original space, if the direction matrix of the output space was not the identity matrix, this was leading to the tensors being misaligned.

### DIFF
--- a/Modules/CLI/ResampleDTIVolume/ResampleDTIVolume.cxx
+++ b/Modules/CLI/ResampleDTIVolume/ResampleDTIVolume.cxx
@@ -590,7 +590,7 @@ int ReadTransform( parameters & list,
 
 // Set Output parameters
 template <class PixelType>
-void SetOutputParameters( const parameters & list,
+itk::Matrix< double , 3 , 3 > SetOutputParameters( const parameters & list,
                           typename itk::DiffusionTensor3DResample<PixelType, PixelType>
                           ::Pointer & resampler,
                           const typename itk::Image<itk::DiffusionTensor3D<PixelType>, 3>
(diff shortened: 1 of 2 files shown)


Modules/CLI/ResampleDTIVolume/ResampleDTIVolume.cxx
```diff
@@ -1011,21 +1012,28 @@ int Do( parameters list )
...
+    outputImageDirection = SetOutputParameters<PixelType>( list, resampler, image );
+    if( !hasMeasurementFrame || list.noMeasurementFrame )
+    {
+      itk::Matrix<double, 3, 3> transposeOutputDirection ( outputImageDirection.GetTranspose() ) ;
+      measurementFrame = transposeOutputDirection * measurementFrame ;
+    }
     resampler->SetDefaultPixelValue( static_cast<PixelType>( defaultPixelValue ) );    // Could be set directly in the
                                                                                        // interpolator, but we keep
                                                                                        // ITKResampleImageFilter design
+    transform->SetMeasurementFrame( measurementFrame ); // This was probably wrong in the old versions
```
jcfr: Since you are updating this file, do you know what we should do about this call ? Is the comment ` This was probably wrong in the old versions` still valid ?

fbudin69500: This comment should be removed. It was added in Slicer3, in commit 12353. This is an old comment that is not relevant anymore.

jcfr:
Integrated in [r24194](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24194)

PR #263, opened 2015-04-28
BUG: Pattern matching for label volume does not account for underscore

rweiss42:
Off by one issue:

>>> master = 'something'
>>> label = 'something_123'
>>> label[len(master):]
'_123'
>>> label[len(master):].isdigit()
False
>>> label[len(master)+1:].isdigit()
True

### DIFF
--- a/Modules/Scripted/EditorLib/HelperBox.py
+++ b/Modules/Scripted/EditorLib/HelperBox.py
@@ -954,7 +954,7 @@ class HelperBox(object):
         nodeName = node.GetName()
         if nodeName.find(name) == 0:
           # prefix matches, is the rest all numbers?
-          if nodeName == name or nodeName[len(name):].isdigit():
+          if nodeName == name or nodeName[len(name)+1:].isdigit():
             return node
       except:
         pass
--- a/Modules/Scripted/EditorLib/HelperBox.py
+++ b/Modules/Scripted/EditorLib/HelperBox.py
@@ -945,21 +945,8 @@ class HelperBox(object):
 
   def getNodeByName(self, name):
     """get the first MRML node that has the given name
-    - use a regular expression to match names post-pended with numbers"""
-
-    slicer.mrmlScene.InitTraversal()
-    node = slicer.mrmlScene.GetNextNode()
-    while node:
-      try:
-        nodeName = node.GetName()
-        if nodeName.find(name) == 0:
-          # prefix matches, is the rest all numbers?
-          if nodeName == name or nodeName[len(name)+1:].isdigit():
-            return node
-      except:
-        pass
-      node = slicer.mrmlScene.GetNextNode()
-    return None
+    - use a regular expression to match names post-pended with addition characters"""
+    return slicer.util.getNode(name+"*")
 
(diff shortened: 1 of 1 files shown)


jcfr:

To account for both cases (with and without underscore), what about the following:

```
if nodeName == name or nodeName[len(name):].lstrip('_').isdigit():
```

Ping @mehrtash

rweiss42:
Is there any reason why a node would be named something like MasterName-label_123_456 that we would NOT want to grab as a label volume?  I don't think so, but I'm not sure.  Both solutions seem fine to me, adding 1 to len(name) is more restrictive in that it will only match the pattern AAAAA-label_### but I have no feelings one way or another as to if this is good/bad.

jcfr:
The idea was to support these cases:
* MasterName-label / MasterName-label123
* MasterName-label / MasterName-label_123

Considering the function is named `getNodeByName`, it also makes sense to be more general. Ideally, the function should be removed from `HelperBox.py` and instead the function `vtkMRMLScene::GetFirstNode` could be adapted to include a `byPrefix` parameter.

@mehrtash Was the previous implementation not working at all ? Or are we trying to support a new use case ?

pieper:
It's a heuristic anyway, so whatever we think is most useful is fine by me.

I was thinking we could also just change the entire method to be

slicer.util.getNode(name+"*")

jcfr:
@pieper :+1:

rweiss42:
you'd want slicer.util.getNode(name+"-label*"), no?

pieper:
Well, yes, but "-label" is added by the caller in current usage.

https://github.com/rweiss42/Slicer/blob/editor-pattern-matching/Modules/Scripted/EditorLib/HelperBox.py#L196-L197

mehrtash:
The previous implementation "if nodeName == name or nodeName[len(name):].isdigit():" could only grab the merge volume label for nodeName == name (the first created label volume for the merge volume e.g MRHead-label.) In case the user or code removes the initial label  and create a new one (MRHead_label_1 ... _n) the code fails to grab the new label and prompt the user to create a new label volume (by popping up the "Create a merge label map for ..." ). This happened in PCampReview module where we needed to remove nodes often and creating new merge labels.

rweiss42:
@pieper indeed.  Ok, I think replacing the guts of getNodeByName with slicer.util.getNode(name+"*") seems sensible.  Want me to go ahead and commit that here?

jcfr:
@rweiss42 Would be great if you could simplify the code and propose a patch where `slicer.util.getNode` is used. Thanks

jcfr:
The suggested fix wasn't working. Instead the method `vtkMRMLScene::GetFirstNode` has been extended with a parameter named `exactNameMatch` defaulting to `True`. If set to `False`, `byName` parameter is considered as a regular expression.

Fixed in https://github.com/Slicer/Slicer/compare/12113f96a6ca22a323cf34f5f17fd9ea4b9a9d43...f0bcf9b2788fc92140d98a4ce412937ec11710dd

pieper:
:+1:  Nice!

PR #268, opened 2015-05-01
Add support for view node visibility

jcfr:
After integrating this topic, calling `SetVisibility()` on a view node will effectively hide/show the associated widget in the layout.

Note that as detailed in 451973b, this set of patches disable the reading of view node visibility attribute from the MRML file. Associated comment is also reported below:

```
+    // XXX Do not read 'visibility' attribute and default to 1 because:
+    // (1) commit r21034 (STYLE: Add abstract class for all view nodes)
+    // changed the default value for 'visibility' attribute from 1 to 0. This
+    // means there are a lot of already saved scene where visibility attribute
+    // value is saved as 0.
+    // (2) support for visibility attribute by the layout manager has been
+    // added.
+    // XXX Support for 'visibility' attribute could be restored by updating
+    // the mrml version. Scene with a newer version number would consider the
+    // serialized attribute whereas older scene would not.
```

### DIFF
--- a/Libs/MRML/Core/vtkMRMLAbstractViewNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLAbstractViewNode.cxx
@@ -30,7 +30,7 @@ vtkMRMLAbstractViewNode::vtkMRMLAbstractViewNode()
 {
   this->LayoutLabel = NULL;
   this->Active = 0;
-  this->Visibility = 0;
+  this->Visibility = 1;
 
   double black[3] = {0.,0.,0.};
   memcpy(this->BackgroundColor, black, 3 * sizeof(double));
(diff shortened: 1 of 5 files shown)


jcfr:
@naucoin @fedorov @pieper @lassoan Thanks for reviewing.

pieper:
Seems reasonable.  Not sure how to test it.

jcfr:
@pieper I updated the last commit to introduce a new test named `qMRMLLayoutManagerVisibilityTest`

This test check that the widget visibility updates as expected when setting the `Visibility` on a view node. 

See https://github.com/jcfr/Slicer/blob/404c9b942cb9103abe357835b462861aca873ff7/Libs/MRML/Widgets/Testing/qMRMLLayoutManagerVisibilityTest.cxx

jcfr:
Integrated in [r24208](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24208) and [r24209](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24209)

PR #269, opened 2015-05-01
BUG: Fixed SlicerApp-real.exe.exe filename on Windows

lassoan:
Removed the double .exe extensions from the main application name.
Before the fix: SlicerApp-real.exe.exe
After the fix: SlicerApp-real.exe

### DIFF
--- a/CMake/SlicerMacroBuildApplication.cmake
+++ b/CMake/SlicerMacroBuildApplication.cmake
@@ -405,11 +405,8 @@ macro(slicerMacroBuildApplication)
   set(executable_name ${SLICERAPP_APPLICATION_NAME})
   if(NOT APPLE)
     set(executable_name ${executable_name}App-real)
-    if(WIN32)
-      set(executable_name ${executable_name}.exe)
-    endif()
   endif()
-  message(STATUS "Setting ${SLICERAPP_APPLICATION_NAME} executable name to '${executable_name}'")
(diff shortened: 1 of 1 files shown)


jcfr:
:+1:  Good catch. Thanks for the fix.

Before integrating to svn, could you also add the following text to the commit message:

```
This commit fixes a regression introduced in r24024 (ENH: Add
slicerMacroBuildAppLibrary and slicerMacroBuildApplication) where
it was wrongly assumed that specifying executable extension when
setting the OUTPUT_NAME target property was required.
```

lassoan:
updated commit message as requested

jcfr:
:+1:  Feel free to `git svn dcommit`

lassoan:
git-svn does not work on my computer, I would need to copy the change manually to SVN

jcfr:
Integrated in [r24206](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24206)

As a side note, extending git interface with `hub` makes it very easy to cherry pick commit from a pull request. See https://hub.github.com/ 

For example I just did:

```
$ git cherry-pick https://github.com/lassoan/Slicer/commit/136d20f10027e2ef169ec4c063c49cc2c7994c48
[...]
$ git log
[...] 
$ git svn dcommit --add-author-from
[...]
```

PR #271, opened 2015-05-05
ENH: Add function to explicitly set label outline state on all nodes

rweiss42:
Feature suggested by @pieper


pieper:
Thanks for the contribution - committed in r24214

PR #272, opened 2015-05-05
BUG: Fix "orphan view" issue in layout manager

jcfr:
**** WORK IN PROGRESS ****
Todo:
* Update test. Test doesn't pass yet. It needs to consider the
"MappedInLayout" attribute when checking visibility
* Update documentation to clearly document MappedInLayout vs Visibility.
**** WORK IN PROGRESS ****

The node visibility is now set considering both the "MappedInLayout"
and the Visibility properties.

It fixes a regression re-introduced in r24209 (ENH: Update layout
view factory to consider view node visibility property) that was
originally fixed in r24175 (BUG: Fixed orphaned views displayed in
secondary layout widgets).

### DIFF
--- a/Libs/MRML/Core/vtkMRMLAbstractViewNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLAbstractViewNode.cxx
@@ -246,3 +246,29 @@ void vtkMRMLAbstractViewNode::RemoveActiveFlagInScene()
       }
     }
 }
+
+//------------------------------------------------------------------------------
+int vtkMRMLAbstractViewNode::IsMappedInLayout()
+{
+  if (!this->GetAttribute("MappedInLayout"))
+    {
+    return 0;
+    }
+  return strcmp(this->GetAttribute("MappedInLayout"), "1") == 0;
+}
+
+//------------------------------------------------------------------------------
+void vtkMRMLAbstractViewNode::SetMappedInLayout(int value)
+{
+  if (this->IsMappedInLayout() == value)
+    {
+    return;
+    }
+  this->SetAttribute("MappedInLayout", value ? "1" : "0");
+}
+
+//------------------------------------------------------------------------------
+bool vtkMRMLAbstractViewNode::IsViewVisibleInLayout()
+{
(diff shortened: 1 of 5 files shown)


jcfr:
@finetjul If you have a chance, let me know what you think of this. 

For view node, here is my understanding of MappedInLayout vs Visibility:
* MappedInLayout: Specify if the view node is associated with the current layout or not
* Visibility: Specify if the widget associated with the view node should be visible when mapped to a layout.

Ping @lassoan

lassoan:
As I see in the code, MappedInLayout attribute is not used. Probably it would be better to remove it to make things simpler.

jcfr:
It is now used in my commit.

--

jcfr:
@lassoan @pieper @finetjul @naucoin 

Here the updated patch I plan on integrating. 

It extends `vtkMRMLAbstractViewNode` API adding:
* `IsMappedInLayout()`
* `SetMappedInLayout()`
* `IsViewVisibleInLayout()`

jcfr:
Integrated in [r24215](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24215)

lassoan:
MappedInLayout stores redundant information (the layout node must know what views it displays and where), so it should be removed from the node. The MappedInLayout information was not used for anything, I think it was not even set correctly (the value was always true), and of course it should not have been stored as a custom node attribute. It must have been some leftover from early prototyping or quick testing.

At some point we'll have to make the vtkMRMLLayoutNode a non-singleton node (having only one layout seriously limits using Slicer as a generic application platform; we cannot even display a secondary window that has a different layout then the main window), so we cannot afford to keep this kind of redundant code around.

jcfr:
Agreed. 

Here are few possible incremental changes:
* Change `MappedInLayout` from attribute store in the xml to a regular ivar. As you mentioned, this should not break any code. 
* Introduce a MRMLReference between `vtkMRMLAbstractViewNode` and `vtkMRMLLayoutNode`
* Change LayoutNode from singleton to regular node

PR #273, opened 2015-05-07
ENH: Added warning when GrowCut is attempted with unsupported image type

lassoan:
If GrowCut editor effect is attempted to be applied on an image or labelmap that is not short then a warning is displayed for the user.
See details here: http://www.na-mic.org/Bug/view.php?id=2873

### DIFF
--- a/Modules/Scripted/EditorLib/GrowCutEffect.py
+++ b/Modules/Scripted/EditorLib/GrowCutEffect.py
@@ -7,6 +7,7 @@ from __main__ import slicer
 from EditOptions import EditOptions
 from EditorLib import EditorLib
 import Effect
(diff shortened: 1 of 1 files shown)


jcfr:
Since developer mode and testing mode are different, both user and developer would see the warning. The popup wouldn't be displayed only when running Slicer using the test suite or after starting Slicer using the `--testing` flag.

jcfr:
Integrated in [r24428](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24428) and [r24429](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24429)

PR #274, opened 2015-05-07
BUG: Fixed labelmap selector in Editor

lassoan:
It was not possible to select any other labelmap than the default.

For example:
* Download MRHead sample
* Go to Editor module, click apply in the colormap selector
* Click "Set..." button to set another labelmap
* Click "Create new...", click apply in the colormap selector
=> ERROR: still the old labelmap is selected
* Click "Set..." button to set another labelmap
* Select the new labelmap
=> ERROR: still the old labelmap is selected

### DIFF
--- a/Modules/Scripted/EditorLib/HelperBox.py
+++ b/Modules/Scripted/EditorLib/HelperBox.py
@@ -110,8 +110,8 @@ class HelperBox(object):
     """select master volume - load merge volume if one with the correct name exists"""
 
     self.master = self.masterSelector.currentNode()
-    self.merge = None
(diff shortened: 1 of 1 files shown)


jcfr:
:+1: 

After checking that the scenario described in 1c08ec55c still work.

I would suggest to add the following text to the commit message:
```
This commit is a follow-up of r24207 (BUG: Editor: Ensure label map is updated each
time a master volume is selected )
```

lassoan:
Thanks, as I see it has been already integrated (rev 24207)

PR #275, opened 2015-05-08
ENH: Editor - Adding ability to delete only the selected structure.

rweiss42:
At @pieper 's suggestion, adding a button to the Editor that allows deletion of only the selected structure instead of deleting all.

Related conversation in https://github.com/fedorov/PCampReview/pull/32


pieper:
Merged http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24223

Thanks for the contribution!

PR #277, opened 2015-05-09
BUG: Resolve multithread issue that caused errors with multiple b val…

ljod:
…ues.

Fixes #3977

http://www.na-mic.org/Bug/view.php?id=3977

### DIFF
--- a/Libs/vtkTeem/vtkTeemEstimateDiffusionTensor.cxx
+++ b/Libs/vtkTeem/vtkTeemEstimateDiffusionTensor.cxx
@@ -78,6 +78,9 @@ vtkTeemEstimateDiffusionTensor::vtkTeemEstimateDiffusionTensor()
   this->SetDiffusionGradient(4,0,1,-1);
   this->SetDiffusionGradient(5,1,-1,0);
   this->SetDiffusionGradient(6,-1,0,1);
+  
+  this->RescaledDiffusionGradients = vtkDoubleArray::New();
+  this->RescaledDiffusionGradients->DeepCopy(this->DiffusionGradients);
 
 }
 
@@ -86,6 +89,7 @@ vtkTeemEstimateDiffusionTensor::~vtkTeemEstimateDiffusionTensor()
 {
   this->BValues->Delete();
   this->DiffusionGradients->Delete();
+  this->RescaledDiffusionGradients->Delete();
 #if (VTK_MAJOR_VERSION <= 5)
   this->Baseline->Delete();
   this->AverageDWI->Delete();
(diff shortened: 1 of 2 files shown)


jcfr:

It has been tested following the instruction in [#3977](http://www.na-mic.org/Bug/view.php?id=3977) and integrated as [r24240](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24240)

Next, I will add a test.

jcfr:
@ljod @pieper While testing with the tutorial dataset didn't show any difference after computing the FAs. Using Helix-DWI dataset shows some difference. See http://www.na-mic.org/Bug/view.php?id=3977#c13039

pieper:
I believe the issue there is that Subtract Scalar Volumes does a resampling step (all the arithmetic CLIs do) and I thing that's noisy for an essentially identity transform.  If that is the issue we should note it in the manual and file a new issue about it.

To test, we could try something like:

a1 = array('fa_1')
a2 = array('fa_2')
(a1-a2).max()

ljod:
Yes I agree. The subtract step was not reliable for detecting changes in
FA. The apparent FA changes I reported from b=1000 data were artifacts of
the subtraction module. The real bug was in multishell.

On Tuesday, May 12, 2015, Steve Pieper <notifications@github.com> wrote:

> I believe the issue there is that Subtract Scalar Volumes does a
> resampling step (all the arithmetic CLIs do) and I thing that's noisy for
> an essentially identity transform. If that is the issue we should note it
> in the manual and file a new issue about it.
>
> To test, we could try something like:
>
> a1 = array('fa_1')
> a2 = array('fa_2')
> (a1-a2).max()
>
> —
> <https://github.com/Slicer/Slicer/pull/277#issuecomment-101242802>.
>
> The information in this e-mail is intended only for the person to whom it
> is
> addressed. If you believe this e-mail was sent to you in error and the
> e-mail
> contains patient information, please contact the Partners Compliance
> HelpLine at
> http://www.partners.org/complianceline . If the e-mail was sent to you in
> error
> but does not contain patient information, please contact the sender and
> properly
> dispose of the e-mail.
>


-- 
Lauren O’Donnell, Ph.D.
Assistant Professor, Harvard Medical School
http://scholar.harvard.edu/laurenjodonnell
1249 Boylston St., Room #210
Boston, Massachusetts 02215
odonnell@bwh.harvard.edu

jcfr:
A self test using @pieper  approach has been added. See [r24248](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24248)

As a side note, using the Subtract module and setting the interpolation order to 0 (instead of the default value being 1), the difference of the two FAs is zero.

PR #279, opened 2015-05-13
ENH: Editor - Ensure a related merge volume is selected after a structure delete

rweiss42:
After a structure delete (one or all), the slice viewers don't get their label layers updated with any sort of logic, they just select the "next" available label volume (determined I think by the order in which the label volumes were added to the scene).  

Adding a call to select() in the delete methods should resolve this issue and will also make sure the Editor is in a good state following the removal of a node from the scene (which is probably a good thing to check for).

@pieper can you take a look?

### DIFF
--- a/Modules/Scripted/EditorLib/HelperBox.py
+++ b/Modules/Scripted/EditorLib/HelperBox.py
@@ -255,8 +255,19 @@ class HelperBox(object):
       self.updateStructures()
 
     if options.find("noEdit") < 0:
-      self.edit( label )
-
+      self.selectStructure(self.structures.rowCount()-1)
+      
+  
+  def selectStructure(self, idx):
+    """programmatically select the specified structure"""
+    selectionModel = self.structuresView.selectionModel()
+    selectionModel.select(qt.QItemSelection(self.structures.index(idx,0), 
+                                            self.structures.index(idx,4)), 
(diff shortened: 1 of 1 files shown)


rweiss42:
I've pushed two more commits that cause the StructuresView to get a row selected on both add and delete structure.  In the case of deleting all, we get the merge volume back and edit(0) called to send the editor to DefaultTool to make it a little more apparent to the user that no structure is currently selected.

This PR also adds a function to HelperBox that selects and activates a given row index in the structuresView.

pieper:
The code looks good and sounds like good features.  But there's a merge
conflict - can you update it to apply on top of the current master (a few
fixes were added in the past couple days).

https://github.com/Slicer/Slicer/commit/86140599d0e0d7b273cf595aced3a504bff7ad01

https://github.com/Slicer/Slicer/commit/c441ba649f54ab52c531f1b1a5399859006bea5d

rweiss42:
Okie dokie.  I *think* I managed managed to rebase this PR (and in the process learned like 12 new git verbs).  I think this should be good to go.

pieper:
Worked for me!

Committed to svn in r24279

(Agreed, the git commands take a while to get used to, but then again sometimes they can do magical things so it's worth getting good with them).

PR #280, opened 2015-05-20
BUG: ensure libvtkGUISupportQtOpenGL is packaged on mac

fedorov:
The library is not packaged, and this appears to be a regression
from the earlier fix here

https://github.com/Slicer/Slicer/commit/17564f2aba3c59c9cc45ee6d38e321cf878caa11

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -561,6 +561,7 @@ if(${VTK_VERSION_MAJOR} GREATER 5)
     vtkFiltersExtraction
     vtkFiltersFlowPaths
     vtkFiltersGeometry
+    vtkGUISupportQtOpenGL
     vtkIOImage
     vtkIOLegacy
     vtkIOPLY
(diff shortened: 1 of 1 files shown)


pieper:
@jcfr is out of email contact for a while.  I'd say go ahead and push this if you think it will fix the issues.

fedorov:
committed in r24268 = 65cfea39f4325505e1b7e2c75b14766c7c5b2c71

fedorov:
Should the following libraries be also packaged?

```
libvtkGUISupportQtWebkit 
libvtkGUISupportQtSQL 
```

This CLI is failing to load without them.

See discussion http://slicer-devel.65872.n3.nabble.com/libvtkGUISupportQtOpenGL-6-2-1-dylib-missing-on-MacOS-td4032757.html

pieper:
I would suspect that yes, they would be good to add.

fedorov:
Thanks Steve!

@cpinter @lassoan do you have any comments? 

do you observe any issues related to these libraries not packaged on mac for SlicerRT?

lassoan:
I don't have any comments on this, I have not tried to run Slicer on MacOS for a long time.

cpinter:
We had the same problem with other, similar libraries. Jc added them to the installer and it fixed it, so I think it's the way to go in this case too.

fedorov:
Thanks for the feedback! Webkit and SQL libraries added in 

r24269 = 5cbd16bed13c6034360b73841e3372ca688a1ddf

PR #283, opened 2015-05-29
ENH: Use vtkMRMLLabelMapVolumeNode class for labelmap volumes

lassoan:
Before vtkMRMLScalarVolumeNode was used for both scalar and label map
volumes and the LabelMap custom MRML node attribute was used for
distinguishing between them (0=scalar; 1=label map volume).
This made conversion between labelmap/scalar volumes very easy but made
it difficult to customize behavior, display, processing of segmentation
information.

Now a new vtkMRMLLabelMapVolumeNode class is used for storing segmentation
information (still using vtkMRMLScalarVolume used as base class for backward
compatibility; but in the future the base class may be changed to reflect
that segmentation can be represented in various ways, not just as volumes).

Notes:

A new "segmentation" node has been developed that is based on vtkMRMLLabelMapVolumeNode,
which can store segmentation information in various ways (not just labelmaps but planar
contours, closed surfaces, etc). To allow using of this new segmentation node,
first the vtkMRMLLabelMapVolumeNode has to be integrated into the Slicer core.

When reading a scene that was saved with an old version of Slicer, label maps are stored
in vtkMRMLScalarVolumeNode with label map attribute set to 1. Added a check for this case
and handle it by converting the vtkMRMLScalarVolumeNode to a vtkMRMLLabelMapVolumeNode during import.


lassoan:
@pieper @jcfr Please check this. This is a prerequisite for the new segmentation node. The segmentation node will be added first to the SlicerRT extension and once it fully matured and polished it can be moved to the trunk.

lassoan:
Integrated into Slicer core (rev 24291)

PR #284, opened 2015-06-02
Update dicomwidget to invoke new file loaded

jcfr:


jcfr:
Integrated in [r24415](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24415) and [r24416](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24416)

PR #285, opened 2015-06-03
BUG: Update extension build system to report packaging error on CDash.

jcfr:
See #3032

This commit adds a new target named "uploadpackage" that will
(1) build the standard 'package' target,
(2) extract the list of generated packages from its standard output and
(3) append the list of generated package filepaths to a file named PACKAGES.txt,
(4) upload the extension packages on midas.

Using this target an extension can be uploaded to midas without
relying on CTest.

Within the ctest script, the target "uploadpackage" is built and then
the URL to download packages from midas are uploaded to CDash based on the
content of the file PACKAGES.txt

This commit removes the now obsolete ExperimentalUploadOnly target.

### DIFF
--- a/CMake/SlicerExtensionCPack.cmake
+++ b/CMake/SlicerExtensionCPack.cmake
@@ -167,28 +167,7 @@ if(APPLE)
 endif()
 
 #-----------------------------------------------------------------------------
-# Set EXTENSION_UPLOAD_ONLY_COMMAND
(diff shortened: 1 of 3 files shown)


jcfr:
Integrated in [r24289](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24289)

PR #287, opened 2015-06-10
Refactor editor helperbox

jcfr:


jcfr:
@cpinter @lassoan @pieper These are some changes that I would like to integrate now. The serve as basis for the topic [SlicerRt/Slicer@segmentation-editor](https://github.com/SlicerRt/Slicer/tree/segmentation-editor) already started by @cpinter. 

A new version of this topic adapted from the topic of @cpinter is now available here: https://github.com/jcfr/Slicer/tree/segmentation-editor-2

Make sure to consider this new topic for future work on segmentation/editor support.

Modules/Scripted/EditorLib/HelperBox.py
```diff
@@ -102,7 +101,7 @@ def createMerge(self):
 
     if not merge:
       merge = self.volumesLogic.CreateAndAddLabelVolume( slicer.mrmlScene, self.master, mergeName )
-      merge.GetDisplayNode().SetAndObserveColorNodeID( self.colorSelector.currentNodeID )
+      merge.GetDisplayNode().SetAndObserveColorNodeID( self.editorLUTNodeID )
```
pieper: I'd prefer to see Color rather than LUT here.

Modules/Scripted/EditorLib/HelperBox.py
```diff
@@ -165,19 +165,17 @@ def setVolumes(self,masterVolume,mergeVolume):
 
   def setMasterVolume(self,masterVolume):
     """select merge volume"""
+    if isinstance(masterVolume, basestring):
+      masterVolume = slicer.mrmlScene.GetNodeByID(masterVolume)
     self.masterSelector.setCurrentNode( masterVolume )
-    self.select()
```
pieper: There's some tricky corner cases around switching master volumes and bringing up corresponding merge and per-structure label volumes.  @fedorov pointed out an issue lately where this was problematic when controlling the editor from another module.  We should look out for any issues that fall out from changing this logic.

Modules/Scripted/EditorLib/HelperBox.py
jcfr: what about `colorNodeID` ? That will be consistent.

Modules/Scripted/EditorLib/HelperBox.py
```diff
@@ -584,8 +582,6 @@ def updateStructures(self,caller=None, event=None):
     self.setMergeButton.setDisabled(not self.master)
 
     # reset to a fresh model
-    self.brushes = []
-    self.items = []
```
pieper: The reason for these lists was to handle memory management -- the QItems are not QObjects so they fall into a gray zone as far as PythonQt.

http://massmail.spl.harvard.edu/public-archives/slicer-devel/2014/016179.html

Has this been fixed?

Modules/Scripted/EditorLib/LabelSelectDialog.py
```diff
@@ -0,0 +1,94 @@
+
+import qt
+import slicer
+
+#-----------------------------------------------------------------------------
+def _map_property(objfunc, name):
```
pieper: Can you add a comment describing this function?

Modules/Scripted/EditorLib/HelperBox.py
jcfr: Agreed. 

On a similar topic, I wonder if it would make more sense to explicitly created the label map if none is available instead of trying to lazily create it. It means that we would check if a merge volume is needed in the `Editor/enter()` function.

Modules/Scripted/EditorLib/HelperBox.py
pieper: Yes, "editor" is redundant, but NodeID is a nice qualifier of the type.

Modules/Scripted/EditorLib/HelperBox.py
jcfr: Indeed. See commit message for more details: https://github.com/jcfr/Slicer/commit/a75da9c562778aeb0c76a38a6ae14ceb5e75439f

Modules/Scripted/EditorLib/HelperBox.py
pieper: In general it is explicitly created on enter() - at least the user is
prompted to create one if one isn't found.

https://github.com/Slicer/Slicer/blob/master/Modules/Scripted/Editor/Editor.py#L131-L151

or did you mean something else?

jcfr: Agreed. That is the observed behaviour after this change is implemented. 

@fedorov: I would be happy to learn more about your use case to improve/fix the existing interface.

Modules/Scripted/EditorLib/LabelSelectDialog.py
jcfr: What about the following:

```
def _map_property(objfunc, attributename):
   """Given ``objfunc`` returning a python object, this function creates a python
   :class:`property` which setter and getter are associated with ``attributename``.
   """
   [...]
```

I guess we could add this to `slicer.util`.

@mwoehlke-kitware Since you are the original author of this function, would appreciate your feedback :)

fedorov:
JC, our main use case is here https://github.com/fedorov/PCampReview

I will test, and I think @rweiss42 will want to look at it. Please do not merge until we test -- figuring out bugs with the Editor selection node is not fun (at least, not for me!)

Modules/Scripted/EditorLib/LabelSelectDialog.py
mwoehlke-kitware: May be worth noting that `objfunc` takes a `self`. The original usage AFAICT is to take something with a Qt UI (i.e. from a `.ui` file) and create a Pythonic `property` on the object owning the UI that maps to a property on a Qt object within the UI. The `objfunc` is a lambda to obtain the appropriate object from the UI.

I'm not sure `attributename` is appropriate, as that parameter refers to a property of a Qt object. (Personally I'd leave it `name` — which I guess I did when I originally wrote it :smile: — referring to the "property" of the function name.)

jcfr: Thanks for your comments.

It seems the function is not specific to QObject, it allows to create a property associated with the attribute of an other object.

Here is an updated description:

```
  """Creates a python :class:`property` associated with a QObject
  property ``name``. ``objfunc`` is a function that takes a ``self`` as
  parameter and returns the QObject to consider.
  """
```

mwoehlke-kitware: > It seems the function is not specific to QObject, it allows to create a property associated with the attribute of an other object.

Yes, using `getattr` and `setattr` should work with any Python entity which supports those in a reasonable manner. (Actually, I guess in that sense, "attributename" would in fact be reasonable. I just used it for QObject properties.) In which case the above should be fine replacing "QObject" with "object" (don't forget the "a" → "an" :grin:).

Nit: capitalize "Python"?

jcfr: Great. Thanks for the review.

```
  """Creates a Python :class:`property` associated with an object
  ``attributename``. ``objfunc`` is a function that takes a ``self`` as
  parameter and returns the object to consider.
  """
```

jcfr:
@pieper @fedorov @cpinter @lassoan This updated topic the editor moving the "Per-structure" buttons and functions into a dedicated widget.

We are getting close to have a modular base to get started with the integration of `qMRMLSegmentsTableView`

Note: I still need to test and confirm it all works as expected

As already pointed out by @cpinter, the *tricky* part will be to minimize the backward incompatible changes so that effect can draw on hidden *temporary* label map

cpinter:
Great progress today, thanks!

jcfr:
@pieper  @cpinter @lassoan  Et voila. This topic is ready for integration.

@rweiss42  @fedorov If you can provide me with some test data, I would be happy to check if it works for [PCampReview](https://github.com/fedorov/PCampReview)

fedorov:
I need to adapt PCampReview to the changes related to LabelNode introduction, it will not work with the current trunk. I put a test dataset here http://slicer.kitware.com/midas3/item/203506, feel free to try, but it may be difficult without explanation/documentation. I am busy with other things right now. Is integration of this topic urgent?

fedorov:
I would prefer to test this, but if this is an urgent matter, go ahead and integrate. I won't have time to work on this today.

jcfr:
It can definitively wait for next week.

jcfr:
I looked at all the existing effect and here is a short summary of each one of them get their input label images:

```
ChangeIslandEffect.py
  => self.getScopedLabelInput() / self.getScopedLabelOutput() / self.applyScopedLabel()

ChangeLabelEffect.py
  => self.getScopedLabelInput() / self.getScopedLabelOutput() / self.applyScopedLabel()

DilateEffect.py
  => self.getScopedLabelInput() / self.getScopedLabelOutput() / self.applyScopedLabel()

DrawEffect.py
  => self.logic.applyPolyMask(self.polyData)

Effect.py
  => Base class

ErodeEffect.py
  => self.getScopedLabelInput() / self.getScopedLabelOutput() / self.applyScopedLabel()

FastMarchingEffect.py
  => EditUtil.getLabelImage()

GrowCutEffect.py
  => self.getScopedLabelInput() / self.applyScopedLabel()

IdentifyIslandsEffect.py
  => self.getScopedLabelInput() / self.applyScopedLabel()

IslandEffect.py
  => Base class for IdentifyIslandsEffect

LabelEffect.py: Base class
  Provides "applyPolyMask" calling "applyImageMask"
  => self.sliceLogic.GetLabelLayer().GetVolumeNode().GetImageData()

LevelTracingEffect.py
  => applyPolyMask(polydata) 

MakeModelEffect.py
  => EditUtil.getLabelVolume().GetID()
     EditUtil.getLabel()

MorphologyEffect.py
  => Base class for erode and dilate

PaintEffect.py
   => self.sliceWidget.sliceLogic().GetVolumeNode().GetImageData()

RectangleEffect.py
  => applyPolyMask(polydata)

RemoveIslandsEffect.py
  => self.getScopedLabelInput()

SaveIslandEffect.py
  => self.getScopedLabelInput()

ThresholdEffect.py
  => self.editUtil.getLabelImage()

WandEffect.py
  => self.sliceLogic.GetLabelLayer().GetVolumeNode()

WatershedFromMarkerEffect.py
  => self.sliceLogic.GetLabelLayer().GetVolumeNode()
```

rweiss42:
Assuming I can get some other code out the door today, I too should be able to look at this next week.

jcfr:
@rweiss42 That would be great

jcfr:
@pieper I noticed that there is the `Effect,scope` parameter. Few remarks:
* Seems the `scope` option is explicitly disabled for morphology based effect since they have been ported to Slicer4. See [here](https://github.com/Slicer/Slicer/blob/9dce0519d08e254bc26a50b624b33816c9ab7cdd/Modules/Scripted/EditorLib/MorphologyEffect.py#L35-L36) and https://github.com/Slicer/Slicer/commit/5127b9342d636944a7f9caf4b5ff8743ea664425
* Tried to enabled the scope for morphology effect but the scope doesn't work for dilate effect. See screenshot below comparing observed result between current Slicer trunk and the version 3.6.

effect-scope-broken-in-slicer4-for-morphology-operator


Since I would like to provide a unified interface to access the label map node and image that is updated by the different effects. I suggest the following plan:

* Step1: remove the support for *scope*
 * remove `Effect.getScopedLayer()`, `Effect.getScopedBackground()` and `Effect.getScopedLabelInput()` functions
 * remove `Effect.getScopedLabelOutput()` function
 * instead use `self.sliceLogic.GetLabelLayer().GetVolumeNode()` and `self.imageBuffer`

* Step2: introduce the `Effect` property `image` that would correspond the label image currently being edited. It could either be the image data of a labelmap node or the one of segment.

* Step3: In `EditUtil`, remove or update function like the following ones `getCompositeNode/getSliceWidget/getSliceLogic/getBackgroundImage/getBackgroundVolume/getBackgroundID/getLabelImage/getLabelID/getLabelVolume/getLabelVolume` to make them context specific. Effect shouldn't assume the volume in the red slice viewer is the one being edited. This should be set at the application level.

pieper:
Here's a related feature request:

http://www.na-mic.org/Bug/view.php?id=2287

The idea of the scope was to make a general stencil-like restriction so that operations only happen within the user-selected region.  

https://github.com/Slicer/Slicer/blob/master/Modules/Scripted/EditorLib/Effect.py#L49-L58

pieper:
Can we discuss this on a hangout?  I want to make sure we are on the same
page about what is already there and the desired behavior.  Maybe this
afternoon.  There are a lot of things going on in the code and I want to
make sure we don't break anything while we add improvements.

lassoan:
I would be interested in the hangout, too

jcfr:
Great. What about scheduling a hangout for 1pm EST today ?

> re: The idea of the scope was to make a general stencil-like restriction so that operations only happen within the user-selected region. 

Make sense to keep the code around then. 

In that case the `labelImage` property could look like this:

```
@property
def labelImage(self):
  if scope == 'Visible':
    return self._scopedLabelImage
  else
    return self._labelImage

@labelImage.setter
def labelImage(self, value):
  if scope == 'Visible':
    self._scopedLabelImage = self.getScopeLayer(value)
  else:
    self._labelImage = value
```

or we could keep things the way they are ... 


I think having a discussion about the interface in general and we want to integrate things with the new segmentation node would be good.

pieper:
One o'clock is a little early for me.  How about three?  @fedorov wants to
join too.

rweiss42:
I'd like to listen in too if that's cool.  I can take a break from coding whenever, just shoot me an invite, weiss.robin.m@gmail.com<mailto:weiss.robin.m@gmail.com>

Robin

--
Robin M. Weiss
Research Application Engineer
Research Computing Center
The University of Chicago
6030 S. Ellis Ave., Suite 289C
Chicago, IL 60637
773.702.9030
robinweiss@uchicago.edu<mailto:robinweiss@uchicago.edu>

On Jun 12, 2015, at 10:32 AM, Steve Pieper <notifications@github.com<mailto:notifications@github.com>>
 wrote:

One o'clock is a little early for me. How about three? @fedorov wants to
join too.

cpinter:
I'll be on the road back from world congress most of the afternoon, but you can do it without me. I think I explained most of my points to Jc already. Please take minutes though :)

jcfr:
@pieper: I have a meeting at 3pm, would 2pm or 4pm be better ?

pieper:
Let's do four unless anyone has a conflict.

jcfr:
It turns out that my other meeting has been moved at 2pm ... I scheduled the discussion for 3pm.

fedorov:
Other extensions that are reusing Editor widgets:
* Reporting
* LongitudinalPETCT
* PET-IndiC
* Chest Imaging Platform (from the screenshots I've seen)
* iGyne (I think, I haven't used it personally)

Extensions that introduce new Editor effects:
* WindowLevel
* PETTumorSegmentations
* Carrera
* FastGrowCut

pieper:
This is another editor effect that is not an extension but demonstrates
another use case.

https://github.com/pieper/SlicerCL/tree/master/GrowCutCL

jcfr:
Integrated in r24310 to r24337

PR #288, opened 2015-06-10
3956 repeated saves of mrbs cause corruption rebase

naucoin:
Starting from lost fiducials when saving MRBs multiple times, I found and fixed a host of bugs related to fiducials, scene views, volumes, storage nodes, the MRML scene, and saving MRBs and MRML files. 
These fixes were mostly necessary due to the fact that once an MRB has been expanded and loaded into Slicer, the files are removed from disk and file paths can become invalid.

Changes of note:
- add missing storage nodes for storable nodes when saving them in a scene view
- more book keeping for file list members so that they can be restored after MRB operations
- use a mini scene for volume loading to avoid adding and removing lots of types of nodes
- warn users if they add storable data after making a scene view (will be lost on restore)
- remove the scene root dir variable on the mrml node as it wasn't staying in synch with the scene's root directory, made needed changes to the XML parser to set the scene on the storage node early so that absolute file paths can be computed correctly.

I can cancel this pull request and make one with one commit, this request is to get feedback on the changes in a more modular fashion.

### DIFF
--- a/Applications/SlicerApp/Testing/Python/CMakeLists.txt
+++ b/Applications/SlicerApp/Testing/Python/CMakeLists.txt
@@ -206,6 +206,7 @@ if(Slicer_USE_QtTesting AND Slicer_USE_PYTHONQT)
   slicer_add_python_unittest(SCRIPT test_tractography_display.py)
   slicer_add_python_unittest(SCRIPT sceneImport2428.py)
   slicer_add_python_unittest(SCRIPT SlicerMRBTest.py)
+  slicer_add_python_unittest(SCRIPT SlicerMRBMultipleSaveRestoreLoopTest.py)
   slicer_add_python_unittest(SCRIPT SlicerMRBMultipleSaveRestoreTest.py)
   slicer_add_python_unittest(SCRIPT Slicer4Minute.py)
   slicer_add_python_unittest(SCRIPT Charting.py)
@@ -229,6 +230,7 @@ if(Slicer_USE_QtTesting AND Slicer_USE_PYTHONQT)
     test_tractography_display.py
     sceneImport2428.py
     SlicerMRBTest.py
+    SlicerMRBMultipleSaveRestoreLoopTest.py
     SlicerMRBMultipleSaveRestoreTest.py
     Slicer4Minute.py
     Charting.py
--- a/Applications/SlicerApp/Testing/Python/CMakeLists.txt
+++ b/Applications/SlicerApp/Testing/Python/CMakeLists.txt
@@ -208,6 +208,7 @@ if(Slicer_USE_QtTesting AND Slicer_USE_PYTHONQT)
   slicer_add_python_unittest(SCRIPT SlicerMRBTest.py)
   slicer_add_python_unittest(SCRIPT SlicerMRBMultipleSaveRestoreLoopTest.py)
   slicer_add_python_unittest(SCRIPT SlicerMRBMultipleSaveRestoreTest.py)
+  slicer_add_python_unittest(SCRIPT SlicerMRBSaveRestoreCheckPathsTest.py)
   slicer_add_python_unittest(SCRIPT Slicer4Minute.py)
   slicer_add_python_unittest(SCRIPT Charting.py)
   slicer_add_python_unittest(SCRIPT SliceLinkLogic.py)
@@ -232,6 +233,7 @@ if(Slicer_USE_QtTesting AND Slicer_USE_PYTHONQT)
     SlicerMRBTest.py
     SlicerMRBMultipleSaveRestoreLoopTest.py
     SlicerMRBMultipleSaveRestoreTest.py
+    SlicerMRBSaveRestoreCheckPathsTest.py
     Slicer4Minute.py
     Charting.py
     SliceLinkLogic.py
--- a/Applications/SlicerApp/Testing/Python/CMakeLists.txt
+++ b/Applications/SlicerApp/Testing/Python/CMakeLists.txt
@@ -206,6 +206,7 @@ if(Slicer_USE_QtTesting AND Slicer_USE_PYTHONQT)
   slicer_add_python_unittest(SCRIPT test_tractography_display.py)
   slicer_add_python_unittest(SCRIPT sceneImport2428.py)
   slicer_add_python_unittest(SCRIPT SlicerMRBTest.py)
+  slicer_add_python_unittest(SCRIPT SlicerMRBMultipleSaveRestoreTest.py)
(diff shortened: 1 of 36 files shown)


jcfr:
Great work. 

I like having multiple commits with a explicit commit message.

I will try the topic asap.

pieper:
Agreed - nicely done!

jcfr:
@naucoin I built the topic and no issue on that side.

That said, I did the following and I wonder what the message means exactly.

* Load MRHead twice. You will then have `MRHead` and `MRHead_1` data loaded.
* Go to Volume rendering module, setup volume rendering for `MRHead`
* Save MRB
* Clear scene
* Load MRB
* In Volume rendering module, if you select `MRHead_1` you get the following message:

screenshot from 2015-06-16 18 42 15

Since I didn't create any sceneview and didn't added any data I wonder what this means. 

As a side note, the message could be formatted with some new lines.

naucoin:
@jcfr : the automatic MRB scene view was a requirement from Ron, it's needed to be the "thumbnail" when the MRB is in the data store. 
For the record, I do check if new storable nodes have "hide from editors" set to 1 or "save with scene" set to 0 and don't pop up the warning then.
Where would you suggest adding newlines to the message?

jcfr:
(1) A new line between each sentence. A minor tweak would also be to use a regular `OK` button 

(2) I still think providing a thumbnail and automatically creating a sceneview are two separate things. The thumbnail could systematically be created when saving a MRB (or even a scene). This will satisfy the requirement of associating MRB with a thumbnail without systematically creating a sceneview. As a user not dealing with sceneview, the message is confusing. I guess this is a minor issue that could be addressed later. The settings `SceneViews/NeverShowLostDataWarning` will allow to easily disable the popup.

fedorov:
+1 JC's comment 2

Users who do not use/know about sceneviews should not suffer from this. I
think majority of users fall into that category.

naucoin:
1. There is a newline after each sentence, do you mean leaving a blank line between them?
I though about Ok vs Acknowledged, it's a message that I feel needs more consideration than an Ok seems to imply - you can lose data if you proceed.

2. It would be non-trivial to manage a screen snapshot without a sceneview. The warning should only come up for nodes that the user is concerned with, I'd vote to change the settings on the volume rendering property node to put it into the "user doesn't need to worry about this" category. I'll defer to Ron wrt the bigger picture about scene views as he's usually the driving factor behind my making sure they work.

fedorov:
Thinking more about this, I am not sure how many users use even MRML scene. I have been having continuous issues with MRML scenes, and opted for task-specific mechanisms for my users instead. So I retract my earlier comment. It will not have any effect on users I work with.

jcfr:
(1) By adding a new line, I meant this:

```
This is a sentence !

This is an other one.

This is a last one.
```

instead of 


```
This is a sentence !
This is an other one
This is a last one.
```

(2) Agreed. This discussion and changes could be done at some later point.

naucoin:
Pushed a commit to add white space to the message.
Hopefully the scene views will get completely revamped with Steve's DB based solution.

jcfr:
Thanks.

Re (2) this is more an issue in the MRB format specification. Independently of sceneview, we will want to associate thumbnail(s) with MRBs.

pieper:
I agree the wording of the message is difficult and can be confusing even
for a user who knows what Scene Views are.  Maybe we can add some more
explanatory text.

naucoin:
Is it just the first sentence? Do you have any suggestions?

pieper:
The biggest issue in my mind is that if the user didn't explicitly create a
Scene View and they load a scene and get that message they would be
confused.  If I understand Jc's workflow steps all he did was save and
restore the scene but then got a dialog talking about Scene Views.

So maybe the real behavior should be:

1) save the scene and there's a default Master Scene View

2) load the scene and don't pay attention to new nodes being added.

3) when the user tries to restore a Scene View, examine the scene and the
Scene View and see if there's a different set of nodes.  If they are
different, ask the user if they want to copy missing nodes from the scene
into the Scene View (with a warning that the new data will otherwise be
lost).

The nice thing about this option is that not only is the dialog shown only
in a context that makes some sense, this behavior would also help fix the
underlying issue of data being lost when restoring scene views.

jcfr:
@pieper I got the message after (1) loading a MRB having two nodes referencing MRHead data and then (2) selecting the second node in the volume rendering module.  It does NOT appear directly after loading the MRB.

pieper:
Right, but you never explicitly created a Scene View, correct?  It only
happened as a byproduct of saving the scene (I believe it happens the same
for mrb or regular mrml).

jcfr:
@pieper Indeed, I never created a scene view hence the confusion created by the message.

pieper:
Thanks @jcfr that makes sense.  I still think that we should investigate point 3 on my list above.  I think it would actually make Scene Views much more useful as a feature.  @naucoin what do you think?

naucoin:
Point 3 does make sense. Any chance we could merge this branch first and then I work on it on a new branch as a new feature request? I can take out the warning from this branch easily but don't want to delay the MRB specific fixes much longer.

jcfr:
+1 Sounds good. Make sense to disable the warning for now and integrate.

On Thu, Jun 18, 2015 at 10:26 AM, Nicole Aucoin <notifications@github.com>
wrote:

> Point 3 does make sense. Any chance we could merge this branch first and
> then I work on it on a new branch as a new feature request? I can take out
> the warning from this branch easily but don't want to delay the MRB
> specific fixes much longer.
>
> —
> <https://github.com/Slicer/Slicer/pull/288#issuecomment-113172533>.
>



-- 
+1 919 869 8849

naucoin:
Warning removed.

pieper:
Excellent - thank you.

-Steve

naucoin:
Can I integrate this today?

pieper:
@fedorov may want another day to see if we can debug the dicom private
dictionary issue.  What do you think Andrey?

fedorov:
I definitely would like to see this private dict issue sorted out. And as I alluded to earlier, I am ambivalent to MRBs.

naucoin:
Looks like the DICOM private dictionary change is on hold for a while, and htis change has bug fixes for scene views and MRML files, not just MRBs. 
Can I go ahead and rebase and integrate?

fedorov:
I confirm it is on hold until @jcfr is back

pieper:
Yes, let's get the MRB fix into the trunk.

naucoin:
Did it via git svn dcommit, checked in as svn 24370 to 24399, git 64f2ebc2f95745c9c7b6c3379b089bf31b0b8305 up to ce8488e644bb40a2af78780db580403578870f20

PR #292, opened 2015-06-18
Node selector multiple output types

lassoan:
Fixes http://www.na-mic.org/Bug/view.php?id=3748 and allows converting between scalar/labelmap node using the "Cast scalar volume" module


pieper:
I haven't tried but this looks great!

jcfr:
Thanks for your contribution.

Integrated in [r24350](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24350) and [r24351](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24351)

PR #293, opened 2015-06-18
BUG: String containing list of dependencies was not split

fbudin69500:
If one was creating an extension that depended on multiple other extensions and was giving, as specified in the s4ext files, a list of extensions separated by spaces, Slicer did not split that list when looking for dependencies upon download, and was trying to find an extension with all the names.
e.g. If your s4ext file contained the following line:
depends     DTIProcess ResampleDTIlogEuclidean
Slicer was trying to resolve the dependencies and was trying to find the extension named "DTIProcess ResampleDTIlogEuclidean", instead of finding both extensions separately.

### DIFF
--- a/Base/QTCore/qSlicerExtensionsManagerModel.cxx
+++ b/Base/QTCore/qSlicerExtensionsManagerModel.cxx
@@ -1289,7 +1289,7 @@ bool qSlicerExtensionsManagerModel::installExtension(
     }
 
   // Gather information on dependency extensions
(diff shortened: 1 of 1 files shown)


jcfr:

Integrated in [24343](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24343)

Wiki page documenting the `depends` property has also been updated. See http://wiki.slicer.org/slicerWiki/index.php/Documentation/Nightly/Developers/Extensions/DescriptionFile

PR #294, opened 2015-06-24
4010 add copy paste linear transform module

adamrankin:
A little feature I added that received positive feedback at NAMIC Summer. It adds the ability to copy/paste from the clipboard to/from the transforms module.

### DIFF
--- a/Modules/Loadable/Transforms/qSlicerTransformsModuleWidget.cxx
+++ b/Modules/Loadable/Transforms/qSlicerTransformsModuleWidget.cxx
@@ -20,6 +20,13 @@
 
 // Qt includes
 #include <QFileDialog>
+#include <QShortcut>
+#include <QApplication>
+#include <QClipboard>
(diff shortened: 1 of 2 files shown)


jcfr:

I applied some minor stylistic changes and added the option `QString::SkipEmptyParts` to `QString::split` call.
See https://github.com/jcfr/Slicer/tree/4010-add-copy-paste-linear-transform-module

It can now support case like this one where each line has some extra spaces at the end:
```
10 2 3 4 
5 60 7 8 
0 0 1 0   
0 0 0   1  
```

Let me know if that works for you.

jcfr:
After some more testing. I will integrate.

jcfr:
Integrated in [r24355](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24355)

Associated issue http://www.na-mic.org/Bug/view.php?id=4010

PR #300, opened 2015-07-08
ENH: Add option to turn on ITK Python wrapping.

thewtex:
This is currently only tested on the build tree with Linux.

This also requires a bump to the ITK Python to get a fix for the WrapITK.pth
file:

  https://github.com/InsightSoftwareConsortium/ITK/commit/13b368b2290ee0ac035e16aa1aed07ca8d03c957


jcfr:

Integrated in [r24405](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24405)

PR #301, opened 2015-07-10
ENH: Convert to using ITK version of MGHIO

hjmjohnson:
The ITK supplied version of MGHIO is a derivative
work that originated in Slicer.  The ITK version
has more widespread use in non-slicer envirnments and
can be used and maintained by a larger community.


hjmjohnson:
THIS IS NOT READY!!! 

When testing, I accidentally compiled the wrong source directory tree, and verified the wrong source code.

This patch does not yet compile  :(.  Changes are being made this weekend.

jcfr:
Hi Hans, thanks for the update.
Jc

jcfr:
And thanks for working on this. :)

PR #303, opened 2015-07-14
ENH: Added node printself output to node inspector

lassoan:
Detailed node content (output of node->PrintSelf) is now available in the node inspector.
If the node is changed then the window is updated automatically (if the node inspector is not visible then no update happens to avoid any potential performance regressions).
The vtkObject's irrelevant list of observers are removed from the output.


lassoan:
See screenshot of node info in the Data module:

nodeinfoscreenshot

jcfr:
:+1:

jcfr:
Is the Node inspector collapsed by default ?

lassoan:
Yes, the node inspector is it's collapsed by default.

jcfr:
Thanks.

Integrated in [24414](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24414)

PR #304, opened 2015-07-15
Update SimpleFilters with patches for LabelMapVolume node changes

blowekamp:
48b7873 Remove input based automatic LabelMap output restriction.
c90a778 Replace LabelMap checkbox with option in output combo box
d3f3bf1 Restore support for label map volumes as input
a560064 Correct nodeType to be a list of one element.
2f392a5 Update to use new MRMLLabelmapNode


jcfr:

Integrated in [r24431](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24431)

PR #306, opened 2015-07-15
STYLE: SimpleMarkupsWidget: Add Markups logic setter/getter.

jcfr:
Allow the widget to be used with any given Markups logic

### DIFF
--- a/Modules/Loadable/Markups/Widgets/qSlicerSimpleMarkupsWidget.cxx
+++ b/Modules/Loadable/Markups/Widgets/qSlicerSimpleMarkupsWidget.cxx
@@ -138,19 +138,6 @@ void qSlicerSimpleMarkupsWidget::enter()
 {
   Q_D(qSlicerSimpleMarkupsWidget);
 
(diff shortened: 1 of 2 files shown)


jcfr:
@mholden8 Could you review this patch ? Thanks

With this change, the user of the widget  will be expected to set the Markups logic.

mholden8:
@jcfr Thanks for doing this. This makes sense to me.

My one comment would be now that the markups logic (and scene) are not automatically set, if the user tries to use the widget without setting the markups logic, there will be a run-time error.

Probably there should be a check before each instance where the markups logic is used to ensure the logic has already been set. If you agree, I can add these checks.

jcfr:
> Probably there should be a check before each instance where the markups logic is used to ensure the logic has already been set.

Makes sense. Thanks for working on this.

jcfr:
Since I am travelling from August 12 to 19, if not already done, I will be integrating this topic when I return.

jcfr:
Closing this PR. It is obsoleted by commit [r24937](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24937)

PR #309, opened 2015-07-27
ENH: units/coordinates display in DataProbe

Punzo:
here I propose some modifications for implementing the two first points of:
http://www.na-mic.org/Bug/view.php?id=4020

main modifications:

1) GetPixelString is now a method of vtkMRMLScalarVolumeDisplayNode. Child class methods have
    been implemented (Label and DiffusionTensor).

2) for enabling customization of the coordinate and units display I have added the following method:
   a) GetCustomizeWorldCoordinates in the vtkMRMLScalarVolumeNode (extension can fill the method
        for showing different coordinate in DataProbe
   b) added in vtkMRMLVolumeDisplayNode the following attributes and relative GET/SET method (and         copy, WriteXML, etc.): 
         HorizontalQuantity = "length";
         VerticalQuantity = "length";
         DepthQuantity = "length";
         sysref = "RAS";

       extension can modify this attributes for displaying different coordinates and units for each
       each dimension (for the moment only for the GUI of DataProbe). DataProbe now use these, in 
       ProcessEvent, to format the string viewInfo.text. The coordinates and units display is set to the 
       first volume found with this order of priority : BackgroundLayer, ForegroundLayer and LabelLayer.

 @finetjul, @pieper may you check this and have your feedback?
Thanks.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLDiffusionTensorVolumeDisplayNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLDiffusionTensorVolumeDisplayNode.cxx
@@ -13,6 +13,7 @@ Version:   $Revision: 1.2 $
 =========================================================================auto=*/
 
 // MRML includes
+#include "vtkMRMLDiffusionTensorDisplayPropertiesNode.h"
 #include "vtkMRMLDiffusionTensorVolumeDisplayNode.h"
 #include "vtkMRMLDiffusionTensorVolumeSliceDisplayNode.h"
 #include "vtkMRMLScene.h"
@@ -35,6 +36,10 @@ Version:   $Revision: 1.2 $
 #include <vtkObjectFactory.h>
 #include <vtkSphereSource.h>
 #include <vtkVersion.h>
+#include <vtkPointData.h>
+#include <vtkTensor.h>
+#include <vtkFloatArray.h>
+#include <vtkDiffusionTensorMathematics.h>
 
 // STD includes
 
@@ -68,6 +73,24 @@ vtkMRMLDiffusionTensorVolumeDisplayNode
  sphere->Delete();
 
  this->AutoScalarRange = 1;
+
+ this->dtimath = vtkSmartPointer<vtkDiffusionTensorMathematics>::New();
+ this->singlepixelimage = vtkSmartPointer<vtkImageData>::New();
+ this->singlepixelimage->SetExtent(0, 0, 0, 0, 0, 0);
+#if (VTK_MAJOR_VERSION <= 5)
+ singlepixelimage->AllocateScalars();
+#endif
+ this->tensordata = vtkSmartPointer<vtkFloatArray>::New();
+
+ this->tensordata->SetNumberOfComponents(9);
+ this->tensordata->SetNumberOfTuples(singlepixelimage->GetNumberOfPoints());
+ this->singlepixelimage->GetPointData()->SetTensors(tensordata);
+
+#if (VTK_MAJOR_VERSION <= 5)
+ this->dtimath->SetInput(singlepixelimage);
+#else
+ this->dtimath->SetInputData(singlepixelimage);
+#endif
 }
 
 //----------------------------------------------------------------------------
@@ -83,6 +106,24 @@ vtkMRMLDiffusionTensorVolumeDisplayNode
   this->ImageCast->Delete();
 }
 
+namespace
+{
+//----------------------------------------------------------------------------
+template <typename T> std::string NumberToString(T V)
+{
+    std::string stringValue;
+    std::stringstream strstream;
+    strstream << V;
+    strstream >> stringValue;
+    return stringValue;
+}
+
+//----------------------------------------------------------------------------
+std::string DoubleToString(double Value)
+{
+    return NumberToString<double>(Value);
+}
+}// end namespace
 
 //----------------------------------------------------------------------------
 void vtkMRMLDiffusionTensorVolumeDisplayNode::WriteXML(ostream& of, int nIndent)
@@ -470,4 +511,47 @@ int vtkMRMLDiffusionTensorVolumeDisplayNode::GetNthScalarInvariant(int i)
   static std::vector<int> modes =
     vtkMRMLDiffusionTensorVolumeDisplayNode::GetSupportedColorModes();
   return modes[i];
+}
+
+//----------------------------------------------------------------------------
+const char *vtkMRMLDiffusionTensorVolumeDisplayNode::getPixelString(double *ijk)
+{
+    vtkImageData *imageData = this->GetVolumeNode()->GetImageData();
+    if(imageData == NULL){
+        return "No Image";
+    }
+
+    vtkIdType point_idx = imageData->FindPoint(ijk[0], ijk[1], ijk[2]);
+
+    if(point_idx == -1)
+      return "Out of bounds";
+
+    if(imageData->GetPointData() == NULL)
+      return "No Point Data";
+
+    if(imageData->GetPointData()->GetTensors() == NULL)
+      return "No Tensor Data";
+
+    double tensor[9];
+    double *a = &tensor[0];
+    imageData->GetPointData()->GetTensors()->GetTuple(point_idx, a);
+
+    int operation;
+
+    operation = this->GetScalarInvariant();
+
+    std::string valueString = this->GetScalarInvariantAsString();
+
+    const float *tensorf = (float*) &tensor[0];
+    this->tensordata->SetTupleValue(0, tensorf);
+    this->tensordata->Modified();
+    this->singlepixelimage->Modified();
+    this->dtimath->SetOperation(operation);
+    this->dtimath->Update();
+    vtkImageData *output = this->dtimath->GetOutput();
+
+    if(output && output->GetNumberOfScalarComponents() > 0)
+        valueString += " " + DoubleToString(output->GetScalarComponentAsDouble(0, 0, 0, 0));
+
+    return valueString.c_str();
 }
--- a/Libs/MRML/Core/vtkMRMLDiffusionTensorVolumeDisplayNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLDiffusionTensorVolumeDisplayNode.cxx
@@ -41,8 +41,6 @@ Version:   $Revision: 1.2 $
 #include <vtkFloatArray.h>
 #include <vtkDiffusionTensorMathematics.h>
 
-// STD includes
-
 //------------------------------------------------------------------------------
 vtkMRMLNodeNewMacro(vtkMRMLDiffusionTensorVolumeDisplayNode);
 
@@ -63,7 +61,7 @@ vtkMRMLDiffusionTensorVolumeDisplayNode
  this->ImageCast = vtkImageCast::New();
  this->ImageCast->SetOutputScalarTypeToUnsignedChar();
 
- this->ImageMath  = vtkImageMathematics::New();
+ this->ImageMath = vtkImageMathematics::New();
  this->ImageMath->SetOperationToMultiplyByK();
  this->ImageMath->SetConstantK(255);
 
@@ -73,23 +71,20 @@ vtkMRMLDiffusionTensorVolumeDisplayNode
  sphere->Delete();
 
  this->AutoScalarRange = 1;
-
- this->dtimath = vtkSmartPointer<vtkDiffusionTensorMathematics>::New();
- this->singlepixelimage = vtkSmartPointer<vtkImageData>::New();
- this->singlepixelimage->SetExtent(0, 0, 0, 0, 0, 0);
+ this->DTIMath = vtkSmartPointer<vtkDiffusionTensorMathematics>::New();
+ this->SinglePixelImage = vtkSmartPointer<vtkImageData>::New();
+ this->SinglePixelImage->SetExtent(0, 0, 0, 0, 0, 0);
 #if (VTK_MAJOR_VERSION <= 5)
- singlepixelimage->AllocateScalars();
+ SinglePixelImage->AllocateScalars();
 #endif
- this->tensordata = vtkSmartPointer<vtkFloatArray>::New();
-
- this->tensordata->SetNumberOfComponents(9);
- this->tensordata->SetNumberOfTuples(singlepixelimage->GetNumberOfPoints());
- this->singlepixelimage->GetPointData()->SetTensors(tensordata);
-
+ this->TensorData = vtkSmartPointer<vtkFloatArray>::New();
+ this->TensorData->SetNumberOfComponents(9);
+ this->TensorData->SetNumberOfTuples(SinglePixelImage->GetNumberOfPoints());
+ this->SinglePixelImage->GetPointData()->SetTensors(TensorData);
 #if (VTK_MAJOR_VERSION <= 5)
- this->dtimath->SetInput(singlepixelimage);
+ this->DTIMath->SetInput(SinglePixelImage);
 #else
- this->dtimath->SetInputData(singlepixelimage);
+ this->DTIMath->SetInputData(SinglePixelImage);
 #endif
 }
 
@@ -99,7 +94,6 @@ vtkMRMLDiffusionTensorVolumeDisplayNode
 {
   this->DTIMathematics->Delete();
   this->DTIMathematicsAlpha->Delete();
-
   this->DiffusionTensorGlyphFilter->Delete();
   this->ShiftScale->Delete();
   this->ImageMath->Delete();
@@ -514,44 +508,45 @@ int vtkMRMLDiffusionTensorVolumeDisplayNode::GetNthScalarInvariant(int i)
 }
 
 //----------------------------------------------------------------------------
-const char *vtkMRMLDiffusionTensorVolumeDisplayNode::getPixelString(double *ijk)
+std::string vtkMRMLDiffusionTensorVolumeDisplayNode::GetPixelString(double *ijk)
 {
-    vtkImageData *imageData = this->GetVolumeNode()->GetImageData();
-    if(imageData == NULL){
-        return "No Image";
+  vtkImageData *imageData = this->GetVolumeNode()->GetImageData();
+  if(imageData == NULL)
+    {
+    return "No Image";
+    }
+  vtkIdType point_idx = imageData->FindPoint(ijk[0], ijk[1], ijk[2]);
+  if(point_idx == -1)
+    {
+    return "Out of bounds";
+    }
+  if(imageData->GetPointData() == NULL)
+    {
+    return "No Point Data";
+    }
+  if(imageData->GetPointData()->GetTensors() == NULL)
+    {
+    return "No Tensor Data";
     }
 
-    vtkIdType point_idx = imageData->FindPoint(ijk[0], ijk[1], ijk[2]);
+  double tensor[9];
+  double *a = &tensor[0];
+  imageData->GetPointData()->GetTensors()->GetTuple(point_idx, a);
+  int operation;
+  operation = this->GetScalarInvariant();
+  const float *tensorf = (float*) &tensor[0];
+  this->TensorData->SetTupleValue(0, tensorf);
+  this->TensorData->Modified();
+  this->SinglePixelImage->Modified();
+  this->DTIMath->SetOperation(operation);
+  this->DTIMath->Update();
+  vtkImageData *output = this->DTIMath->GetOutput();
+  std::string valueString = this->GetScalarInvariantAsString();
 
-    if(point_idx == -1)
-      return "Out of bounds";
+  if(output && output->GetNumberOfScalarComponents() > 0)
+    {
+    valueString += " " + DoubleToString(output->GetScalarComponentAsDouble(0, 0, 0, 0));
+    }
 
-    if(imageData->GetPointData() == NULL)
-      return "No Point Data";
-
-    if(imageData->GetPointData()->GetTensors() == NULL)
-      return "No Tensor Data";
-
-    double tensor[9];
-    double *a = &tensor[0];
-    imageData->GetPointData()->GetTensors()->GetTuple(point_idx, a);
-
-    int operation;
-
-    operation = this->GetScalarInvariant();
-
-    std::string valueString = this->GetScalarInvariantAsString();
-
-    const float *tensorf = (float*) &tensor[0];
-    this->tensordata->SetTupleValue(0, tensorf);
-    this->tensordata->Modified();
-    this->singlepixelimage->Modified();
-    this->dtimath->SetOperation(operation);
-    this->dtimath->Update();
-    vtkImageData *output = this->dtimath->GetOutput();
-
-    if(output && output->GetNumberOfScalarComponents() > 0)
-        valueString += " " + DoubleToString(output->GetScalarComponentAsDouble(0, 0, 0, 0));
-
-    return valueString.c_str();
+  return valueString;
 }
--- a/Libs/MRML/Core/vtkMRMLDiffusionTensorVolumeNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLDiffusionTensorVolumeNode.cxx
@@ -64,5 +64,5 @@ vtkMRMLDiffusionTensorVolumeDisplayNode* vtkMRMLDiffusionTensorVolumeNode
 //----------------------------------------------------------------------------
 vtkMRMLStorageNode* vtkMRMLDiffusionTensorVolumeNode::CreateDefaultStorageNode()
 {
-  return vtkMRMLNRRDStorageNode::New();
+    return vtkMRMLNRRDStorageNode::New();
 }
--- a/Libs/MRML/Core/vtkMRMLDiffusionTensorVolumeNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLDiffusionTensorVolumeNode.cxx
@@ -64,5 +64,5 @@ vtkMRMLDiffusionTensorVolumeDisplayNode* vtkMRMLDiffusionTensorVolumeNode
 //----------------------------------------------------------------------------
 vtkMRMLStorageNode* vtkMRMLDiffusionTensorVolumeNode::CreateDefaultStorageNode()
 {
-    return vtkMRMLNRRDStorageNode::New();
+  return vtkMRMLNRRDStorageNode::New();
 }
--- a/Libs/MRML/Core/vtkMRMLDiffusionTensorVolumeDisplayNode.h
+++ b/Libs/MRML/Core/vtkMRMLDiffusionTensorVolumeDisplayNode.h
@@ -186,7 +186,7 @@ class VTK_MRML_EXPORT vtkMRMLDiffusionTensorVolumeDisplayNode : public vtkMRMLGl
 
   ///
   /// Given a volume node, create a human readable string describing the contents
-  virtual const char* getPixelString(double *ijk);
+  virtual std::string GetPixelString(double *ijk);
 
 protected:
   vtkMRMLDiffusionTensorVolumeDisplayNode();
@@ -225,10 +225,10 @@ protected:
    /// Scalar display parameters
   int ScalarInvariant;
 
-  ///smart Pointer needed for calculatating PixelString
-  vtkSmartPointer<vtkDiffusionTensorMathematics> dtimath;
-  vtkSmartPointer<vtkImageData> singlepixelimage;
-  vtkSmartPointer<vtkFloatArray> tensordata;
+  /// Smart Pointer needed for calculatating PixelString
+  vtkSmartPointer<vtkDiffusionTensorMathematics> DTIMath;
+  vtkSmartPointer<vtkImageData> SinglePixelImage;
+  vtkSmartPointer<vtkFloatArray> TensorData;
 
 
 };
--- a/Libs/MRML/Core/vtkMRMLDiffusionTensorVolumeDisplayNode.h
+++ b/Libs/MRML/Core/vtkMRMLDiffusionTensorVolumeDisplayNode.h
@@ -184,6 +184,10 @@ class VTK_MRML_EXPORT vtkMRMLDiffusionTensorVolumeDisplayNode : public vtkMRMLGl
   static int GetNumberOfScalarInvariants();
   static int GetNthScalarInvariant(int i);
 
+  ///
+  /// Given a volume node, create a human readable string describing the contents
+  virtual const char* getPixelString(double *ijk);
+
 protected:
   vtkMRMLDiffusionTensorVolumeDisplayNode();
   ~vtkMRMLDiffusionTensorVolumeDisplayNode();
@@ -220,6 +224,11 @@ protected:
 
    /// Scalar display parameters
   int ScalarInvariant;
+
+  ///smart Pointer needed for calculatating PixelString
+  vtkSmartPointer<vtkDiffusionTensorMathematics> dtimath;
+  vtkSmartPointer<vtkImageData> singlepixelimage;
+  vtkSmartPointer<vtkFloatArray> tensordata;
 
 
 };
--- a/Libs/MRML/Core/vtkMRMLLabelMapVolumeDisplayNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLLabelMapVolumeDisplayNode.cxx
@@ -15,6 +15,9 @@ Version:   $Revision: 1.2 $
 #include "vtkMRMLLabelMapVolumeDisplayNode.h"
 #include "vtkMRMLProceduralColorNode.h"
 #include "vtkMRMLScene.h"
+#include "vtkMRMLColorNode.h"
+#include "vtkMRMLDisplayNode.h"
+#include "vtkMRMLVolumeNode.h"
 
 // VTK includes
 #include <vtkImageData.h>
@@ -26,6 +29,7 @@ Version:   $Revision: 1.2 $
 
 // STD includes
 #include <cassert>
+#include <sstream>
 
 //----------------------------------------------------------------------------
 vtkMRMLNodeNewMacro(vtkMRMLLabelMapVolumeDisplayNode);
@@ -46,6 +50,25 @@ vtkMRMLLabelMapVolumeDisplayNode::~vtkMRMLLabelMapVolumeDisplayNode()
 {
    this->MapToColors->Delete();
 }
+
+namespace
+{
+//----------------------------------------------------------------------------
+template <typename T> std::string NumberToString(T V)
+{
+    std::string stringValue;
+    std::stringstream strstream;
+    strstream << V;
+    strstream >> stringValue;
+    return stringValue;
+}
+
+//----------------------------------------------------------------------------
+std::string IntToString(int Value)
+{
+    return NumberToString<int>(Value);
+}
+}// end namespace
 
 //----------------------------------------------------------------------------
 void vtkMRMLLabelMapVolumeDisplayNode::SetDefaultColorMap()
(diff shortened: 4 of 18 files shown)


Libs/MRML/Core/vtkMRMLDiffusionTensorVolumeDisplayNode.cxx
```diff
@@ -68,6 +73,24 @@ ::vtkMRMLDiffusionTensorVolumeDisplayNode()
  sphere->Delete();
 
  this->AutoScalarRange = 1;
+
+ this->dtimath = vtkSmartPointer<vtkDiffusionTensorMathematics>::New();
```
pieper: Fix instance caps

Libs/MRML/Core/vtkMRMLDiffusionTensorVolumeDisplayNode.cxx
```diff
@@ -471,3 +512,46 @@ int vtkMRMLDiffusionTensorVolumeDisplayNode::GetNthScalarInvariant(int i)
     vtkMRMLDiffusionTensorVolumeDisplayNode::GetSupportedColorModes();
   return modes[i];
 }
+
+//----------------------------------------------------------------------------
+const char *vtkMRMLDiffusionTensorVolumeDisplayNode::getPixelString(double *ijk)
```
pieper: Capitalize method name

pieper:
Hi Davide - 

Overall great, but a few picky things  : )

Be sure to follow the style guide [1] which basically says we follow the coding conventions of the superclass library, in most cases VTK.

So that means two-space indentations and camelcase instance variables with an initial capital.  Also acronyms either full capitalized or non capitalized.  So this->DTIMath and this->SincelPixelImage.

Also VTK-style brackets are always used, like:

if (boolean)
  {
  statement;
  }

never:

if (boolean)
  statement;

and never:

if (boolean) {
  statement;
}

Same rules for 'for'.


I'll add some notes in line.

Easy stuff to fix!

[1] http://www.slicer.org/slicerWiki/index.php/Documentation/Nightly/Developers/Style_Guide

Libs/MRML/Core/vtkMRMLDiffusionTensorVolumeDisplayNode.cxx
```diff
@@ -471,3 +512,46 @@ int vtkMRMLDiffusionTensorVolumeDisplayNode::GetNthScalarInvariant(int i)
...
+{
+    vtkImageData *imageData = this->GetVolumeNode()->GetImageData();
+    if(imageData == NULL){
+        return "No Image";
+    }
+
+    vtkIdType point_idx = imageData->FindPoint(ijk[0], ijk[1], ijk[2]);
+
+    if(point_idx == -1)
```
pieper: add vtk-style brackets

Libs/MRML/Core/vtkMRMLDiffusionTensorVolumeNode.cxx
```diff
@@ -64,5 +64,5 @@ ::GetDiffusionTensorVolumeDisplayNode()
 //----------------------------------------------------------------------------
 vtkMRMLStorageNode* vtkMRMLDiffusionTensorVolumeNode::CreateDefaultStorageNode()
 {
-  return vtkMRMLNRRDStorageNode::New();
+    return vtkMRMLNRRDStorageNode::New();
```
pieper: two spaces, not four

Libs/MRML/Core/vtkMRMLDiffusionTensorVolumeDisplayNode.cxx
```diff
@@ -471,3 +512,46 @@ int vtkMRMLDiffusionTensorVolumeDisplayNode::GetNthScalarInvariant(int i)
...
+    this->tensordata->Modified();
+    this->singlepixelimage->Modified();
+    this->dtimath->SetOperation(operation);
+    this->dtimath->Update();
+    vtkImageData *output = this->dtimath->GetOutput();
+
+    if(output && output->GetNumberOfScalarComponents() > 0)
+        valueString += " " + DoubleToString(output->GetScalarComponentAsDouble(0, 0, 0, 0));
+
+    return valueString.c_str();
```
finetjul: The returned value is deleted before being returned

pieper:
Also not sure about 'sysref' - maybe should be 'space' like in teem [1].  Or ReferenceSpace might be the best wording.  It should be the same term in all places (i.e. in the XML and in the name of the method).  So better would be GetReferenceSpace and not GetCustomizeReferenceSpace (it's not customized it's a first class property).  Also I wonder if this should actually be at the transformable node [2] level since the same concept could apply broadly.

More pickiness, but instead of 'horizonal' and 'vertical' which only have meaning in the context of a camera (or gravity) we could refer to the axis index or just have one attribute which is a list of tokens, again like in teem [1].

[1] http://teem.sourceforge.net/nrrd/format.html#space

[2] http://slicer.org/doc/html/classvtkMRMLTransformableNode.html

Punzo:
Hi, thanks for the feedback.

@pieper : 
1) sorry for the style errors, I'll fix them;
2) I'll change the name of the attribute 'sysref' ('CoordinateSystem' in the second commit) with 'space';
3) With GetReferenceSpace and GetCustomizeReferenceSpace do you refer at the method GetCustomWorldCoordinates(double* ijk, double* SlicerWorld) in vtkMRMLScalarVolumeNode, don't you? I can cahnge to GetReferenceSpace(double* ijk, double* SlicerWorld). Shifting it up at the transformable node is also fine for me. I'll wait your confirmation for this;
4)I agree with you about the axis names, I'll change Horizontal,etc. in {space units: "<unit[0]>" "<unit[1]>" ... "<unit[dim-1]>"} (which means in principle also a generalization on the number of axes/quantities).

@finetjul may you specify a little bit more the problem? I am sorry, but I don't know DiffusionTensor datasets and vtkDiffusionTensorMathematics class either. I have just translated the python code in DataProbe in the C++ method. I checked the results with the DTI dataset in the Sample Data, which was showing always [ColorOrientation 0] both with the Python and my implementation in C++.

Libs/MRML/Core/vtkMRMLDiffusionTensorVolumeDisplayNode.cxx
finetjul:
```
const char* foo = getPixelString(...);
// here because valueString is destroyed, foo is pointing to memory that has been destroyed.
```
You might want to return a std::string instead

Punzo: I got it. you are completely right, it's very dangerous to return the pointer of the string defined in the method. Thanks!

Punzo:
@pieper  modification done.

A Note about the vtkMRMLScalarNode method GetCustomWorldCoordinates(double* ijk, double* SlicerWorld) now -> GetReferenceSpace(const double *ijk, const char *Space, double *SpaceCoordinates).

The idea behind this method is to retrieve different coordinates respect the coordinates used to for displaying the data (which are the RAS setted through the RAStoIJK Matrix during the loading of the datacube). From a point of view of SlicerAstro this method is in the MRMLAstroVolume (child of MRMLScalarVolume) because this coordinates output are charaterized by variables stored in MRMLAstroVolume (the wcs struct, i.e. projection, etc.).

This coordinates are not the display coordinates, but the relative "mathematical" coordinates to the ijk pixel.

Here the implementation in SlicerAstro:

void vtkMRMLAstroVolumeNode::GetReferenceSpace(const double *ijk, const char *Space, double *SpaceCoordinates)
{
  if (ijk != NULL && SpaceCoordinates != NULL)
    {
    if (!strcmp(Space, "WCS"))
      {
      double phi[1], imgcrd[3], theta[1];
      int stati[1];

      if((WcsStatus = wcsp2s(wcs, 1, 3, ijk, imgcrd, phi, theta, SpaceCoordinates, stati)))
        {
        vtkErrorMacro("wcsp2s ERROR "<<WcsStatus<<": "<<wcshdr_errmsg[WcsStatus]<<"\n");
        }
      }
    else if (!strcmp(Space, "IJK"))
      {
      SpaceCoordinates[0] = ijk[0];
      SpaceCoordinates[1] = ijk[1];
      SpaceCoordinates[2] = ijk[2];
      }
    }
}

Said this, @pieper if you think that can be useful to move the method to TrasnformableNode for me it is fine.


P.S.: a proper vtk6 implementation of A proper implementation of GetSpaceQuantitiesList will be:

vtkSmartPointer<vtkStringArray> vtkMRMLVolumeDisplayNode::GetSpaceQuantitiesList()
{
  vtkSmartPointer<vtkStringArray> Tokens = vtkSmartPointer<vtkStringArray>::New();
  Tokens->SetNumberOfComponents(1);
  Tokens->SetName("tokens");
  std::istringstream f(this->SpaceQuantities);
  std::string s;
  while (std::getline(f, s, ';'))
    {
    Tokens->InsertNextValue(s);
    }
  return Tokens;
}

This is much more elegant (Tokens is now a local variable and automatic memory handling).
However in python when I call it I get the error that returned class vtkSmartPointer<vtkStringArray>
is not knowed. Have I done some error or is not possible to return to the python wrapper a SmartPointer?

P.S.2: How can I link part of git code in the comments?

pieper:
Hi Davide -

Can you think of a use case for ReferenceSpace in non-volume nodes?  I.e. would it be the case that you would want to report fiducial locations in a non-RAS space?  How about LabelStatistics?  What about for Models, would they also have a use for ReferenceSpace?  I'm just trying to think if this is really an aspect of the volume itself or more of a property of the scene as a whole.  

Regarding ownership of memory allocated inside a method this can be very tricky, particularly with wrapping.  SmartPointers is a C++ concept (python pointers are naturally smart) and can be good in the use of vtk classes, but not in their API.   Best thing is not to allocate new classes in Get methods.  I think the best would be for you to keep the SpaceQuantities as a vtkStringArray instance variable (created at the time you read it) and then return a pointer to that class.  For example do this like GetRenderers here:

https://github.com/Kitware/VTK/blob/e60bfb5fe6166b334d488bd02996735eb2fc62ab/Rendering/Core/vtkRenderWindow.h#L107

That way the caller can call Register on the object if they want to ensure it stays in scope (or assign it to a SmartPointer if they want it automatically managed).  Otherwise they can just use the pointer.  This works well for python and is consistent with the rest of VTK.

Regarding your question note 4 above, I'd say aiming to use VTK classes like vtkStringArray is better than keeping a semicolon separated string.

HTH,
Steve

Punzo:
Hi Steve

1) ReferenceSpace: I think yes (see new implementation, 53c629c,  in TrasformableNode). The 
    method "knows" IJK, RAS, LAS and LPS systems assuming that the Slicer use RAS in default (i.e.  
    GetMatrixTransformToWorld() return a matrix from ijk to RAS).
     
    However doesn't work. 


2)  Tokens memory: thanks for the clarification on smartpointer and python wrapper.
      I removed the Tokens variable and I switched SpaceQuantities from cons char* to vtkStringArray*.  
     This solved the issue.

Punzo:
Hi Steve

1) ReferenceSpace: I think can be generalized (see new implementation, 53c629c, in TrasformableNode). The method "knows" IJK, RAS, LAS and LPS systems assuming that the the Parent tranformationNode is to RAS in default (i.e. GetMatrixTransformToWorld() return a matrix from ijk to RAS). I am not sure if you were meaning something like this and if this is useful (or if leaving blanked is better). However, then the method is reimplemented in ScalarVolume in which it uses the IJKtoRAS matrix stored in ScalarVolume. Other classes, such as model, can in principle have the same 
an overload of the method like the ScalarVolume one.

NOTE: Extensions have to implement the method in child classes for adding other systems of reference (such as vtkMRMLAstroVolumeNode is a child class of vtkMRMLScalarVolume and it reimplements the method for WCS calculation);

2) Tokens memory: thanks for the clarification on smartpointer and python wrapper. I removed the Tokens variable and I switched SpaceQuantities from cons char* to vtkStringArray*. This solved the issue.

I'll wait a your feedback on (1).

pieper:
It's interesting Davide - I'm still not sure if the ReferenceSpace is actually a property of the individual mrml nodes or if it should be a property of the whole scene, for example have a vtkMRMLReferenceSpaceNode as a singleton in the scene like the layout node or even as a instance variable of vtkMRMLScene.  If that were the case then it would make sense to have IJKToWorld rather than IJKToRAS in things like transforms, fiducials, volumes, etc and only use the ReferenceSpace information for formatting I/O, messages to the user, and inside calculations where the space might be meaningful.  That would certainly be a big change and would require some thought.

Unless we have a comprehensive reference space solution I wouldn't want options for LPS, LAI and other anatomical spaces inside slicer.  Right now there are many places where slicer is hard-coded for RAS and that will be hard to change without breaking backward compatibility (so it would be for a future Slicer5, which isn't on anyone's radar right now).

For the current idea of putting the reference space in the transformable node it raises the question of mixing transforms with different reference spaces in the same transform hierarchy, which seems problematics (or at least like an extra layer of transforms will be added when calculating the overall transform).   

For right now I do think that it might be easiest for SlicerAstro purposes to create a vtkMRMLReferenceSpaceNode and add it to the scene as a singleton and then it can be used in any astro related calculations.

Actually this may be a stupid question but would it ever make sense to have astro volumes with two different spaces in the same scene?

Punzo:
ok. Not stupid at all, actually I spent a lot of time for understanding the WCS. So the problem is indeed, that every datacube is characterize by this wcsstruct (an ensemble of ~50 variables)  which then are needed by the C-style function wcsp2s that return the WCS coordinates from the IJK. So, Indeed in my case the GetReferenceSpace is really a property of each volume. For example, each volume has different resolutions, sky projection and so on. When two volumes needs to be compared operation of  the sky projection are needed. This is done only for the display of the coordinates (like in DataProbe) and on the axis labels (the image is visualized in IJK coordinates). So for example if I want to overlay two volumes (background and foreground), I need to access to the wcs struct at the same time for both Volumes and I have to do different computations for getting under the mouse the right, precise, WCS coordinates at given pixel IJK.  
That is why I was thinking at it as a method of Volumes classes. Then my plan was to use the method in DataProbe (as implemented in these commits), on the slider widgets and on the scaling ruler of the 2-D views (this last two with future submission). The method should return the RAS coordinates if a ScalarVolume (or child of it) and WCS if it is a AstroVolume.

I think that a singleton can't help me to much for achieving this, can it?

Punzo:
Another example: I load an AstroVolume1 -> I generate a LabelVolume1 . I Load an AstroVolume2. 
I set AstroVolume2 as backgroundVolume and I want to display the LabelVolume1. This has the same issues of example1 and probably it needs the implementation of a AstroLabelVolume (which copy the wcs struct from  AstroVolume1 in the constructor). However, I was waiting the new release of the SegmentationNode and editor even to plan any work to make compatible LabelVolume compatible with WCS coordinates.

Punzo:
visual example: here is an example using 2-D visualization with kvis (karma). 

u12808-overlay

The optical image is the master and the HI 0th moment map (projection of all the pixel along the z-axis) is the slave. The WCS (coordinates on the axis labels) are calculated using the the wcs struct of the master. Thenfor that WCS coordinates is calculated where the slave resides on the "pixel coordinates" (so here is necessary to access at the metadata (sky position, resolution, etc.) of the slave WCS but using the same sky projection type used by the master).

So, at the moment, what I'd like to do is something similar with Slicer on the Background and Foreground Volumes with the workflow explained before.

Sorry for this long speech and the many massages, but I hope I made clear the issue.

pieper:
Okay, yes, things are starting to be a bit clearer to me.  Maybe I can try to restate things a slightly different way and you can see if I'm on the right track.  You are saying that each WCS configuration is like a different 'camera' on the data and that in general it is possible to resample the slave data into the WCS view of any given master.  To me this seems analogous to medical imaging where, for example, an MR can be obtained with various spacings and orientations but any particular slice viewer can show all datasets superimposed in a common space.  If this analogy holds, then maybe the thing to do is to let the user select the desired space and then have the AstroVolume display nodes perform the mapping to that selected space.  In Slicer this is already a button on the slice viewer called 'rotate to acquisition plane' which defines that the viewer should match the background.  Couldn't this same concept be applied to the WCS?

Punzo:
Hi Steve, I think you got it. To make even simpler let's say that you have two volumes with different center and spacings values. So the point is to give at the second volume the same center and spacings of the first one. With Astrovolume is just a little bit more complicated (we have the sky location, i.e. the center, the resolution, i.e. the spacing, and the sky projection; to this add that the systems of reference are not linear). 

Regarding  the  'rotate to acquisition plane'
dasasd
is the button (rotate to acquisition plane) the one highlighted in yellow in picture?
correct me if I am wrong but if I click there, It is only a change of the view of the camera (to fit the Background volume). It is not a changing of the spacing, the center or the orientations of the Foreground Volume. Instead for my purpose I need to change the "spacing" and the "origin" using the WCS.

I was thinking that in general if an user will use a Foreground Volume is because he/she wants to compare with a Background one. That's way I was thinking to make it all automatic: when there is a Background and a Foreground the Logic immediately change the Foreground Volume proprieties. I think with the Observers pattern in the AstroLogic this can be done (like observing the CompositeNode maybe?). 

Every suggestion is very welcome, but, maybe, I am bringing the discussion too much offtopic respect the pull request, which was related only to DataProbe and display the coordinates under the cursor.
So if I can summarize:

1) the string Space and SpaceUnits are stored in the DisplayNodes (whicth are needed by DataProbe for writing the string viewInfo.text). SET/GET method (copy and XML) also added.

2) GetPixelString has been implemented as a method class in the various Nodes.

3) What about the method GetSpaceReference (given ijk return Worldcoordinates). Maybe the best implementation was the first one ( 6ab5bd2 )? or you don't like it? In case, how  can I proceed on this? I understand that for you the RAS coordinates is something global, but for AstroVolume to go from ijk to WCS is needed to computation using the wcs struct stored in the AstroVolumeNode (so actually it has sense that is a method of a VolumeNode). Moreover, in order to not insert directly any Astro code in Slicer, for me the easiest thing that I could think is, indeed, this method GetSpaceReference::vtkMRMLScalarVolumeNode then overloaded in AstroVolume. The I can use in DataProbe and display the right coordinates.

pieper:
No, I meant the 'Rotate' button on this page:

http://www.slicer.org/slicerWiki/index.php/Documentation/4.4/SlicerApplication/MainApplicationGUI#Slice_Viewers

I have to admit I'm having trouble making concrete suggestions on the pull request -- I just don't have time to think through all the Astro requirements.   In general it seems like the right direction but I don't want to introduce changes that will be problematic for the rest of slicer.  Maybe @jcfr or @naucoin can also have a look and provide some feedback?

Punzo:
Rotate button sync the orientation of the 2-D slice view, indeed, but does not seem to resample the Foreground.

OK, I'll wait for further feedback, 

Thanks.

jcfr:
Also adding @lassoan . Andras has done a  lot of work to improve the transform support in Slicer. Including, for example, the visualization of transforms. See http://wiki.slicer.org/slicerWiki/index.php/Documentation/Nightly/Modules/Transforms

jcfr:
@Punzo Thanks again for your time, patience and perseverance. This is very much appreciated :+1:

lassoan:
vtkMRMLScalarVolumeNode is data set is always in the world coordinate system (with some optional spatial transformation between the volume's physical coordinate system and the world coordinate system), so there should be no 'ReferenceSpace' attribute in transformable nodes.
You may rename world coordinate system axes, for example: RAS -> ADt (even dynamically), but not the axes of individual scalar volumes.

If you have higher dimensional data that has to be converted into the dimensions of the world coordinate system then you can do something similar to what is done in the Sequences module. Store original high-dimensional data in a custom node ('Sequence' node), you store conversion parameters in 'SequenceBrowser' node (defines how to convert the 'Sequence' node to a 3D volume node), and the extracted 3D data is stored in 'Data' node (which would be a regular scalar volume node in the world coordinate system). The sequence logic module observes browser nodes and updates the Data nodes dynamically based on the conversion parameters stored in SequenceBrowser nodes.

Punzo:
jcfr: no thanks to you all for listening to me. 

lassoan:  I understand your perplexity. I understand that in Slicer there is a 1:1 correspondence between the ijk to RAS coordinates of ScalarVolumes, because in the system is set up a "global" system of reference for all the Nodes. From my point of view the 'problem' is indeed that in Slicer is assumed that the World Coordinates System is global and a 3-axis length. This is great, but in this logic, there is no rumor for Nodes that lives in other system of reference (such as 2 length + 1 velocity). 

It will be great to implement other 'global coordinate system of reference' as Steve also noted. However, I think that, at the moment, the RAS assumption is too hardcoded in all Slicer that it will be impossible to be realizable just as an extension. 

So I thought to the ReferenceSpace attributes allocated in the DisplayNodes. In my mind they are just like descriptors of the VolumeNodes. If this descriptors are the default, the units and coordinates display will be the 'proper' ones. For example: 

  b

Instead, if this descriptors are different by the default, we are not using the Slicer World Coordinate System anymore. Example for AstroVolume:
 
a

 (but then it is up to the Volume to provide the for example the right coordinates at a given ijk, beacuse oesn't exist anymore any relation between the "display" coordiantes and the real coordinates. For that I need a GetReferenceSpace method).

lassoan:
We can redefine the world coordinate system to be anything, for example (deg,deg,km/s), it's not a problem at all. Of course it requires some developments in the Slicer core, but it is needed for many people anyway, for example for microCT.

What does not seem to make sense for me is to create scalar volumes that don't fit into the world coordinate system. If you have volumes that are defined in some other (potentially higher-dimensional) space then those data structures should be stored in a custom node and converted to volumes that can exist in the world coordinate space by a custom module (this is done for 4D volumes in the Sequences extension).

Punzo:
Maybe I didn't stress here that: for SlicerAstro the 'display coordinates' are not the Astronomical WCS. 

The system for displaying the data (we call it image coordinates) is the same and global (and linear of course) for all the volumes, meanwhile the "coordinates display" (get the numbers under the cursor or on the axis labels) is "local" and it depends on the information derived from the header of the file (as explained before).
This means that in Slicer the volume is stored in "image coordinates" (in practice the ijk); the display and all the operations are done in this system of coordinates. This is done because the Astronomical WCS are highly non-linear and because the value of flux stored in the pixel depends from the beam (kind of observation resolution). Transforming in a non linear way the data means transform the beam and this has not a lot of sense by point of view of our data. 

Said this, for each dataset what we have is: a volume loaded in ijk coordinates (and actually lives in the ijk coordinates" from point of view of Slicer -> so the ijk->RAS matrix is identity), then every datacube has as attribute a wcs struct which is a propriety of the datacube (it is an ensemble of variables needed by wcslib to operate properly the coordinate transformations and display instead the coordinates). Therefore, each Volume is in a different system of reference in this sense. 
This means that the system of reference for displaying the data and the coordinates are different (instead in Slicer the coordinates of the display are also the real one!! This is not true for SlicerAstro).

This is in general the reason why I need a method GetReferenceSpace (returning worldcoordinates [3] from ijk[3]).

lassoan:
I understand your use case better now.

It would be very difficult to correctly implement configurable coordinate system for all the transformable nodes, so I don't think this could happen in the near future.

Do you only need these coordinates in the data probe? If yes, then I would suggest to add alternative coordinate systems support to astro volume nodes only and customize the data probe to use that information (the DataProbe is implemented in Python, so you can overwrite any methods from an extension; or allow adding custom behavior similarly to DICOM importer) or create a new AstroDataProbe widget that is optimized for your volumes.

Punzo:
I agree with you, doing a general implementation that will return custom coordinates for all the Nodes is major chage (I mean I don't feel at all that I can do it in a reasonable time and with success either).

So I am proposing this GetReferenceSpace (at beginning was called GetCustomCoordinates) as a C++ method of ScalarNode because I need to apply it in:

1) DataProbe (current pull request, Mantis http://www.na-mic.org/Bug/view.php?id=4020).
2) axis labels in DataProbeLib.py (axis need to be like this http://www.atnf.csiro.au/computing/software/karma/tool-screen-snapshots/kvis.gif).
3) to the coordinates showed in the 2-D viewer slices (up-right, on the right-side of the widget to control the slice position).
4) ruler widget.
(I'll open a Mantis ticket for each one and do pull request one by one)
 
Some of these are in C++ and can not be overloaded runtime so I need this GetReferenceSpace (6ab5bd2) that I was proposing. As I told Steve, I don't see any proper better way to get my custom coordinates, at least for the moment. Of course if you have any idea is very welcome.

jcfr:
Hi @Punzo , @lassoan , @pieper ,

Based on the different comments, I suggest the following changes:

1. DataProbe: 
 * Within Slicer core `vtkMRMLVolumeDisplayNode` API would be augmented with a virtual method named `GetPixelDescription(double ijk[3])`. Basically most of the changes in 6ab5bd2c02 related to `GetPixelString` (excluding changes to `vtkMRMLScalarVolumeNode` )
 * Next, two options:
   * if we all agree on adding the virtual method `GetReferenceSpace()` to  `vtkMRMLScalarVolumeNode` (or even `vtkMRMLVolumeNode`) that would return `RAS` by default. a other patch based on your original work could be done. What do you think @lassoan @pieper ?
    * if not, SlicerAstro could dynamically add MRML a node attribute named `ReferenceSpace` to `vtkMRMLAstroVolume`, then we would have to provide a way for extension to customize any of the DataProbe info lines. The custom implementation would make sure of the `ReferenceSpace` attribute. Or if don't want to maintain such an API, SlicerAstro could simply provide its own version of the DataProbe widget.

2. axis labels in DataProbeLib/SlicerViewAnnotation.py: this should be discussed based on change implemented in (1)

3. Slice position showed in the 2-D viewer slice viewer (up on the right of the slice slider widget): This corresponds to the `SliceOffset` (See `vtkMRMLSliceNode::Set/GetSliceOffset()`).

 * Within SlicerAstro, you would create a `vtkMRMLAstroSliceNode` where you would override the virtual method `vtkMRMLSliceNode::Set/GetSliceOffset()` to perform the expected computation specific to SlicerAstro coordinate system.

 * Within SlicerAstro you would create a custom layout associated with `vtkMRMLAstroSliceNode` and the corresponding view factory. You would then register the factory in the `setup()` method of a module provided by the SlicerAstro extension. Few hints:

    * you could derive from `qMRMLLayoutSliceViewFactory` implemented [here](https://github.com/Slicer/Slicer/blob/4d3e2d414a66624a5293ec628d90500d4d35fb65/Libs/MRML/Widgets/qMRMLLayoutManager_p.h#L173-L194) and simply override method `qMRMLLayoutSliceViewFactory::viewClassName()` to return `vtkMRMLAstroSliceNode`)
    * See [here](https://github.com/Slicer/Slicer/blob/4d3e2d414a66624a5293ec628d90500d4d35fb65/Libs/MRML/Widgets/qMRMLLayoutManager.cxx#L278-L280) for an example of layout factory registration.
    * Looking at test [qMRMLLayoutManagerWithCustomFactoryTest.cxx](https://github.com/Slicer/Slicer/blob/master/Libs/MRML/Widgets/Testing/qMRMLLayoutManagerWithCustomFactoryTest.cxx) is also useful.


4. ruler widget: Are you talking about ruler annotation (similar to fiducial or ROI) or the ruler widget used by the `DataProbeLib/SlicerViewAnnotation.py` for the Axis label discussed in point (3) ?

Punzo:
Hi jcfr, thanks for the perfect summary and hints. For point 4, I meant the ruler annotation (similar to fiducial or ROI).  In general for point 1 (and consequently for point 2) will be great if it's possible to implement such API. Of course I can also replace the DataProbe.py with an AstroDataProbe.py and check periodically if anything will break up.

Punzo:
In general point 4 can be extended to all the annotation (ruler, fiducial, roi). In fact I found all of them very useful also in my user case, but I never investigate their implementation, so a their cootdinates customisation  for the moment it's only a hypothetical idea.

pieper:
I don't mind adding a few virtual methods to allow flexibility to support
Davide's requirements.  I do like the idea that the core be made more
flexible because there are other times when non-anatomical coordinates
would be useful even in medical applications.

lassoan:
I have two major issues with adding GetPixelDescription() and GetReferenceSpace() to core MRML nodes.

1. If we just return strings then the methods are very limited: a particular formatting is hardcoded, we cannot do computations (we would need parse the string to retrieve components, etc). If we add axis and voxel description to general MRML nodes then also the methods should be general enough: we would need to define axis type, unit, etc. and define voxel type and physical unit (similarly to what is done in NRRD). Unfortunately, even if we define a nice representation of all these data in MRML nodes, it would be huge work to make Slicer actually use this information. I would be happy if we could just statically switch from RAS to LPS everywhere, but even that would be a huge undertaking.

2. Even if MRML nodes can store complete axis and voxel information, widgets and general logic classes would not be able to use them: How the ruler would compute anything in a mixed non-linear spatial/temporal space? It is not an Euclidean space, so the distance that it would compute would be meaningless. What would the slice offset mean in a mixed space? We would need pluggable logic at many places anyway to be able to properly interpret and use the axis and voxel information.

Proposed solution:

As we need pluggable logic in (or custom variants of) probe widget, slice viewer, ruler, etc. anyway, implement that first: make logics pluggable in the core (e.g., allow a module to register a custom method to create pixel description, etc. for a particular node type) or add custom variants of widgets in the extension.

Don't pollute core MRML classes with purely defined virtual methods that would be used by very few but would cause a lot of confusion. Instead, axis and voxel description could stay in custom MRML attributes or in methods defined in custom MRML nodes.

pieper:
Glad we have you in the discussion Andras!  These are exactly the sort of
discussions and considerations that we need to sort out as we think about
how to generalize things.

Punzo:
Hi, so if I understood, you would like to go for a full implementations of customization through  factories (such as the one in the Volumeslogic when allocating the type of volumes) for C++ and 'overloading' the Python file with custom ones.

In practice, for me, this should mean:

1) I implement all the proposed modification of this pull request only in my extension.

2) I do an AstroDataProbe module and i replace it to the original on (How can I make that they don't go in conflict? Is there any way for an extension to unload DataProbe.py without the use of the GUI? I mean before to start Slicer? So what I am thinking is that if a certain point I'll upload SlicerAstro as extension, when the user download it, will result in two 'DataProbe' modules which it can be a problem, can't it? A way can be to give binary using CustomSlicerGenerator instead to provide an extension?). 

3) same of (2) for DataProbeLib/SlicerViewAnnotation.py

4) using the factory in SliceNode for implementing customization of 2-D viewer layout as suggested by jcfr

5) For the coordinates of the annotations may we, in future, think some kind of factory also? (I don't know their implementation so for the moment I can't say anything).

In this way we can avoid virtual confusing methods and attribute in the core MRML, but then some factory support may be needed somewhere. In future, when everything will be set up, a kind of documentation will probably be nice for developers.  

What do you think about all of this?

Davide.

lassoan:
1 - yes

2 - You can do something like this to customize the DataProbe:
```Python
from DataProbe import DataProbeInfoWidget

def getPixelStringAstro(self,volumeNode,ijk):
  # you may call the original getPixelString implementation here for non-Astro volumes
  return "Astro!"

dataProbeInstance = slicer.modules.DataProbeInstance 
funcType = type(dataProbeInstance.infoWidget.getPixelString)
dataProbeInstance.infoWidget.getPixelString = funcType(getPixelStringAstro, dataProbeInstance.infoWidget, DataProbeInfoWidget)
```
3 - Similar to 2. We may factor out some methods that is convenient to override if you don't find one or few methods that you can easily override.

4 - yes

5 - probably it's better to create custom widgets, based on markups (see for example Markup fiducial widget)

jcfr:
Finally, if you would like to customize any VTK object across the entire application (except MRML node), you could leverage the VTK factory override.
See http://www.vtk.org/doc/nightly/html/classvtkObjectFactory.html#a8466fa22fcf9a907d8d46a459dda4f5c

We will have to investigate the details but the object factory approach combined with some re-factoring in Slicer core should allow you to provide you own implementation of for exampe a VTK widget used in the annotation module.

Punzo:
ok thank you very much for the help. I'll close the pull request.

pieper:
Thank you Jc and Andras for jumping in to this discussion and Davide for
your patience and flexibility.  I think these ideas sound like excellent
directions.

Punzo:
Hi @jcfr,
If I may I have a question regarding the MRMLSliceViewNode:
 I looked at the implementation of the methods relative to sliceOffset (both MRML andin the widgets). The fact is I don't need to change the  coordinates of the slice internal to slicer ( I use the ras coordinates as my image coordinates as discussed earlier). What I like to do is something similar that we achieved with dataprobe. I need only to personalize the display of the coordinate. Which means some changing in the Slider widgets. Is any way to overload or use factory for the widgets of MRML node? I mean it will be nice if I can do my slider widget and then overload in some way the init in the SliceController widget. I don't know how to achieve something like this, I'd like a your feedback. 

Thanks,

Davide.

Punzo:
In addition, I'd like also to change the name of the slice orietation. For the MRMLSliceViewNode I can use the factory that you described, but for the in the interface again I need to change some part of SliceCrontroller. Is there actually some way that I can provide my full implementtion of the SliceController and then register my class for the layouts and so on?

jcfr:
> some changing in the Slider widgets

An option would be to:

1. create `qAstroSliceWidget` deriving from `qMRMLSliceWidget`. Internally, this class would instantiate its custom version of slice controller.
2. Within the Astro module, you would then unregister the existing slice factory and register a custom one:

```
   qMRMLLayoutSliceViewFactory* mrmlSliceViewFactory =
    qobject_cast<qMRMLLayoutSliceViewFactory*>(
      app->layoutManager()->mrmlViewFactory("vtkMRMLSliceNode"));

  qAstroLayoutSliceViewFactory* astroSliceViewFactory =
    new qAstroLayoutSliceViewFactory(app->layoutManager());
  sliceViewFactory->setSliceLogics(mrmlSliceViewFactory->sliceLogics());

  app->layoutManager()->unregisterViewFactory(mrmlSliceViewFactory);
  app->layoutManager()->registerViewFactory(astroSliceViewFactory);
```

For reference, here is the comment where I first talked about the view factory: https://github.com/Slicer/Slicer/pull/309#issuecomment-127484847


Alternatively, if the existing `vtkMRMLSliceViewNode` could probably be customized to have a virtual method returning the offset to be displayed ? you would then create a `vtkMRMLAstroSliceNode` specializing that method. That node would be registered as an override for `vtkMRMLSliceViewNode`.

Punzo:
OK thank you very much for the info. 

Davide.

Punzo:
Hi @jcfr,
using the ViewFactory and with the addition of the constructors in the pull request https://github.com/Slicer/Slicer/pull/361#issue-109319280, I was able to modify the widgets of qMRMLSliceWidget and qMRMLSliceControllerWidget.

However I have a problem with the customization of the vtkMRMLSliceNode. I made a child class of it (vtkMRMLAstroSliceNode) and I override the method qMRMLLayoutSliceViewFactory::viewClassName() in my ViewFactory. 
My problem is that in the logic class vtkMRMLSliceLogic, I could not find a factory for the registration of my vtkMRMLAstroSliceNode. 

See vtkMRMLSliceLogic::UpdateSliceNode(). Here SliceLogic allocate only vtkMRMLSliceNodes. 

Is it possible to do a registration factory such as the one in the VolumeLogic? or is there any other way?

I can otherwise do my implementation of the SliceLogic and overriding the method UpdateSliceNode, but SliceLogics is heavily used along many classes and I am worried about breaking down things.

jcfr:
Hi Davide,
It should be possible to override the slice node in the entire application
by registering a class override at the VTK level. I will provide more
details later when I am back near a computer.
Hth
Jc

Punzo:
ok thanks!!!

Punzo:
Hi @jcfr, 
I tried to look for info regarding the vtk object override and factory, but I didn't find a lot of documentation.
May you point me to some implementation of it, as example?
I look forward to hear from you.
Thanks,

Davide.

Punzo:
@jcfr I investigated more and I just want to confirm that the issue is indeed the vtkMRMLSliceLogic::UpdateSliceNode() 
(now the layout factory support only AstroSliceNode, but only SliceNode are allocated in the scene).
I think a good option may be a registration factory to customize UpdateSliceNode(). I looked around for the vtkoverride, but I didn't get it. I'll wait your return and feeback, before to try to implement the registration factory.

PR #312, opened 2015-08-03
Move to itkmghio

hjmjohnson:
@jcfr @pieper 

This set of patches allow building and using the ITK version of MGHIO that has recent bug fixes:  

https://github.com/Slicer/itkMGHImageIO/commit/48411ea73ac25a777ab2e038431d032b85f77330

====
Previous versions of BRAINSTools required special #pragma code when built inside of  Slicer to  properly use the Slicer version of MGHIO.  That code caused conflicts with the standard ITK loading mechanisms.  The pragmas are no longer needed, so the code is simpler now.

=====
BRAINSTools also fixes problems with conversion to/from nii and nrrd files when some image orientations were present.  Those bugs have been resolved.

Hans


pieper:
+1

jcfr:
Thanks @hjmjohnson :+1: I will review and integrate in the coming days.

For reference, associated issue is http://www.na-mic.org/Bug/view.php?id=3410

jcfr:
Update: Since I am travelling from August 12 to 19, I will be integrating this topic on my return.

hjmjohnson:
@pieper @jcfr 

I hope that the new version of BRAINSTools can be incorporated soon.

Hans

pieper:
@jcfr do you still want to do it or should I?

jcfr:
@pieper This is still in my plate, I will take care of it. See http://www.slicer.org/slicerWiki/index.php/Developer_Meetings/20150901

jcfr:
Hi @pieper @hjmjohnson , 

I looked into integrating 17ad219 (Convert to using ITK version of MGHIO). 

It turns out that the Slicer version as a special _hack_ to avoid every CLI shared libraries to register all Image IO factories. See https://github.com/Slicer/Slicer/commit/03b896142b82728f0e5a6e374b72d7ce9ca8a6c1#diff-d2f21c6c290e012fc65ecfd3172a7ff1

Waiting the ITK factory mechanism is updated, I would suggest we continue to build the MGHIOFactory within Slicer. 

Thoughts ? Comments ?

Ping @thewtex @blowekamp

pieper:
I remember that issue - we definitely want to have the factory registration
properly sorted out at some point.  That said, does it matter if the
library is part of Slicer or ITK?  Wouldn't itkmghio be handled the same as
all the other ITK IOs?

jcfr:
@pieper 

> re: That said, does it matter if the library is part of Slicer or ITK? Wouldn't itkmghio be handled the same as all the other ITK IOs?

Good point. I may have overlooked the issue. I will have an other look during my coming :airplane:

hjmjohnson:
The primary reason for moving mghio was to ensure that it would be handled just like all the other IO mechanisms. 

MGHIO should no longer require the “hack” any longer.

Hans


From:  Steve Pieper
Reply-To:  Slicer/Slicer
Date:  Wednesday, September 9, 2015 at 4:51 PM
To:  Slicer/Slicer
Cc:  Hans Johnson
Subject:  Re: [Slicer] Move to itkmghio (#312)

I remember that issue - we definitely want to have the factory registration
properly sorted out at some point.  That said, does it matter if the
library is part of Slicer or ITK?  Wouldn't itkmghio be handled the same as
all the other ITK IOs?

jcfr:
Integrated in [r24539](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24539)

jcfr:
@hjmjohnson  Thanks again for your help with this

hjmjohnson:
Did you need to update BRAINSTools as well?  I thought that was a requirement in order to make ITKMGHIO work?  

Perhaps BRAINSTools was updated earlier.

jcfr:
@hjmjohnson Thanks for the note.

`ENH: Update BRAINSTools to version 4.5.0` integrated in [r24540](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24540)

jcfr:
Update 1:  Following integration of [PR#312 Move to itkmghio](https://github.com/Slicer/Slicer/pull/312), test `py_nomainwindow_test_slicer_mgh` is failing on [linux](http://slicer.cdash.org/viewTest.php?onlydelta&buildid=637261) and [windows](http://slicer.cdash.org/viewTest.php?buildid=637262&onlydelta).

Update 2: Thanks to a suggestion of @blowekamp during today's [developer hangout](http://wiki.slicer.org/slicerWiki/index.php/Developer_Meetings/20150915#Conclusion), I will have a fix pushed shortly.

For future reference, the  associated mantis issue is [3410: Depends on MGH IO module provided by ITK ](http://www.na-mic.org/Bug/view.php?id=3410)

jcfr:
Fixed in [r24546](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24546)

PR #313, opened 2015-08-04
ENH: check for data that will get deleted on scene view restore

naucoin:
Added a flag to the scene view node restore call that will cancel if
any nodes come up as being in the main Slicer scene but not in the
scene view scene about to be restored.
Added a method to add nodes to the scene view scene that are in the
main Slicer scene but not in the scene view.
Updated the GUI for restoring scenes via the tool bar and the Scene
Views module to pop up message boxes to give the user options to
discard the data, save it to the scene view, or cancel the restore.
Updated the test to include a failure case when removing nodes
is not allowed.
Unify scene view node to use SafeDownCast instead of a mix of
dynamic_cast and SafeDownCast.

Issue #3956

### DIFF
--- a/Libs/MRML/Core/vtkMRMLSceneViewNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLSceneViewNode.cxx
@@ -313,7 +313,7 @@ void vtkMRMLSceneViewNode::UpdateStoredScene()
   // prevent data read in UpdateScene
   for (n=0; n<nnodesSanpshot; n++)
     {
-    node  = dynamic_cast < vtkMRMLNode *>(this->SnapshotScene->GetNodes()->GetItemAsObject(n));
+    node  = vtkMRMLNode::SafeDownCast(this->SnapshotScene->GetNodes()->GetItemAsObject(n));
     if (node)
       {
       node->SetAddToSceneNoModify(0);
@@ -323,7 +323,7 @@ void vtkMRMLSceneViewNode::UpdateStoredScene()
   // update nodes in the snapshot
   for (n=0; n<nnodesSanpshot; n++)
     {
-    node  = dynamic_cast < vtkMRMLNode *>(this->SnapshotScene->GetNodes()->GetItemAsObject(n));
+    node  = vtkMRMLNode::SafeDownCast(this->SnapshotScene->GetNodes()->GetItemAsObject(n));
     if (node)
       {
       node->UpdateScene(this->SnapshotScene);
@@ -334,7 +334,7 @@ void vtkMRMLSceneViewNode::UpdateStoredScene()
   // update nodes in the snapshot
   for (n=0; n<nnodesSanpshot; n++)
     {
-    node  = dynamic_cast < vtkMRMLNode *>(this->Nodes->GetNodes()->GetItemAsObject(n));
+    node  = vtkMRMLNode::SafeDownCast(this->Nodes->GetNodes()->GetItemAsObject(n));
     if (node)
       {
       node->SetAddToSceneNoModify(1);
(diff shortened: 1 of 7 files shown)


jcfr:
Few warnings:

```
/home/jcfr/Projects/Slicer/Libs/MRML/Core/vtkMRMLSceneViewNode.cxx: In member function ‘virtual void vtkMRMLSceneViewNode::WriteNodeBodyXML(std::ostream&, int)’:
/home/jcfr/Projects/Slicer/Libs/MRML/Core/vtkMRMLSceneViewNode.cxx:83:15: warning: unused variable ‘sceneURL’ [-Wunused-variable]
   const char *sceneURL = this->SnapshotScene->GetURL();
               ^
/home/jcfr/Projects/Slicer/Libs/MRML/Core/vtkMRMLSceneViewNode.cxx:84:15: warning: unused variable ‘rootDir’ [-Wunused-variable]
   const char *rootDir = this->SnapshotScene->GetRootDirectory();
               ^
```

jcfr:
```
/home/jcfr/Projects/Slicer/Libs/MRML/Widgets/qMRMLSceneViewMenu.cxx: In member function ‘void qMRMLSceneViewMenuPrivate::restoreSceneView(const QString&)’:
/home/jcfr/Projects/Slicer/Libs/MRML/Widgets/qMRMLSceneViewMenu.cxx:205:18: warning: unused variable ‘continueButton’ [-Wunused-variable]
     QPushButton *continueButton = missingNodesMsgBox.addButton(QMessageBox::Discard);
                  ^
/home/jcfr/Projects/Slicer/Libs/MRML/Widgets/qMRMLSceneViewMenu.cxx:206:18: warning: unused variable ‘addButton’ [-Wunused-variable]
     QPushButton *addButton = missingNodesMsgBox.addButton(QMessageBox::Save);
                  ^
[ 43%] Built target 
```

jcfr:
```
/home/jcfr/Projects/Slicer/Modules/Loadable/SceneViews/GUI/qSlicerSceneViewsModuleWidget.cxx: In member function ‘void qSlicerSceneViewsModuleWidget::restoreSceneView(const QString&)’:
/home/jcfr/Projects/Slicer/Modules/Loadable/SceneViews/GUI/qSlicerSceneViewsModuleWidget.cxx:306:18: warning: unused variable ‘continueButton’ [-Wunused-variable]
     QPushButton *continueButton = missingNodesMsgBox.addButton(QMessageBox::Discard);
                  ^
/home/jcfr/Projects/Slicer/Modules/Loadable/SceneViews/GUI/qSlicerSceneViewsModuleWidget.cxx:307:18: warning: unused variable ‘addButton’ [-Wunused-variable]
     QPushButton *addButton = missingNodesMsgBox.addButton(QMessageBox::Save);
                  ^
```

jcfr:
Thinking more about the text in the dialog, here are few remarks:

* Icon is a question mark. Since user can loose data. I suggest to switch to warning. (`missingNodesMsgBox.setIcon(QMessageBox::warning);`)

* I suggest we improve the `ctkMessageBox` to be able to customize the text associated with the checkbox. Could you add a property named named `dontShowAgainText` to [ctkMessageBox.h](https://github.com/commontk/CTK/blob/master/Libs/Widgets/ctkMessageBox.h) ? That way the text would be set to _Don't show this message again and always save data before restoring._

* There is inconsistency between the text _Continue restoring and Discard data, Save missing data to this scene, or Cancel_ and the name of the button. Here is a possible text:

Dialog title: SceneViews

Dialog content:
```
Save data to scene view "<NameOfSceneView>" before restoring ?

If you don't save, data not already saved to file or added in an other scene views will be permanently lost.

[ ] Don't show this message again and always save data before restoring.

 [Restore without saving]  [Cancel]  [Save and Restore]
```

Current dialog:

screenshot from 2015-08-04 23 22 15

Gedit (text editor) dialog I inspired from:

screenshot from 2015-08-04 23 37 03

naucoin:
re: the message box text
Point 1: done
Point 2: I think it would make more sense to automate that by using the text on the button that has the accept role defined.
Point 3: Thanks for the feedback, I had meant to ask for specific suggestions about the text of the box first to make it as clear as possible. The default text on the buttons, as well as the order, is slightly different on my machine:
restoresceneviewwarning

naucoin:
Was able to fairly easily get at the don't show again option's text, this is what it looks like now with CTK + Slicer changes:
restoresceneviewwarningnew

jcfr:
Looking good :+1: 

That said, since clicking on "Restore scene view" does not check if added data are already present in other scene view. I would prefer the sentence in the middle to look like this:

```
If you don't save, data not already saved to file or added in an other scene views will be permanently lost.
```

Minor tweak. I suggest we keep only the question in bold.

pieper:
I had to read back through the discussion to understand the meaning of "save" in this context.  It means add the new data into the older scene view, and not save to disk, right?  Should we use "add" instead of "save" ?

Also the latest dialog seems to have lost the first sentence about "Data has been found..." and without that it wasn't clear to me what the rest of the message means.

naucoin:
@jcfr Is this what you mean for getting to a state when data is in a scene view and the current scene but not in the scene view to be restored:
start slicer, add data
make scene view 1
add data
make scene view 2
try to restore scene view 1
If the user doesn't save the data into view 1, it will most likely be lost, since restoring after that from scene view 2, the data wasn't saved to disk so it might not be saved fully in the scene view (still working out bugs in this situation, MRB files being deleted from disk after loading make it tricky). 
I changed the wording to try and emphasise that only data saved to disk is safe unless it gets added to this scene view, but I can add "to file" explicitly.

naucoin:
@pieper : yes, it means adding the data to the scene view. I'm okay with either using add or save, whatever the consensus for least confusing is.
I can add back the top level explanation.

jcfr:
Good point.

Dialog title: `SceneViews - New of modified data has been found

Dialog content:
```
Add data to scene view "Foo" before restoring ? 

If you don't add, data not already saved to disk or included in an other scene view will be permanently lost.

[ ] Don't show this message again and always <i>Add and Restore</i>

[Add and Restore]  [Restore without adding] [Cancel]
```

naucoin:
This check is only for new data, not modfied. 
I still would only like to include "saved to disk" rather than implying that if it's in a scene view it's safe.

jcfr:
> re: Is this what you mean for getting to a state when data is in a scene view and the current scene but not in the scene view to be restored:

The scenario you outlined is exactly what I did.

>re : add or save

Since creating a sceneview doesn't necessarily imply all data and state are saved/serialized to disk.  I think `Add` makes sense here.

naucoin:
Trying to fix the bold/not bold text led me to put in a bug report:
https://github.com/commontk/CTK/issues/585

pieper:
Nice catch!

naucoin:
CTK change on my topic branch for feedback:
https://github.com/naucoin/CTK/commit/151b41e8f396fb9b1ab47477ea2496c27303b3f9
Until the CTK bug is fixed, all text is bold, but I have it ready to split out into not bold informative text.
I'm still trying to figure out where my windowTitle setting is getting unset (I'm using "Data missing from Scene View"), but here's the current message box with the situation summary text:
restoresceneviewwarningnewer

jcfr:
> re: naucoin/CTK@151b41e
See https://github.com/naucoin/CTK/commit/151b41e8f396fb9b1ab47477ea2496c27303b3f9#commitcomment-12552992

> re: commontk/CTK#585
See https://github.com/commontk/CTK/issues/585#issuecomment-128170380

> re informative text
`.. be permanently lost.` -> `.. be permanently lost !`

Thanks
Jc

pieper:
It seems some platforms ignore the dialog title so better not to put
important info in there.

naucoin:
New commit on the CTK message box: https://github.com/naucoin/CTK/commit/f3294326cf91ae92eeb1e028994f9bb69e00ef11

naucoin:
CTK message box patches have been committed, updated to that point on the CTK master.

jcfr:
Integrated in [r24493](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24493)

PR #315, opened 2015-08-05
Fiber bundle label select bug

zhangfanmark:
MANTIS issue link: http://www.na-mic.org/Bug/view.php?id=4025

1) Fixing the bug from ADD operation for ‘Label to include’
2) Fixing the module so that it takes both VTK and VTP when being used as CLI

@pieper @jcfr, would you check the changes? 

Thanks,
Fan

### DIFF
--- a/Modules/CLI/FiberBundleLabelSelect/FiberBundleLabelSelect.cxx
+++ b/Modules/CLI/FiberBundleLabelSelect/FiberBundleLabelSelect.cxx
@@ -34,6 +34,10 @@
 #include <vtkXMLPolyDataWriter.h>
 #include <vtkVersion.h>
 
+// added for taking VTK and VTP as input and output
+#include <vtkPolyDataReader.h>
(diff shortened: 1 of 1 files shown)


pieper:
@zhangfanmark - Thanks for the fixes! Looks like really good with one small comment about style.  Generally we like the comments to describe the current state of the code and not refer back to old bugs or previous state of the code unless it helps explain why a particular implementation was used.  E.g. you could include a link to the bug tracker issue or you could explain why two different readers are used (if you think that's important to help future readers understand the code).  But comments like the ones marked CHANGE won't really be informative to future readers of the code since they won't know the context. Similarly we try to avoid leaving commented out code in the file - referring to the issue tracker or git history is a cleaner way to preserve the story line without cluttering the code.

zhangfanmark:
@pieper Thanks! 
Ok, I'll update the code and submit another pull request later. 

Regards,
Fan

PR #316, opened 2015-08-07
BUG: FiberBundleLabelSelect

zhangfanmark:
1) Fixing the bug from ADD operation for ‘Label to include’
2) Fixing the module so that it takes both VTK and VTP when being used
as CLI

MANTIS issue link: http://www.na-mic.org/Bug/view.php?id=4025

@pieper @jcfr Would you check this pull request please? 

Thanks,
Fan


pieper:
Tested and merged as r24489.  Thanks!

PR #318, opened 2015-08-11
BUG: fix creating new extension

pieper:
Wrong variable was being checked for directory existence.


pieper:
@jcfr @lassoan have a look at this - I tested and it works for me.

pieper:
Actually this version introduces an extra layer of directory /tmp/ext/ext when you expect /tmp/ext

PR #319, opened 2015-08-11
ENH: Adding boolean property to control behavior

adamrankin:
By default the qSlicerSimpleMarkupsWidget entered place mode when a fiducial list was chosen in the drop down. This behavior was not desirable to me so I made it controllable with a boolean set to True by default.


jcfr:

Integrated in [r24495](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24495)

PR #321, opened 2015-08-13
BUG: Fix running self tests for scripted modules generated by Extension Wizard

msmolens:
An error occurs when running the self tests for certain scripted modules through the SelfTests module GUI. For example, running the SurfaceToolbox self test results in:

    NameError: name 'SurfaceToolboxTest' is not defined

The affected modules are those that inherit from `ScriptedLoadableModule`. That includes all modules generated by Extension Wizard; see the [template](https://github.com/Slicer/Slicer/blob/master/Utilities/Templates/Modules/Scripted/TemplateKey.py).

This commit fixes how `ScriptedLoadableModule` finds and instantiates the self test.

Additionally, for consistency with the above change, this commit changes how `ScriptedLoadableModuleWidget` runs the self test when the user clicks "Reload and Test".

This error was also noted [here](https://github.com/Slicer/Slicer/pull/164#issuecomment-52069238) and [here](https://github.com/Slicer/Slicer/commit/671595bbdaddb90ad2772bcff96686466126da4e#commitcomment-7374819). @naucoin, can you check this?

### DIFF
--- a/Base/Python/slicer/ScriptedLoadableModule.py
+++ b/Base/Python/slicer/ScriptedLoadableModule.py
@@ -2,6 +2,7 @@ import os, string
 import unittest
 from __main__ import qt, ctk, slicer
 import logging
+import importlib
 
 __all__ = ['ScriptedLoadableModule', 'ScriptedLoadableModuleWidget', 'ScriptedLoadableModuleLogic', 'ScriptedLoadableModuleTest']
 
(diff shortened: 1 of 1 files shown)


naucoin:
Will check this out today or tomorrow, thanks!

naucoin:
I checked the color scalar bar test as well as the surface tool box one and one from the Reporting extension - all work with this change, thanks! 
This is ready to merge.

msmolens:
Thanks!
Integrated in [r24498](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24498)

PR #299, opened 2015-06-30
ENH: add support of DCMTK private dictionary

fedorov:
See issue [4013](http://www.na-mic.org/Bug/view.php?id=4013) and related discussion here:

http://slicer-devel.65872.n3.nabble.com/Enabling-support-for-DCMTK-private-dictionary-td4034305.html

This change installs private.dic of DCMTK in a runtime location,
and sets DCMDICTPATH launcher variable. This will allow for more robust
interpretation of DICOM private tags based on the information
contained in the DCMTK private dictionary.

PR #322, opened 2015-08-15
ENH: adding test of DCMTK private dictionary

fedorov:
See issue 4013 and #299 

Philips phantom DWI dataset is included. The test runs dcmdump command, and looks
for the description of tag (2001,1003), which should be DiffusionBFactor, and can
only be recognized when private dictionary support is enabled in DCMTK.

@jcfr this test succeeds when I run --python-script from the binary install, but not from the build tree.


fedorov:
The env variable is not set for the built Slicer, only for installed. Do you know what is wrong JC?

jcfr:
@fedorov The variable set in the launcher settings is `DCMDICTHOME`. Should it be changed to `DCMDICTPATH` ?

fedorov:
@jcfr that was wrong indeed. I committed a fix in r24507 = 4e54d1fc0ec7171cf2987bf1daac57c09a65ed92

fedorov:
I rebuilt Slicer with this change, but now for unknown to me reason the test times out. No print statements other than this

```
[fedorov@beat] ctest -VVV -R DCMTK
Test project /Users/fedorov/local/builds/Slicer4-Debug/Slicer-build
    Start 545: py_nomainwindow_DCMTKPrivateDictionaryTest
```

jcfr:
Closing this PR. It is superseded by #327 

@fedorov Let me know if I misunderstood something and if this PR should be re-opened and integrated.

fedorov:
JC, I think this still might make sense to integrate. The test is verifying the private dictionary is enabled. This can get broken if the compile flags are not set correctly, or if there is a regression in DCMTK.

jcfr:
@fedorov Very good point. I just re-open the PR.
Sorry for the noise.

jcfr:
Integrated in [r24521](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24521)

@fedorov I tweaked the commit:
* now using `slicer_add_python_test` passing `SLICER_ARGS`
* also using `print()` function instead of the statement (forward compatible with python 3)
* and finally using exception instead of `exit()` .. otherwise Slicer is not exited cleanly and there are leaks

PR #323, opened 2015-08-17
BUG: Fixed crash caused in Model display node widget

lassoan:
Crashed when updated a filter pipeline that generated a polydata that was used by the model node.
There was an infinite loop (and stack overflow) caused by automatically selecting the first scalar attribute when the polydata is regenerated by the pipeline.

__Integration of CTK fix (https://github.com/commontk/CTK/pull/589) is required before merging this!__


lassoan:
Committed to SVN rev 24502

PR #324, opened 2015-08-17
BUG: remove extra node added event

naucoin:
When switching to using a mini scene to try various ways of
loading volumes, the final node added event wasn't removed
when I switched to using AddNode to move the final set of
nodes into the main scene. This change removes the extra
node added event triggered by adding a volume.

Issue #4033


jcfr:
This has been integrated in [r24509](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24509)

PR #326, opened 2015-09-01
BUG: Default parameters allowed incorrect usage

hjmjohnson:
Slicer/Modules/Loadable/Models/qSlicerModelsModule.cxx:165:44: warning: 'this' pointer cannot be null in
      well-defined C++ code; pointer may be assumed to always convert to true [-Wundefined-bool-conversion]
      QStringList() << "vtkMRMLModelNode", this));
                                           ^~~~

The 'this' pointer was interpreted as boolean.  To ensure complete proper usage, the default
parameters were removed, thus forcing compiler errors that were easy to identify and fix.

### DIFF
--- a/Base/QTGUI/qSlicerNodeWriter.h
+++ b/Base/QTGUI/qSlicerNodeWriter.h
@@ -39,8 +39,8 @@ public:
   qSlicerNodeWriter(const QString& description,
                     const qSlicerIO::IOFileType& fileType,
                     const QStringList& nodeTags,
-                    bool useCompression = true,
-                    QObject* parent = 0);
+                    bool useCompression,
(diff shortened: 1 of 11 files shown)


hjmjohnson:
@pieper @jcfr 

The use of default parameters caused the pointer to "this" to be interpreted as boolean (always true!).

Building Slicer with clang and c++11 found this.

Hans

pieper:
Interesting - I certainly never would have found that!  Okay to merge as far as I can see.

jcfr:
Good catch. Thanks :+1: 

Integrated in [r24515](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24515)

PR #327, opened 2015-09-02
ENH: update DCMTK used by Slicer

fedorov:
The important improvements provided by the current DCMTK is the introduction of
dcmiod, dcmfg and dcmseg classes that will be helpful in creating multiframe
DICOM objects, such as segmentation and parametric map objects. This updated
version also includes the feature to embed standard and private DICOM dictionaries.
This makes it unnecessary to use DCMDICTPATH for loading dictionaries from external
file, thus that functionality is removed.


fedorov:
@jcfr per discussion on the last Slicer hangout, this is the pull request updating DCMTK version. I understand it may need to wait until other updates are integrated and tested, but as discussed, it is ready from my side whenever you can review and merge. I confirmed the embedded dictionary functionality works as expected.

jcfr:
Great. Thanks for the detailed note :+1: 
 
Should the file `SlicerBlockInstallDCMTKApps.cmake` be updated to list additional apps ?

I will integrate ASAP.

fedorov:
JC, about additional apps, there are no new ones that I can think of that we need to include. The introduced modules are adding new API, no new command line tools of interest. Let's also check with @pieper in case he wanted to add any apps. For the reference, here are the ones that are currently included:

storescu storescp dcmdump dump2dcm img2dcm dcmdjpeg dcmqrscp dcm2xml dsr2xml dsrdump echoscu

pieper:
I'm thinking that dsr2xml, dsr2html and xml2dsr would also be helpful in
the future.  They are pretty small (typically 40-100k bytes installed, at
least on mac).

What do you think @fedorov?  I'm okay with leaving them out if people think
they are just bloat, but I lean toward including them.

fedorov:
dsr2xml is already included, I agree dsr2html will be useful, and I guess I can see how xml2dsr might be useful too.

jcfr:
Thanks for the feedback.

I guess. Since these are not directly related to the new dcmiod, dcmfg and dcmseg classes. A separate PR  could be done to explicitly package them ...

fedorov:
Makes sense. Should I issue a new PR?

pieper:
Sure, if you don't mind.

fedorov:
needed to roll back, so see #329 instead

PR #328, opened 2015-09-02
ENH: Create function GetValue()

agirault:
Until now vtkMRMLDoubleArrayNode values could only be obtained
using GetXYValue(int, double*, double*), by passing the index
as a copy and the values we want to get as pointers. In python
such a method can not be wrapped if it passes pointers.

The function double GetValue(int, int, int&) allows to return
one value by giving its index, and specifying the component
number. The last parameter is a reference to indicate if the
method succeeded or not, so that we can use GetValue() within
GetXYValue(). We also apply more robust conditions on the index
and the component number.

We were interested in creating this method for vtkMRMLDoubleArrayNode
because this node can be used to transfer double vectors between
CLIs and Slicer [1]. The needs we had were :
(1) Getting the returned values in a python module running the CLIs
(2) Being able to get a specified component (not limited to X and Y)

 [1] https://github.com/Slicer/Slicer/blob/866900a/Modules/CLI/ExecutionModelTour/ExecutionModelTour.xml#L183-L199

### DIFF
--- a/Libs/MRML/Core/Testing/vtkMRMLDoubleArrayNodeTest1.cxx
+++ b/Libs/MRML/Core/Testing/vtkMRMLDoubleArrayNodeTest1.cxx
@@ -10,18 +10,237 @@
 
 =========================================================================auto=*/
 
-#include "vtkMRMLDoubleArrayNode.h"
-
-
+// MRML includes
 #include "vtkMRMLCoreTestingMacros.h"
+#include "vtkMRMLDoubleArrayNode.h"
+#include "vtkMRMLDoubleArrayStorageNode.h"
 
-int vtkMRMLDoubleArrayNodeTest1(int , char * [] )
+// VTK includes
+#include <vtkNew.h>
+
+// STD includes
+#include <vector>
+#include <sstream>
+
+//---------------------------------------------------------------------------
+bool TestSetValues();
+bool TestSetValue();
+bool TestSetXYValue();
+bool TestSetXYerrValue();
+bool TestAddValues();
+bool TestAddValue();
+bool TestAddXYValue();
+bool TestAddXYerrValue();
+
+//---------------------------------------------------------------------------
+int vtkMRMLDoubleArrayNodeTest1(int argc, char * argv[])
 {
-  vtkSmartPointer< vtkMRMLDoubleArrayNode > node1 = vtkSmartPointer< vtkMRMLDoubleArrayNode >::New();
+  vtkNew<vtkMRMLDoubleArrayNode> node1;
 
-  EXERCISE_BASIC_OBJECT_METHODS( node1 );
+  EXERCISE_BASIC_OBJECT_METHODS(node1.GetPointer());
 
-  EXERCISE_BASIC_MRML_METHODS(vtkMRMLDoubleArrayNode, node1);
+  EXERCISE_BASIC_MRML_METHODS(vtkMRMLDoubleArrayNode, node1.GetPointer());
 
-  return EXIT_SUCCESS;
+  bool res = true;
+  res = res && TestSetValues();
+  res = res && TestSetValue();
+  res = res && TestSetXYValue();
+  res = res && TestSetXYerrValue();
+  res = res && TestAddValues();
+  res = res && TestAddValue();
+  res = res && TestAddXYValue();
+  res = res && TestAddXYerrValue();
+
+  return res ? EXIT_SUCCESS : EXIT_FAILURE;
+}
+
+//---------------------------------------------------------------------------
+bool doubleArrayMatch(int line, double array1[], double array2[],
+	unsigned int sizeOfArray)
+{
+  for(unsigned int i = 0; i < sizeOfArray; i++)
+    {
+    std::ostringstream s1;
+    std::ostringstream s2;
+    s1 << array1[i];
+    s2 << array2[i];
+    if( s1.str() != s2.str() )
+      {
+      std::cerr << "Line " << line
+	      << " - Array do not match !\n"
+	      << array1[i] << " not equal to " << array2[i]
+	      << std::endl;
+      return false;
+      }
+    }
+  return true;
+}
+
+//---------------------------------------------------------------------------
+bool TestSetValues()
+{
+  vtkNew<vtkMRMLDoubleArrayNode> doubleArray;
+  doubleArray->SetSize(2);
+  double testArray0[3] = { 12.1, 425.576817, -454};
+  double testArray1[3] = { 8796325425.01, 0.0, -1};
+  double testArray2[3];
+  double testArray3[3];
+
+  bool res = true;
+  res = res && doubleArray->SetValues(0, testArray0);
+  res = res && doubleArray->SetValues(1, testArray1);
+  res = res && doubleArray->GetValues(0, testArray2);
+  res = res && doubleArray->GetValues(1, testArray3);
+  res = res && doubleArrayMatch(__LINE__, testArray0, testArray2, 3);
+  res = res && doubleArrayMatch(__LINE__, testArray1, testArray3, 3);
+
+  return res;
+}
+
+//---------------------------------------------------------------------------
+bool TestSetValue()
+{
+  vtkNew<vtkMRMLDoubleArrayNode> doubleArray;
+  doubleArray->SetSize(2);
+  double testSet[3] = { 12.1, 4255444465.576817, -454};
+  double testGet[3];
+
+  bool res = true;
+  res = res && doubleArray->SetValue(0, 0, testSet[0]);
+  res = res && doubleArray->SetValue(0, 1, testSet[1]);
+  res = res && doubleArray->SetValue(1, 1, testSet[2]);
+  int success0;
+  testGet[0] = doubleArray->GetValue(0, 0, success0);
+  int success1;
+  testGet[1] = doubleArray->GetValue(0, 1, success1);
+  int success2;
+  testGet[2] = doubleArray->GetValue(1, 1, success2);
+  res = res && doubleArrayMatch(__LINE__, testSet, testGet, 3);
+
+  return (res && success0 && success1 && success2);
+}
+
+//---------------------------------------------------------------------------
+bool TestSetXYValue()
+{
+  vtkNew<vtkMRMLDoubleArrayNode> doubleArray;
+  doubleArray->SetSize(2);
+  double testArray0[2] = { 12.1, -454};
+  double testArray1[2] = { 8796325425.01, -1};
+  double testArray2[2];
+  double testArray3[2];
+
+  bool res = true;
+  res = res && doubleArray->SetXYValue(0, testArray0[0], testArray0[1]);
+  res = res && doubleArray->SetXYValue(1, testArray1[0], testArray1[1]);
+  res = res && doubleArray->GetXYValue(0, &testArray2[0], &testArray2[1]);
+  res = res && doubleArray->GetXYValue(1, &testArray3[0], &testArray3[1]);
+  res = res && doubleArrayMatch(__LINE__, testArray0, testArray2, 2);
+  res = res && doubleArrayMatch(__LINE__, testArray1, testArray3, 2);
+
+  return res;
+}
+
+//---------------------------------------------------------------------------
+bool TestSetXYerrValue()
+{
+  vtkNew<vtkMRMLDoubleArrayNode> doubleArray;
+  doubleArray->SetSize(2);
+  double testArray0[3] = { 12.1, 425.576817, -454};
+  double testArray1[3] = { 8796325425.01, 0.0, -1};
+  double testArray2[3];
+  double testArray3[3];
+
+  bool res = true;
+  res = res && doubleArray->SetXYValue(0, testArray0[0], testArray0[1], testArray0[2]);
+  res = res && doubleArray->SetXYValue(1, testArray1[0], testArray1[1], testArray1[2]);
+  res = res && doubleArray->GetXYValue(0, &testArray2[0], &testArray2[1], &testArray2[2]);
+  res = res && doubleArray->GetXYValue(1, &testArray3[0], &testArray3[1], &testArray3[2]);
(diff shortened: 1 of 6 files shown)


jcfr:
In the second paragraph of the commit message, I suggest the following:
```
To overcome this limitation, this commit add the function `double GetValue(int, int, int&)` allowing to return ... 
```

You could start the third paragraph with
```
Rational:

We were interested in creating ... 
```

agirault:
Amended the commit message and pushed back in my branch

agirault:
Added a few commits.
Still need to :
(1) create SetValue()
(2) update SetXYValue()
(3) create tests

jcfr:
:+1: Thanks for the update.

agirault:
- We abandoned the idea of creating SetValue() and update SetXYValue() because this would require a lot more calls to vtkDoubleArray:: SetTupleValue().
- I updated the DoubleArrayStorage Writer/Reader in the last commit to be able to handle the case were no labels were given. Does this look good to you?
- I will be doing upgrading the tests this week.

jcfr:
> We abandoned the idea of creating SetValue() and update SetXYValue() because this would require a 
> lot more calls to vtkDoubleArray:: SetTupleValue().

I still think it make sense to add `bool SetValue(int index, int component, double value)`

Internally, the function could use:

```
bool SetValues(int index, double* values)
```

where `values` is an array expected to have the number of components support by the underlying `vtkDoubleArray`. 

`SetXYValue` would be updated to use `SetValues`

The doxygen of `SetValues` should explicitly mention that the function is not allocating/managing memory, the caller is responsible to pass an array of the expected size. See [vtkDataArray::SetTuple](http://www.vtk.org/doc/nightly/html/classvtkDataArray.html#a3c220749574343904d380e3b428a497f) documentation for an example.

Notes:
* Ideally `vtkIdType` should be used for index instead of `int`, but that is a different issue.
* Internally, VTK nicely wrap method of the form `SetValue()` .. but that mechanism is not exposed to class different from . See [Python_Wrapper_Enhancement](http://www.vtk.org/Wiki/VTK/Python_Wrapper_Enhancement#GetTuple.2FSetTuple_.28done_as_of_Aug_6.2C_2010.29) and [vtkWrap.c#L661-L697](https://github.com/Kitware/VTK/blob/077fcb394f32c76b4c05a2456b61d000563a19d5/Wrapping/Tools/vtkWrap.c#L661-L697)

> I updated the DoubleArrayStorage Writer/Reader in the last commit to be able to handle the case were 
> no labels were given. Does this look good to you?

Great. I posed comments. See above.

> I will be doing upgrading the tests this week.

Great. As mentioned in https://github.com/Slicer/Slicer/pull/328#commitcomment-13533350, consider using measurement file saved using the older format and the new one in the tests.

Libs/MRML/Core/vtkMRMLDoubleArrayNode.cxx
```diff
@@ -259,6 +260,37 @@ unsigned int vtkMRMLDoubleArrayNode::GetSize()
...
+//----------------------------------------------------------------------------
+double vtkMRMLDoubleArrayNode::GetValue(int index, int component, int& success)
+{
+  if (component < 0 ||
+    component >= this->Array->GetNumberOfComponents() ||
+    index < 0 ||
+    index >= this->Array->GetNumberOfTuples())
+    {
+    success = 0;
+    return -1;
```
jcfr: To be consistent with other method. Returns `0`

jcfr: Forget my previous comment. That will work well.

agirault:
Test finished :
```
Tests the main methods from vtkDoubleArrayNode, as well as the
reader/writer of vtkDoubleArrayStorageNode. Also tests the reader
given the old writter convention (which was assuming that there
always were labels in the .mcsv files) to show backward compatibility.
```

Libs/MRML/Core/Testing/vtkMRMLDoubleArrayNodeTest1.cxx
```diff
@@ -10,18 +10,413 @@
...
+    std::cerr << "Line " << __LINE__
+	      << " - Missing parameters !\n"
+	      << "Usage: " << argv[0] << " /path/to/temp"
+	      << std::endl;
+    return EXIT_FAILURE;
+    }
+
+  const char* tempDir = argv[1];
+
+  std::string fileNameOldFile = std::string(tempDir) + "/vtkMRMLDoubleArrayNodeTest1_OldFile.mcsv";
```
jcfr: Seems the `.mcsv` file are missing ?

Libs/MRML/Core/Testing/vtkMRMLDoubleArrayNodeTest1.cxx
```diff
@@ -10,18 +10,413 @@
...
+  res = res && TestAddXYerrValue();
+  res = res && TestReadFileWithLabels(fileNameWithLabels);
+  res = res && TestReadFileWithoutLabels(fileNameWithoutLabels);
+  res = res && TestReadOldFile(fileNameOldFile);
+  
+  return res ? EXIT_SUCCESS : EXIT_FAILURE;
+}
+
+//---------------------------------------------------------------------------
+bool doubleArrayMatch(double array1[], double array2[], unsigned int sizeOfArray)
```
jcfr: Consider adding a `int line` parameter

Libs/MRML/Core/Testing/vtkMRMLDoubleArrayNodeTest1.cxx
```diff
@@ -10,18 +10,413 @@
...
+{
+  for(unsigned int i = 0; i < sizeOfArray; i++)
+    {
+    std::ostringstream s1;
+    std::ostringstream s2;
+    s1 << array1[i];
+    s2 << array2[i];
+    if( s1.str() != s2.str() )
+      {
+      std::cout << "ERROR : " << array1[i] << " not equal to " << array2[i] << std::endl;
```
jcfr: that you would use here to do something like: 

```
std::cerr << "Line " << line << " - Array do not match. " << array1[i] << " not equal to " << array2[i] << std::endl
```

Consider using `std::cerr` instead of `std::cout`

Libs/MRML/Core/Testing/vtkMRMLDoubleArrayNodeTest1.cxx
```diff
@@ -10,18 +10,413 @@
...
+  double testArray1[3] = { 8796325425.01, 0.0, -1};
+  double testArray2[3];
+  double testArray3[3];
+  
+  bool res = true;
+  res = res && doubleArray->SetValues(0, testArray0);
+  res = res && doubleArray->SetValues(1, testArray1);
+  res = res && doubleArray->GetValues(0, testArray2);
+  res = res && doubleArray->GetValues(1, testArray3);
+  res = res && doubleArrayMatch(testArray0, testArray2, 3);
```
jcfr:
```
res = res && doubleArrayMatch(__LINE__, testArray0, testArray2, 3);
```

Libs/MRML/Core/Testing/vtkMRMLDoubleArrayNodeTest1.cxx
lassoan: Using vtkErrorMacro would be even better, as it also logs the line number so it's easier to find where the error actually occurred.

Libs/MRML/Core/vtkMRMLDoubleArrayNode.h
```diff
@@ -100,7 +100,16 @@ class VTK_MRML_EXPORT vtkMRMLDoubleArrayNode : public vtkMRMLStorableNode
   /// int GetXYAxisValue(int index, double* x, double* y, double*  yerr);
 
   ///
-  /// Set X and Y values at he data point specified by 'index'
+  /// Set values at the data point specified by 'index'
```
jcfr:
```
/// Set array component values at the specified 'index'
/// \sa GetValues
```

Libs/MRML/Core/vtkMRMLDoubleArrayNode.h
```diff
@@ -100,7 +100,16 @@ class VTK_MRML_EXPORT vtkMRMLDoubleArrayNode : public vtkMRMLStorableNode
...
+  int SetValues(int index, double* values);
+
+  ///
+  /// Set value of the component specified by 'component'
+  /// at the data point specified by 'index'
```
jcfr:
```
/// Set value for the specified 'component' at the given 'index'
```

Libs/MRML/Core/vtkMRMLDoubleArrayStorageNode.cxx
```diff
@@ -221,16 +228,19 @@ int vtkMRMLDoubleArrayStorageNode::WriteDataInternal(vtkMRMLNode *refNode)
...
-      vtkErrorMacro("WriteData: unable to cast input node " << refNode->GetID() << " to a known double array node");
-      return 0;
-      }
+    {
+	vtkErrorMacro("WriteData: unable to cast input node " << refNode->GetID() << " to a known double array node");
+	return 0;
+    }
+    if (doubleArrayNode->GetSize() == 0)
+    {
+	vtkErrorMacro("WriteData: " << refNode->GetID() << " is emppty, no data to write in " << fullName.c_str());
```
jcfr: emppty -> empty

Libs/MRML/Core/vtkMRMLDoubleArrayNode.h
jcfr: Even better, you are right, it makes sense to keep the notion of data point. Forget my comment. What you wrote make perfect sense.

Libs/MRML/Core/vtkMRMLDoubleArrayNode.h
jcfr: /// Set value for the specified data point 'component' at the given 'index'

Libs/MRML/Core/vtkMRMLDoubleArrayNode.h
```diff
@@ -109,7 +118,16 @@ class VTK_MRML_EXPORT vtkMRMLDoubleArrayNode : public vtkMRMLStorableNode
   int SetXYValue(int index, double x, double y, double yerr);
 
   ///
-  /// Get X and Y values at he data point specified by 'index'
+  /// Get values at the data point specified by 'index'
+  int GetValues(int index, double* values);
+
+  ///
+  /// Get value of the component specified by 'component'
+  /// at the data point specified by 'index'
```
jcfr:
```
/// Get value for the specified data point 'component' at the given 'index'
```

Libs/MRML/Core/Testing/vtkMRMLDoubleArrayNodeTest1.cxx
jcfr: In that particular case, the idea is to display the line where the function `doubleArrayMatch` is used and not where the `vtkErrorMacro` is used. That allow to quickly know in which test and at which line the comparison failed.

Libs/MRML/Core/Testing/vtkMRMLDoubleArrayNodeTest1.cxx
agirault: They do not exist at the beginning, they are only created by the storage node writer [1][2] or manually to replicate the former writer.
I could maybe create a .mcsv file for testing the backward compatibility in `TestReadOldFile` [3] instead of writing it to test the backward compatibility?

  [1] https://github.com/Slicer/Slicer/pull/328/files#diff-ecf042206b75393c4c26a631735fc225R281
  [2] https://github.com/Slicer/Slicer/pull/328/files#diff-ecf042206b75393c4c26a631735fc225R323
  [3] https://github.com/Slicer/Slicer/pull/328/files#diff-ecf042206b75393c4c26a631735fc225R361

jcfr: > I could maybe create a .mcsv file for testing the backward compatibility in TestReadOldFile

I think it makes sense to provide a version of the old file.

Libs/MRML/Core/vtkMRMLDoubleArrayNode.h
agirault: Wouldn't this be more adequate :
```
/// Set value for the specified data point 'index' at the given 'component'
```
or to be consistent with other comments:
```
/// Set value at the data point specified by 'index' at the given 'component'
```
?

jcfr: :+1: for second option.

agirault:
Corrections applied and branch squashed into three commits :
(1) d1c04cd - ENH: Improve vtkMRMLDoubleArrayNode set/get method
(2) b113db3 - BUG: Fix support for missing labels in vtkMRMLDoubleArrayStorage reader 
(3) ad1281b - ENH: Add test for vtkMRMLDoubleArrayNode and its storage node

Thank you for your reviews and your help @jcfr

jcfr:
Thanks for working on this.  Looking good. Few more things to clarify and it should good. :smile:

agirault:
Corrections applied and branch squashed into four commits :
(1) d1c04cd - ENH: Improve vtkMRMLDoubleArrayNode set/get methods
(2) 4e69c73 - ENH: Add test for vtkMRMLDoubleArrayNode set/get methods
(3) 880ced8 - BUG: Fix support for missing labels in vtkMRMLDoubleArrayStorage reader 
(4) fd48947 - ENH: Add test for vtkMRMLDoubleArrayStorageNode reader/writer

Tests succeed before and after modifications in commit fd48947.

jcfr:
First two commits integrated in [r24613](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24613) and [r24614](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24614)

Closing this PR, it is superseded by #371

PR #330, opened 2015-09-03
ENH: add extra DCMTK apps

fedorov:
* dsr2html can be useful for creating human-readable representations of DICOM
structured reports
* xml2dcm and xml2dsr are reverse operations for existing converters to XML,
which is using DCMTK-specific schema


jcfr:
Integrated in [r24519](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24519)

PR #302, opened 2015-07-10
ENH: adding extra units support 

Punzo:
As talked with Steve during the Summer Project week, It will be nice to add support for other units in Slicer. Attached there is a screenshot with a preview. Feedback will be very appreciated.

screenshot from 2015-07-10 16 19 36

tested on linux. 

see http://na-mic.org/Mantis/view.php?id=4017

PR #332, opened 2015-09-08
BUG: few fix in Unit Module and MRML

Punzo:
few minor bug fix

### DIFF
--- a/Libs/MRML/Core/vtkMRMLUnitNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLUnitNode.cxx
@@ -198,10 +198,10 @@ const char* vtkMRMLUnitNode::GetDisplayStringFromValue(double value)
 }
 
 //----------------------------------------------------------------------------
-const char* vtkMRMLUnitNode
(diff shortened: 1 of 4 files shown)


Punzo:
Hi jcfr, 

I should have done two different commits (sorry!). 
the changes in vtkMRMLUnitNode are not really BUG.
1) strstream.setf(ios::fixed,ios::floatfield; guarantee that the returned string will have a number of decimals equal to precision (instead of a number of digit equal to precision). This prevents the cut of decimals when we have big numbers.
2) The spaces is just a style. I actually inserted them in ENH: adding extra units support #302. I thought simply will be better to leave them and I could add the spaces in my extension.

jcfr:
@Punzo Thanks for clarification :) I now have enough information to move forward and integrate, it all makes sense.

Punzo:
you are welcome :). Thanks for integrating.

jcfr:
Note that most of this PR has been integrated in r24534, r24535 and r24536. 

The rest will be integrated has part of #337

PR #335, opened 2015-09-12
ENH:DataProbe:  splitting processEvent method for override

Punzo:
As discussed in ENH: units/coordinates display in DataProbe #309
I need to override part of the method processEvent in DataProbe. 
I'd like to propose to split the method eventProcess so I can override only 
what is needed.

@pieper , @jcfr  , @lassoan. may you take a fast look at it? Thank you.

### DIFF
--- a/Modules/Scripted/DataProbe/DataProbe.py
+++ b/Modules/Scripted/DataProbe/DataProbe.py
@@ -252,45 +252,8 @@ class DataProbeInfoWidget(object):
     #  1: decimal point:
     #  1: number of digits after decimal point
 
-    spacing = "%.1f" % sliceLogic.GetLowestVolumeSliceSpacing()[2]
-    if sliceNode.GetSliceSpacingMode() == slicer.vtkMRMLSliceNode.PrescribedSliceSpacingMode:
-      spacing = "(%s)" % spacing
 
-    self.viewInfo.text = \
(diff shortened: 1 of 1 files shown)


lassoan:
Looks good except GetInfo is a very bad method name. Please rename it to something that explains what the method actually does.

Punzo:
GetCursorPositionInfo?

lassoan:
Much better. Maybe UpdateCursorPositionDisplayText would be even more accurate.

pieper:
Agreed - I like UpdateCursorPositionDisplayText

Punzo:
thanks for the feedback, I changed the name,

Davide.

Modules/Scripted/DataProbe/DataProbe.py
```diff
@@ -331,6 +294,52 @@ def _roundInt(value):
...
+              spacing=spacing
+              )
+
+    def _roundInt(value):
+      try:
+        return int(round(value))
+      except ValueError:
+        return 0
+
+    hasVolume = True
```
jcfr: @Punzo  I am integrating the changes. Note that I will keep this as `hasVolume = False` to match the existing behaviour.

jcfr:
Closing. Superseded by #338

PR #336, opened 2015-09-14
ENH: Update CTK to include PythonQt cleanup fixes

msmolens:
Fixes crash on exit in py_nomainwindow_DCMTKPrivateDictTest.

    $ git shortlog 3884436e0e089360c900e786513f821281a00805..0f2e0bec3a6d1c3be5b7924485b4fab933c797fb --no-merges
    Max Smolens (1):
          ctkAbstractPythonManager: revert change to clean up PythonQt before finalizing the Python interpreter

### DIFF
--- a/SuperBuild/External_CTK.cmake
+++ b/SuperBuild/External_CTK.cmake
@@ -58,7 +58,7 @@ if(NOT DEFINED CTK_DIR AND NOT ${CMAKE_PROJECT_NAME}_USE_SYSTEM_${proj})
   ExternalProject_Add(${proj}
     ${${proj}_EP_ARGS}
     GIT_REPOSITORY "${git_protocol}://github.com/commontk/CTK.git"
-    GIT_TAG "3884436e0e089360c900e786513f821281a00805"
+    GIT_TAG "0f2e0bec3a6d1c3be5b7924485b4fab933c797fb"
     SOURCE_DIR ${CMAKE_BINARY_DIR}/${proj}
     BINARY_DIR ${proj}-build
     CMAKE_CACHE_ARGS


jcfr:
:+1: 

Since there is only commit associated with the CTK update, to give more context, could  you change the message to be something like this ?

```
ENH: Update CTK to include PythonQt cleanup fixes

Fixes crash on exit in py_nomainwindow_DCMTKPrivateDictTest.

// -------------------------------------------
commit commontk/CTK@ffc7f7a9
Author: Max Smolens <max.smolens@kitware.com>
Date:   Fri Sep 11 16:52:23 2015 -0400

    ctkAbstractPythonManager: revert change to clean up PythonQt before finalizing the Python interpreter
    
    This commit reverts commontk/CTK@77a24f2
    
    The original change worked in conjunction with fixes in PythonQt to avoid
    various crashes while destroying the ctkAbstractPythonManager, for example by
    calling Py_DECREF after the Python interpreter was finalized.
    
    Further testing shows that a better fix is indeed to finalize first to ensure
    clean destruction of PythonQt objects, and to make PythonQt's clean up step more
    robust by handling a finalized interpreter.
    
    An example of code which necessitates this change is creating an instance of
    class Foo:
    
        class Foo(object):
    
          def __init__(self):
        self.timer = qt.QTimer()
    
          def __del__(self):
        self.timer.setSingleShot(True)
    
    During Py_Finalize(), the Foo instance's __del__ method is called, which in turn
    calls a method on a Qt object. For this to succeed, PythonQt's data must still
    be in place.
// -------------------------------------------
```

jcfr:
LGTM :8ball: 

Please integrate

msmolens:
Thanks! Integrated in [24538](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24538).

PR #338, opened 2015-09-15
STYLE: Refactor DataProbe for easier customization

jcfr:
This PR is a collection of changes simplifying the code in the `processEvent()` function by introducing these functions:
* `generateViewDescription(self, xyz, ras, sliceNode, sliceLogic)`
* `generateLayerName(self, slicerLayerLogic)`
* `generateIJKPixelDescription(self, ijk, slicerLayerLogic)`
* `generateIJKPixelValueDescription(self, ijk, slicerLayerLogic)`
* `_createMagnifiedPixmap()`

It is based on the original PR #335 from @Punzo. The difference is that in this PR, instead of adding a function named `updateCursorPositionDisplayText` that would be expected to explicitly update all DataProbe widgets, it introduces `generate[...]Description()` functions returning text used internally by the `DataProbe`. That extensions customizing the text used in the `DataProbeInfoWidget` are less likely to break because of refactoring associated with the `DataProbeInfoWidget`.

Additionally, it also removes some redundant code (eb650dd), simplifies the DataProbeInfoWidget constructor removing the extra `type` parameter set to `False` (a9f7fbf) and renamed function `createSmall()` to `_createSmall()` (630c7bf)

### DIFF
--- a/Modules/Scripted/DataProbe/DataProbe.py
+++ b/Modules/Scripted/DataProbe/DataProbe.py
@@ -63,7 +63,6 @@ See <a>http://www.slicer.org</a> for details.  Module implemented by Steve Piepe
       print("No Data Probe frame - cannot create DataProbe")
       return
     self.infoWidget = DataProbeInfoWidget(parent,type='small')
-    self.infoWidget.onShowImage(False)
     parent.layout().insertWidget(0,self.infoWidget.frame)
 
   def showZoomedSlice(self, value=False):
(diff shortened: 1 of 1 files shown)


jcfr:
@pieper @lassoan  @Punzo Let me know what think.

@Punzo In particular, does it make sense to pass the `xyz` parameter to `generateViewDescription()` ? It is currently not used in the Slicer implementation. Is it useful in your case ?

Punzo:
Hi jcfr, thank you for the refactorization. Yes, I need the xyz for getting the ijk in generateViewDescription() (and from the ijk then I get the WCS).

pieper:
I didn't test, but looks reasonable to me.

jcfr:
Thanks for the review.

Integrated in [r24547](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24547) to [r24552](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24552)

PR #339, opened 2015-09-15
STYLE: Facilitate integration with python IDE

jcfr:
This PR introduces `slicer/kits.py` facilitating integration with IDE like
PyCharm when simply adding the "Slicer/Base/Python" folder as dependent
project for scripted module. In this case, the `slicer` module, while incomplete
because of unavailable kits, will still be importable.

### DIFF
--- a/Base/Python/CMakeLists.txt
+++ b/Base/Python/CMakeLists.txt
@@ -1,5 +1,6 @@
 
 set(Slicer_PYTHON_SCRIPTS
+  slicer/__init__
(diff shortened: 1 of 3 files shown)


jcfr:
Cc: @Slicer/slicer-core-team

jcfr:
Integrated in [r24553](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24553)

PR #341, opened 2015-09-15
BUG: fix volume property file extensions for saving

naucoin:
When fixing the repeated file extensions problem, the
volume property files were missed. The problem
came up when loading in a volume property file
via add data (rather than a scene file) then
trying to save it again. The node was named with
the .vp file extension and then the file name
was adjusted to add another .vp to the end. This
fix changes the the name of the node to exclude
the extension so it will save correctly.

Added testing for the AddVolumePropertyFromFile
method and added some checks to it for missing
scene and file names. Added the supported
write types as well.

Issue #3956

### DIFF
--- a/Base/QTGUI/qSlicerSaveDataDialog.cxx
+++ b/Base/QTGUI/qSlicerSaveDataDialog.cxx
@@ -113,7 +113,7 @@ QString qSlicerFileNameItemDelegate::fixupFileName(const QString& fileName, cons
       }
     }
 
-  QString stippedFileName = fixup;
+  QString strippedFileName = fixup;
   if(mrmlScene)
     {
     vtkObject * object = mrmlScene;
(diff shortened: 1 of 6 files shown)


pieper:
LGTM :+1:

naucoin:
I was poking at the save data dialogue and found a spot there that could generally prevent doubled extensions by checking if the stripped file name ends with another instance of the extension in stripKnownExtension. It would still fail if someone had already saved a file that had changed formats and had different extensions piled on each other though. 
https://github.com/Slicer/Slicer/blob/master/Base/QTGUI/qSlicerSaveDataDialog.cxx#L744

naucoin:
I added the fix to check for doubled extensions to this branch:
https://github.com/naucoin/Slicer/commit/079a417891d351cee9300ae2b0d8d628713167d8

naucoin:
Integrated:
https://github.com/Slicer/Slicer/commit/4f65338f4bb26d5ca789483e460f1e122ddd3abe
https://github.com/Slicer/Slicer/commit/c44f73b3e79f1ef10b876a963bcd472e488145f4

PR #345, opened 2015-09-18
BUG: Fix py_nomainwindow_DCMTKPrivateDictTest on Windows

msmolens:
Test py_nomainwindow_DCMTKPrivateDictTest fails on Windows because of a line
ending issue. Although the test passed on a local build after r24557 [1], the
test still fails on factory-south-win7.kitware.

This commit changes the test to use Python's universal newlines [2] when parsing
dcmdump's output.

[1] http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24557
[2] https://docs.python.org/2/glossary.html#term-universal-newlines


jcfr:
Good catch !

LGTM

msmolens:
Thanks! Integrated in [24563](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24563)

fedorov:
Thank you!

PR #348, opened 2015-09-20
ENH: adding more Suffixes in Units

Punzo:
Hi,

in order to enable proper format for units with more suffixes (see figure below), I'd like to propose the following changes:

1) adding SecondSuffix and ThirdSuffix in MRML, Logic and Qt classes of Units.

2)  Overload of GetDisplayStringFromValue

3) Overload of GetDisplayStringFromDisplayValueString

capture

@pieper, @jcfr and @finetjul may you kindly check the pull request?

Thanks ,

Davide.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLUnitNode.h
+++ b/Libs/MRML/Core/vtkMRMLUnitNode.h
@@ -112,7 +112,7 @@ public:
 
   ///
   /// Set/Get the unit suffix. For example, the suffix
-  /// for the unity metre would be "m".
+  /// for the unity meter would be "m".
   /// Default is "".
   /// \sa SetSuffix(), GetSuffix()
   vtkGetStringMacro(Suffix);
@@ -120,7 +120,7 @@ public:
 
   ///
   /// Set/Get the unit secondsuffix. For example, the secondsuffix
-  /// for the unity degree would be "\x27".
+  /// for the unity degree would be "arcminute (')".
   /// Default is "".
   /// \sa SetSecondSuffix(), GetSecondSuffix()
   vtkGetStringMacro(SecondSuffix);
@@ -128,7 +128,7 @@ public:
 
   ///
   /// Set/Get the unit thirdsuffix. For example, the thirdsuffix
-  /// for the unity degree would be "\x22".
+  /// for the unity degree would be "arcsecond (")".
   /// Default is "".
   /// \sa SetThirdSuffix(), GetThirdSuffix()
   vtkGetStringMacro(ThirdSuffix);
--- a/Libs/MRML/Core/vtkMRMLUnitNode.h
+++ b/Libs/MRML/Core/vtkMRMLUnitNode.h
@@ -58,6 +58,7 @@ public:
   /// Get node XML tag name (like Volume, Model)
   virtual const char* GetNodeTagName() {return "Unit";};
 
+  ///
   /// Reimplemented to prevent reset if unit node is a singleton.
   virtual void Reset();
 
@@ -67,19 +68,30 @@ public:
   const char* GetQuantity();
   void SetQuantity(const char* quantity);
 
+  ///
   /// Return the value multiplied by DisplayCoefficient and summed by
   /// DisplayOffset.
   /// \sa GetValueFromDisplayValue(), GetDisplayStringFromValue(),
   /// GetDisplayCoefficient(), GetDisplayOffset()
   virtual double GetDisplayValueFromValue(double value);
+
+  ///
   /// Return the value subtracted from DisplayOffset and divided by
   /// DisplayCoefficient.
   /// \sa GetDisplayValueFromValue(), GetDisplayStringFromValue()
   virtual double GetValueFromDisplayValue(double value);
+
+  ///
   /// Return the display value with prefix and suffix.
   /// \sa GetDisplayValueFromValue(), GetValueFromDisplayValue()
   const char* GetDisplayStringFromValue(double value);
 
+  ///
+  /// Return the display value with prefix and suffixes.
+  /// \sa GetDisplayValueFromValue(), GetValueFromDisplayValue()
+  const char* GetDisplayStringFromValue(double value[3]);
+
+  ///
   /// Return the display string format to use with printf/sprintf.
   /// Note that the value passed to the format must be the DisplayValue.
   /// For example: "%#6.3g mm" if the precision is 3 and the suffix mm.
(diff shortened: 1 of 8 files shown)


Libs/MRML/Core/vtkMRMLUnitNode.h
```diff
@@ -99,14 +111,30 @@ class VTK_MRML_EXPORT vtkMRMLUnitNode : public vtkMRMLNode
   vtkSetStringMacro(Prefix);
 
   ///
-  /// Set/Get the unit suffix. For example, the suffix for the unity
-  /// meter would be "m".
+  /// Set/Get the unit suffix. For example, the suffix
+  /// for the unity metre would be "m".
```
pieper: We typically like American English spelling.

Libs/MRML/Core/vtkMRMLUnitNode.h
```diff
@@ -99,14 +111,30 @@ class VTK_MRML_EXPORT vtkMRMLUnitNode : public vtkMRMLNode
...
+  /// Set/Get the unit secondsuffix. For example, the secondsuffix
+  /// for the unity degree would be "\x27".
+  /// Default is "".
+  /// \sa SetSecondSuffix(), GetSecondSuffix()
+  vtkGetStringMacro(SecondSuffix);
+  vtkSetStringMacro(SecondSuffix);
+
+  ///
+  /// Set/Get the unit thirdsuffix. For example, the thirdsuffix
+  /// for the unity degree would be "\x22".
```
pieper: Not sure I follow this comment and the one above - are both x22 and x27 'unity degree'?  What is 'unity degree' ?  Does that mean one degree as in an angle?

pieper:
Thanks for the contribution @Punzo !  I had a quick look - but someone who knows about the Units code should look through too.  @jcfr @finetjul maybe?

Libs/MRML/Core/vtkMRMLUnitNode.h
Punzo: Sorry it is just a typo, I'll correct tomorrow.

Libs/MRML/Core/vtkMRMLUnitNode.h
Punzo: X27 and x22 are primes, ',  (60' =  1°) and arcseconds, ", (60" = 1'). For us the unit degree is indeed the angle of view on the sky. I can change the  description, if you think that is not explanatory.

Punzo: https://en.wikipedia.org/wiki/Minute_and_second_of_arc

jcfr:
@Punzo Thanks for the PR.

Instead of introducing `SecondSuffix` and `ThirdSuffix` in the API, what about teaching the framework how to understand when a suffix is a [derived unit](https://en.wikipedia.org/wiki/SI_derived_unit) ? 

For example, if `mJy/beam` is passed, it would be split by `/` and `.` to get the _base_ units ... 
If characters different from _letter_, _digit_, `^`, `.` and `/` are passed, an error would be reported. 

Looking at this package is also interesting .. https://pypi.python.org/pypi/units/

Punzo:
Hi @jcfr , thanks for revising.

But maybe I did some confusion with the screenshot. The pull request is not aimed for derived units Jy/beam or km/s) and relative operation (which will be amazing, for example for astronomical units: http://amusecode.org/doc/reference/quantities_and_units.html). 
My aim for this pull request is the display for coordinates which are represented by more numbers with different Suffixes, i.e.:

                       114° 10' 3.869" 

This is a single quantity (coordinate) but has this particular representation for which the decimals are expressed with "subunits" of the degree [ degree ( ° ), arcminute, ( ' ), (60' = 1°) and arcseconds, ( " ), (60" = 1') https://en.wikipedia.org/wiki/Minute_and_second_of_arc ]. This kind of displaying the sky coordinates is very typical in astronomical visualization package.
 
The Overload of GetDisplayStringFromValue takes an array[3] (e.g.: 114, 10, 3.86887261) and return the string.

lassoan:
I'm not sure this is the best approach to solve this problem.

1. Suffix, SecondSuffix, ThirdSuffix forces copy-pasting everywhere, because we cannot iterate through them. Also, the situation just gets even worse if we need additional suffices (somebody could ask for day, hours, months, sec - do we add a FourthSuffix then?).

2. Now that we actually format multiple numbers, we have to specify a separate precision for each. For example, '3 feet 3.31 in' (=100cm). Neither '3.00 feet 3.31' in nor '3 feet 3 in' would work.

3. What if we need different separators? For example 15d 13:24:11.

We could of course come up with some custom solution, but why not just use [printf](http://www.cplusplus.com/reference/cstdio/printf/) (or some printf-like) format specifier? These printf format specifiers are used throughout VTK for most label formatting.

Punzo:
Hi lassoan, thanks for your feedback.

1) I dont' understand what do you mean with copy and paste. I mean, calling the Overload of GetDisplayStringFromValue  one can have the string formatted with all the suffixes. I agree that a FourthSuffix and so on can be a problem.

2) on my implementation, the attribute precision is used on the last number and the first two use precision zero.  So with the current implementation of the overload of GetDisplayStringFromValue one will get  114.1677413888888889  °   =   114° 10' 3.869".   
Similarly can be  implemented a  GetDisplayStringFromValue which takes 2 values instead of 1 or 3 returning the string '3 feet 3.31 in'. 

One can use also printf or stream but I think many widgets at the moment use the GetDisplayStringFromValue method with one suffix.

lassoan:
I've just realized that there are much bigger issues with all this:
* Value editing: Unit-aware widgets use the unit node, so they should all be updated to support the changes. For example, you should be able to set image origin in the Volumes module as 114deg 15' 3.21''. Not just precision but also range has to be specified separately for each component.
* Value storage: it would enormous work to allow storing values with multiple components everywhere; and in most standard file formats there is simply no way to do that (e.g., in images the origin and spacing must be defined as single-component scalars)

What you actually need is a configurable label formatting in the Data probe.
If you need editing as well then you would need to implement editor widgets that do formatting/parsing on-the-fly, so that the user can enter 114deg 15' 3.21' but you would actually store 114.264233.

lassoan:
Maybe what you could add is just a simple "DisplayHint" string attribute to the Unit node, which could be used by special widgets, such as the Data probe, which support special formatting of the value for display.

For example, setting DisplayHint="FractionsAsMinutesSeconds" could indicate to the Data probe to show value as (integerpart)suffix (fractionpart-minutes)' (fractionpart-seconds)'': 15h 12' 34'' or 15deg 12' 34''.

Punzo:
You are right. Trying to generalize everything was a bad idea. In fact, for my purpose I need this coordinates formatting just for display outputs. 

I like the idea of this DisplayHint attribute. So the workaround will be:
 
1)for this pull request I add just this DisplayHint attribute in the MRMLUnitsNode.
2)then in my extension checking the new attribute I can reformat the output string from 
GetDisplayStringFromValue as desired.

Thanks a lot as usual for the very useful feedback.

Punzo:
@pieper , @jcfr and @lassoan  I'll close this pull request and open one for the DisplayHint attribute if for is it fine for everyone.

jcfr:
> open one for the DisplayHint attribute

:+1:

PR #350, opened 2015-09-21
BUG: Fix unitialized variable that causes py_LabelStatistics test failure

msmolens:
In a release build running the py_LabelStatistics test intermittently results in
a hang or memory corruption.

This commit fixes the problem by initializing a member variable in
vtkSlicerVolumesLogic. The member variable, and therefore the regression, was
introduced in r23934 ("BUG: avoid false positives on geometry check"). The
initial value matches what's described in the member variable's comment.

The unit test is updated to check the initial value of the member variable.


jcfr:
:+1: LGTM

Good catch.

msmolens:
Integrated in [r24568](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24568)

PR #352, opened 2015-09-21
BUG: Fix libpng warnings in console when loading images with Qt 4.8.7

msmolens:
This commit eliminates the following libpng warning seen when loading certain
PNG images using libpng 1.6.x, which is included in Qt 4.8.7:

    libpng warning: iCCP: known incorrect sRGB profile

To fix the images I ran them through ImageMagick, which removes the incorrect
profiles, as suggested in [1]:

    convert image.png image.png

libpng 1.6.17 describes the warning in png.c line 2308:

    /* These profiles are known to have bad data that may cause
     * problems if they are used, therefore attempt to
     * discourage their use, skip the 'have_md5' warning below,
     * which is made irrelevant by this error.
     */

[1] http://stackoverflow.com/questions/22745076/libpng-warning-iccp-known-incorrect-srgb-profile


jcfr:
:+1:  LGTM

msmolens:
Integrated in [24571](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24571)

PR #353, opened 2015-09-22
ENH:adding DisplayHint attribute in MRML, Logic and Qt of Units

Punzo:
Implementation of DisplayHint attribute in MRMLUnitNode classe as discussed in 
ENH: adding more Suffixes in Units #348. Plus added also in widgets and logic for user input.

@pieper @jcfr @lassoan 

Can you check it, please?

Thanks,

Davide.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLUnitNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLUnitNode.cxx
@@ -39,6 +39,7 @@ vtkMRMLUnitNode::vtkMRMLUnitNode()
 
   this->Prefix = 0;
   this->Suffix = 0;
+  this->DisplayHint = 0;
   this->Precision = 3;
   this->MinimumValue = VTK_DOUBLE_MIN;
   this->MaximumValue = VTK_DOUBLE_MAX;
@@ -49,6 +50,7 @@ vtkMRMLUnitNode::vtkMRMLUnitNode()
   this->SetQuantity("");
   this->SetPrefix("");
   this->SetSuffix("");
+  this->SetDisplayHint("");
 }
 
 //----------------------------------------------------------------------------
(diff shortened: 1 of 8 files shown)


Punzo:
screenshot from 2015-09-22 18 55 17

then it is apt to my extension to recognize the attribute and in case change the formatting of the units output string.

jcfr:
@Punzo Thanks for updating the PR.

For this use case, wouldn't be sufficient to set a not  node attribute using `SetAttribute` ?  See http://slicer.org/doc/html/classvtkMRMLNode.html#a8b989c19474d97618f97e47c2439b6ad

Punzo:
In the Logic of AstroVolume now I set DisplayHint ="FractionsAsArcMinutesArcSeconds"  with the SetAttribute as soon an AstroVolume is loaded. I tought, it can be useful if the user can chose different formatting inserting for example FractionsAsArcMinute or  FractionsAsArcMinuteArcSeconds.

However rethinking at it will be better if I do dedicated interface in my AstroDataProbe module.
I'll remove the logic and widgets modification. Sorry for the mess.

Punzo:
For MRML still I need Get and Set Macro, and copying the value between nodes and in XML files

jcfr:
> it will be better if I do dedicated interface in my AstroDataProbe module

Agreed. Moving forward, when we will face more use case we could still rethink and generalized the unit framework.

> Sorry for the mess.

No problem @Punzo :smile:  PRs, mailing list and issue tracker are the perfect ground to discuss and present prototype. 

> For MRML still I need Get and Set Macro, and copying the value between nodes and in XML files

GetAttribute / SetAttribute should work for this.

Attributes are also serialized to XML and copied.
See https://github.com/Slicer/Slicer/blob/e4ebd3b2cb90a228871474bfa679cba453fb841d/Libs/MRML/Core/vtkMRMLNode.cxx#L455-L470

Punzo:
@jcfr  I just looked better at SetAttribute. Yes It will be sufficient. Thank you very much for the info. I'll close again the pull request (:

PR #354, opened 2015-09-24
ENH: Show build information in Windows shortcut names for nightly builds

msmolens:
Previously, multiple installations of nightly builds would result in
indistinguishable shortcuts named "Slicer" on the Windows 8 start screen/Windows
10 start menu. This commit adds the build information to the shortcut name for
nightly builds. Released builds will still show "Slicer".

An example of a nightly build shortcut name is now:

```
Slicer 4.4.0-2015-09-24
```

Fixes #2702
Fixes #2829


fedorov:
so +1!

pieper:
Yes!  +1

CMake/SlicerCPack.cmake
```diff
@@ -210,7 +210,11 @@ set(CPACK_NSIS_MODIFY_PATH OFF)
 
 set(APPLICATION_NAME "${Slicer_MAIN_PROJECT_APPLICATION_NAME}")
 set(EXECUTABLE_NAME "${Slicer_MAIN_PROJECT_APPLICATION_NAME}")
-set(CPACK_PACKAGE_EXECUTABLES "..\\\\${EXECUTABLE_NAME}" "${APPLICATION_NAME}")
+set(PACKAGE_APPLICATION_NAME "${APPLICATION_NAME}")
+if(Slicer_VERSION_TWEAK GREATER 0)
```
jcfr: Consider the definition of release vs non-release defined [here](https://github.com/Slicer/Slicer/blob/6297a07267540e972d42cddfdefa7f2b7dc8d1e7/CMakeLists.txt#L473-L475), it seems we would like the full version to be associated with the application name in that situation:
```
if(NOT DEFINED Slicer_VERSION_TWEAK)
```

jcfr:
Looking again at the issue, what I mentioned in my previous comment is not correct. Looking at the [entire](https://github.com/Slicer/Slicer/blob/6297a07267540e972d42cddfdefa7f2b7dc8d1e7/CMakeLists.txt#L473-L497) snippet setting version variables, `Slicer_VERSION_TWEAK` is always set. 

For non-release, it is set to the build date, and for release it is set to an integer greater or equal to 0.

This means that if we want the following shortcut:

* non-release: `Slicer 4.4.0-2015-09-24`

* release: `Slicer`  (what we have currently)

We would have to introduce a way to clearly know when we are building a release.

Similarly what is one in CMake, we could introduce a variable named `Slicer_VERSION_IS_RELEASE`

I would suggest to look at what is done in CMake proper. See [here](https://github.com/Kitware/CMake/blob/master/Source/CMakeVersionCompute.cmake#L1)

lassoan:
Nightly release is a release, too.Also, ... _IS_RELEASE can be also interpreted as not Debug.

It would be more clear to introduce a Slicer_RELEASE_TYPE variable which could be set to STABLE, NIGHTLY, etc.

It would be also more clear to not use the ...TWEAK variable for different things (actual tweak version or build date) but introduce a variable with a descriptive name: Slicer_RELEASE_DATE.

Then, wherever the shortcut name is constructed there could be a nice and clear logic that would construct the name based on the Slicer_RELEASE_TYPE, using Slicer_RELEASE_DATE, Slicer_VERSION_..., etc.

pieper:
I would like to see the stable releases binaries also get tagged with the
version.

For example, in my /Applications folder I have manually renamed Slicer.app
to be Slicer-4.4.0.app and I also have a Slicer-4.3.1.app.  It would be
much cleaner and less confusing if the stable releases were distributed
with that naming convention.

On windows the installer creates versioned menu entries, but on mac/linux
we currently don't have an easy way for users to know for sure which
version of Slicer they are running.

-Steve

fedorov:
I am struggling the same way as Steve. I also often want to have multiple versions of Slicer, and currently it is tricky to manually keep track of the versions in the Mac Applications folder.

jcfr:
Great to see the discussion moving in that direction :+1: 

To summarize, we will plan on doing three changes:

1. add a variable named `Slicer_VERSION_IS_RELEASE_PACKAGE` (or `Slicer_VERSION_IS_STABLE`) (would prefer to keep a boolean option and avoid introducing a free form variable like `Slicer_RELEASE_TYPE`). Since there is already the variable `Slicer_BUILD_DATE` set [here](https://github.com/Slicer/Slicer/blob/6297a07267540e972d42cddfdefa7f2b7dc8d1e7/CMakeLists.txt#L479), I don't think `Slicer_RELEASE_DATE` is needed.
2. ensure windows shortcut is either `Slicer X.Y.Z` or `Slicer X.Y.Z-YYYY-MM-DD `
3. ensure MacOSX `.app` folder is either `Slicer-X.Y.Z.app` (for release) or `Slicer-X.Y.Z-YYYY-MM-DD.app` (for non-release). This now possible thanks to 9b261143 and 8046783c


It would also be good to review and update this wiki page: http://www.slicer.org/slicerWiki/index.php/Documentation/Nightly/Developers/Versioning#Slicer_Package_naming_scheme

lassoan:
Slicer_VERSION_IS_RELEASE_PACKAGE  is not self-explaining. What if it's FALSE? It's a debug package? Or nightly or experimental?

Encoding release type using boolean flags is not very good because they are redundant/contradicting when there are more than two types (it would be very limiting to support only two release types). You can define enums in CMake GUI and also easily verify in CMake if an invalid value is defined.

Build and release date are very different! Stable version is rebuilt many times on different dates.

jcfr:
> Slicer_VERSION_IS_RELEASE_PACKAGE is not self-explaining. What if it's FALSE? It's a debug 
> package? Or nightly or experimental?

As of today:

* defining the variable `Slicer_VERSION_TWEAK` in the `CMakeLists.txt` allows to indicate that the version of the source code will generate a release/stable package of the application. This only translates into having a different naming convention for install folder, shortcut name, ... 

* it is assumed the individual in charge of releasing the application will:
  * commit the modified `CMakeLists.txt` (with few other updates) specifying `ENH: Slicer vX.Y`
  * do a `Release` build by:
    * settings `CMAKE_BUILD_TYPE` to `Release` when using single-configuration build tool
    * or by explicitly selecting `Release` when using multi-configuration build tool like Visual Studio

The existing release process is documented [here](http://wiki.slicer.org/slicerWiki/index.php/Documentation/Nightly/Developers/ReleaseProcess#Release)

It means the fact we are doing a `Debug` or `Release` build can *not*  be _fully_ captured by the a version of the source tree. (single-config IDE vs multi-config IDE)

Regarding `Experimental` vs `Nightly`, this can not be captured within the source tree alone. Currently, the release package are manually created by running the release script manually created and are currently logged as `Experimental` build on CDash. The dashboard scripts do *not* have a knowledge of the version of the source they are building.

That is why I think adding the variable `Slicer_VERSION_IS_RELEASE` is an improvement , when set to `TRUE` by modifying the top level `CMakeLists.txt` and committing the change using the message `ENH: Slicer v4.Y`, it will configure the application name, install folder, shortcut name to have a more concise name. As a result, the built/generated package will be name in a way that indicates the associated version of the source was tagged as release. It is more explicit that setting `Slicer_VERSION_TWEAK` and will @msmolens to correctly set the windows shortcut.  

Ideally, the fact of building a release vs a non-release should solely be defined by the associated commit message like `ENH: Slicer v4.5` (and/or the fact it is associated with a git tag like `v4.5`). We shouldn't have to update the CMakeLists.txt to indicate this. 


> Encoding release type using boolean flags is not very good because they are redundant/contradicting 
> when there are more than two types (it would be very limiting to support only two release types). 

The role of that boolean variable is to represent a boolean state:
1.  the package is generated from a source tree considered stable / released
2.  the package is generated from a source tree considered in-development

> You can define enums in CMake GUI and also easily verify in CMake if an invalid value is defined.

Agreed. If needed, CMake enum are the way to go. My original choice wasn't guided by CMake limitation per se.  

> Build and release date are very different! Stable version is rebuilt many times on different dates.

Couldn't agree more.

Currently `Slicer_BUILD_DATE` is set by extracting the date of the commit associated with the build. It should probably be renamed to `Slicer_VERSION_DATE` to better represent this.


### How to deal with different type of release ? 

Now we have a way to know if a source tree is considered stable or not (either by using the variable `Slicer_VERSION_TWEAK` / `Slicer_VERSION_IS_RELEASE` or by extracting the git tag / commit message info), we still need a way to describe the type of releases. 

Currently:
  * the infrastructure supports one type of release package described by the option associated with the release scripts hosted on https://github.com/Slicer/DashboardScripts
  * The website http://download.slicer.org works under the assumption there only type of release.

Moving forward, we could have these type of releases:

 `Slicer-4.4.0-py27-win-amd64.exe`
 `Slicer-4.4.0-py3k-win-amd64.exe`
 `Slicer-4.4.0-pyconda27-win-amd64.exe`
 `Slicer-4.4.0-pyconda3k-win-amd64.exe`

To achieve this, few things would be needed:

1) We need to have a way to indicate to our user that the generate package has been created using given set of configure options. This could be done by having a variable like `Slicer_VERSION_QUALIFIER` or `Slicer_VERSION_CONTEXT` set based on the configure option.

2) Update http://download.slicer.org to support different type of release


Keeping in mind the provenance idea summarized [here](https://github.com/QIICR/ProjectIssuesAndWiki/wiki/Software-provenance-for-SR#for-3d-slicer-based-data-provenance-we-should-have) is also good.

lassoan:
I agree that Slicer_VERSION_IS_RELEASE would be a great improvement. I would recommend two further improvements to that:

1. Use a string enum because we already have NIGHTLY, RELEASE_CANDIDATE, and RELEASE official factory builds and it would be useful to distinguish private EXPERIMENTAL builds that I build on my computer. Adding several boolean flags is bad, for example we've almost forget about the Slicer_VERSION_RC flag. We could replace all these binary flags with one enum.

2. Slicer_RELEASE_TYPE (or Slicer_VERSION_IS_RELEASE_PACKAGE) should not be hardcoded into CMakeLists.txt but a configuration option (initialized to EXPERIMENTAL by default). On factory machines it would be set to NIGHTLY or RELEASE.

lassoan:
Py27, py3k, pyconda27, etc. variations... This terrifies me. As an extension developer, it's enough to worry about supporting 6 versions of Slicer (3 platforms - win/mac/linux and two branches - stable/nightly). I would not want to support several different official release types as well. Can we do with Python that we do with all other libraries (i.e., we decide to switch from VTK5 to VTK6 and we do it at once; we don't add a new release type, not even temporarily)?

jcfr:
>I agree that Slicer_VERSION_IS_RELEASE would be a great improvement.

:+1: 

> for example we've almost forget about the Slicer_VERSION_RC flag.

Since we are not doing release candidate, we could remove the corresponding code.

> it would be useful to distinguish private EXPERIMENTAL builds that I build on my computer. 

Since enum are basically list of predefined values, that would not allow us to distinguish different type of packages.

What about introducing an option named `Slicer_VERSION_QUALIFIER` that could be set to any string.

 `Slicer_VERSION_QUALIFIER` would then be used to set the `Slicer_VERSION_FULL` (see [here](https://github.com/Slicer/Slicer/blob/75174d59b51f15dc81f856939f1dbd63f50afd89/CMakeLists.txt#L490-L499))

For custom/private release, this variable could be set with the content of `BUILD_OPTIONS_STRING` defined [here](https://github.com/Slicer/Slicer/blob/75174d59b51f15dc81f856939f1dbd63f50afd89/CMake/SlicerDashboardScript.TEMPLATE.cmake#L111) in the dashboard script.


>  Slicer_RELEASE_TYPE (or Slicer_VERSION_IS_RELEASE_PACKAGE) should not be hardcoded 
>  into CMakeLists.txt 

If someone grab at the source corresponding to a release, it should be possible to create the corresponding release package without having to explicitly pass the option `-DSlicer_RELEASE_TYPE:STRING=RELEASE`. that is why, we currently modify the `CMakeLists.txt` when doing the commit `ENH: Slicer vX.Y`.

To avoid hardcoding anything in the CMakeLists.txt, I am thinking about two options:
* Extract the commit message or git tag when source are within a VCS
* Add a file named `RELEASE` at the root when we will create Source distribution archive (.tar.gz, .zip)



> Py27, py3k, pyconda27, etc. variations... This terrifies me. 

Agreed. I mentioned it because the concept of _different type of release_ was not clear to me.

lassoan:
> If someone grab at the source corresponding to a release, it should be possible to create the corresponding release package without having to explicitly pass the option -DSlicer_RELEASE_TYPE:STRING=RELEASE

There is a high chance a "release" that a developer builds from the source distribution archive on his own computer is not be compatible with the official release build by the factory (due to different compiler version, qt version, build options, cmake version, etc). Therefore, by default I would not set release type = STABLE for packages that are build from the stable version's source on a random computer. Note that this also simplifies things, as this means that EXPERIMENTAL type can be set in the source code by default. STABLE and NIGHTLY are basically reserved for factory machines.

Getting revision information from source control would be nice (MAJOR/MINOR could come from branch name; SVN revision would work great as TWEAK; I'm not sure how to determine PATCH).

jcfr:
> his own computer is not be compatible with the official release build by the factory

Agreed.  Having a working *release* go beyond the version of the source code. 

You convinced me. Hardcoding that information in the CMakeLists.txt does not make sense.

Declaring a package as a "release" should then be explicitly and intentionally  set when configuring the sources.

Considering that a package generated doing Nightly/Experimental/Continuous build (as currently defined [here](https://github.com/Slicer/Slicer/blob/75174d59b51f15dc81f856939f1dbd63f50afd89/Extensions/CMake/SlicerExtensionsDashboardDriverScript.cmake#L193-L209)) is not related to the stability of the package, i still think it makes sense to have a boolean option: Stable vs non-stable. 

What about having an option named `Slicer_VERSION_STABLE:BOOL` set to `OFF` by default ? 


To summarize:

* Introduction of `Slicer_VERSION_STABLE:BOOL` set to `OFF` by default. When set it would generate package having shorter name.
* Introducing of  `Slicer_VERSION_QUALIFIER:STRING`: useful to identify custom package generated using different options.

lassoan:
I still feel that "stability" status is often more complex than a boolean yes/no and using a string enum would be more future-proof without making things more complex, but otherwise what you described sounds good and much better than what we have now.

jcfr:
Instead of `Slicer_VERSION_STABLE:BOOL`, an idea would to use `Slicer_VERSION_SIGNED:BOOL` anticipating the application will be signed.

jcfr:
For reference, see http://www.na-mic.org/Bug/view.php?id=3275

jcfr:
Closing. This PR is superseded by https://github.com/Slicer/Slicer/pull/806

PR #356, opened 2015-09-27
ENH: Built-in module categories are listed separately in modules list

cpinter:
Modules that are loaded from the Slicer build/install tree are marked as built-in, modules from different folders are not built-in. If a non-built-in module adds a category, then it is shown above the built-in categories, divided by a separator.

### DIFF
--- a/Base/Logic/vtkSlicerApplicationLogic.cxx
+++ b/Base/Logic/vtkSlicerApplicationLogic.cxx
@@ -1728,6 +1728,34 @@ bool vtkSlicerApplicationLogic::IsPluginInstalled(const std::string& filePath,
   return true;
 }
 
+//----------------------------------------------------------------------------
+bool vtkSlicerApplicationLogic::IsPluginBuiltIn(const std::string& filePath,
+                                                  const std::string& applicationHomeDir)
+{
+  if (filePath.empty())
+    {
+    vtkGenericWarningMacro( << "filePath is an empty string !");
+    return false;
+    }
+  if (applicationHomeDir.empty())
+    {
+    vtkGenericWarningMacro( << "applicationHomeDir is an empty string !");
+    return false;
+    }
+
+  std::string path = itksys::SystemTools::GetFilenamePath(filePath);
+  std::string canonicalPath = itksys::SystemTools::GetRealPath(path.c_str());
+
+  // On MacOSX extensions are installed in the "<Slicer_EXTENSIONS_DIRBASENAME>-<slicerRevision>"
+  // folder being a sub directory of the application dir, an extra test is required to make sure the
+  // tested filePath doesn't belong to that "<Slicer_EXTENSIONS_DIRBASENAME>-<slicerRevision>" folder.
+  // Since package name can be rename from "Slicer.app" to "Something.app", let's compare
+  // using ".app/Contents/" instead of "Slicer_BUNDLE_LOCATION" which is "Slicer.app/Contents/"
+  bool macExtension = (canonicalPath.find(".app/Contents/" Slicer_EXTENSIONS_DIRBASENAME "-") != std::string::npos);
+
+  return itksys::SystemTools::StringStartsWith(canonicalPath.c_str(), applicationHomeDir.c_str()) && !macExtension;
+}
+
 namespace
 {
 //----------------------------------------------------------------------------
--- a/Base/QTGUI/qSlicerModulesMenu.cxx
+++ b/Base/QTGUI/qSlicerModulesMenu.cxx
@@ -39,8 +39,8 @@ public:
   void init();
   void addDefaultCategories();
 
-  void addModuleAction(QMenu* menu, QAction* moduleAction, bool useIndex = true);
-  QMenu* menu(QMenu* parentMenu, QStringList subCategories);
+  void addModuleAction(QMenu* menu, QAction* moduleAction, bool useIndex = true, bool builtIn = true);
+  QMenu* menu(QMenu* parentMenu, QStringList subCategories, bool builtIn = true);
 
   QAction* action(const QVariant& actionData, const QMenu* parentMenu)const;
   QAction* action(const QString& text, const QMenu* parentMenu)const;
@@ -90,8 +90,10 @@ void qSlicerModulesMenuPrivate::addDefaultCategories()
   q->addSeparator();
   // between the 2 separators are the top level modules (with no category)
   q->addSeparator();
-  // after the separator are the predefined categories followed by the custom
-  // ones.
+  // after the top level modules are the non-built-in categories (from
+  // extensions or any user defined external folder)
+  q->addSeparator();
+  // after the separator are the predefined categories
 }
 
 //---------------------------------------------------------------------------
@@ -137,7 +139,7 @@ QAction* qSlicerModulesMenuPrivate::action(const QString& text, const QMenu* par
 }
 
 //---------------------------------------------------------------------------
-void qSlicerModulesMenuPrivate::addModuleAction(QMenu* menu, QAction* moduleAction, bool useIndex)
+void qSlicerModulesMenuPrivate::addModuleAction(QMenu* menu, QAction* moduleAction, bool useIndex, bool builtIn)
 {
   Q_Q(qSlicerModulesMenu);
   QList<QAction*> actions = menu->actions();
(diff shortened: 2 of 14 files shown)


cpinter:
20150926_builtinmodulecategories

Base/QTGUI/qSlicerModulesMenu.cxx
```diff
@@ -39,8 +39,8 @@ class qSlicerModulesMenuPrivate
   void init();
   void addDefaultCategories();
 
-  void addModuleAction(QMenu* menu, QAction* moduleAction, bool useIndex = true);
-  QMenu* menu(QMenu* parentMenu, QStringList subCategories);
+  void addModuleAction(QMenu* menu, QAction* moduleAction, bool builtIn = true, bool useIndex = true);
```
lassoan: The behavior is different for the same inputs with the old and new implementation (if only one optional parameter is defined then it's the useIndex in earlier versions but builtIn in the new one). If we keep the useIndex argument as the third argument then there is no backward compatibility issue.

lassoan:
This is great!

Just one thing: I'm not sure if the separators are needed. There are no separators between built-in/installed modules or between built-in/installed subcategories, so adding them in the category list is somewhat inconsistent. Also, there are a lot of categories, so losing two lines might be an issue on small or low-resolution screens.

Maybe simply removing the separators would be better.

Base/QTGUI/qSlicerModulesMenu.cxx
cpinter: Good point, I'll change this.

cpinter:
Andras, there has been a consensus regarding the separator both at the Slicer hangout and in the Mantis issue, so I implemented it accordingly.

lassoan:
So, what do you all think about the inconsistency of the separators present in the top level category list but missing in subcategories and missing between modules?

Extensions are supposed to be good quality enough so that modules or categories that they create should not be necessary to distinguish from the built-in modules. Maybe category 3 extensions could have a small "work-in-progress icon", but it would not be very easy to add, as it might need to be mixed with the module's icon.

lassoan:
I know that it may be frustrating but often it's difficult to see all the good and bad aspects of ideas until they are actually implemented.
If you want to just highlight newly installed modules (it could be useful), then we could consider implement something similar to the Windows application menu's temporary "NEW" labels (and/or highlighted background) on recently installed applications.

cpinter:
I think the separator makes it even easier to identify which categories came from extensions and user-added module paths.
I don't see inconsistency here, I just see a new group of modules that we show to the user.

cpinter:
Also look at comments in the ticket to get others' opinions
http://www.na-mic.org/Bug/view.php?id=3844

lassoan:
Thanks for the bringing the ticket to my attention, I'll see an important issue here: how much modules provided by extensions are considered part of Slicer. It may be a longer discussion, so I have no problem if you integrate this as is, we can easily change presentation of extension modules based on the outcome of the discussion. I'll comment in the mantis item 3844.

cpinter:
If your suggestions/issues are about the deeper concept of the modules, then I think it will be a post-4.5 thing. So I'd suggest keeping this topic and pull request to the scope of this particular issue, integrating it if nobody has further comments/objections (which are still welcome), and starting a new discussion about what Andras has in mind.

jcfr:
Since on MacOSX, extension modules are also installed under `Slicer.app` folder, the `IsPluginBuiltIn` function should be tweaked to consider this. 

See https://github.com/Slicer/Slicer/blob/4d3e2d414a66624a5293ec628d90500d4d35fb65/Base/QTCore/qSlicerCoreApplication.cxx#L626-L640

cpinter:
@jcfr I amended the commit, see vtkSlicerApplicationLogic::IsPluginBuiltIn. Is this satisfactory?

Base/Logic/vtkSlicerApplicationLogic.cxx
```diff
@@ -1728,6 +1728,34 @@ bool vtkSlicerApplicationLogic::IsPluginInstalled(const std::string& filePath,
...
+  std::string canonicalPath = itksys::SystemTools::GetRealPath(path.c_str());
+
+  // On MacOSX extensions are installed in the "<Slicer_EXTENSIONS_DIRBASENAME>-<slicerRevision>"
+  // folder being a sub directory of the application dir, an extra test is required to make sure the
+  // tested filePath doesn't belong to that "<Slicer_EXTENSIONS_DIRBASENAME>-<slicerRevision>" folder.
+  // Since package name can be rename from "Slicer.app" to "Something.app", let's compare
+  // using ".app/Contents/" instead of "Slicer_BUNDLE_LOCATION" which is "Slicer.app/Contents/"
+  bool macExtension = (canonicalPath.find(".app/Contents/" Slicer_EXTENSIONS_DIRBASENAME "-") != std::string::npos);
+
+  return itksys::SystemTools::StringStartsWith(canonicalPath.c_str(), applicationHomeDir.c_str()) || macExtension;
```
jcfr: Should it be 
```
return itksys::SystemTools::StringStartsWith(canonicalPath.c_str(), applicationHomeDir.c_str()) && !macExtension;
```

To confirm, could you add a test similar to the one for `isPluginInstalled` ? 
See https://github.com/cpinter/Slicer/blob/70196a7d2c1414c2cc51a5ce805eea0642923134/Base/QTCore/Testing/Cxx/qSlicerUtilsTest1.cxx#L271-L273

Base/Logic/vtkSlicerApplicationLogic.cxx
```diff
@@ -1728,6 +1728,34 @@ bool vtkSlicerApplicationLogic::IsPluginInstalled(const std::string& filePath,
...
+    }
+
+  std::string path = itksys::SystemTools::GetFilenamePath(filePath);
+  std::string canonicalPath = itksys::SystemTools::GetRealPath(path.c_str());
+
+  // On MacOSX extensions are installed in the "<Slicer_EXTENSIONS_DIRBASENAME>-<slicerRevision>"
+  // folder being a sub directory of the application dir, an extra test is required to make sure the
+  // tested filePath doesn't belong to that "<Slicer_EXTENSIONS_DIRBASENAME>-<slicerRevision>" folder.
+  // Since package name can be rename from "Slicer.app" to "Something.app", let's compare
+  // using ".app/Contents/" instead of "Slicer_BUNDLE_LOCATION" which is "Slicer.app/Contents/"
```
jcfr: :+1:

Base/Logic/vtkSlicerApplicationLogic.cxx
cpinter: You're right, it's IsBuiltIn, not IsExternal. I'll add a test section. Thanks!

cpinter:
@jcfr Issue fixed, test added.
One comment is that directory defined in https://github.com/cpinter/Slicer/blob/3844-builtin-module-categories/Base/QTCore/Testing/Cxx/qSlicerUtilsTest1.cxx#L422 is not used. Is it an oversight?

jcfr:
Good catch.

Instead of 

```
isPluginInstalledTest(__LINE__, false, tmp1.path() + "/" + debug + "/plugin.txt", tmp1.path());
```

it should be 

```
isPluginInstalledTest(__LINE__, false, tmp1.path() + "/" + debug + "/plugin.txt", tmp4.path());
```

Let me know if you would like me to fix this in a separate commit.

In the test, it would also be good to consider the MacOSX case.

cpinter:
OK, I'll add a section for the MacOS case. I'll also make the change that you suggest regarding tmp4. (Maybe not today, I have deadlines tomorrow)

cpinter:
I made the changes, please merge if looks OK.

Sorry I didn't make it to the hangout today, I see you talked about this. I agree that it will be needed to make changes to this later, and that this is only a first step.
In the minutes what does "Module should indicate the category they belong to" mean? Indicate where?

jcfr:
> I made the changes, please merge if looks OK.

Great. I will have a look and integrate.

> In the minutes what does "Module should indicate the category they belong to" mean? Indicate where?

I updated the minute to make things more explicit:

```
 Idea: Module could indicate the category they belong to. This could be done by
 dynamically updating the Help text by indicating the category. 
```

cpinter:
@jcfr If there are no more comments, then I think we can integrate.

Thanks everyone for the thoughts! Certainly there will be more work along these lines, maybe we can discuss at the project week.

jcfr:
Integrated in [r24594](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24594)

cpinter:
Thank you very much!

PR #357, opened 2015-09-28
ENH: serialise singleton tags

naucoin:
Singleton nodes were not having their singleton tag serialized to MRML,
so on loading a scene back in, two nodes would be present in the scene
if the node logic is relying on the tag to update the node that was originally in
the scene.
This fix serializes the singleton tag to MRML, adds a utility method
to uniformly turn the property on or off, adds a test, and expands
the node documentation.

Issue #4006

### DIFF
--- a/Libs/MRML/Core/vtkMRMLNode.h
+++ b/Libs/MRML/Core/vtkMRMLNode.h
@@ -339,12 +339,19 @@ public:
   /// The SingletonTag is used by the scene to build a unique ID.
   ///
   /// If the there can only be one instance of a given node class in the scene,
-  /// then the singleton tag should be Singleton.
+  /// then the singleton tag should be Singleton. For example, the interaction and
+  /// selection nodes are named Selection and Interaction, with Singleton tags set to
+  /// Singleton, and with IDs set to vtkMRMLSelectionNodeSingleton and
+  /// vtkMRMLInteractionNodeSingleton.
   /// If the singleton node is associated with a specific module it should be
-  /// named for the module, which already needs to be unique.
+  /// named for the module, which already needs to be unique. The Editor module
+  /// uses this naming convention, with a parameter node that has a singleton tag
+  /// of Editor and a node ID of vtkMRMLScriptedModuleNodeEditor.
   /// If the there is more than one instance of the node class then the
   /// singleton tag should be Singleton post-pended with a unique identifier
-  /// for that specific node (e.g. the name).
+  /// for that specific node (e.g. the name). Any new color nodes should use this
+  /// convention, with a name of NewName, a Singleton tag of SingletonNewName, leading
+  /// to an ID of vtkMRMLColroTableNodeSingletonNewName.
   /// \sa vtkMRMLScene::BuildID
   vtkSetStringMacro(SingletonTag);
   vtkGetStringMacro(SingletonTag);
--- a/Libs/MRML/Core/vtkMRMLNode.h
+++ b/Libs/MRML/Core/vtkMRMLNode.h
@@ -333,13 +333,33 @@ public:
   ///
   /// If set to NULL, multiple instances of this node class are allowed.
   ///
-  /// If set to a non-NULL string, the node will be a singletong and
+  /// If set to a non-NULL string, the node will be a singleton and
   /// the scene will replace this node instead of adding new instances.
   ///
   /// The SingletonTag is used by the scene to build a unique ID.
+  ///
+  /// If the there can only be one instance of a given node class in the scene,
+  /// then the singleton tag should be Singleton.
+  /// If the singleton node is associated with a specific module it should be
+  /// named for the module, which already needs to be unique.
+  /// If the there is more than one instance of the node class then the
+  /// singleton tag should be Singleton post-pended with a unique identifier
+  /// for that specific node (e.g. the name).
   /// \sa vtkMRMLScene::BuildID
   vtkSetStringMacro(SingletonTag);
   vtkGetStringMacro(SingletonTag);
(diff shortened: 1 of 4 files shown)


Libs/MRML/Core/vtkMRMLNode.h
```diff
@@ -333,13 +333,33 @@ class VTK_MRML_EXPORT vtkMRMLNode : public vtkObject
...
   ///
   /// The SingletonTag is used by the scene to build a unique ID.
+  ///
+  /// If the there can only be one instance of a given node class in the scene,
+  /// then the singleton tag should be Singleton.
+  /// If the singleton node is associated with a specific module it should be
+  /// named for the module, which already needs to be unique.
+  /// If the there is more than one instance of the node class then the
+  /// singleton tag should be Singleton post-pended with a unique identifier
+  /// for that specific node (e.g. the name).
```
lassoan: Please add examples for each case, it would help a lot.

lassoan:
What's the status of the hack in vtkMRMLScene::GetSingletonNode(vtkMRMLNode* n)? Is it still needed? I guess only for legacy scenes. It would be nice to add a comment there to explain why it is still needed and when the hack can be removed.

naucoin:
That method wasn't being used by the Editor, it was doing it's own version of getting it's singleton parameter node. The mrml scene hack is needed for legacy scenes, I can add a comment there pointing to the update in the singleton tag in the mrml node.

jcfr:
> I can add a comment there pointing to the update in the singleton tag in the mrml node.

:+1:

naucoin:
I was just going to say "legacy scenes from before October 2015" since I don't have an commit to reference yet on the main Slicer tree. Any better suggestions for the cross reference?

Libs/MRML/Core/vtkMRMLNode.h
naucoin: The current mrml nodes don't really follow the convention of the third case, the color nodes specifcally do not include the 'Singleton' string in the singleton tag, just the color node names generally, leading to IDs like:
name= MediumChartColors id = vtkMRMLColorTableNodeFileMediumChartColors.txt singleton tag =  FileMediumChartColors.txt
The updated docs I'm about to commit mention any new future nodes with an example of the convention to be followed in the future.

lassoan:
Giving examples in the documentation for tag names that developers would need to use (for new developments) would be useful. We can mention that existing don't always follow these guidelines but they are kept unchanged for preserving backward compatibility.

>    /// the scene will replace this node instead of adding new instances.
>    ///
>    /// The SingletonTag is used by the scene to build a unique ID.
> +  ///
> +  /// If the there can only be one instance of a given node class in the scene,
> +  /// then the singleton tag should be Singleton.
> +  /// If the singleton node is associated with a specific module it should be
> +  /// named for the module, which already needs to be unique.
> +  /// If the there is more than one instance of the node class then the
> +  /// singleton tag should be Singleton post-pended with a unique identifier
> +  /// for that specific node (e.g. the name).
The current mrml nodes don't really follow the convention of the third case, the color nodes specifcally do not include the 'Singleton' string in the singleton tag, just the color node names generally, leading to IDs like:
name= MediumChartColors id = vtkMRMLColorTableNodeFileMediumChartColors.txt singleton tag = FileMediumChartColors.txt
The updated docs I'm about to commit mention any new future nodes with an example of the convention to be followed in the future.

naucoin:
@lassoan  New revision of the documentation committed to the branch.

lassoan:
Current version is perfect, thank you.

Libs/MRML/Core/vtkMRMLNode.h
```diff
@@ -333,13 +333,42 @@ class VTK_MRML_EXPORT vtkMRMLNode : public vtkObject
...
   ///
   /// The SingletonTag is used by the scene to build a unique ID.
+  ///
+  /// If the there can only be one instance of a given node class in the scene,
+  /// then the singleton tag should be Singleton. For example, the interaction and
+  /// selection nodes are named Selection and Interaction, with Singleton tags set to
+  /// Singleton, and with IDs set to vtkMRMLSelectionNodeSingleton and
+  /// vtkMRMLInteractionNodeSingleton.
+  /// If the singleton node is associated with a specific module it should be
+  /// named for the module, which already needs to be unique. The Editor module
```
jcfr:
```
/// If the singleton node is associated with a specific module it should be
/// named for the module ....
```
-> 

```
/// If the singleton node is associated with a specific module it should be
/// named like the module ....
```

Beside of `for` -> `like` tweak.

:+1:

naucoin: I'm not sure that "like" is more clear than "for", maybe "after"?

jcfr: Good point. Let's use `after` and also add `name` 
```
/// If the singleton node is associated with a specific module it should be
/// named after the module name, which already needs to be unique.
```

naucoin: How about this:
  /// If the singleton node is associated with a specific module it should use
  /// the module name, which already needs to be unique, as the tag. The Editor module

jcfr: :+1:

naucoin: Change pushed.

naucoin:
Integrated:
http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24582
https://github.com/Slicer/Slicer/commit/808c8e68438c8fff6e77f4886b67facb77659a9b

PR #358, opened 2015-09-29
Two clis in a row issue

vovythevov:
There is an issue when running the same CLI twice in a row, the second CLI never completes.
This implements a test and a fix.

### DIFF
--- a/Applications/SlicerApp/Testing/Python/CMakeLists.txt
+++ b/Applications/SlicerApp/Testing/Python/CMakeLists.txt
@@ -235,6 +235,7 @@ if(Slicer_USE_QtTesting AND Slicer_USE_PYTHONQT)
   slicer_add_python_unittest(SCRIPT FiducialLayoutSwitchBug1914.py)
   slicer_add_python_unittest(SCRIPT SubjectHierarchyGenericSelfTest.py)
   slicer_add_python_unittest(SCRIPT SubjectHierarchyCorePluginsSelfTest.py)
+  slicer_add_python_unittest(SCRIPT TwoCLIsInARowTest.py)
 
   # add as hidden module for use at run time
   set(KIT_PYTHON_SCRIPTS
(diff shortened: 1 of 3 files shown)


pieper:
Nice fix :+1:

lassoan:
Looks good to me.

Have you tested if CLI auto-run still works fine?

vovythevov:
No actually, good call.
I'll test that tomorrow and report back.

On Tuesday, September 29, 2015, Andras Lasso <notifications@github.com>
wrote:

> Looks good to me.
>
> Have you tested if CLI auto-run still works fine?
>
> —
> <https://github.com/Slicer/Slicer/pull/358#issuecomment-144226420>.
>


-- 
Sent from Gmail Mobile

vovythevov:
@lassoan,  @pieper :
Tested the CLI auto-run by hand and it seems to run to fine.
Thanks for the (very quick) feedback !

jcfr:
Nitpicks: `VTKObservationMixin` could be used to observe events in the test. That said, it looks good the way it is and can definitively be integrated.
See http://wiki.slicer.org/slicerWiki/index.php/Documentation/Nightly/Developers/Python_scripting#How_can_I_access_callData_argument_in_a_VTK_object_observer_callback_function

vovythevov:
Actually, I think it would be even better if ScriptedLoadableModuleLogic ( https://github.com/Slicer/Slicer/blob/master/Base/Python/slicer/ScriptedLoadableModule.py#L130) inherited directly from VTKObservationMixin. We would get all those (very useful) methods in the CLIs.

What do you think ?

jcfr:
Very good point. They will update the templates and existing code in a different commit.

vovythevov:
Alexis Girault reported an issue when the same CLIs is ran twice in parallel. 

After looking at it with @jcfr, we suggest that the observations of the application logic should be made in the SetApplicationLogic() instead of everytime a CLI is executed.

@agirault: Would you mind looking into it ?

jcfr:
More specifically, it seems the solution will be to add/remove the observation associated with `vtkSlicerApplicationLogic::RequestProcessedEvent` in the `SetApplicationLogic()` setter.

vovythevov:
I'll close this pull request in favor of #458.

PR #329, opened 2015-09-03
ENH: update DCMTK to use snapshot DCMTK-3.6.1_20150924

fedorov:
The important improvements provided by the current DCMTK is the introduction of
dcmiod, dcmfg and dcmseg classes that will be helpful in creating multiframe
DICOM objects, such as segmentation and parametric map objects. This updated
version also includes the feature to embed standard and private DICOM dictionaries.
This makes it unnecessary to use DCMDICTPATH for loading dictionaries from external
file, thus that functionality is removed.

PR #359, opened 2015-09-30
ENH: Update DMCTK to use version DCMTK-3.6.1_20150924

fedorov:
This PR is the update on #329, with the commits squashed. 

@jcfr I confirm the test is passing after removing the check for env DCMDICTPATH variable, which is included in this commit. I would appreciate if you let me know as soon as feasible if there is any other issue with it. 

Copying all participants of #329: @pieper @michaelonken @msmolens 

---

The important improvements provided by the new version of DCMTK
is the support for new API that in particular greatly simplify creation of and
interaction with DICOM Segmentation image objects.

More specifically, it introduces the dcmiod, dcmfg and dcmseg classes
helpful in creating multiframe DICOM objects, such as segmentation and
parametric map objects.

Considering that standard and private dictionaries can now be  embedded
in DCMTK library, this commit partially reverts r24473 (ENH: Add support
for DCMTK private dictionary). The test DCMTKPrivateDictTest.py was updated to
not expect DCMDICTPATH environment variable that is no longer needed and is not
initialized.

List of DCMTK changes:

$ git shortlog 3366181..DCMTK-3.6.1_20150924 --no-merges

Andreas Thiel (1):
      Updated man pages for new development snapshot.

Jan Schlamelcher (86):
      Some workarounds for VC6 compiler peculiarity.
      Introduced OFunique_ptr as a fallback if std::unique_ptr is not available.
      Make use of  unique pointer (free memory if C-MOVE/C-FIND is canceled).
      Fixed dll build on windows.
      Correction of inconsistent previous commit.
      Fixed DcmSCPPool test case failing on older systems.
      Enabled shutdown functionality of DcmSCPPool.
      Fixed building with mingw-w64.
      Fixed most of the warnings in dcmjpeg.
      Fixed potential bug in dcmjpeg 12 and 16 bit variants.
      Marked SCPPool-Test (dcmnet_scp_pool) as exhaustive.
      Fixed most of the warnings in dcmdata.
      Fixed a mistake in CHANGES.361.
      Minor fixes in dcmdata.
      Added support for some C++11 types, literals and traits.
      Added documentation for the new macro "DOXYGEN".
      Fixed most of the warnings in ofstd and oflog.
      Added support for some additional C++11 traits.
      Introduced the new explicit boolean type OFExplicitBool.
      Fixed possibly wrong overload resolution in DSRCodedEntryValue.
      Minor fixes (whitespace and line ending errors).
      Modified DcmQuantColorHashTable to perform lazy initialization.
      Introduced OFoptional, implementing optional types.
      Renamed DoxygenLayout.xml to comply with filename conventions.
      Minor fixes and workarounds for Microsoft's compilers.
      Fixed ofstd_tatof error on older MSVC versions.
      Fixed DCMTK-Exports for custom module setups.
      Fixed and suppressed some warnings.
      Added 'DCMTK_USE_CXX11_STL' and 'DCMTK_CXX11_FLAGS'.
      Updated manpages.
      Updated DCMTK_ABI_VERSION for new development snapshot.
      Clarified an ambiguous configure message.
      Removed misleading comment in dcmjpeg.
      Fixed a wrong type-alias for OFintegral_constant.
      Introduced C++11-like alignment support.
      Fixed an alignment problem in OFoptional.
      Fixed building 32 bit targets with mingw-w64.
      Refined definition of numeric types and traits.
      Fixed a problem with Microsoft Safe CRT in oflog.
      Fixed a bug in dcmdata.
      Introduced DcmOutputBufferStream::filled() method.
      Fixed a problem in storescp.
      Modified some of OFoptional's comparison operators.
      Introduced OFtuple implementing generic tuples.
      Introduced move emulation for pre C++11 compilers.
      Small fixes for a previous commit about OFtuple.
      Several fixes and additions for yesterday's commits.
      Reintroduced bugfix-comment in dcmjpeg.
      Fixed autoconf install regarding ofstd variadic.
      Introduced OFnumeric_limits.
      Several fixes regarding OFnumeric_limits.
      Fixed possible arith.h generation problem on MinGW.
      Introduced basic support for assisted cross compiling.
      Fixed some problems regarding DCMTK's C++11 support.
      Added support for trusted client certificates.
      Adjusted DCMNET's default socket buffer length.
      Introduced cross compiling support for Android.
      Fixed a const correctness problem within DCMTLS.
      Introduced check-exhaustive targets to Autoconf.
      Corrections regarding recent commits.
      Modified CMake minimum version as required.
      Revised cleanup of temporary unit test files.
      Updated Autoconf dependencies.
      Updated man pages for new development snapshot.
      Updated DCMTK_ABI_VERSION for new development snapshot.
      Fixed a CMake problem with paths containing spaces.
      Moved some utility functions inside OFStandard.
      Modified fallback implementation of OFtuple.
      Added additional configuration methods for DCMTLS.
      Updated Makefile dependencies.
      Updated DCMTK_ABI_VERSION for new development snapshot.
      Moved CHANGES.361 to /docs.
      Updated Doxygen configuration files.
      Updated man pages.
      Updated default compiler flags.
      Refactored handling of some <cmath> functions.
      Fixed problems with previous commit.
      Fixed cross compiling via MinGW and Wine.
      Fixed building on FreeBSD.
      Fixed building on FreeBSD with C++11 enabled.
      Fixed inconsistent defines in CMake's osconfig.h.
      Fixed a problem in DcmAttributeFilter's unit test.
      Fixed building via GCC 4.2 on OpenBSD.
      Updated man pages for new development snapshot.
      Updated DCMTK_ABI_VERSION for new development snapshot.
      Updated CHANGES.361 for new development snapshot.

Joerg Riesmeier (511):
      Added support for CP-1236 to data dictionary.
      Added support for CP-1236 to checkAndUpdateVR() method.
      Added support for Supplement 154 to the data dictionary.
      Added support for new SOP Class UIDs from Supplement 154.
      Added support for directory record type "SURFACE SCAN".
      Added new SOP Class UID from Supplement 162.
      Added missing assignment of member variable.
      Added full support for the new Comprehensive 3D SR IOD.
      Fixed various issues in syntax usage output.
      Minor fixes in comments, e.g. typos.
      Minor fixes in comments. Also made some input parameters "const".
      Use non-throwing new and delete operator.
      Fixed two possible memory leaks.
      Added comment to variable that is currently unused.
      Fixed issue with odd-length pixel data.
      Added missing sub-directory to "make install" target.
      Added some debug output to lookForXfer() method.
      Added some debug output to read() method.
      Added new print options for SOP class output.
      Added new command line options for print output.
      Fixed bug in checkSOPClassUID() method.
      Added note on support for shared libraries.
      Added support for CP-1246 and CP-1249 to data dictionary.
      Fixed some tag key constants of recently retired attributes.
      Fixed issue with unnecessary call to clearNodeCursorStack().
      Added missing support for the new print options.
      Enhanced output to trace logger.
      Fixed inconsistencies introduced with last commit.
      Various fixes to comments. Removed trailing spaces.
      Added support for recently approved Correction Items.
      Enhanced documentation of complete flag methods.
      Added support for Supplement 157 to data dictionary.
      Added SOP Class UIDs from Supplement 157.
      Minor fixes to comments on VOI-related methods.
      Fixed bug in getTimeZoneFromString() method.
      Fixed issue with misplaced log output.
      Replaced "OFString::length() == 0" by "OFString::empty()".
      Replaced "OFString::length() != 0" by "!OFString::empty()".
      Replaced "OFString::length() > 0" by "!OFString::empty()".
      Added support for new VR "Other Double String" (OD)
      Added private tags for AGFA IMPACS 6.5.x Solution.
      Fixed wrong VR for private AGFA IMPAX tag.
      Updated version number of snapshot compilers.
      Added comment on inappropriate use of EC_IllegalCall.
      Fixed various oddities and inconsistencies:
      Replaced tabulator by space characters.
      Fixed various oddities and inconsistencies.
      Really call shouldStopAfterCurrentSOPInstance():
      Fixed further documentation issues.
      Removed trailing space characters.
      Fixed various typos.
      Added releaseAssociation() and abortAssociation()
      Fixed various typos and other Doxygen issues.
      Enable WITH_THREADS for thread-related classes.
      Added new condition constants.
      Fixed various typos.
      Use optional fields in sendSTOREResponse().
      Reimplemented sendSTOREResponse() method.
      Modified log output in loadAssociationCfgFile().
      Fixed issues with optional message fields.
      Removed unused fields from sendACTIONResponse().
      Added file missing for last commit :(
      Added progress notification to DcmSCP.
      Removed trailing spaces.
      Added global helper function createDirectory().
      Fixed some minor Doxygen issues.
      Added new class DcmStorageSCP.
      Added new command line tool "dcmrecv".
      Added comments to unused exit codes.
      Rebuilt Makefile dependencies.
      Updated man pages.
      Function createDirectory() also works on Windows:
      Updated CMake version.
      Fixed various typos.
      Deprecated closeAssociation() method.
      Added support for Supplement 163 to data dictionary.
      Added meta header to stored DICOM files.
      Fixed various typos in comments.
      Pass existing dataset to handleSTORERequest().
      Fixed various inconsistencies in comments.
      Do not copy received dataset in memory.
      Removed class name from derived method calls.
      Removed "&" from constant integer parameters.
      Fixed wrong log message in handleMOVERequest().
      Fixed various typos in comments.
      Restructured implementation of handleSTORERequest().
      Introduced dataset storage mode.
      Fixed various typos in comments.
      Receive C-STORE request dataset directly to file.
      Added new dataset storage mode.
      dded new storage mode options.
      Updated copyright dates (where appropriate).
      Added support for generating a Doxygen tag file.
      Fixed wrong Doxygen reference to a method.
      Minor fixes, e.g. added Doxygen markup.
      Fixed issue with CP-246 files and private tags.
      Slightly changed output of print() method.
      Added new print flag for showing invalid codes.
      Added new option for printing invalid codes.
      Added missing references to closed DCMTK features.
      Fixed typos.
      Added support for <InlineBinary> XML element.
      Clarified use of writeXML() flags.
      Fixed typo copied from Supplement 163.
      Output a warning message if the VR is unknown.
      Output currently processed line of input file.
      Removed superfluous footnote symbol "(*)".
      Fixed various coding style issues.
      Added trace log message on VR replacement.
      Added two new global flags for VR generation.
      Made use of new global flags for VR generation:
      Updated our email address for bug reports.
      Updated to autoconf 2.69.
      Added definition of DCMTK_PACKAGE_URL.
      Added Doxygen tag file to .gitignore.
      Added *.patch files to .gitignore.
      Also updated autoconf support files.
      Updated automake script to version 1.14.
      Fixed typos.
      Fixed various typos.
      Fixed various typos in comments.
      Added method to determine SpecificCharacterSet.
      Added 'charset' parameter to checkStringValue().
      Added new variant of checkElementValue().
      Avoid unnecessary copies of sequence elements.
      Enhanced implementation of checkCode() method.
      Report warning on unsupported character set.
      Use helper function findAndGetSequence().
      Updated comment on current character set issue.
      Fixed memory leak in log4cplus code.
      Minor fixes in comments.
      Minor changes to comments.
      Updated Doxygen configuration files.
      Removed outdated custom CSS file.
      Updated URL of DCMTK's home page.
      Fixed timeout issue in handleAssociation().
      Fixed source code formatting and various comments.
      Fixed wrong content type of XML element in DTD.
      Do not output group length in Native DICOM Model.
      Fixed outdated comment on 1st PN component group.
      Removed redundant API comment.
      Added support for CP-1268 to data dictionary.
      Added support for Supplement 168 to data dictionary.
      Added support for new UID from Supplement 168.
      Never output NULL bytes to markup string/stream.
      Adapted test cases for markup conversion.
      Iterate more efficiently over sequence items.
      Use helper method for deleting data elements.
      Added comments on missing features.
      Fixed small typos in comments.
      Added support for Real World Value Mapping objects.
      Added comment on getFrameList() method.
      Moved checkCurrentValue() to "protected" section.
      Slightly enhanced comments.
      Check return value of write() methods.
      Added support for Referenced Segment Number.
      Updated CMake version.
      Fixed small issue reported by CMake 2.8.12.
      Added Corneal Topography Map Storage SOP Class.
      Fixed bug that could cause a crash.
      Fixed doubled output of XML start tag for VR=OD.
      Added comment on wrong color model for icon images.
      Fixed issue with color icon images.
      Removed trailing white spaces from man pages files.
      Fixed build issue with shared libraries (Win32).
      Avoid CMake warning on non-installable files.
      Added missing break statements to switch cases.
      Removed redundant and outdated Word document.
      Added global flag for disabling the VR checker.
      Added new option for disabling the VR checker.
      Fixed minor Doxygen-related issue in documentation.
      Fixed name of PN component in Native DICOM Model.
      Fixed minor Doxygen-related issue in documentation.
      Fixed issue with test "ofstd_markup_6" (Win32).
      Fixed various typos in man pages.
      Fixed output of AT values in Native DICOM Model.
      Do not check valueLoaded() for Native DICOM Model.
      Completed support for <InlineBinary> XML element.
      Second try to fix the issue with "ofstd_markup_6".
      Fixed typos, comments and coding style.
      Replaced deprecated DcmSCU::closeAssociation().
      Removed check on number of command line arguments.
      Cleaned up source code in various aspects.
      Again, fixed comments to be consistent with code.
      Fixed comments and documented default values.
      Added support for recently approved CPs.
      Added support for Supplement 165 to data dictionary.
      Added support for new UIDs from Supplement 165.
      Fixed wrong/missing comment on method parameter.
      Fixed various minor documentation issues.
      Added missing module dcmrt to include directories.
      Removed comment on UID_HangingProtocolStorage.
      Added support for approved but still missing CPs.
      Added comments on some ACR/NEMA attributes.
      Fixed timeout issue in DIMSE_moveUser().
      Fixed typos in comments.
      Moved ACR/NEMA data dictionary to a separate file.
      Fixed issue with wrong check on stream compression.
      Fixed wrong VR for some retired command fields.
      Consistently use upper case letters for DICOM tags.
      Added new test cases for testing helper functions.
      Fixed typo and added comment.
      Added OFFilename versions to some helper functions.
      Added new helper function for renaming files.
      Added function for appending filename extension.
      Added wide character support to this tool.
      Added new helper function for copying files.
      Changed file mode bits back from "0755" to "0644".
      Fixed typo in comment.
      Added new command line option --create-backup.
      Avoid passing invalid values to rename().
      Avoid passing NULL to OFString, which is deprecated.
      Added support for wide chars to DICOMDIR code.
      Fixed issue when compiled with wide char support.
      Added new condition constant and fixed typo.
      Fixed issue with missing terminating NULL byte.
      Always use OFFilename() instead of "" as a default.
      Use printError() instead of OFLOG_FATAL().
      Fixed further issues with wide char support.
      Allow empty file-set ID in DICOMDIR (type 2).
      Fixed some comments.
      Fixed wrong tag in CodingSchemeIdentificationSequence.
      Made installation of tag file optional.
      Added another addContentItem() method.
      Replaced tabs by spaces and fixed indention.
      Fixed various issues in Doxygen comments.
      Fixed various minor issues in new helper class.
      Rebuilt Makefile dependencies.
      Added missing data elements to BD profiles.
      Added check for Shared Functional Groups Sequence.
      DIMSE_createFilestream() now uses an OFFilename.
      Fixed various typo.
      Pass OFString to DIMSE_createFilestream().
      Added support for CP-1090 to the data dictionary.
      Fixed issue with automatically generated classes.
      Removed superfluous quotations marks.
      Fixed small typos in comments.
      Added readPatientData() and readStudyData().
      Updated outdated comments on "log stream".
      Made sure that header file is parsed by Doxygen.
      Made sure that "Files" section is available.
      Added further header files to Doxygen documentation.
      Added support for recently approved CPs.
      Fixed some issues with wide character support.
      Added test for OFStandard::appendFilenameExtension().
      Fixed issue with untyped NULL reported by MSVC++.
      Fixed various typos in comments.
      Enhanced removeRootDirFromPathname() function.
      Enhanced createDirectory() function.
      Use non-throwing new and delete operator.
      Fixed typo in comment.
      Output to debug logger if memory allocation fails.
      Simplified creation of temporary DICOMDIR file.
      Optionally disable isFilenameValid() check.
      Added support for new UIDs from Supplement 124.
      Added support for Supplement 124 to data dictionary.
      Removed superfluous member initialization.
      Fixed typo in comments.
      Added support for determining the FrameTime.
      Access the (total) number of frames in an image.
      Fixed inconsistent parameter type.
      Added explicit type casts to avoid warnings.
      Fixed inconsistent parameter type.
      Updated data dictionary for DICOM 2013 edition.
      Fixed typos in comments.
      Added further error handling options.
      Fixed typos in comments.
      Fixed typos, added missing documentation, etc.
      Various minor fixes for development version.
      Fixed typos in comments.
      Made sure that the move originator is cleared.
      Fixed issue with text alignment in PatternLayout.
      Fixed some Doxygen related issues (and a typo).
      Consistently use "presInfo" for variables/parameters.
      Added support for Supplement 159 to data dictionary.
      Added support for new UID from Supplement 159.
      Added support for new SR IOD from Supplement 159.
      Added DcmSCP::abortAssociation() method.
      Fixed typo and updated copyright date.
      Removed meaningless layout file setting.
      Renamed Doxygen's layout file again.
      Made sure that header file is parsed by Doxygen.
      Group condition constants in Doxygen output.
      Avoid full directory path in Doxygen footer.
      Replaced CMake variable in Doxygen config file.
      Removed outdated comment on DICOM 2008 standard.
      Added missing empty line.
      Fixed typos and other minor issues.
      Updated reference to DICOM standard edition.
      Updated DCMTK's copyright date.
      Added support for CP-1125 to SR implementation.
      Fixed incomplete #include comment.
      Added support for recently approved CPs.
      Updated comment based on CP-1335.
      Added support for new Colon CAD SR constraints.
      Fixed various typos and minor Doxygen issues.
      Updated data dictionary for DICOM 2014a edition.
      Removed year component from version field.
      Minor fixes to file's comment header.
      Updated copyright date and other minor changes.
      Implemented copy constructors and added clone().
      Added explaining comment on new copy constructor.
      Removed outdated CMake exclude pattern.
      Excluded "*.md5" files from installation.
      Made sure that all top-level nodes are deleted.
      Added method counting the number of nodes.
      Added two new convenience functions.
      Fixed issue with tests run in different order.
      Fixed various Doxygen-related issues.
      Introduced new class DSRDocumentSubTree.
      Implemented copy constructors for tree classes.
      Added getPreviousNode() and getChildNode().
      Added further checking methods to cursor class.
      Added gotoFirst() and gotoLast() methods.
      Minor fixes on "const" declaration of pointers.
      Added cloneSubTree() method to tree classes.
      Changed return type of addContentItem(node*).
      Reimplemented various getXXX() methods.
      Cleaned up handling of "unknown" relationships.
      Restructured classes for nodes and trees.
      Fixed a bug in convertNewlineCharacters().
      Added insertSubTree() method to subtree class.
      Added checking of IOD-specific constraints.
      Fixed two memory leaks in recently added code.
      Clarified use of template identification.
      Removed DLL specifier from template classes.
      Slightly enhanced layout of Doxygen documentation.
      Added public method to change SR document type.
      Fixed wrong term in Doxygen documentation.
      Added removeSubTree() method to subtree class.
      Added extractSubTree() method to tree classes.
      Enhanced support for by-reference relationships.
      Further enhanced by-reference relationships.
      Insert a subtree into an empty document tree.
      Added dcmIsImageStorageSOPClassUID() function.
      Added check on valid "IMAGE" SOP Classes.
      Added support for optional JPEG fill byte.
      Removed misleading comment from checkStringValue().
      Fixed typos in comments and other minor issues.
      Added support for new value representation "UR".
      Updated data dictionary for DICOM 2014b edition.
      Added functions to enable/disable post-1993 VRs.
      Enhanced rewriting rules in DcmVR::getValidEVR().
      Use new VR "UR" for Pixel Data Provider URL.
      Fixed typo in method name.
      RLE images should not use multiple fragments.
      Added missing documentation for --propose-deflated.
      Attribute (0040,4005) has not been retired.
      Fixed wrong term "Device Driving Level".
      Updated "dcmrt" classes based on DICOM 2014b.
      Added "data correction" options to findscu.
      Fixed issue when scaling certain images.
      Removed unneeded header include.
      Fixed typo in comment.
      Fixed wrong validation of PN, LO and SH values.
      Added check for requirements clarified by CP-1401.
      Updated data dictionary for DICOM 2014c edition.
      Added support for new UIDs from DICOM 2014c.
      Fixed typo in comment.
      Added new temporarily generates files.
      Use "try and catch" block around new[].
      Fixed error handling in detachValueField().
      Fixed error handling in changeValue().
      Added new test case for VR=FD.
      Changed attribute used for the "FD" test.
      Added support for new value representation "UC".
      Added support for recently approved CPs.
      Added Mapping Resource UID from CP-1417.
      Removed outdated name suffix from EXS_xxx constant.
      Added missing standard include for assert().
      Fixed another CMake 3.1.x quotation issue.
      Updated latest tested CMake version.
      Fixed documentation of default TCP buffer length.
      Updated data dictionary for DICOM 2015a edition.
      Fixed typos in comments and other minor issues.
      Updated Makefile dependencies again.
      Updated outdated API documentation.
      Output message to logger on wrong pixel data class.
      Enhanced use of DSRCodingSchemeIdentificationList.
      Added support for recently approved CPs.
      Added support for Supplement 173 to dictionary.
      Added support for new UIDs from Supplement 173.
      Added support for recently approved CP-1425.
      Fixed test after after implementing CP-1425.
      Added further test cases for VR checker.
      Added read flag to all readXXX() methods.
      Optionally accept invalid content item values.
      Distinguish b/w invalid content items and values.
      Enhanced output to debug logger.
      Fixed wrong position counter in tree node.
      Added regression test for DSRTree::getPosition().
      Added method for counting number of child nodes.
      Fixed issue with VR checker for PN values.
      Fixed warning C4800 reported by Visual Studio.
      Updated data dictionary for DICOM 2015b edition.
      Ignore empty value for XML element 'charset'.
      Added support for optional Mapping Resource UID.
      Pass OFList parameter by-reference not by-value.
      Fixed issue with interpolation algorithm.
      Added hasTemplateIdentification() method.
      Replace the document tree of an SR document.
      Added initial support for SR templates.
      Added helpers for setting a value from dataset.
      Set observation date/time from dataset.
      Moved various methods to protected section.
      Added helpers for setting a value from element.
      Set observation date/time from element.
      Avoid using "check" parameter of SR code class.
      Introduced new error code "Unknow VR".
      Fixed source code and/or comment formatting.
      Do not check invalid UID values for IMAGE node.
      Added support for long code values (CP-1031).
      New class for "Basic Coded Entry Attributes".
      Fixed bug in time normalization code.
      Added missing empty() method to OFMap.
      Added new option for print() method.
      Added new comparison method to DSRCodedEntryValue.
      Made getAndRemoveRootNode() public.
      Fixed issue with addNode() in basic tree class.
      Added new and enhanced existing swap() function.
      Check whether a template is extensible.
      Added initial support for Supplement 186.
      Added initial support for Supplement 180.
      Made constructor and destructor "protected".
      Removed explicit name of base class from call.
      Added mechanism for remembering node positions.
      Added base class for context groups.
      Added initial version of DCM code definitions.
      Added first specific Templates and Context Groups.
      Updated "DCM" codes for DICOM 2015c edition.
      Updated data dictionary for DICOM 2015c edition.
      Fixed issue with wrong quotation marks.
      Added static initialization function (MT-safe).
      Fixed issue in OFStandard::combineDirAndFilename().
      Enhanced check on empty dataset.
      Improved documentation of disableXXXCheck().
      Consistently use "file-set" and "file format".
      Allow for disabling check on "DICOM file format".
      Renamed ANNOUNCE file (removed version suffix).
      Enhanced documentation of createNewSOPInstance().
      Added conversion operator to context group class.
      Added support for CID 42 (Numeric Value Qualifier).
      Added function that initializes all context groups.
      Updated copyright date.
      Fixed wrong documentation of return value.
      Removed superfluous OFStringStream_ends statement.
      Fixed wrong exit code on invalid input file.
      Added support for further Context Groups.
      Fixed various typos in comments.
      Enhanced handling of empty Measured Value Sequence.
      Added check for CID 42 to NUM values.
      Added support for Context Group CID 29.
      Added general information on coding scheme "DCM".
      Added easier method to set a composite reference.
      Fixed issue reported by Clang 3.5 on OpenBSD.
      Introduced condition constant EC_NotYetImplemented.
      Do not set empty concept name codes.
      Added missing DCMTK_DCMDATA_EXPORT macro.
      Added new helper method gotoEntryFromNodeList().
      Added method to look up a context group entry
      Added new mode for adding nodes to the tree.
      Fixed typos in comment.
      Added comment on reserved error codes (for "cmr").
      Fixed wrong comment.
      Introduced new flag to customize the print output.
      Added new command line options to dsrdump.
      Added specific logger for submodule "dcmsr/cmr".
      Added parameter "pos" to getStringValueFromDataset.
      Added static method setEnhancedEncodingMode().
      Clarified use of invalid coded entries.
      Added support for Context Group CID 244.
      Made sure that IOD constraint checker is used.
      Added "Extensible SR" to known document types.
      Added enhanced versions of CID 29, 244 and 4031.
      Added cleanup() function to context group classes.
      Fixed typo and other minor issues in comments.
      Added name prefix to class ContentMappingResource.
      Fixed copyright header in module "dcmsr".
      Added new optional parameter "vm" to readSequence.
      Added methods to read/write to/from sequence items.
      Fixed inconsistencies regarding the use of OFMap.
      Updated Makefile dependencies.
      Added enhanced versions of CID 10013 and 10033.
      Added initial version of SRT and UCUM codes.
      Introduced type for list of CID 7445 values.
      Updated Makefile dependencies.
      Added initial version of TID 1600 (Image Library).
      Slightly changed conversion operator.
      Extracted general information from template.
      Avoid seperator if time component is empty.
      Added support for annotating content items.
      Added support for Context Group CID 100 and 7021.
      Added support for clearing tree node annotations.
      Made sure that tree node annotations are copied.
      Made sure that "dicom.dic" is loaded for tests.
      Fixed various issues in test case.
      Fixed "cardinality" for SQ elements in comments.
      Added comment that "Big Endian" is retired.
      Removed version suffix "3" from dictionary entries.
      Check logger for debug output of template tree.
      Introduced global definitions for certain tags.
      Added CREDITS file.
      Added "YXLON" to the new CREDITS file.
      Added support for recently approved CPs.
      Added support for Supplement 156 to dictionary.
      Added support for new UIDs from Supplement 156.
      Check for recently approved multi-frame images.
      Updated currently supported compiler versions.
      Fixed various minor issues in module descriptions.
      Fixed order of modules in main HTML file.
      Added three new DCMTK modules to README.

Marco Eichelberg (5):
      Introduced OFStandard::dropPrivileges().
      Tools now use OFStandard::dropPrivileges().
      Fixed minor bug in DICOM upper layer FSM.
      Always use current dicom.dic for dcmdata unit tests.
      Removed unsupported wwwapps and perl scripts from module dcmwlm.

Michael Onken (84):
      Added (optional) originating C-MOVE AE title and message ID to C-STORE message.
      Make DcmStorageSCU make use of new MoveOriginator feature.
      Added optional status detail when sending C-MOVE response.
      Make sure memcpy does not copy to NULL destination. Thanks to Francesco Gigante     for the report.
      Added option to discard illegal characters when converting between     DICOM character sets.
      Add "Discard illegal character" support to dcmconv and classes like DcmItem.
      Make sure SCU and SCP request/response data structures are initially zeroed.
      Fixed typo (coersion -> coercion) in DIMSE status constant.
      Added missing kudos.
      Fixed too small buffer for UID in Q/R server.
      Fixed issues where JPEG marker map was not cleared if more than one JPEG     was converted using the same plugin instance.
      Fixed possible doubled destruction of association.
      Make sure oflog also compiles if threads are disabled.
      Always write meta header when receiving files. Removed Winsock comment.
      Added missing CHANGES entry.
      Warning to re-create index.dat after 20130705:
      Create DcmFileFormat from DcmDataset w/o copying.
      Do not copy DcmDataset during C-GET receive.
      Worklist SCP support for Responsible Person/Role.
      Replaced tabs with spaces.
      Support explicit item length denoting too many bytes.
      Corrected misleading transfer syntax error message.
      Fixed Comments.
      Described movescu storage SCP timeout problem.
      Make sure CMAKE_BUILD_TYPE has a default value.
      Fixed doubled tag output in XML output of VR OF.
      Avoid second OpenSSL initialization in constructor.
      Documentation for CMAKE_BUILD_TYPE default value.
      Fixed typo.
      Fixed CHANGES entries.
      Fixed some inconsistencies in documentation.
      Make sure to handle setuid() return code properly.
      Add libjpeg dependency for libtiff and minor changes.
      Fixed shared library build under Unix (configure).
      Enhanced DCMTK CMake exports for external projects.
      Capitalized all CMake key words for consistency.
      Ensure CMake v2.6 still writes DCMTKConfig.cmake.
      DCMTK.cmake and DCMTKTargets.cmake documentation.
      Added helpers to copy common modules between items.
      Fixed Makefile.dep for new files from last comit.
      Switched windows line endings to unix line endings.
      Added further module helpers and fixed small issues.
      Fixed comments and error message.
      Updated module helpers to latest DICOM 2013 draft.
      Fixed wrong ifdef syntax.
      Add connection timeout setting to SCU class.
      Fixed possible memory leak.
      Specific error if no presentation context proposed.
      Reverted false retirement status of some SOP Classes.
      Fix last commit by including implementation file.
      New compare() function for dcmdata object values.
      Make sure dcmdata tests compile with autoconf.
      Make new compare() method more reliable.
      Fixed conversion of color to palette color mode.
      Documented DCMTK's C++ cast macros.
      Fixed installation and added CHANGES file.
      Disable built-in dictionary for CMake Unix builds.
      Make sure presentation contexts do not interfer.
      Fixed iterator type.
      Introduced new modules dcmiod, dcmfg and dcmseg.
      Removed accidentially checked in test files.
      Removed more accidentially checked in test files.
      Fixed iterator dereference.
      Fixed iterator constness.
      Fixed possible DICOM VR problems.
      Removed unused method.
      Fixed mostly constness for some compilers.
      Fixed various doxygen formatting issues.
      Fixed compilers errors (mostly STL-specific).
      Removed unused link libraries.
      Made copyright header consistent in new modules.
      Build flags for internal and external dictionary.
      Updated dictionary build and code documentation.
      Added shared library export maro to IODModule.
      Added more files to be ignored by git.
      Added mising "check-exhaustive" targets.
      Replaced tab characters with spaces.
      Use static_cast instead of dynamic_cast.
      Renamed dictionary configuration build options.
      Simplified builtin dictionary infrastructure.
      Fixed typo in CMake script that can cause warnings.
      Fixed typo and made documentation more consistent.
      Fixed another typo.
      Enable private tags in builtin dictionary on CMake.

Nikolas Goldhammer (1):
      Replaced deprecated CMake expressions.

Uli Schlachter (13):
      Fixed wrong definition of Value Representation Person Name (PN)
      Removed unused local variable which shadowed a variable with the same name.
      Use ".NOTPARALLEL:" to make parallel builds (e.g. -j4) possible.
      Update rootconf for changes which were done directly to generated files.
      Fixed various issues found via cppcheck.
      Use DcmTagKey's output operator where possible.
      Don't include intrin.h on Visual Studio 2005 because of build errors.
      Fixed bug in OFString::copy() and improved unit tests.
      Fixed compiler warnings in ConvertIntegerToString with unsigned integers.
      Worked around compiler warnings on 32bit systems for too large shifts.
      Fixed crash if less than four transfer syntaxes are enabled.
      Silence some more compiler warnings about implicit casts.
      Don't set the defines for thread local storage support incorrectly.


jcfr:
:+1:

jcfr:
That is great. Thanks for the updated PR. 

Assuming the build issue you initially mentioned was due to a non-clean build tree, feel free to integrate. :8ball:

fedorov:
Thanks JC! Regarding the build error, it must have been non-clean build tree. All I can say I could not reproduce the issue while building from scratch.

I will merge today.

jcfr:
> Thanks JC! Regarding the build error, it must have been non-clean build tree. 
> All I can say I could not reproduce the issue while building from scratch.

Makes sense.

Could you send a short note on the dev list to indicate that user may have to do a clean build if they experience build error following the update.

> I will merge today.

:+1:

fedorov:
Merged in r24586 = 3c0c211a290f499de67f05ed42ec6f7ca2e4bd46

fedorov:
Very unfortunately, this broke the build ... This is on Mac Snow Leopard. I tested on Yosemite, and did not have build errors....

http://trunk.cdash.org/viewBuildError.php?buildid=662460

```
Error while building C++ shared library " ../../lib/libcmr.3.6.1.dylib" in target cmr
Command	
[+] "/usr/bin/clang++"
Directory	/Users/kitware/Dashboards/Continuous/Slicer-0-build/DCMTK-build/dcmsr/libcmr

Undefined symbols for architecture x86_64:
  "DSRTree<DSRDocumentTreeNode>::getRoot() const", referenced from:
      vtable for TID1001_ObservationContext in tid1001.cc.o
      vtable for TID1204_LanguageOfContentItemAndDescendants in tid1204.cc.o
      vtable for TID1600_ImageLibrary in tid1600.cc.o
ld: symbol(s) not found for architecture x86_64
```

fedorov:
@michaelonken do you know OFFIS tested DCMTK on Mac SnowLeopard?

jcfr:
@fedorov Looking at the error on the factory now.

michaelonken:
From the Dashboard I can see it is darwin 14.4.0 which seems to belong to Mac OS X 10.10 "Yosemite". This is the only Apple platform tested. Probably you need a patch; I am happy to integrate it into DCMTK upstream.

jcfr:
The error happen when building the target `cmr`:

```
$ make cmr/fast  VERBOSE=1
make -f dcmsr/libcmr/CMakeFiles/cmr.dir/build.make dcmsr/libcmr/CMakeFiles/cmr.dir/build
Linking CXX shared library ../../lib/libcmr.dylib
cd /Users/kitware/Dashboards/Continuous/Slicer-0-build/DCMTK-build/dcmsr/libcmr && "/Users/kitware/Dashboards/Support/CMake 3.0.1.app/Contents/bin/cmake" -E cmake_link_script CMakeFiles/cmr.dir/link.txt --verbose=1
"/Users/kitware/Dashboards/Support/CMake 3.0.1.app/Contents/bin/ctest" --launch --target-name cmr --build-dir /Users/kitware/Dashboards/Continuous/Slicer-0-build/DCMTK-build/dcmsr/libcmr --output ../../lib/libcmr.3.6.1.dylib --target-type SHARED_LIBRARY --language CXX -- /usr/bin/clang++    -fvisibility=hidden -D_XOPEN_SOURCE_EXTENDED -D_BSD_SOURCE -D_BSD_COMPAT -D_OSF_SOURCE -D_DARWIN_C_SOURCE -fPIC -O3 -DNDEBUG -arch x86_64 -isysroot /Developer/SDKs/MacOSX10.6.sdk -mmacosx-version-min=10.6 -dynamiclib -Wl,-headerpad_max_install_names   -compatibility_version 7.0.0 -current_version 3.6.1 -o ../../lib/libcmr.3.6.1.dylib -install_name /Users/kitware/Dashboards/Continuous/Slicer-0-build/DCMTK-build/lib/libcmr.7.dylib CMakeFiles/cmr.dir/init.cc.o CMakeFiles/cmr.dir/logger.cc.o CMakeFiles/cmr.dir/srnumvl.cc.o CMakeFiles/cmr.dir/tid1001.cc.o CMakeFiles/cmr.dir/tid1204.cc.o CMakeFiles/cmr.dir/tid1600.cc.o CMakeFiles/cmr.dir/cid11.cc.o CMakeFiles/cmr.dir/cid29.cc.o CMakeFiles/cmr.dir/cid29e.cc.o CMakeFiles/cmr.dir/cid42.cc.o CMakeFiles/cmr.dir/cid100.cc.o CMakeFiles/cmr.dir/cid244.cc.o CMakeFiles/cmr.dir/cid244e.cc.o CMakeFiles/cmr.dir/cid4020.cc.o CMakeFiles/cmr.dir/cid4021.cc.o CMakeFiles/cmr.dir/cid4031.cc.o CMakeFiles/cmr.dir/cid4031e.cc.o CMakeFiles/cmr.dir/cid5000.cc.o CMakeFiles/cmr.dir/cid5001.cc.o CMakeFiles/cmr.dir/cid7021.cc.o CMakeFiles/cmr.dir/cid7445.cc.o CMakeFiles/cmr.dir/cid7452.cc.o CMakeFiles/cmr.dir/cid7453.cc.o CMakeFiles/cmr.dir/cid10013.cc.o CMakeFiles/cmr.dir/cid10013e.cc.o CMakeFiles/cmr.dir/cid10033.cc.o CMakeFiles/cmr.dir/cid10033e.cc.o ../../lib/libdcmsr.3.6.1.dylib ../../lib/libdcmimage.3.6.1.dylib ../../lib/libdcmimgle.3.6.1.dylib ../../lib/libdcmdata.3.6.1.dylib ../../lib/liboflog.3.6.1.dylib ../../lib/libofstd.3.6.1.dylib -lpthread 
Undefined symbols for architecture x86_64:
  "DSRTree<DSRDocumentTreeNode>::getRoot() const", referenced from:
      vtable for TID1001_ObservationContext in tid1001.cc.o
      vtable for TID1204_LanguageOfContentItemAndDescendants in tid1204.cc.o
      vtable for TID1600_ImageLibrary in tid1600.cc.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[1]: *** [lib/libcmr.3.6.1.dylib] Error 1
make: *** [cmr/fast] Error 2
```

```
$ /usr/bin/clang++ --version
clang version 3.1 (tags/RELEASE_31/final)
Target: x86_64-apple-darwin10.8.0
Thread model: posix
```

fedorov:
@jcfr yes, that's right, but the function is defined.

@michaelonken this seems to be some fancy C++ magic - do you have any idea what might be incompatible?

```
virtual T *getRoot() const;
...
template<typename T>
T *DSRTree<T>::getRoot() const
{
    return RootNode;
}
```

fedorov:
http://tiny-qiicr.dipr.partners.org:8080/source/xref/dcmtk/dcmsr/include/dcmtk/dcmsr/dsrtree.h#349

jcfr:
Here is a link accessible from outside: https://github.com/commontk/DCMTK/blob/8fc3fc0857101fb29ff2a5fb908143eee3267a69/dcmsr/include/dcmtk/dcmsr/dsrtree.h#L889-L893

michaelonken:
We saw the same error on 32 Bit Linux using Clang 3.0-6.2. We blamed it on an error in that clang version, later versions (obviously not all...) of clang we tested handled it flawlessly. So we decided not to fix it (unitl now:).

fedorov:
@jcfr sorry, here's the version that should be accessible from the outside (it is cross-referenced source code, might be easier to track the issue): http://tiny-qiicr.partners.org/source/xref/dcmtk/dcmsr/include/dcmtk/dcmsr/dsrtree.h#349

michaelonken:
One way to work around the issue is to build without libcmr; it's new and you don't use it so far.

fedorov:
True, but I did not enable it explicitly, so we will still need to patch dcmtk... Let's see if Jorg has any suggestions (I emailed him).

fedorov:
Adding @jriesmeier!

fedorov:
This is my configuration on which this code compiles fine:

```
$ clang++ --version
Apple LLVM version 7.0.0 (clang-700.0.72)
Target: x86_64-apple-darwin14.5.0
Thread model: posix
```

jcfr:
We could configure DCMTK explicitly setting the list of modules we would like to use:

```
-DDCMTK_MODULES:STRING="ofstd;oflog;dcmdata;dcmimgle;dcmimage;dcmjpeg;dcmjpls;dcmtls;dcmnet;dcmsign;dcmwlm;dcmqrdb;dcmpstat;dcmrt;dcmiod;dcmfg;dcmseg"
```
Alternatively we could also introduce an option named `DCMTK_DISABLED_MODULES` allowing to set a list of modules to disable ...

michaelonken:
Unfortunately libcmr is not a module, but more a sub library of the dcmsr module and thus is not directly configurable via CMake (there could be such an option, though; the same applies to the libi2d library inside dcmdata which Jörg used as a template in terms of code structure).

fedorov:
And to add to what Michael said, we would prefer not to disable dcmsr altogether. Let's wait a bit to see what dashboards show for win and linux, and see if there's any response from Jörg.

I will also investigate if I can set up a linux virtual machine with that version of clang to try to debug/test the issue ...

jcfr:
> linux virtual machine with that version of clang 

More details about the compiler used on MAcOSX are reported here: http://wiki.slicer.org/slicerWiki/index.php/Documentation/Nightly/Developers/Factory#factory.kitware

fedorov:
Not reproducible on Ubuntu 12.04 x86 with clang 3.0 :(((

```
[100%] Built target dcmsr
fedorov@ubuntu:~/Downloads/dcmtk-build$ clang++ --version
Ubuntu clang version 3.0-6ubuntu3 (tags/RELEASE_30/final) (based on LLVM 3.0)
Target: i386-pc-linux-gnu
Thread model: posix
fedorov@ubuntu:~/Downloads/dcmtk-build$ uname -a
Linux ubuntu 3.13.0-32-generic #57~precise1-Ubuntu SMP Tue Jul 15 03:50:54 UTC 2014 i686 i686 i386 GNU/Linux
```

jcfr:
On Linux:

```
$ nm ./dcmsr/libcmr/CMakeFiles/cmr.dir/tid1001.cc.o  | grep getRoot | c++filt 
0000000000000000 W DSRTree<DSRDocumentTreeNode>::getRoot() const
```

On MacOSX:

```
$ nm ./dcmsr/libcmr/CMakeFiles/cmr.dir/tid1001.cc.o  | grep getRoot | c++filt 
                 U DSRTree<DSRDocumentTreeNode>::getRoot() const
```

jcfr:
Fixed in [r24587](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24587)

@michaelonken @jriesmeier The associated patch to integrate in DCMTK master is commontk/DCMTK@ade7e93

fedorov:
:+1:  thank you JC!

jriesmeier:
The library "cmr" (Content Mapping Resource) is an optional extension to the DCMTK module "dcmsr" (DICOM Structured Reporting), which adds support for specific Templates and Context Groups, e.g. those that are needed for QIICR. Optional means that not all programs using "dcmsr" also need to link "cmr". In my opinion, it makes no sense to introduce another (CMake) configuration option to enable/disable compilation of this submodule.

With regard to the compilation issue on Mac OS X: as Michael already wrote, we experienced the same problem with Clang 3.0.x (and "share libs" enabled) - but only with this particular version. Newer versions of Clang (e.g. 3.3 and 3.4 I've installed temporarily on my Linux machine) work as expected. Also Clang 3.0-6.2 with static linkage works out of the box (according to our Dashboard). Not sure whether this is an issue with the source code or with the compiler/linker.

Since our own Dashboard does not yet allow us to start a compilation process on a particular machine with a particular version of the source code, I don't know how I/we should find a solution. I don't want to have "test commits" in the DCMTK repository and we don't want to have "merge commits" either.

Any other suggestion?

jriesmeier:
@jcfr Done. Now in DCMTK's "testing" branch. Will show up in "master" probably tomorrow after the nightly build on all other tested platforms was successful (or at least not worse ;).

Thanks!

jcfr:
Great. Thanks.
Jc

fedorov:
@michaelonken @jriesmeier there is another set of errors with VS2008: http://trunk.cdash.org/buildSummary.php?buildid=663433

```
39>6>12>C:\D\N\Slicer-1-build\DCMTK\ofstd\include\dcmtk/ofstd/oftypes.h(111) : error C2146: syntax error : missing ';' before identifier 'Sint64'
39>6>12>C:\D\N\Slicer-1-build\DCMTK\ofstd\include\dcmtk/ofstd/oftypes.h(111) : error C4430: missing type specifier - int assumed. Note: C++ does not support default-int
39>6>12>C:\D\N\Slicer-1-build\DCMTK\ofstd\include\dcmtk/ofstd/oftypes.h(111) : error C4430: missing type specifier - int assumed. Note: C++ does not support default-int
39>6>12>C:\D\N\Slicer-1-build\DCMTK\ofstd\include\dcmtk/ofstd/oftypes.h(125) : error C2146: syntax error : missing ';' before identifier 'Uint64'
39>6>12>C:\D\N\Slicer-1-build\DCMTK\ofstd\include\dcmtk/ofstd/oftypes.h(125) : error C4430: missing type specifier - int assumed. Note: C++ does not support default-int
39>6>12>C:\D\N\Slicer-1-build\DCMTK\ofstd\include\dcmtk/ofstd/oftypes.h(125) : error C4430: missing type specifier - int assumed. Note: C++ does not support default-int
39>6>12>C:\D\N\Slicer-1-build\DCMTK\ofstd\include\dcmtk/ofstd/oftypes.h(145) : error C2146: syntax error : missing ';' before identifier 'OFintptr_t'
39>6>12>C:\D\N\Slicer-1-build\DCMTK\ofstd\include\dcmtk/ofstd/oftypes.h(145) : error C4430: missing type specifier - int assumed. Note: C++ does not support default-int
39>6>12>C:\D\N\Slicer-1-build\DCMTK\ofstd\include\dcmtk/ofstd/oftypes.h(145) : error C2378: 'Sint64' : redefinition; symbol cannot be overloaded with a typedef
39>6>12>        C:\D\N\Slicer-1-build\DCMTK\ofstd\include\dcmtk/ofstd/oftypes.h(111) : see declaration of 'Sint64'
39>6>12>C:\D\N\Slicer-1-build\DCMTK\ofstd\include\dcmtk/ofstd/oftypes.h(145) : error C4430: missing type specifier - int assumed. Note: C++ does not support default-int
39>6>12>C:\D\N\Slicer-1-build\DCMTK\ofstd\include\dcmtk/ofstd/oftypes.h(150) : error C2146: syntax error : missing ';' before identifier 'OFuintptr_t'
39>6>12>C:\D\N\Slicer-1-build\DCMTK\ofstd\include\dcmtk/ofstd/oftypes.h(150) : error C4430: missing type specifier - int assumed. Note: C++ does not support default-int
39>6>12>C:\D\N\Slicer-1-build\DCMTK\ofstd\include\dcmtk/ofstd/oftypes.h(150) : error C2378: 'Uint64' : redefinition; symbol cannot be overloaded with a typedef

<many more>
```

fedorov:
@jcfr do you have a suggestion how to proceed with debugging this issue? I do not have a system set up with a VS2008 compiler to work on a patch.

Do we have to revert this commit until it is confirmed DCMTK can be compiled with VS2008?

jriesmeier:
In fact, DCMTK - also the latest snapshot - can be compiled with VS 2008 (according to DCMTK'S Dashboard). Maybe, your problems result from a mixture of different source code packages? Something that is related to the particular CMake configuration or the superbuild feature, which is not used on DCMTK's Dashboard?

fedorov:
@jriesmeier these look like purely DCMTK compile errors. I can't think of a DCMTK dependency that is compiled by Slicer superbuild. However, maybe it works differently on the Slicer factory because we use different build options? The ones used in Slicer are here: https://github.com/Slicer/Slicer/blob/master/SuperBuild/External_DCMTK.cmake#L39-L67

Looking at my mac build, ```DCMTK_CXX11_FLAGS``` is set to ```-std=c++11``` -- is this how you tested the build on the OFFIS factories?

@jcfr are there any more detailed options about VS2008 compiler configuration that can be relevant? I looked at http://wiki.slicer.org/slicerWiki/index.php/Documentation/Nightly/Developers/Factory#VM:_Windows_7, but did not find much detail.

jcfr:
Interestingly, going in the build directory on the factory and clicking on "Build all", it completes without error. See below.

I suspect this is an error happening only when doing a clean build tree ... will report back.

```
1>------ Skipped Build: Project: RUN_TESTS, Configuration: Release x64 ------
1>Project not selected to build for this solution configuration 
2>------ Skipped Build: Project: mkvrscan, Configuration: Release x64 ------
2>Project not selected to build for this solution configuration 
3>------ Build started: Project: mkdictbi, Configuration: Release x64 ------
4>------ Build started: Project: mkdeftag, Configuration: Release x64 ------
4>Compiling...
4>dctypes.cc
4>cmdlnarg.cc
4>dchashdi.cc
4>dcvr.cc
4>dcdict.cc
4>dcdicent.cc
5>------ Skipped Build: Project: INSTALL, Configuration: Release x64 ------
5>Project not selected to build for this solution configuration 
4>dctagkey.cc
4>mkdeftag.cc
4>Performing Pre-Link Event...
3>Compiling...
4>Linking...
3>dctypes.cc
3>cmdlnarg.cc
3>dchashdi.cc
3>dcvr.cc
4>Embedding manifest...
4>Build log was saved at "file://c:\D\N\Slicer-1-build\DCMTK-build\dcmdata\libsrc\mkdeftag.dir\Release\BuildLog.htm"
4>mkdeftag - 0 error(s), 0 warning(s)
6>------ Skipped Build: Project: updatedeftag, Configuration: Release x64 ------
6>Project not selected to build for this solution configuration 
3>dcdict.cc
3>dcdicent.cc
3>dctagkey.cc
3>mkdictbi.cc
3>Performing Pre-Link Event...
3>Linking...
3>Embedding manifest...
3>Build log was saved at "file://c:\D\N\Slicer-1-build\DCMTK-build\dcmdata\libsrc\mkdictbi.dir\Release\BuildLog.htm"
3>mkdictbi - 0 error(s), 0 warning(s)
7>------ Skipped Build: Project: updatebuiltindict, Configuration: Release x64 ------
7>Project not selected to build for this solution configuration 
========== Build: 2 succeeded, 0 failed, 88 up-to-date, 5 skipped ==========
```

lassoan:
The error happens when you build the target CTKDICOMCorePythonQt, not when building DCMTK.

32>3>3>------ Build started: Project: CTKDICOMCorePythonQt, Configuration: Release x64 ------
32>3>3>Compiling...
32>3>2>   Creating library F:\SAB\Slicer-Nightly-build\CTK-build\CTK-build\bin\Release\CTKVisualizationVTKCore.lib and object F:\SAB\Slicer-Nightly-build\CTK-build\CTK-build\bin\Release\CTKVisualizationVTKCore.exp
32>3>3>org_commontk_CTKDICOMCore_init.cpp
32>3>1>moc_ctkPopupWidget_p.cxx
32>3>3>F:\SAB\Slicer-Nightly-build\DCMTK-build\config\include\dcmtk/config/osconfig.h(175) : warning C4005: 'HAVE_FTIME' : macro redefinition
32>3>3>        f:\sab\slicer-nightly-build\python-install\include\pyconfig.h(582) : see previous definition of 'HAVE_FTIME'
32>3>3>F:\SAB\Slicer-Nightly-build\DCMTK-build\config\include\dcmtk/config/osconfig.h(217) : warning C4005: 'HAVE_GETPID' : macro redefinition
31>10>sitkAntiAliasBinaryImageFilter.cxx
32>3>3>        f:\sab\slicer-nightly-build\python-install\include\pyconfig.h(593) : see previous definition of 'HAVE_GETPID'
31>7>Build log was saved at "file://f:\SAB\Slicer-Nightly-build\SimpleITK-build\Code\BasicFilters\src\SimpleITKBasicFilters5.dir\Release\BuildLog.htm"
32>3>3>F:\SAB\Slicer-Nightly-build\DCMTK-build\config\include\dcmtk/config/osconfig.h(274) : warning C4005: 'HAVE_IO_H' : macro redefinition
31>7>SimpleITKBasicFilters5 - 3 error(s), 0 warning(s)
32>3>3>        f:\sab\slicer-nightly-build\python-install\include\pyconfig.h(58) : see previous definition of 'HAVE_IO_H'
31>9>Generating Code...
32>3>3>F:\SAB\Slicer-Nightly-build\DCMTK-build\config\include\dcmtk/config/osconfig.h(622) : warning C4005: 'HAVE_STRERROR' : macro redefinition
32>3>3>        f:\sab\slicer-nightly-build\python-install\include\pyconfig.h(64) : see previous definition of 'HAVE_STRERROR'
32>3>3>F:\SAB\Slicer-Nightly-build\DCMTK-build\config\include\dcmtk/config/osconfig.h(718) : warning C4005: 'HAVE_SYS_UTIME_H' : macro redefinition
32>3>3>        f:\sab\slicer-nightly-build\python-install\include\pyconfig.h(59) : see previous definition of 'HAVE_SYS_UTIME_H'
32>3>3>F:\SAB\Slicer-Nightly-build\DCMTK-build\config\include\dcmtk/config/osconfig.h(727) : warning C4005: 'HAVE_TEMPNAM' : macro redefinition
32>3>3>        f:\sab\slicer-nightly-build\python-install\include\pyconfig.h(60) : see previous definition of 'HAVE_TEMPNAM'
32>3>3>F:\SAB\Slicer-Nightly-build\DCMTK-build\config\include\dcmtk/config/osconfig.h(736) : warning C4005: 'HAVE_TMPNAM' : macro redefinition
32>3>3>        f:\sab\slicer-nightly-build\python-install\include\pyconfig.h(62) : see previous definition of 'HAVE_TMPNAM'
32>3>3>F:\SAB\Slicer-Nightly-build\DCMTK-build\config\include\dcmtk/config/osconfig.h(974) : warning C4005: 'HAVE_LONG_LONG' : macro redefinition
32>3>3>        f:\sab\slicer-nightly-build\python-install\include\pyconfig.h(304) : see previous definition of 'HAVE_LONG_LONG'
32>3>1>moc_ctkPathListButtonsWidget.cxx
32>3>3>F:\SAB\Slicer-Nightly-build\DCMTK\ofstd\include\dcmtk/ofstd/oftypes.h(111) : error C2146: syntax error : missing ';' before identifier 'Sint64'
32>3>3>F:\SAB\Slicer-Nightly-build\DCMTK\ofstd\include\dcmtk/ofstd/oftypes.h(111) : error C4430: missing type specifier - int assumed. Note: C++ does not support default-int


By the way, according to the change log, it's been removed from DCMTK long time ago (see CHANGES.360):
- Removed Sint64 and Uint64 since there is no 64bit int available everywhere.
  Affects: ofstd/include/dcmtk/ofstd/oftypes.h

fedorov:
Ah! Thanks a lot Andras! :+1:  

This is a partial relief :)

jriesmeier:
> Looking at my mac build, DCMTK_CXX11_FLAGS is set to -std=c++11 -- is this how you tested the build on the OFFIS factories?"

@fedorov No, I don't think so, and I personally would not use DCMTK with "C++11" support enabled -- at least not, until the implementation proved to work on more than a few systems. This is still "work in progress" or "experimental", whatever you'd like to call it. I'm not the author of these things.

lassoan:
The issue is that int64_t is not defined and it is expected because HAVE_INT64_T is defined by PythonQt.h (I don't know why). We can fix the build issue by adding a #undef HAVE_INT64_T in CTK\CMake\ctkWrapPythonQt.py to:
...
#include <PythonQt.h>
// XXX Avoid  warning: "HAVE_STAT" redefined
#undef HAVE_STAT
#undef HAVE_INT64_T
#include "${namespace}_${target}.h" 
...

Note that there is already a similar hack there for HAVE_STAT.

jcfr:
Few updates:

1. About the build error detailed by @lassoan  previous comment, the macro `HAVE_INT64_T` is defined as a side-effect of including `python.h`. A workaround would be undefined the macro. More about the macro in the next bullet point.

2. regarding the redefined warnings. I have a PR in CTK to workaround these. See https://github.com/commontk/CTK/pull/605. This approach is not sustainable since we have to replicate the workaround in few other location ...   @jriesmeier @michaelonken  Would it make sense to update [CMake/osconfig.h.in](https://github.com/commontk/DCMTK/blob/patched-DCMTK-3.6.1_20150924/CMake/osconfig.h.in) (and eventually [config/include/dcmtk/config/osconfig.h.in](https://github.com/commontk/DCMTK/blob/patched-DCMTK-3.6.1_20150924/config/include/dcmtk/config/osconfig.h.in) undefining the macro if it as already been defined ? `osconfig.h` would basically include a private header named `ospreconfig.h` that would undef the macro. 

3. about cxx11 -  the current CMake code does not seem to give the option to disable the support. It will systematically enable it compiler support it. See [here](https://github.com/commontk/DCMTK/blob/ade7e936e362ba219df0e982d07807c6c7b73881/CMake/dcmtkPrepare.cmake#L436)  and [here](https://github.com/commontk/DCMTK/blob/ade7e936e362ba219df0e982d07807c6c7b73881/CMake/dcmtkPrepare.cmake#L399-L411)

jriesmeier:
@jcfr Documentation for C++11 support in DCMTK is here: http://support.dcmtk.org/docs-snapshot/cxx11_support.html

Unless you activate DCMTK_USE_CXX11_STL, C++11 support should be disabled (even though the compiler supports it). The naming conventions of all these xxx_USE_xxx, xxx_WITH_xxx options will probably change since they do not yet follow the DCMTK conventions.

Regarding 2. Could you please send your proposal to "bugs/at/dcmtk/dot/org" so that the other DCMTK development team members can also have a look at it?

lassoan:
Long-term solution may be to rename HAVE_INT64_T in DCMTK (to something more specific, such as DCMTK_HAVE_INT64_T), because undefining HAVE_INT64_T may cause side effects elsewhere in the software (for example it may break PythonQt).

jcfr:
@fedorov Having the variable `DCMTK_CXX11_FLAGS` set is independent of having C++11 support enabled, if you didn't explicitly enabled the option `DCMTK_USE_CXX11_STL`, there are no problem on that said.

fedorov:
@jcfr thank you for the clarification. No, I didn't enable anything manually. I just wanted to mention the settings superbuild set for DCMTK on my machine automatically.

jcfr:
@fedorov Makes sense.  While we are at it, could you also check that (1) `DCMTK_USE_CXX11_STL` is disabled in the CMakeCache.txt and (2) `CMAKE_CXX_FLAGS` does NOT contain `-std=c++11` ?

fedorov:
> @fedorov Makes sense. While we are at it, could you also check that (1) DCMTK_USE_CXX11_STL 
> is disabled in the CMakeCache.txt and (2) CMAKE_CXX_FLAGS does NOT contain -std=c++11 ?

Yes, I confirm DCMTK_USE_CXX11_STL is disabled in the CMakeCache.txt and CMAKE_CXX_FLAGS does NOT contain -std=c++11.

michaelonken:
To my understanding Python.h (or the included pyconfig.h?) is setting HAVE_INT64_T wrongly, since it does NOT seem to be supported on the system, right? So is that a problem with the python package used? 

I then would not work around it in DCMTK but in the code using Python.h and report that to the python package maintainers if not already fixed in later versions. 

I think renaming the macro in DCMTK would solve the problem, sure, but actually the macro name uses the common convention (?) for type checks with the idea that one can reuse them in code once detected. Actually the macro documentation in pyconfig.h says it should exactly signal whether int64_t is available or not.

lassoan:
Both Python's and DCMTK's use of HAVE_INT64_T is valid they are just different. If you do a quick web search you can see that HAVE_INT64_T is used by many software packages and almost all of them compute the value differently. The best is to keep yourself away from this name.

jcfr:
Issue has been fixed in [r24588](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24588)

That said, I still think DCMTK should be fixed. Here is the proposal I would  like to share on "bugs/at/dcmtk/dot/org":

Relying on macro starting with `HAVE_` in public header cause problem when DCMTK is integrated with other libraries doing the same (ie python). In addition to have undesired "undefined warning" (see commontk/CTK@7aaf1ac), it can also cause build error because the meaning associated with a given macro can be different across libraries (See commontk/CTK@281ffb7 and [here](https://github.com/Slicer/Slicer/pull/359#issuecomment-144832497))

The plan would be the following:
* Update all DCMTK macros to be prefixed with `DCMTK_`
* By default, the original `HAVE_XXXX` will also be defined to ensure code relying on them can still build.
* An option named `USE_GENERIC_HAVE_MACROS` (or similar name) will be added to allow project to build with/without setting the `HAVE_XXXX` macros.

@Slicer/slicer-core-team  Let me know what you think. Thanks

fedorov:
@jcfr having read the discussion - yes, what you propose makes sense to me. I don't think it is feasible to expect different packages to agree on a consistent definition of ```HAVE_XXXX```.

pieper:
+1

michaelonken:
Thanks for the fix jc and clarification, Andras. We discuss the issue with the dcmtk team and see and let you know what we do about it.

fedorov:
@pieper reported a problem below. Does this make we cannot make both new and old mac compilers happy?

https://github.com/Slicer/Slicer/commit/d83b7edd805638e127235b4add9441f6e0a2a5df#commitcomment-13598280

> Just updated Slicer on my Mac.
> This commit
https://github.com/Slicer/Slicer/commit/d83b7edd805638e127235b4add9441f6e0a2a5df

```
Produces this error
Linking CXX shared library ../../lib/libdcmsr.dylib
ld: duplicate symbol DSRTree<DSRDocumentTreeNode>::isEmpty() const  in
CMakeFiles/dcmsr.dir/dsrdocst.cc.o and
CMakeFiles/dcmsr.dir/dsrdoc.cc.o for architecture x86_64
collect2: ld returned 1 exit status
make[2]: *** [lib/libdcmsr.3.6.1.dylib] Error 1
make[1]: *** [dcmsr/libsrc/CMakeFiles/dcmsr.dir/all] Error 2

My gcc compiler version is:
i686-apple-darwin11-llvm-gcc-4.2 (GCC) 4.2.1 (Based on Apple Inc.
build 5658) (LLVM build 2336.1.00)
```

jcfr:
@fedorov It should be possible to keep both compiler happy. Let's move the discussion here: https://github.com/commontk/DCMTK/commit/ade7e936e362ba219df0e982d07807c6c7b73881#commitcomment-13535781

michaelonken:
We sat together with the DCMTK team today and also discussed the macro naming issue. As a result, we'll keep the HAVE_... macro names for now as they are, since this is the only time (we remember) a problem with them occurred. However we see it could make sense to shift them to their own "namespace" in the future, so we added that as a [feature ticket to the DCMTK issue tracker](http://support.dcmtk.org/redmine/issues/668).

In this particular case, I'm (stil) not quite sure that the #define in the python header is actually correct (related to its own documentation) but it doesn't matter for the general case, anyway ;)

jcfr:
Hi Michael,

Thanks for the update. It all makes sense.

On my side, I will discuss the HAVE_* issue with the CMake team to get
their feedback.

Note that I will be travelling this coming week and will provide some more
feedback when I return.

Thanks
Jc

On Fri, Oct 9, 2015 at 11:49 AM, Michael Onken <notifications@github.com>
wrote:

> We sat together with the DCMTK team today and also discussed the macro
> naming issue. As a result, we'll keep the HAVE_... macro names for now as
> they are, since this is the only time (we remember) a problem with them
> occurred. However we see it could make sense to shift them to their own
> "namespace" in the future, so we added that as a feature ticket to the
> DCMTK issue tracker <http://support.dcmtk.org/redmine/issues/668>.
>
> In this particular case, I'm (stil) not quite sure that the #define in the
> python header is actually correct (related to its own documentation) but it
> doesn't matter for the general case, anyway ;)
>
> —
> <https://github.com/Slicer/Slicer/pull/359#issuecomment-146908573>.
>



-- 
+1 919 869 8849

PR #360, opened 2015-09-30
ENH: Add method cli.runSync

jcfr:
Add method cli.runSync

This is a work-in-progress, test still need to be added.

### DIFF
--- a/Base/Python/slicer/cli.py
+++ b/Base/Python/slicer/cli.py
@@ -46,6 +46,15 @@ def setNodeParameters(node, parameters):
     else:
       print "parameter ", key, " has unsupported type ", value.__class__.__name__
 
+def runSync(module, node=None, parameters=None, delete_temporary_files=True):
+  '''Run a CLI synchronously, optionally given a node with optional parameters,
+  returning the node (or the new one if created)
+  node: existing parameter node (None by default)
+  parameters: dictionary of parameters for cli (None by default)
+  delete_temporary_files: remove temp files created during exectuion (True by default)
+  '''
+  return run(module, node=node, parameters=parameters, wait_for_completion=True, delete_temporary_files=delete_temporary_files)
+
 def run(module, node = None, parameters = None, wait_for_completion = False, delete_temporary_files = True):
   '''Runs a CLI, optionally given a node with optional parameters, returning
   back the node (or the new one if created)
@@ -60,7 +69,7 @@ def run(module, node = None, parameters = None, wait_for_completion = False, del
   else:
     node = createNode(module, parameters)
     if not node:
-      return
+      return None
 
   logic = module.logic()
 


jcfr:
Integrated in [24661](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24661)

After loading `MRHead` sample data into Slicer, manually tested using:

```
slicer.cli.runSync(slicer.modules.gaussianblurimagefilter, parameters={'inputVolume':getNode('MRHead'), 'outputVolume':getNode('MRHead'), 'sigma':6})
```

lassoan:
The runSync method name is not very intuitive (only long-time Slicer programmers or hardcore programmers may be able to guess what it may mean). Instead, I think it would be much more clear to use 'run' (wait for completion) and 'runBackground' (don't wait for completion) names instead. I know that changing meaning of 'run' to wait for completion by default would mean a change in behavior, but I don't think it could cause any issue if 'run' only returned when the processing is completed (changing the default behavior from 'wait' to 'don't wait' would be problematic, but fortunately we don't need that).

jcfr:
Thanks for reviewing the code.

While I like the original idea of using the `Sync` suffix (similar to what is is done in `node.js`) proposed by @mhalle:

* `run` 
* `runSync`  
* `runAsync`

other suffix would work. May be:

* `run`
* `runInBackground`
* `runInForeground`

That said `Background` / `Foreground` are not adjective or adverb and it would mean the action verb `run`  would target the actual `Background` / `Foreground` noun. That is why I added the preposition `in`

> but I don't think it could cause any issue if 'run' only returned when the processing is completed
could
I guess it would _only_ prevent module from running in parallel.  

I don't feel comfortable changing the meaning before the coming release. 


Cc: @Slicer/slicer-core-team

lassoan:
Looking at this again, the only issue is that 'run' does not wait for completion by default. Nobody expects this behavior and so it causes trouble very often. It would be enough to just have a single 'run' method that waits for completion by default.
In the extremely rare case when somebody wants to run a CLI in the background it should be no problem to set wait_for_completion=False, as it would only be used by very advanced developers.

jcfr:
> the only issue is that 'run' does not wait for completion by default.

Exactly. 

Then should I revert my commit and wait after the `4.5` release to change the default ?

lassoan:
Yes, I think that would be a good solution.

pieper:
While I'm not fanatical about it, I tend to disagree with Andras - I think
keeping the "run" semantics unchanged is appropriate for backwards
compatibility, even if the initial default behavior isn't what we would
prefer now.  The "*Sync" convention, as Mike suggested, is well established
and understandable to any programmer who does a little research on the
topic.

lassoan:
Many slicer scripting users are very inexperienced programmers. Threads and parallelism may be foreign for more experienced Python programmers. I don't mind very much if runSync and runAsync are added but I don't think it's needed.

mhalle:
How much existing codebase uses this method and why/how?

Running it async allows monitoring of progress, right?

—Mike

jcfr:
> How much existing codebase uses this method and why/how?

I am collecting this information. I will get back later with more details.

> Running it async allows monitoring of progress, right?

Not only, it also allows to run CLIs in parallel.

vovythevov:
I think changing the flag default makes sense. Most Slicer users will want to run their CLI
synchronously.
Personally, it's the first thing I do when you start putting together a module (first make it work synchronously and then asynchronously).

I would push to add the runAsync function too if you add the runSync function.
I would expect to find either both (runSync/runASync) functions or none at all (and just rely on the wait_for_completion option).

dzenanz:
Most projects I work with use it. One example is [SlicerPET](http://public.kitware.com/Wiki/PET-CT).

Just changing the default might have confusing effect. I think it would be better to not do that. Either replace run by runSync and runASync or just add runSync.

pieper:
I like the idea of leaving run as-is with the existing default but then
adding both a runSync and a runAsync.

PR #362, opened 2015-10-02
ENH: Introduce autoDeleteSubjectHierarchyChildren property

jcfr:
Flag determining whether children of subject hierarchy nodes are automatically
deleted upon deleting a parent subject hierarchy node.

By default, a pop-up question asking the user to confirm the deletion of
children nodes will be shown.

### DIFF
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginLogic.cxx
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginLogic.cxx
@@ -66,6 +66,10 @@ public:
   /// Flag determining whether subject hierarchy nodes are automatically created upon
   /// adding a supported data node in the scene, or just when entering the module.
   bool AutoCreateSubjectHierarchy;
+
+  /// Flag determining whether subject hierarchy children nodes are automatically
+  /// deleted upon deleting a parent subject hierarchy node.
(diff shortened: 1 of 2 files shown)


jcfr:
Cc: @cpinter  

Let me know what you think.

cpinter:
I have no objections.

The pop-up question is already shown I think, so the only difference would be an additional "Yes always" and "No always" option right?
Also I recently added qSlicerSubjectHierarchySettingsPanel, so one more option can be added for this.

jcfr:
@cpinter To clarify, this commit does not add a new popup, the existing one will show up by default (`AutoDeleteSubjectHierarchyChildren == false`) and ask the user to confirm. If  `AutoDeleteSubjectHierarchyChildren` is `true`, nodes will be automatically deleted (the behaviour I want)

cpinter:
Sorry, Jc I didn't pay enough attention in the rush. I see the commit now. I think it would make sense to add the flag as an option in the settings panel.

jcfr:
Great. Will add the setting in a separate commit and integrate then. Thanks for reviewing.

cpinter:
Sounds good. Thanks!

jcfr:
Integrated in [r24609](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24609)  and [r24610](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24610)

PR #363, opened 2015-10-05
BUG: FixedParametersValueType -> FixedParametersType

hjmjohnson:
Incorrect copy-n-paste error for ITK after 4.8.1.

### DIFF
--- a/Libs/MRML/Core/vtkITKTransformConverter.h
+++ b/Libs/MRML/Core/vtkITKTransformConverter.h
@@ -623,7 +623,7 @@ template <typename BSplineTransformType> bool vtkITKTransformConverter::SetITKBS
   // * mesh spacing X, Y, Z
   // * mesh direction 3x3 matrix (first row, second row, third row)
 #if defined(ITK_FIXED_PARAMETERS_ARE_DOUBLE) // After 4.8.1
-  typename BSplineTransformType::FixedParametersValueType transformFixedParamsItk;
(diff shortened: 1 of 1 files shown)


jcfr:
@hjmjohnson Thanks for you contribution 

Which problem is this fixing ? How can be done to reproduce the problem fixed by this commit ?

Cc: @lassoan

hjmjohnson:
@Jcfr,

I testing a build with the newest version of ITK.  Basically preparing for future.

I have a few more patches as well.  These should all be backwards compatible for many many revisions of ITK.

Hans


From:  Jean-Christophe Fillion-Robin
Reply-To:  Slicer/Slicer
Date:  Monday, October 5, 2015 at 11:08 PM
To:  Slicer/Slicer
Cc:  Hans Johnson
Subject:  Re: [Slicer] BUG: FixedParametersValueType -> FixedParametersType (#363)


@hjmjohnson Thanks for you contribution 

Which problem is this fixing ? How can be done to reproduce the problem fixed by this commit ?

Cc: @lassoan

jcfr:
> preparing for future.

:+1: 

I tweaked the commit message and integrated in [r24605](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24605)

jcfr:
> I have a few more patches as well. These should all be backwards compatible for many many revisions of ITK.

This is great. Thanks for your help with Slicer.

PR #365, opened 2015-10-06
BUG: CR and LF replaced in URL when uploading extensions

fbudin69500:
CMake string variables can contain new lines [1][2]. The Slicer extension mechanism uses string CMake to create extension description files and to create the URL containing meta-data and the package file name to upload extensions to the Midas server. New lines need to be handled correctly, otherwise the user will receive a message that the Midas server did not accept the submission of the extension. Carriage Return (CR) and Line Feed (LF) need to be replace by equivalent URL-encoding characters.

[1] http://slicer-devel.65872.n3.nabble.com/Problem-with-the-factory-not-uploading-some-extensions-td4035069.html
[2] https://github.com/Slicer/ExtensionsIndex/pull/1112#issuecomment-145706433


jcfr:

Integrated in [24600](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24600)

PR #368, opened 2015-10-06
BUG: Remove codebase for old GrowCut

hjmjohnson:
Removed old grow cut based on conversation in:
https://www.na-mic.org/Mantis/view.php?id=2873

This patch set removes code that has been abandoned,

lassoan (developer)
2015-05-07 18:01
Yes, FastGrowCut works well. It may be also an option to retire the
basic GrowCut.

pieper  (administrator)
2015-05-07 19:08
Yes, the FastGrowCut developers have been suggesting that we could
retire the old GrowCut - it would be a good project for someone. I
believe they do not give exactly the same answer, so it might be an
option to include both versions and let the user decide which to use
(assuming the crash scenarios are excluded).


fedorov:
Considering that GrowCut is one of the main segmentation workhorses, with a number of publications that used the original version, I would not remove the original version from the codebase. Especially considering that testing is very very limited. At the very least, there should be a set of various examples that at least confirm the new version does not crash.

Of note, we have a bunch of (again, in my opinion) totally useless "legacy" modules that are still in the base code. And now we are going to retire something that is actually used and is working? No, my opinion is that it is not a good idea.

fedorov:
This suggested removal should also be announced on the user and developer list, to increase our chances of catching potential problem, and get an idea if people actually tried FastGrowCut. Based on a communication with a collaborator who is using regular GrowCut quite heavily, when asked about FastGrowCut he said something along the lines "yes, we heard about it, it should be faster", but they never actually tried it out.

jcfr:
@hjmjohnson Thanks for your help cleaning up the code. 

Seems we will have to get a little more feedback from the community before moving forward.

>  announced on the user and developer list

I will send an email on the list.

fedorov:
Another idea to consider, to smooth the transition, is to "promote" FastGrowCut to a regular Editor effect (just another button in the GUI) and remove the old one after some time if there are no complaints.

fedorov:
Publications I know of that evaluated/validated current GrowCut

Parmar C., Rios Velazquez E., Leijenaar R., Jermoumi M., Carvalho S., Mak R.H., Mitra S., Shankar B.U., Kikinis R., Haibe-Kains B., Lambin P., Aerts H.J.W.L. Robust Radiomics Feature Quantification using Semiautomatic Volumetric Segmentation. PLoS One. 2014 Jul 15;9(7):e102107. PMID: 25025374. PMCID: PMC4098900. http://www.spl.harvard.edu/publications/item/view/2631

Egger J., Kapur T., Fedorov A., Pieper S., Miller J.V., Veeraraghavan H., Freisleben B., Golby A.J., Nimsky C., Kikinis R. GBM Volumetry using the 3D Slicer Medical Image Computing Platform. Sci Rep. 2013 Mar 4;3:1364. PMID: 23455483. PMCID: PMC3586703. http://www.spl.harvard.edu/publications/item/view/2331

hjmjohnson:
This has been a Mantis bug for several years, and has routinely been discussed at NAMIC week.  I was attempting to find a way to clear the bug report using the recommendations of several people that have previously tried to make the code more robust.

I was just trying to help.  I don't use this feature becasue it routinly failed with garbage results, and it routinely caused hours of lost time.  I was trying to prevent other users from having such an unpleasant experience.

Please reject this patch set.

pieper:
I agree with Andrey and will add that I've discussed this with Allen and I
believe we agree that that "Fast"GrowCut should really be called something
like "Approximate"GrowCut since the first name implies the a faster
calculation of the same result while the latter is more descriptive of this
algorithm since it may give a different result.

The different result may be equally or even more useful than the original
GrowCut result, but if it is different then we can't simply swap one for
the other.

hjmjohnson:
This patch set is REJECTED!

Thanks all for the comments.  I've updated the mantis page.  I am closing that mantis ticket as it is in a terminal state now.  It will not be fixed for new problems, but it seems to work well in some important cases.

jcfr:
> we agree that that "Fast"GrowCut should really be called something like "Approximate"GrowCut since 
> the first name implies the a faster calculation of the same result while the latter is more descriptive of this 
> algorithm since it may give a different result. 

Make sense. I created an issue to document the fact tests done using known input volume & expert ground truth should be added. See https://github.com/ljzhu/FastGrowCut/issues/4

The same data should be used with the current GrowCut.

> The different result may be equally or even more useful than the original GrowCut result, but if it is 
> different then we can't simply swap one for the other.

Indeed. Would it make sense to bundle FastGrowCut and GrowCut into the same extension ? 

What should we do regarding the failing GrowCut ? See https://www.na-mic.org/Mantis/view.php?id=2873

jcfr:
Note also that there are currently to implementation of FastGrowCut. See https://github.com/ljzhu/FastGrowCut/issues/5

pieper:
I agree - if we have two similar algorithms with similar functionality it
makes sense to put them together into the same tool.  I was thinking we
could have one GrowCut effect that would have a radio button picking the
implementation to use.  We can choose which one to have as the default
based on our testing of common use scenarios.

PR #370, opened 2015-10-07
STYLE: Update python script to use "is None" instead of "== None"

jcfr:
Rational copied from https://www.python.org/dev/peps/pep-0290/#testing-for-none

// -----------------
 Since there is only one None object, equality can be tested with identity. Identity tests are slightly faster than equality tests. Also, some object types may overload comparison, so equality testing may be much slower.

Pattern:

if v == None  -->  if v is None:
if v != None  -->  if v is not None:
// -----------------

### DIFF
--- a/Applications/SlicerApp/Testing/Python/SubjectHierarchyCorePluginsSelfTest.py
+++ b/Applications/SlicerApp/Testing/Python/SubjectHierarchyCorePluginsSelfTest.py
@@ -99,8 +99,8 @@ class SubjectHierarchyCorePluginsSelfTestTest(ScriptedLoadableModuleTest):
     # Make sure subject hierarchy auto-creation is on for this test
     subjectHierarchyWidget = slicer.modules.subjecthierarchy.widgetRepresentation()
     subjectHierarchyPluginLogic = subjectHierarchyWidget.pluginLogic()
-    self.assertTrue( subjectHierarchyWidget != None )
-    self.assertTrue( subjectHierarchyPluginLogic != None )
+    self.assertTrue( subjectHierarchyWidget is not None )
+    self.assertTrue( subjectHierarchyPluginLogic is not None )
     subjectHierarchyPluginLogic.autoCreateSubjectHierarchy = True
 
     # Set constants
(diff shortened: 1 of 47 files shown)


jcfr:
Cc: @Slicer/slicer-core-team

Applications/SlicerApp/Testing/Python/SubjectHierarchyCorePluginsSelfTest.py
```diff
@@ -126,7 +126,7 @@ def section_MarkupRole(self):
     studyNode = vtkMRMLSubjectHierarchyNode.CreateSubjectHierarchyNode(slicer.mrmlScene, patientNode, self.studyName, slicer.vtkMRMLSubjectHierarchyConstants.GetDICOMLevelStudy())
     markupsShNode = vtkMRMLSubjectHierarchyNode.CreateSubjectHierarchyNode(slicer.mrmlScene, studyNode, slicer.vtkMRMLSubjectHierarchyConstants.GetDICOMLevelSeries(), self.sampleMarkupName, markupsNode)
 
-    self.assertTrue( markupsShNode != None )
+    self.assertTrue( markupsShNode is not None )
```
msmolens: This could go further by writing ```self.assertIsNotNone(markupsShNode)```. But there are a lot of other cases, so this might beyond scope of this PR. For example, in the lines below, where we could write ```assertEqual(a, b)``` in place of ```assertTrue(a == b)```. Using the more specific functions usually gives more useful error messages.

cpinter:
+1

Also in case of numpy objects, "x == None" results in an error.

I agree with using the different varieties of assert. I did the same change recently, see
https://github.com/SlicerRt/SlicerRT/commit/8a33caa1c82292fa5e833d32f384a2e8f4c38434

jcfr:
Integrated in [r24758](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24758)

Assert changed to more specific one in [r24759](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24759)

PR #371, opened 2015-10-08
Fix vtk mrml double array storage node reader writer

jcfr:
The reader in `vtkMRMLDoubleArrayStorage` was always setting the
first line of values in the `.mcsv` file as labels. If the stored
`vtkMRMLDoubleArray` had not labels set, then the retrieved double
array labels would be set to the first line of numerical values,
associated with the first data point. This resulted in a missing
data point as well as shifted indices for all remaining points,
and incorrect labels.

The correction this commit offers is to write the line '# nolabels'
in the file if labels exist. If the reader finds this line, it will
set the first line of values as the first data point, instead of
setting it as label values.

Examples :

Here are four examples with (u,v,w) as labels, (12.1,425.577,-454)
for the first data point, and (8.79633e+09,0,-1) for the second one:

(1) New writer: vtkMRMLDoubleArrayNode with labels

```
   # measurement file filename.mcsv
   u,v,w
   12.1,425.577,-454
   8.79633e+09,0,-1
```

(2) New writer: vtkMRMLDoubleArrayNode with no labels set

```
   # measurement file filename.mcsv
   # nolabels
   12.1,425.577,-454
   8.79633e+09,0,-1
```

(3) Old writer: vtkMRMLDoubleArrayNode with labels

```
   # measurement file filename.mcsv
   # columns = x,y,yerr
   u,v,w
   12.1,425.577,-454
   8.79633e+09,0,-1
```

(4) Old writer: vtkMRMLDoubleArrayNode with no labels set

```
   # measurement file filename.mcsv
   # columns = x,y,yerr
   12.1,425.577,-454
   8.79633e+09,0,-1
```

Observations :

The line '# columns = x,y,yerr' was hardcoded [1] in the former
writer and had no meaning related to the actual double array
node labels. The inspiration problably came from .mcsv or .acsv
writers/readers [2][3] but had no effect here.

[1] https://github.com/Slicer/Slicer/blob/6e63b4bf5dbc65a9973160de7397dfc47882e038/Libs/MRML/Core/vtkMRMLDoubleArrayStorageNode.cxx#L244
[2] https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsFiducialStorageNode.cxx#L159-L185
[3] https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/Annotations/MRML/vtkMRMLAnnotationFiducialsStorageNode.cxx#L245-L268

### DIFF
--- a/Libs/MRML/Core/Testing/CMakeLists.txt
+++ b/Libs/MRML/Core/Testing/CMakeLists.txt
@@ -108,6 +108,7 @@ set_target_properties(${KIT}CxxTests PROPERTIES FOLDER ${${PROJECT_NAME}_FOLDER}
 
 #-----------------------------------------------------------------------------
 set(TEMP "${CMAKE_BINARY_DIR}/Testing/Temporary")
+set(DATAPATH "${CMAKE_CURRENT_SOURCE_DIR}/TestData")
 
 #-----------------------------------------------------------------------------
 simple_test( vtkMRMLBSplineTransformNodeTest1 )
@@ -127,7 +128,7 @@ simple_test( vtkMRMLDiffusionWeightedVolumeDisplayNodeTest1 )
 simple_test( vtkMRMLDiffusionWeightedVolumeNodeTest1 )
 simple_test( vtkMRMLDisplayableNodeTest1 )
 simple_test( vtkMRMLDisplayNodeTest1 )
-simple_test( vtkMRMLDoubleArrayNodeTest1 )
+simple_test( vtkMRMLDoubleArrayNodeTest1 ${TEMP} ${DATAPATH})
 simple_test( vtkMRMLFiducialListNodeTest1 )
 simple_test( vtkMRMLFiducialListStorageNodeTest1 )
 simple_test( vtkMRMLFreeSurferModelOverlayStorageNodeTest1 )
--- a/Libs/MRML/Core/Testing/TestData/vtkMRMLDoubleArrayNodeTest1_OldFile.mcsv
+++ b/Libs/MRML/Core/Testing/TestData/vtkMRMLDoubleArrayNodeTest1_OldFile.mcsv
@@ -1,0 +1,5 @@
+# measurement file /path/to/Slicer/Libs/MRML/Core/Testing/TestData/vtkMRMLDoubleArrayNodeTest1_OldFile.mcsv
+# columns = x,y,yerr
+x,y,yerr
+12.1,425.577,-454
+8.79633e+09,0,-1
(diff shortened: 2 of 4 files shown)


jcfr:
Cc: @Slicer/slicer-core-team  @naucoin

pieper:
It looks like this will be fully backward compatible, so +1.

jcfr:
@naucoin Since you did work related to `.mcsv` and `.fcsv` support,  would also be great to also get your feedback. Thanks

lassoan:
I think we should move to standard csv format (https://tools.ietf.org/html/rfc4180, https://en.wikipedia.org/wiki/Comma-separated_values). It would mean:
* Have a single header line and get rid of # prefixed lines.
* Use CSV file extension. To satisfy this requirement but at the same time keep the current way of determining file type solely based on the file name, we could use a file extension such as *.array.csv - resulting in filename such as mydataset.array.csv. Similarly we could have somedata.markup.csv, somethingelse.annotation.csv, myspreadsheet.table.csv.

If we need to store metadata for the CSV file then we can store it in the MRML scene file for now. In the future, we may decide to write a separate data set information file (see http://dataprotocols.org/tabular-data-package/).

Also note that with introduction of the new table node (https://github.com/lassoan/slicer/tree/3912-Add-new-vtkMRMLTableNode-and-corresponding-storagenode-classes, soon will be ready for integration), the vtkDoubleArray node will be mainly useful for management of long arrays that change frequently (for example, signal from a 1D sensor). Usual measurements, quantification results will be much more convenient to be stored in a table node, as in a table node you can have arbitrary number and type of columns, can be shown in a spreadsheet view, etc.

naucoin:
Double extensions might break a few things, I recently did a pass to try and fix a lot of bugs with the default extensions and writers and readers, with the nii.gz extension as my test case, and hopefully I caught everything.
The Markups storage nodes moved away from parsing a comment line for the field information (it's there for information/hand crafting the file), but the Annotation storage nodes still do that. There is metadata that is needed in the Markups file on disk to allow it to be loaded as a stand alone file rather than referenced from MRML - the coordinate system can't be inferred from the data. 
I'm willing to move forward to a more standard CSV format, but would prefer to leave the Slicer 3 fiducials format alone as there are already backward compatibility cases built in. The Annotations are actually vertical bar separated and I consider them legacy as well.

jcfr:
Thanks all for your feedback.

Integrated in [r24675](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24675) and [r24676](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24676)

PR #373, opened 2015-10-09
Fix slice view annotations

jcfr:
In addition to applying various style fixes, removing dead code and re-factoring the code to improve its readability, this PR (more specifically  c89bfdc) fixes SlicerProstate/SliceTracker#33 to avoid blank rectangle from showing up in SliceViewAnnotations

Cc: @fedorov @cpinter @mehrtash @che85 @pieper @Slicer/slicer-core-team

### DIFF
--- a/Modules/Scripted/DataProbe/DataProbeLib/SliceViewAnnotations.py
+++ b/Modules/Scripted/DataProbe/DataProbeLib/SliceViewAnnotations.py
@@ -475,32 +475,24 @@ class SliceAnnotations(VTKObservationMixin):
     self.scalarBarCollapsibleButton.enabled = enabled
     self.restorDefaultsButton.enabled = enabled
 
-    if enabled:
-      for sliceViewName in self.sliceViewNames:
-        sliceWidget = self.layoutManager.sliceWidget(sliceViewName)
-        sl = sliceWidget.sliceLogic()
+    for sliceViewName in self.sliceViewNames:
+      sliceWidget = self.layoutManager.sliceWidget(sliceViewName)
+      sl = sliceWidget.sliceLogic()
+      self.updateRuler(sl)
+      self.updateScalarBar(sl)
+      self.updateOrientationMarker(sl)
+
+      if enabled:
         self.makeAnnotationText(sl)
-        self.updateRuler(sl)
-        self.updateScalarBar(sl)
-        self.updateOrientationMarker(sl)
-    else:
-      # Remove Observers
-
-      for sliceViewName in self.sliceViewNames:
-        sliceWidget = self.layoutManager.sliceWidget(sliceViewName)
-        sl = sliceWidget.sliceLogic()
-        self.updateRuler(sl)
-        self.updateScalarBar(sl)
-        self.updateOrientationMarker(sl)
-
-      # Clear Annotations
-      for sliceViewName in self.sliceViewNames:
+      else:
+        # Clear Annotations
         self.sliceCornerAnnotations[sliceViewName].SetText(0, "")
         self.sliceCornerAnnotations[sliceViewName].SetText(1, "")
         self.sliceCornerAnnotations[sliceViewName].SetText(2, "")
         self.sliceCornerAnnotations[sliceViewName].SetText(3, "")
         self.sliceViews[sliceViewName].scheduleRender()
 
+    if not enabled:
       # reset global variables
       self.sliceCornerAnnotations = {}
 
--- a/Modules/Scripted/DataProbe/DataProbeLib/SliceViewAnnotations.py
+++ b/Modules/Scripted/DataProbe/DataProbeLib/SliceViewAnnotations.py
@@ -1111,7 +1111,8 @@ class SliceAnnotations(VTKObservationMixin):
 
       self.drawCornerAnnotations()
 
-  def updateScalarBarRange(self, sliceLogic, volumeNode, scalarBar, selectedLayer):
+  @staticmethod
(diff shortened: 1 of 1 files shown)


lassoan:
Thanks for the fixes, Jc. They are helpful, but unfortunately, sooner or later we have to clean up this module much more thoroughly. The most important design change that is absolutely essential is to properly use displayable managers to show the actors. Also, such core module would greatly benefit from being implemented in C++, so I would not invest too much time trying to fix this Python implementation.

naucoin:
I'd reported this as a bug as well:
http://www.na-mic.org/Bug/view.php?id=4043

fedorov:
@jcfr thank you for working on this!

@lassoan 

> The most important design change that is absolutely essential is to properly use displayable managers to show the actors.

is it possible to properly use displayable managers from python? I have not developed anything at that level

> Also, such core module would greatly benefit from being implemented in C++

Can you explain why? There are quite a few core modules implemented in python. My opinion is that we should encourage python as much as feasible at different levels, unless C++ is absolutely necessary, to attract more participation from the community in contributing to the code base.

jcfr:
> is it possible to properly use displayable managers from python? I have not developed anything at that level

There are some attempt to create a `ScriptedDisplayableManager`

That said, before moving in that direction, we should refactor the existing displayable to reuse a base/generic manager abstracting the management.

> core module would greatly benefit from being implemented in C++. Can you explain why? 

Here is my rational:

Since the mechanism to properly implement such core feature is not exposed in python, and there are much work required to provide a pythonic, maintainable and tested interface. Implementing DisplayableManager in c++ is the recommended way for now.

fedorov:
>  Implementing DisplayableManager in c++ is the recommended way for now.

Makes sense.

From my side, as far as implementing this in C++, I cannot prioritize this work given the outstanding tasks on the funded projects. If there is a serious concern that this functionality does not follow the recommended implementation approach, we should move it out of the core into extension.

As aside, the pull requests implementing this functionality underwent the standard procedures and developer community review. You can see from the history, @mehrtash invested a lot of time into implementing this: https://github.com/Slicer/Slicer/pulls?utf8=%E2%9C%93&q=is%3Apr+mehrtash+annotations+is%3Aclosed+. I would estimate this took between 3 and 6 month of dedicated effort to get it to where this is right now.

lassoan:
There are many reasons for implementing Slicer core in C++, I just list a few reasons that are relevant for this specific case:
* C++ rigor and extensive compile-time checks increases the chance of catching errors earlier. C++ enforces many things that in Python are managed by good programming practices (that are not enforced and in general not followed by Slicer developers).
* C++ memory management is much more explicit, which is more difficult to learn than Python, but allows much better efficiency (you can prevent memory fragmentation, optimize memory reallocations, etc), which is useful for code that may run hundreds of times per second (this code does).
* There are lots of good examples of solid, fast C++ displayable managers. Python displayable managers are more at the experimental level.
* Accessibility of Python classes from C++ modules is cumbersome. Accessibility of C++ classes is very convenient from both Python and C++ modules.
* Some projects don't use Python and don't want to use it (Python brings in huge amount of additional code, which is a problem if you want to minimize the amount of code that goes through regulatory approval)
* Profiling and optimization of mixed C++/Python code is very difficult (both C++ and Python profilers treat the "other side" as black hole). Since all the rest of the Slicer core is implemented in C++ (with a few small exceptions), it is much easier to implement this in C++, too.

I love Python programming, do it extensively (nowadays I spend much more with Python programming than with C++), but implementing Slicer core features in Python has just too many disadvantages.

jcfr:
> Re @lassoan 

Great summary. I will add your comment to the developer FAQ.


> @fedorov  as far as implementing this in C++, I cannot prioritize this work given the outstanding
> tasks on the funded projects. If there is a serious concern that this functionality does not follow the 
> recommended implementation approach, we should move it out of the core into extension.

There are no problem. The change will happen over time and there are no particular rush.


> As aside, the pull requests implementing this functionality underwent the standard procedures and 
> developer

Work of @mehrtash  remains very important and provide some very useful features. Considering perfect is the enemy of the good, integrating the feature made sense. We will _just_ to iterate over the foundation setup by @mehrtash and improve it together. :smile:

lassoan:
The Python implementation can serve as a prototype for the C++ implementation, so Alireza's efforts were very useful. It would probably take only about a week to port it to C++. It's on my list to get to it sometime, but I have only limited time for infrastructure work and this one is currently not on the critical path.

fedorov:
@lassoan I agree with your arguments, and the point about importance of performance in this instance is critical. I wish we had this discussion at the time Alireza started implementation. I am pretty sure the proposal was presented to the developer community at one of the weekly hangouts, and communicated via "labs" page in addition to the discussions of the PRs. But, good to know this transition to C++ will require small effort and Alireza's implementation will be useful.

jcfr:
> I wish we had this discussion

@fedorov @lassoan I added a new entry named `Should core feature be implemented in C++ or Python ?` in the Developer FAQ. 

See http://wiki.slicer.org/slicerWiki/index.php?title=Documentation/Nightly/Developers/FAQ

jcfr:
Integrated in [r24632](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24632) to [r24649](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24649)

pieper:
Hi Guys -

I added a preamble to the FAQ entry with some additional considerations.

-Steve

jcfr:
:+1:

PR #374, opened 2015-10-14
BUG: return value of "downloadExtension()" not always handled if 0

fbudin69500:
The function downloadAndInstallExtension(const QString& extensionId) did not handled the case of downloadExtension() returning 0. We used the same approach as in scheduleExtensionForUpdate(const QString& extensionName) to handle that case correctly in downloadAndInstallExtension(const QString& extensionId).

### DIFF
--- a/Base/QTCore/qSlicerExtensionsManagerModel.cxx
+++ b/Base/QTCore/qSlicerExtensionsManagerModel.cxx
@@ -1130,6 +1130,11 @@ void qSlicerExtensionsManagerModel::downloadAndInstallExtension(const QString& e
     }
 
   qSlicerExtensionDownloadTask* const task = d->downloadExtension(extensionId);
+  if (!task)
+  {
+    d->critical("Failed to retrieve metadata for extension " + extensionId);
+    return;
+  }
   connect(task, SIGNAL(finished(qSlicerExtensionDownloadTask*)),
           this, SLOT(onInstallDownloadFinished(qSlicerExtensionDownloadTask*)));
 }
--- a/Base/QTCore/qSlicerExtensionsManagerModel.cxx
+++ b/Base/QTCore/qSlicerExtensionsManagerModel.cxx
@@ -1355,14 +1355,18 @@ bool qSlicerExtensionsManagerModel::installExtension(
     if (result == QMessageBox::Yes)
       {
       // Install dependencies
+      QStringList extensionsNotInstalled ;
       foreach (const ExtensionMetadataType& dependency, dependenciesMetadata)
         {
-        bool res = this->downloadAndInstallExtension(dependency.value("extension_id").toString());
+        bool res = this->downloadAndInstallExtension(dependency.value("extension_id").toString() ) ;
         if( !res )
           {
-          QString msg = QString("<p>Error while installing %1</p><ul>").arg(dependency.value("extensionname").toString());
-          QMessageBox::warning(0, "Error installing dependencies", msg);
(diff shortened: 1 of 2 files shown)


lassoan:
What happens if downloadExtension fails? I see that an error is logged but probably the user should be notified, too (maybe there is a mechanism for that in place already).
I think it may be also useful to return with false in case of failure - it may be useful for detecting error when extension is installed programmatically (not directly by the user).

fbudin69500:
Good idea. In the other function that has the same call (scheduleExtensionForUpdate(const QString& extensionName)), it does return false. This function was a void function, and I did not want to change the function type without feedback, but there is no reason that I can think of not to make that function a bool.
Additionally, as I was browsing this file, I saw that there are some calls for "this->critical(...)" and other calls with "qCritical() << ...". I wonder it this should be made uniformed or if these calls are accomplishing two different tasks.

lassoan:
qCritical() only logs. this->critical might do more or something else - maybe it also displays the message on screen for the user?

fbudin69500:
I guess to know more precisely about the difference between qCritical() and this->critical() we may have to wait @jcfr 's input. I was just curious anyway.
Let me know if my last commit looks better. Thanks!

jcfr:
> I guess to know more precisely about the difference between qCritical() and this->critical() we may have to wait @jcfr 's input. I was just curious anyway.

Within the `qSlicerExtensionsManagerModel` using the dedicated log function like `this->critical`, `this->info`, ... ensures the signal `messageLogged` is emitted while logging message to the terminal. See [here](https://github.com/Slicer/Slicer/blob/87e81f68abd1c33451ede8573907077e48d6e761/Base/QTCore/qSlicerExtensionsManagerModel.cxx#L321-L342)

Then:
* within the `qSlicerExtensionsInstallWidget` a html-like dialog with a fading delay is shown to the user. See [here](https://github.com/Slicer/Slicer/blob/87e81f68abd1c33451ede8573907077e48d6e761/Base/QTGUI/qSlicerExtensionsInstallWidget.cxx#L274-L290). 
* within `qSlicerExtensionsManagerWidget`, a regular `QMessageBox` is shown. See [here](https://github.com/Slicer/Slicer/blob/87e81f68abd1c33451ede8573907077e48d6e761/Base/QTGUI/qSlicerExtensionsManagerWidget.cxx#L426-L442)

Base/QTCore/qSlicerExtensionsManagerModel.cxx
```diff
@@ -1351,7 +1357,12 @@ bool qSlicerExtensionsManagerModel::installExtension(
       // Install dependencies
       foreach (const ExtensionMetadataType& dependency, dependenciesMetadata)
         {
-        this->downloadAndInstallExtension(dependency.value("extension_id").toString());
+        bool res = this->downloadAndInstallExtension(dependency.value("extension_id").toString());
+        if( !res )
+          {
+          QString msg = QString("<p>Error while installing %1</p><ul>").arg(dependency.value("extensionname").toString());
+          QMessageBox::warning(0, "Error installing dependencies", msg);
```
jcfr: Instead of showing a `QMessageBox` for each dependency that failed to be installed. Could you instead accumulate the error in a `QStringList` and then call `this->critical(errors.join("\n"))` outside of the loop ? If not, the user will have to acknowledge the popup for each dependencies ...

fbudin69500:
@jcfr : It should do as you requested now.

jcfr:
Thanks for the update. :)

Tested and integrated in [r24666](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24666)

PR #375, opened 2015-10-14
ENH: Adaptor classes facilitating python scripted subject hierarchy p…

cpinter:
…lugins

Assembla ticket for this task:
https://www.assembla.com/spaces/slicerrt/tickets/614
Example plugin available for VolumeClip modules:
https://subversion.assembla.com/svn/slicerrt/trunk/VolumeClip/src

### DIFF
--- a/Modules/Loadable/SubjectHierarchy/Widgets/CMakeLists.txt
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/CMakeLists.txt
@@ -32,6 +32,8 @@ set(${KIT}_SRCS
   qSlicerSubjectHierarchyPluginHandler.h
   qSlicerSubjectHierarchyDefaultPlugin.cxx
   qSlicerSubjectHierarchyDefaultPlugin.h
+  qSlicerSubjectHierarchyScriptedPlugin.cxx
+  qSlicerSubjectHierarchyScriptedPlugin.h
   qSlicerSubjectHierarchyCloneNodePlugin.cxx
   qSlicerSubjectHierarchyCloneNodePlugin.h
   qSlicerSubjectHierarchyChartsPlugin.cxx
@@ -55,6 +57,7 @@ set(${KIT}_MOC_SRCS
   qSlicerSubjectHierarchyAbstractPlugin.h
   qSlicerSubjectHierarchyPluginHandler.h
   qSlicerSubjectHierarchyDefaultPlugin.h
+  qSlicerSubjectHierarchyScriptedPlugin.h
   qSlicerSubjectHierarchyCloneNodePlugin.h
   qSlicerSubjectHierarchyChartsPlugin.h
   qSlicerSubjectHierarchyParseLocalDataPlugin.h
@@ -101,8 +104,8 @@ SlicerMacroBuildModuleWidgets(
   )
 
 #-----------------------------------------------------------------------------
-if(BUILD_TESTING)
-  #add_subdirectory(Testing)
+if(Slicer_USE_PYTHONQT)
+  add_subdirectory(Python)
 endif()
 
 #-----------------------------------------------------------------------------
--- a/Modules/Loadable/SubjectHierarchy/Widgets/Python/AbstractScriptedSubjectHierarchyPlugin.py
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/Python/AbstractScriptedSubjectHierarchyPlugin.py
@@ -1,0 +1,36 @@
+import os
+import vtk, qt, ctk, slicer, logging
+
+#
+# Abstract class of python scripted subject hierarchy plugins
+#
+
+class AbstractScriptedSubjectHierarchyPlugin():
+  """ Abstract scripted subject hierarchy plugin for python scripted plugins
+
+      USAGE: Instantiate scripted subject hierarchy plugin adaptor class from
+        module (e.g. from setup function), and set python source:
+
+      from SubjectHierarchyPlugins import *
+      ...
+      class [Module]Widget(ScriptedLoadableModuleWidget):
+        ...
+        def setup(self):
+          ...
+          scriptedPlugin = slicer.qSlicerSubjectHierarchyScriptedPlugin(None)
+          scriptedPlugin.setPythonSource(VolumeClipSubjectHierarchyPlugin.filePath)
+          ...
+
+      Example can be found here: https://subversion.assembla.com/svn/slicerrt/trunk/VolumeClip/src
+  """
+
+  def __init__(self, scriptedPlugin):
+    self.scriptedPlugin = scriptedPlugin
+
+    # Register plugin on initialization
+    self.register()
+
+  def register(self):
+    pluginHandlerSingleton = slicer.qSlicerSubjectHierarchyPluginHandler.instance()
+    pluginHandlerSingleton.registerPlugin(self.scriptedPlugin)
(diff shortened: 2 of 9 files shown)


cpinter:
@jcfr @pieper Can you please take a look at this commit? It is the implementation of the python scripted subject hierarchy plugin that we talked about in SLC (https://www.assembla.com/spaces/slicerrt/tickets/614).
Thanks!

pieper:
@cpinter - nice!

Once question I can't tell from the code: are the plugins fully reloadable?  That is, I see the VolumeClip plugin has the python testing - is it possible to edit the logic of the plugin and then click Reload & Test?

cpinter:
@pieper Thanks for the feedback! I didn't consider reloading from GUI a requirement, as SH plugins can be independent of modules. Even the example plugin is shared between two (very similar) modules, although indeed most "role" type plugins are associated to one module. "Function" type plugins, however, are independent from any module. This is why the plugin instances are not directly accessible from modules. Of course I'm open to suggestions!

pieper:
Thanks for the description Csaba.  I guess this is a valid use case and I'm
glad you support it.

For my part, without the reload function I'm just as happy to write code in
C++.  Reloading is really a key motivation for using python (or tcl or
javascript) because it enables rapid development and debugging, for me at
least.  I don't know if it's easy to add to your current implementation.  I
know that in the past I wasn't able to reload with other C++ interfaces to
scripted plugins because they were too dependent on one-time initialization
behaviors.

lassoan:
The key advantage of Python plugins is that people can develop it who cannot build Slicer.

Reloading of these plugins would work the same way as the reloading of Python scripted module logic classes that the module manager instantiates at Slicer startup. We still have to figure out how to do that exactly.

pieper:
Yes, let's figure out the reloading part.  We have identified two key
features of writing plugins in python: no compiling and ability to reload.
Ideally we should support both and then copy that pattern to any plugin
situation.

cpinter:
I used to be able to reload files that were not reloaded by the reload button by default using the call
GelDosimetryAnalysisLogic = reload(GelDosimetryAnalysisLogic)
If this still works then we can use it for development. I don't know how important the GUI option is, but this could be a possibility.
Regarding the GUI, it would be not trivial to add reloading as we don't have a place where the plugins are listed (yet). Should there be such a list? If yes, where?

cpinter:
By the way the scripted SH plugins are really similar to DICOM plugins, which also don't have reload function yet as far as I know.

pieper:
It is something of a passion to get reloading working in python, but I've
set up landmark registration and editor effect plugins to work so utter can
be done.  It would be wonderful if this were a first class property the
language.

See def onReload(self) in
https://github.com/pieper/CommonGL/blob/master/Isobrush/Isobrush.py

cpinter:
I am not debating its utility. However I need further input, because it is not a straightforward issue. What do you think about my comments about
- whether it's needed to reload from GUI and the option of reloading from interactor (if you need it for development then running one command form the interactor should not be a blocking issue)
- lack of plugin and module correspondence (i.e. where to put the reload button)
?

fedorov:
Yes, I am curious about this too in the context of DICOM plugins. Steve, do you have a working example for a DICOM plugin, and an idea how that reload would be triggered?

cpinter:
It seems to be a rather complex question. Maybe we can talk about it at the hangout on Tuesday.

If there are no issues with this pull request then I'd integrate. Adding GUI reload if we decide to do it will be possible later.

fedorov:
I didn't review the code, I am not familiar enough with the details to give a good review. But I think it is not fair to hold integration of this feature due to the lack of reload functionality, since it is not clear how it would fit or be implemented, and also DICOM plugins (perhaps also Editor plugins?) are part of the core and don't have reload.

lassoan:
Yes, I completely agree, it should be merged as is and we can figure out reloading later.

cpinter:
Thanks! I'm merging this now and we can discuss the reloading on Tuesday.

@fedorov Can you attend? I have a vague memory that your schedule conflicts with the Slicer dev hangout.

cpinter:
I merged and two commits showed up. Should I have done something differently?

fedorov:
> Can you attend? I have a vague memory that your schedule conflicts with the Slicer dev hangout.

Yes, this overlaps with weekly meeting I have for our prostate group, so I usually cannot make it. This coming week I will be at a conference Mon-Wed.

> I merged and two commits showed up. Should I have done something differently?

Ouch.... Did you merge from github GUI? This is not the right procedure, since Slicer main repository is still SVN. See instructions in http://www.slicer.org/slicerWiki/index.php/Slicer:git-svn#Committing_back_to_svn. But this is a good stress test. I wonder what is the procedure now. Maybe you can just revert with the github button above, and then follow the wiki?

cpinter:
Thanks, I'll revert then. Sorry about this!

cpinter:
No, it's even worse now :( One additional commit, instead of actually reverting.

lassoan:
I think the git repository is just a mirror of the SVN repo. All changes should be committed into SVN. If you have a git repository with git-svn then you can use dcommit to somehow do the SVN commit. Jc uses this dcommit but I couldn't set it up on my computer, so I manually merge changes to my SVN working copy and commit from there.

cpinter:
I see. Is there anything I can do now? If I commit to SVN do you think it would override the git mess I made?

lassoan:
I would not touch the git repo, just use SVN. We'll see if the git repo is going to recover on its own on the next SVN sync or some manual fix is needed.

fedorov:
I think we can also do a hard reset of the master back to the revision before the first merge, but maybe better wait until @jcfr is back. At that point, we should document the proper procedure to follow in this situation.

cpinter:
I made the SVN commit. Let's see if git recovers...

cpinter:
It does not seem to.
@jcfr When you get back can you please help restoring the git repo? Sorry!

fedorov:
@cpinter @lassoan please let me know if you want me to do a hard reset, otherwise will wait for JC to return.

cpinter:
If it preserves the tags, then it would probably be a good solution I guess.

fedorov:
I think I will stay on the safe side. There's no high urgency, so let's
wait for JC. I am 99% sure it will work, but don't want to risk.

On Fri, Oct 16, 2015 at 2:09 PM Csaba Pinter <notifications@github.com>
wrote:

> If it preserves the tags, then it would probably be a good solution I
> guess.
>
> —
> <https://github.com/Slicer/Slicer/pull/375#issuecomment-148788495>.
>
-- 

Sent from on the go

pieper:
Hi from vacation -

I'll be happy to chat about all this Wednesday or later in the week.
Agreed with the merge and I'm sure the repo can get sorted out when Jc
returns.

Regards from Vermont!
Steve

cpinter:
It seems that the github repo has been successfully restored.

fedorov:
Whoever restored it - can you add instructions here and/or on the wiki how this should be properly done?

jcfr:
After fixing the problem and updating the scripts so that similar issue _auto-correct_ on their own. 

I just documented where and how the scripts are run on the [SVN2GitHub](http://wiki.slicer.org/slicerWiki/index.php/Documentation/Nightly/Developers/Build_system/SlicerBot/SVN2GitHub) page link of the [SlicerBot]( http://wiki.slicer.org/slicerWiki/index.php/Documentation/Nightly/Developers/Build_system/SlicerBot#SVN2GitHub_synchronization) page.

Cc: @msmolens

PR #377, opened 2015-10-15
BUG + ENH:fixing ROI visibility and adding two signals in qSlicerVolumeRenderingModuleWidget

Punzo:
I'd like to add a slot setDisplayROIEnabled in qSlicerVolumeRenderingModuleWidget. 
For me it is necessary to enable the ROI display from my widget.

animation

### DIFF
--- a/Modules/Loadable/Annotations/Widgets/qMRMLAnnotationROIWidget.cxx
+++ b/Modules/Loadable/Annotations/Widgets/qMRMLAnnotationROIWidget.cxx
@@ -26,6 +26,7 @@
 
 // MRML includes
 #include <vtkMRMLAnnotationROINode.h>
+#include <vtkMRMLDisplayNode.h>
 
 // 0.001 because the sliders only handle 2 decimals
 #define SLIDERS_EPSILON 0.001
@@ -100,11 +101,17 @@ vtkMRMLAnnotationROINode* qMRMLAnnotationROIWidget::mrmlROINode()const
 void qMRMLAnnotationROIWidget::setMRMLAnnotationROINode(vtkMRMLAnnotationROINode* roiNode)
 {
   Q_D(qMRMLAnnotationROIWidget);
-  qvtkReconnect(d->ROINode, roiNode, vtkCommand::ModifiedEvent,
+
+  this->qvtkReconnect(d->ROINode, roiNode, vtkCommand::ModifiedEvent,
                 this, SLOT(onMRMLNodeModified()));
 
+  this->qvtkReconnect(d->ROINode, roiNode, vtkMRMLDisplayableNode::DisplayModifiedEvent,
+                      this, SLOT(onMRMLDisplayNodeModified()));
+
   d->ROINode = roiNode;
+
   this->onMRMLNodeModified();
+  this->onMRMLDisplayNodeModified();
   this->setEnabled(roiNode != 0);
 }
 
@@ -125,9 +132,6 @@ void qMRMLAnnotationROIWidget::onMRMLNodeModified()
     }
 
   d->IsProcessingOnMRMLNodeModified = true;
-
-  // Visibility
-  d->DisplayClippingBoxButton->setChecked(d->ROINode->GetDisplayVisibility());
 
   // Interactive Mode
   bool interactive = d->ROINode->GetInteractiveMode();
@@ -191,8 +195,8 @@ void qMRMLAnnotationROIWidget::setExtent(double minLR, double maxLR,
 void qMRMLAnnotationROIWidget::setDisplayClippingBox(bool visible)
 {
   Q_D(qMRMLAnnotationROIWidget);
-  d->ROINode->SetDisplayVisibility(visible);
-  emit displayClippingBoxChanged(visible);
+  d->ROINode->SetDisplayClassVisibility
+        ("vtkMRMLAnnotationLineDisplayNode", visible);
 }
 
 // --------------------------------------------------------------------------
@@ -228,4 +232,19 @@ void qMRMLAnnotationROIWidget::updateROI()
                            0.5*(bounds[3]-bounds[2]),
                            0.5*(bounds[5]-bounds[4]));
   d->ROINode->EndModify(wasModifying);
+}
+
+// --------------------------------------------------------------------------
+void qMRMLAnnotationROIWidget::onMRMLDisplayNodeModified()
+{
+  Q_D(qMRMLAnnotationROIWidget);
+
+  if (!d->ROINode)
+    {
+    return;
+    }
+
+  // Visibility
+  d->DisplayClippingBoxButton->setChecked(d->ROINode->
+          GetDisplayClassVisibility("vtkMRMLAnnotationLineDisplayNode"));
 }
--- a/Modules/Loadable/Annotations/Widgets/qMRMLAnnotationROIWidget.cxx
+++ b/Modules/Loadable/Annotations/Widgets/qMRMLAnnotationROIWidget.cxx
@@ -195,8 +195,19 @@ void qMRMLAnnotationROIWidget::setExtent(double minLR, double maxLR,
 void qMRMLAnnotationROIWidget::setDisplayClippingBox(bool visible)
 {
   Q_D(qMRMLAnnotationROIWidget);
-  d->ROINode->SetDisplayClassVisibility
-        ("vtkMRMLAnnotationLineDisplayNode", visible);
+
+  int n = d->ROINode->GetNumberOfDisplayNodes();
+  int wasModifying [n];
+  for(int i = 0; i < n; i++)
+    {
+    wasModifying[i] = d->ROINode->GetNthDisplayNode(i)->StartModify();
+    }
+  d->ROINode->SetDisplayVisibility(visible);
+  for(int i = 0; i < n; i++)
+    {
+    d->ROINode->GetNthDisplayNode(i)->EndModify(wasModifying[i]);
+    }
+
 }
 
 // --------------------------------------------------------------------------
@@ -245,6 +256,12 @@ void qMRMLAnnotationROIWidget::onMRMLDisplayNodeModified()
     }
 
   // Visibility
-  d->DisplayClippingBoxButton->setChecked(d->ROINode->
-          GetDisplayClassVisibility("vtkMRMLAnnotationLineDisplayNode"));
+  if(d->ROINode->GetDisplayVisibility())
+    {
+    d->DisplayClippingBoxButton->setChecked(true);
+    }
+  else
+    {
+    d->DisplayClippingBoxButton->setChecked(false);
+    }
 }
--- a/Modules/Loadable/VolumeRendering/Resources/UI/qSlicerVolumeRenderingModuleWidget.ui
+++ b/Modules/Loadable/VolumeRendering/Resources/UI/qSlicerVolumeRenderingModuleWidget.ui
@@ -555,7 +555,7 @@
   </customwidget>
   <customwidget>
    <class>ctkCheckablePushButton</class>
-   <extends>QPushButton</extends>
+   <extends>ctkPushButton</extends>
    <header>ctkCheckablePushButton.h</header>
   </customwidget>
   <customwidget>
@@ -584,6 +584,11 @@
    <class>ctkExpandButton</class>
    <extends>QToolButton</extends>
    <header>ctkExpandButton.h</header>
+  </customwidget>
+  <customwidget>
+   <class>ctkPushButton</class>
+   <extends>QPushButton</extends>
+   <header>ctkPushButton.h</header>
   </customwidget>
   <customwidget>
    <class>ctkSliderWidget</class>
@@ -817,38 +822,6 @@
     <hint type="destinationlabel">
      <x>230</x>
      <y>732</y>
-    </hint>
-   </hints>
-  </connection>
-  <connection>
-   <sender>ROICropDisplayCheckBox</sender>
-   <signal>toggled(bool)</signal>
-   <receiver>ROIWidget</receiver>
-   <slot>setDisplayClippingBox(bool)</slot>
-   <hints>
-    <hint type="sourcelabel">
-     <x>238</x>
-     <y>193</y>
-    </hint>
-    <hint type="destinationlabel">
-     <x>114</x>
-     <y>541</y>
-    </hint>
-   </hints>
-  </connection>
-  <connection>
-   <sender>ROIWidget</sender>
-   <signal>displayClippingBoxChanged(bool)</signal>
-   <receiver>ROICropDisplayCheckBox</receiver>
-   <slot>setChecked(bool)</slot>
-   <hints>
-    <hint type="sourcelabel">
-     <x>114</x>
-     <y>541</y>
-    </hint>
-    <hint type="destinationlabel">
-     <x>238</x>
-     <y>193</y>
     </hint>
    </hints>
   </connection>
--- a/Modules/Loadable/Annotations/Widgets/qMRMLAnnotationROIWidget.h
+++ b/Modules/Loadable/Annotations/Widgets/qMRMLAnnotationROIWidget.h
@@ -75,6 +75,8 @@ protected slots:
   void onMRMLNodeModified();
   /// Internal function to update the ROI node based on the sliders
   void updateROI();
+  /// Internal function to update the ROIDisplay node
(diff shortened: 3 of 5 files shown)


pieper:
Looks good.  Only question might be that this communication should happen via a mrml node rather than directly manipulating the widget, but this slot looks consistent with the rest of the widget design so no problem to add it.

@jcfr and I are both traveling until next week.  @naucoin if you want to merge that's fine as far as I can see.

lassoan:
That's a very good point: Module widgets must not be manipulated from another module.

Acceptable approaches:

* Modify the ROI MRML node's visibility. There may be helper functions in the volume rendering logic to get the ROI node for a particular volume node. You may add such a helper function if it doesn't exist.

* Insert the MRML widget that contains ROI show/hide into your module widget. If ROI show/hide is not in a separate reusable widget then you need to split it out into a separate reusable MRML widget.

Punzo:
thanks for the feedback.
I am actually modifying the volumeRenderingWidget directly from mine (is the problem that MRML nodes will not get updated in this way?). I'll redo everything using the MRML nodes. If some helper functions will be needed I'll add commits here (I'll leave the pull request open for the moment).

naucoin:
I dug into it more before posting my comment, the GUI checkbox works directly on the widget, there's no MRML setting that can be changed right now. The GUI's mrmlDisplayNode() volume rendering dipslay node accessor could be made invokable from python and it's Visibility flag could be used to toggle the ROI widget's display visibility, but I did't dig deeper for more repercussions. The GUI has enabled and visible as two separate toggles.

Punzo:
I fixed. I should have done more commit for clarity (sorry). Here I summarize the modification:

1) STYLE: now qSlicerVolumeRenderingModuleWidget modify ROIwidget through the MRML 

2) BUG: toggling ROICropDisplayCheckBox in ROIwidget was not updating ROICropDisplayCheckBox in  qSlicerVolumeRenderingModuleWidget) . 

3) ENH: I have added two signals: 
    a) newCurrentMRMLVolumeNode(vtkMRMLNode* node): 
    b) newCurrentDisplayNode(vtkMRMLNode* node):

Libs/MRML/Core/vtkMRMLDisplayableNode.cxx
```diff
@@ -292,6 +292,7 @@ void vtkMRMLDisplayableNode::SetDisplayVisibility(int visible)
       displayNode->SetVisibility(visible);
       }
     }
+  this->Modified();
```
lassoan: If only the visibility is changed then the displayable node should not be modified, because that may be a very expensive operation. For example, if volume rendering is enabled then a lot of computation is done when the volume is changed, while nothing is recomputed if only the visibility is changed. It's convenient but misleading to to have this SetDisplayVisibility method in the displayable node.

Modules/Loadable/VolumeRendering/MRML/vtkMRMLVolumeRenderingDisplayNode.h
```diff
@@ -157,6 +162,8 @@ class VTK_SLICER_VOLUMERENDERING_MODULE_MRML_EXPORT vtkMRMLVolumeRenderingDispla
 
   int CroppingEnabled;
 
+  int ROIDisplayEnabled;
```
lassoan: I really appreciate your work that you've put into this, but unfortunately this solution is not very clean. The issue is that ROI visibility is already stored in the scene (for example, you can access it by getNode('AnnotationROI').GetDisplayVisibility()). If we add a variable that stores this status then we have to implement a synchronization mechanism, which is an unnecessary complexity. Instead, the GUI should just observe the displayable node vtkMRMLDisplayableNode::DisplayModifiedEvent and update the GUI if the event is invoked.

Libs/MRML/Core/vtkMRMLDisplayableNode.cxx
Punzo: You are right, that was quite bad ( I should have understood when I saw recompiling many things)! I moved the Modified event in the slot onROICropDisplayCheckBoxToggled, now should be fine.
Thanks!

Modules/Loadable/VolumeRendering/MRML/vtkMRMLVolumeRenderingDisplayNode.h
Punzo: ok, I'll see how to remove it and use only the visibility varibale from the ROI

Punzo: I have done (see last commit)

in practice I used this in the slot of qSlicerVolumeRenderingModuleWidget for enabling the visibility:

d->ROIWidget->mrmlROINode()->SetDisplayVisibility(toggle);
d->ROIWidget->mrmlROINode()->Modified();

Punzo:
Hi @lassoan now all the communication between the two widgets are going through the MRML.
However the observation

this->qvtkConnect(roiNode, vtkCommand::ModifiedEvent,
              this, SLOT(updateFromMRMLROINode()));

cause crash when I close the scene.

I saw that the observation to the displayNode is done with qvtkReconnect. Is there any reason for that?
thanks for your feedback.

Punzo:
I fixed the crash issue. I tested and it works. @naucoin, @lassoan, @pieper may you review again?
(sorry for the mess about many commits).

Thanks.

P.S. @lassoan I used the  vtkCommand::ModifiedEvent instead of the vtkMRMLDisplayableNode::DisplayModifiedEvent because the ROI widget is updated on a a ModifiedEvent. Thar why I use also d->ROIWidget->mrmlROINode()->Modified(); in the slot of qSlicerVolumeRenderingModuleWidget.

lassoan:
It is a bug in the qSlicerVolumeRenderingModuleWidget that changing visibility of the clipping ROI does not update the state of the ROI visibility 'eye' icon. This has to be fixed properly (by observing the vtkMRMLAnnotationLineDisplayNode display node of the ROI). Anything else would just mask the problem in certain workflows and increase the complexity of the application without actually solving the issue.

How to reproduce the issue:
* load MRHead sample
* activate volume rendering
* Click "Display ROI" eye icon
* Hide the ROI by entering in the Python interactor:
getNode('vtkMRMLAnnotationLineDisplayNode1').SetVisibility(False)
getNode('vtkMRMLAnnotationPointDisplayNode1').SetVisibility(False)
getNode('vtkMRMLAnnotationTextDisplayNode1').SetVisibility(False)
=> ERROR: the ROI is hidden in the 2D/3D viewers but the "Display ROI" eye icon in the volume rendering GUI is still active

lassoan:
There are many changes in the qSlicerVolumeRenderingModuleWidget.ui file. Some of them are functionally equivalent changes (for example lines are moved up/down, some xy values are slightly changed). It would help the code review a lot if you only committed intended changes and not those that are modified as a side effect of using the Qt designer (you can just go through all the changes and revert all those that you did not intend to do).

Punzo:
OK, thanks for the feedback. Tomorrow I'll add to the RoiWidget an observation to the the RoiDisplay node, so I can use the event display modified instead of the vtkmodified (this is the right fix, isn't it?). I'll also clean the ui commit.

lassoan:
Yes, I think display modified event observation of the displayed (ROI) node should work.

Punzo:
Hi, I did it. I cleaned all the old commits (it was a mess) and pushed the new ones.

1) 429cb7b and d05235a regards the bug about the eye icon.

2) 3d9ba2e regards to two signals that I would like to add.

I'll wait your feeback,

thanks.

Punzo:
Commits : d2df268, f618712, 69c7854

are just a revert and reapplication of d05235a for removing another pollution of Qtdesigner on 
qMRMLAnnotationROIWidget.ui

Modules/Loadable/Annotations/Widgets/qMRMLAnnotationROIWidget.cxx
```diff
@@ -100,11 +101,22 @@ vtkMRMLAnnotationROINode* qMRMLAnnotationROIWidget::mrmlROINode()const
 void qMRMLAnnotationROIWidget::setMRMLAnnotationROINode(vtkMRMLAnnotationROINode* roiNode)
 {
   Q_D(qMRMLAnnotationROIWidget);
-  qvtkReconnect(d->ROINode, roiNode, vtkCommand::ModifiedEvent,
+
+  if(!roiNode)
+    {
+    return;
+    }
```
jcfr: I don't think checking if `roiNode` is null is required. Removing the check will ensure the slots will be disconnected.

Modules/Loadable/Annotations/Widgets/qMRMLAnnotationROIWidget.cxx
```diff
@@ -229,3 +238,12 @@ void qMRMLAnnotationROIWidget::updateROI()
                            0.5*(bounds[5]-bounds[4]));
   d->ROINode->EndModify(wasModifying);
 }
+
+// --------------------------------------------------------------------------
+void qMRMLAnnotationROIWidget::onMRMLDisplayNodeModified()
+{
+  Q_D(qMRMLAnnotationROIWidget);
```
jcfr: Instead, you should add the check here. Similarly to what is done for `onMRMLNodeModified()` here: https://github.com/Punzo/Slicer/blob/69c78545742fe9dd258bafbe3fb44f008583c07c/Modules/Loadable/Annotations/Widgets/qMRMLAnnotationROIWidget.cxx#L134-L137

Modules/Loadable/VolumeRendering/Resources/UI/qSlicerVolumeRenderingModuleWidget.ui
```diff
@@ -105,15 +105,15 @@
...
-        </property>
         <property name="noneEnabled">
          <bool>true</bool>
         </property>
         <property name="addEnabled">
          <bool>false</bool>
         </property>
+        <property name="renameEnabled">
+         <bool>true</bool>
+        </property>
```
jcfr: These changes are not needed.

Modules/Loadable/VolumeRendering/Resources/UI/qSlicerVolumeRenderingModuleWidget.ui
```diff
@@ -518,16 +518,16 @@
...
    <class>qMRMLCheckableNodeComboBox</class>
    <extends>qMRMLNodeComboBox</extends>
    <header>qMRMLCheckableNodeComboBox.h</header>
   </customwidget>
   <customwidget>
+   <class>qMRMLDisplayNodeViewComboBox</class>
+   <extends>qMRMLCheckableNodeComboBox</extends>
+   <header>qMRMLDisplayNodeViewComboBox.h</header>
+  </customwidget>
+  <customwidget>
```
jcfr: Idem

Modules/Loadable/Annotations/Widgets/qMRMLAnnotationROIWidget.cxx
Punzo: done

Punzo:
Hi  @naucoin, @lassoan, @pieper, @jcfr,

I'd like to know if in your opinion this is ready. Thanks,

Davide.

naucoin:
I'm having a hard time following the changes on this pull request (squashing everything down to one commit would simplify things), but the one thing that stood out to me was why do you need DisplayClassVisibility? We already have the ROI node methods to set and get DisplayVisibility that deal with the multiple display nodes. A quick python test confirms that the ROI displayable manager only toggles the ROI's visibility according to
roiNode->GetDisplayVisibility()
The visibility settings of the three display nodes are only used in aggregate (all on or all off).

Punzo:
Hi @naucoin, 

thanks you for your feedback. Sorry for the mess, I'll squash the commits!

the problem with roiNode->Set/GetDisplayVisibility() is this:

using SetDisplayVisibility() in https://github.com/Slicer/Slicer/pull/377/files#diff-c07106f88f2fdcd38e56d866084ae28bL191, it should modify the 3 display visibilities. However if you print  the visibility values in https://github.com/Slicer/Slicer/pull/377/files#diff-c07106f88f2fdcd38e56d866084ae28bR238:

with GetDisplayVisibility it returns:
1) if it is a vtkMRMLAnnotationPointDisplayNode -> always 2 (which is a problem!)
2) if it is a vtkMRMLAnnotationTextDisplayNode  -> always 2 (which is a problem!)
3) if it is a vtkMRMLAnnotationLineDisplayNode -> 0 or 1 

P.S.: everytime you use d->ROINode->SetDisplayVisibility(0 or 1) you fire all the 3 vtkMRMLDisplayableNode::DisplayModifiedEvent 

Due this behavior, this means that if I use GetDisplayVisibility() in onMRMLDisplayNodeModified()
I have to mask the cases of return 2, e.g.:

  if(d->ROINode->GetDisplayVisibility() != 2)
    {
    d->DisplayClippingBoxButton->setChecked(d->ROINode->
               GetDisplayVisibility());
    }

I didn't like this, so I prefered to use theDisplayClassVisibility to modify only vtkMRMLAnnotationLineDisplayNode following the first @lassoan 's suggestion.

Of course for me both the solutions are fine, let me know.

naucoin:
@Punzo Thanks for the clarification. If ROI node SetDisplayVisibility isn't setting the visibiity on all of the display nodes, that's a bug that should be fixed. You can avoid having three events fired (modfieid on three display nodes) by enclosing the call in a StartModify/EndModify block. 
Only setting the visibilty flag on the line display node is relying on this test in the displayable manager that shows the whole widget as long as one of the display nodes is set to visible:
https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/Annotations/MRMLDM/vtkMRMLAnnotationROIDisplayableManager.cxx#L405
I think what you're seeing is that the three display nodes are set in sequence by the ROI node's set DisplayVisibility and you're getting the display node modified events as that progresses, so the overall visibility isn't 0 or 1 unless all the display nodes have been set to 1, it's 2 when some are set and some are not. Using a binary check box for this GUI element works best if you link it to the binary visibility setting from the ROI node. I do see that the current trunk is missing a single event at the displayable node level in SetDisplayVisibility, but it was set up as a utility function meant to toggle all the display nodes in a way that anything watching for a particular display node's setting changed can react to it (the annotations GUI for example for the Ruler has separate visibility icons for the text). 

Short version: I'd prefer using ROINode->SetDisplayVisibility and either use the Start/EndModify to reduce the number of events, or make the test above more clear that 2 is a valid value, you're just using a binary check box to display a non binary state, and technically a partially visible state means that the ROI is displayed anyway. So you could use
if (d->ROINode->GetDisplayVisibility() != 0)
set it checked, otherwise unchecked

Punzo:
@naucoin thanks for your explanation. 

Using the Start/EndModify I fixed it.

Punzo:
few lines for explanation of commit  1a5c895:

in 1a5c895 I changed from using the DisplayClassVisibility  to the DisplayVisibility. 
https://github.com/Slicer/Slicer/pull/377#issuecomment-151469446 was an event issue: the GUI was getting updated while the different MRMLDisplay nodes where still getting updated.

I used the Start/EndModify in 
https://github.com/Punzo/Slicer/commit/1a5c8951ba79cd698350cf4eca8bd09a72ba7258#diff-c07106f88f2fdcd38e56d866084ae28bR203

https://github.com/Punzo/Slicer/commit/1a5c8951ba79cd698350cf4eca8bd09a72ba7258#diff-cf65d96a6fd47f822c9fa8e4de33e6baR848

this makes sure that all the 3 visibilities are set to the value in input from the GUI. Therefore,
now in  onMRMLDisplayNodeModified() the return of GetDisplayVisibility is always 0 or 1.

let me know if this is fine. Thanks!

NOTE: all the visibility should be in synch. Otherwise it results in strange behavior on the GUI (for example to toggle off the "eye" you have to press twice, because the first time two visibility are still on true and the eye is stay still on true and the ROI actually will be still visible).

jcfr:
:+1: Integrated in [r24803](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24803)

Note that I did not integrate the addition of the new signals (there were not used anywhere). 

If there are still needed, could you create a new PR ?  I pushed the corresponding change on branch [jcfr/Slicer@add-VR-module-widget-signal](https://github.com/jcfr/Slicer/commits/add-VR-module-widget-signal). See 57854080cde1

PR #380, opened 2015-10-18
BUG: Estimated output baseline image incorrect when using "DWI Volume Masking"

orbe77:
Fixes bug reported on Mantis with ID 0003855 on 2014-12-17.

The updated line calculates the magnitude of each of the gradients, and determines that an image is "baseline" if the magnitude is smaller than a threshold. The use of abs() on double values caused the magnitude to always be 0, and therefore all images were considered as "baseline" and included in the average computed by the filter and presented as "Output baseline volume" to the user.


jcfr:
Thanks for fixing this issue :+1: 

Would it be possible to update the associated test adding a test case (using a small dataset) that would have fail before your fix and is now passing ? 
See https://github.com/Slicer/Slicer/blob/30636dce872c3f51c8f70d0af2b311a5eaf601e9/Modules/CLI/DiffusionWeightedVolumeMasking/Testing/CMakeLists.txt

Cc: @ljod

orbe77:
I'm afraid I'm not familiar with the syntax of these tests. But if you
could forward me a small DTI dataset, I can calculate the correct output
for comparison in the unit-tests.

Also, I'm at 1249 Boylston this week and forward so I'd be happy to meet if
you want to show me how to code these types of tests.

---
Ørjan

jcfr:
@orbe77 Could you work with @ljod who is also based at 1249 Boylston ?  I am sure she can provide you with the requested information. 

Considering we are planning to release Slicer very soon, would be ideal to address this later this afternoon.

On my side, I am base in North Carolina @kitware 

Thanks for the help, 
Jc

Cc: @pieper

orbe77:
Hi

Now I’ve submitted a small update to the test case (#389), and included here are the expected outputs after the bug fix. Please let me know if I should update further.


—
Ørjan

jcfr:
Thanks. 

Regarding  #389 , what do you exactly mean by `Test remove islands functionality separately` ? 

> included here are the expected outputs after the bug fix

It seems you forgot to include files and described what are the expected output.

Ideally you should provide:

1) The command line to invoke and where to find the input data

2) the expected output image (you could either send an email or include it to this issue after adding a `.txt` to the file)

You repeat this for every cases you would like to test.

Mention also what should be the name of the test.

Thanks for your help. We are getting close.

Cc: @ljod  @pieper

orbe77:
Thank you for your quick reply, before I was trying to address too many concerns before without properly explaining myself.

1) So the already existing test case ran the Otsu segmentation with the --removeislands flags, which triggers a separate issue. In #389 I updated this test-case to not use this flag, so that the unit test was only testing one thing (and not the "remove island" functionality ). This was the reason for the (in hindsight) somewhat cryptic name.

2) A proper name for the unit test could be e.g. "verify computed DWI Otsu baseline output"

3) The command line I wanted to run was the one in the https://github.com/Slicer/Slicer/blob/30636dce872c3f51c8f70d0af2b311a5eaf601e9/Modules/CLI/DiffusionWeightedVolumeMasking/Testing/CMakeLists.txt , without the "--removeislands" flag.

4) The files I wanted to use for test were included in the email I sent before, but now I see they were stripped from the web interface since they were not of the supported types. The .txt extension trick did not seem to work, but not I've included them here as .png files. The helix-dwi dataset is already included elsewhere with the test-datasets in Slicer (it is referenced in the link in item (3) above). The expected output from the algorithm are included here as "otsu_*.png"

I hope this clarifies matters, thanks for your patience!

---
orbe77

helix-dwi nhdr
helix-dwi raw gz
otsu_baseline nhdr
otsu_baseline raw gz
otsu_mask nhdr
otsu_mask raw gz

jcfr:
Awesome. Thanks for clarifying :smile: 

One last question, considering the updated test command line and new baseline/mask, is the test failing without your fix in https://github.com/orbe77/Slicer/commit/05e9aabed7b3ae2ad45a248afb4efe34583116ff and passing after integrating your fix ?

I am assuming it is the case. But I wanted to check with you before starting integrating the changes.

orbe77:
Yes, without the fix you will get different (and bad) baseline output. 

---
Ørjan

jcfr:
Great. I now have all the information to move forward with this. Thanks again for your patience with all my questions :smile:

jcfr:
Integrated in [r24673](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24673)

Thanks for your help. :+1: 

Bonus: Now the test harness for this module has been updated, would be great to add test for the `removeisland` feature :smile:

PR #382, opened 2015-10-19
Fix FindGit for CMake 3.4rc01.

thewtex:


jcfr:
Thanks for the fix.

Out of curiosity, do you have more details on the error fix by this commit ? Is there a change in the latest CMake rc that is causing trouble ?

thewtex:
The use of GIT_VERSION_STRING in ExternalProject.cmake was added to upstream in b04c38159 and 7c5b6ed5c.  But, Slicer's FindGit diverges from upstream's.

msmolens:
Also, the comments in CMake bug [15791](https://cmake.org/Bug/view.php?id=15791) give more details.

jcfr:
Commit 5e2ede3 integrated in [r24629](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24629)

jcfr:
Integrated in [r24631](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24631)

Thanks @thewtex and @msmolens

PR #384, opened 2015-10-23
PERF: Generation of .s4ext extension description file identical to th…

jbvimort:
…e ones generated by CMake

The extension wizard now generate extension description files in the same way that the ones generated by configuring your extension with CMake.


jcfr:
Thanks for the patch.

I propose to postpone the integration and review of this for after the Slicer release.

jcfr:
@jbvimort Thanks for the patch.

I add an other look at it and I would prefer we don't integrate it the way it is.

If you would like to have similar description file, I would suggest that you:

* Create a reusable template named (for example) `extension_description.s4ext.in` by removing the text currently hardcoded in `SlicerFunctionGenerateExtensionDescription`  (See [here](https://github.com/Slicer/Slicer/blob/0698939f60b374dd1344f5e9cc0a692744908b19/Extensions/CMake/SlicerFunctionGenerateExtensionDescription.cmake#L112-L156)) and adding it to `Slicer/Extensions/CMake/extension_description.s4ext.in`. 

* Update `SlicerFunctionGenerateExtensionDescription` to configure that file using

```
configure_file(
  ${CMAKE_CURRENT_LIST_DIR}/extension_description.s4ext.in
  ${filename}
  @ONLY)
```

* The wizzard could then reuse that same template by replacing the same variable.


This approach would avoid duplication of text and ensure the generated description file are the same.

jcfr:
@jbvimort  Let us know if you are planning on working on this or if you have any questions.

jbvimort:
Yes, I'm working on it, but I don't really understand how to replace the variables, do I need to do it from ExtensionDescription.py by calling some macros?

jcfr:
> Yes, I'm working on it

Great

> how to replace the variables

There are two cases:

* Configuration from CMake. this will be done using `configure_file` which will take care of replacing all variable of the form `@VARNAME@`

* Configuration from the extension wizard. For this case, you could:
 * have a map `VAR_NAME` -> `Value`. You will have to make use `self.__dict__` and ensure you can map the existing key with the one expected in the template.
 * for each line of the template, look over the map and proceed to the replacement
 * write down the result

jbvimort:
Just to be sure that I understand everything correctly:
-For the configuration with the CMake everything already work perfectly?
-For the configuration from the extension wizard, I will need to create a map in ExtensionDescription.py with taking care of putting the key as expected in SlicerFunctionGenerateExtensionDescription? And I also need to modify SlicerFunctionGenerateExtensionDescription to create a function that will write the good result ( an .s4ext extension file) by reading the map previously created?

jcfr:
>  For the configuration with the CMake everything already work perfectly?

* Create a reusable template named (for example) `extension_description.s4ext.in` by removing the text currently hardcoded in `SlicerFunctionGenerateExtensionDescription`  (See [here](https://github.com/Slicer/Slicer/blob/0698939f60b374dd1344f5e9cc0a692744908b19/Extensions/CMake/SlicerFunctionGenerateExtensionDescription.cmake#L112-L156)) and adding it to `Slicer/Extensions/CMake/extension_description.s4ext.in`. 

* Update `SlicerFunctionGenerateExtensionDescription` to configure that file using

```
configure_file(
  ${CMAKE_CURRENT_LIST_DIR}/extension_description.s4ext.in
  ${filename}
  @ONLY)
```


> For the configuration from the extension wizard,

(( `SlicerFunctionGenerateExtensionDescription` is used only in the CMake case. )) 

-> will need to create a map in ExtensionDescription.py

Yes


In other word, in both the CMake and Python case, the file need `extension_description.s4ext.in` will have to be configured. In the CMake case, it should quite easy. For the Python case, you will have to write the code reading the  `extension_description.s4ext.in`, updating the variable and writing the new file.

jbvimort:
Ok, thank you! Its more clear now, I will try to do that!

jcfr:
Great. Let me know if you have any other questions.

jbvimort:
@jcfr I finished to modify the Cmake version and it worked perfectly. I'm now trying to modify the python version, and I can't find a way to have the path to the extension_description.s4ext.in as this file is only in the source files? Do you think that I will need to copy it somewhere during the compilation of slicer, or it exist a better solution?

jcfr:
Looking good.

> a way to have the path to the extension_description.s4ext.in

I suggest the following:

* Extend the API of `ExtensionDescription` adding a class level property named `TEMPLATE`. If not set, the description file should be generated the way it is. It it is set, it will then generate a description file similar to the one done using CMake code.

* Add an install rule for `extension_description.s4ext.in`

* Update both `slicerExtensionWizard.sh.in` and `slicerExtensionWizard-install.sh.in` so that they set `ExtensionDescription.TEMPLATE`

* Update `Modules/Scripted/ExtensionWizard.py` to set `TEMPLATE`. To distinguish between installed vs built tree case ... `slicer.app.isInstalled`, `slicer.app.slicerHome` could be used.

jbvimort:
I'm sorry but I'm lost again:
What do you mean by Extend the API of ExtensionDescription adding a class level property named TEMPLATE?

jcfr:
> by Extend the API of ExtensionDescription 

Within the python extension wizard, the description file is written in the `_write` function. The one you originally modified. The idea would be to have the  `_write` function generating (by default) the current description file and if a class level property named `DESCRIPTION_FILE_TEMPLATE` is set to an existing file, it would be used to generate the description file.

```
class ExtensionDescription(object):
  """Representation of an extension description.
[...]
  """

 DESCRIPTION_FILE_TEMPLATE = None  # This is the class level property 

  _reParam = re.compile(r"([a-zA-Z][a-zA-Z0-9_]*)\s+(.+)")

[...]

def _write(self, fp):
  # CODE YOU WILL UPDATE to use 

  if self.DESCRIPTION_FILE_TEMPLATE is not None:
    # Generate using template
    # TODO
  else:
   # Generate without template
   # Existing code

```

jbvimort:
Thank you! I will try to do that!

jbvimort:
Hello @jcfr !

I finished to work on the python code, that work pretty well for now!
Therefore my knowledge on CMAKE are really poor and I have to difficulties:
- I can't find where to add the install rule for extension_description.s4ext.in?
- How can I modify ExtensionDescription.TEMPLATE from slicerExtensionWizard.sh.in?

I'm sorry to bother you with that, have a good day!

JB

jcfr:
> finished to work on the python code

Great :+1: 

Would  you mind pushing a commit with the current code ?


> I can't find where to add the install rule for extension_description.s4ext.in?

Thinking more about it, consider adding the file `extension_description.s4ext.in` into the source directory `Slicer/Utilities/Templates/Extensions`.

That way, it will be installed by the already existing rule in [Slicer/Utilities/Templates/CMakeLists.txt](https://github.com/Slicer/Slicer/blob/64a4c7da2502e12a12372b8bac16d8b1dcc8ea77/Utilities/Templates/CMakeLists.txt)


> How can I modify ExtensionDescription.TEMPLATE from slicerExtensionWizard.sh.in?

Having an other look, it seems the easiest will be to update [this](https://github.com/Slicer/Slicer/blob/64a4c7da2502e12a12372b8bac16d8b1dcc8ea77/Utilities/Scripts/SlicerWizard/ExtensionWizard.py#L608-L620) code  so that it looks like:

```
candidateBuiltInTemplatePaths = [
  os.path.join(scriptPath, "..", "..", "Templates"), # Run from source directory
  os.path.join(scriptPath, "..", "..", "..", "share", # Run from install
               "Slicer-%s.%s" % tuple(__version_info__[:2]),
               "Wizard", "Templates")
]
builtInTemplatePath = None
for candidate in candidateBuiltInTemplatePaths:
  if os.path.exists(candidate):
    self._templateManager.addPath(candidate)

ExtensionDescription.TEMPLATE = os.path.join(builtInTemplatePath, "Extensions", "extension_description.s4ext.in")
```

Notes:
* You will have to think about exception cases.
* Consider also updating the method `setupTemplates()` in `Modules/Scripted/ExtensionWizard/ExtensionWizard.py`. See [here](https://github.com/Slicer/Slicer/blob/3d48cf0de0cfda91b9a6d705c4e94092ace20e59/Modules/Scripted/ExtensionWizard/ExtensionWizard.py#L169)

jbvimort:
Thank you so much!

My current version should be online right now!
I will try to implement it with all these tips!

jbvimort:
Obviously not, sorry, this should be better! :)

jbvimort:
Hello JC!

I think I finally finished to make everything working well! I made some tests with a build version and an installed version, and everything work correctly.
Do you think that everything is ok, or should I reopen an other pull request with an updated version of Slicer to make things more clear?

jcfr:
Great job on this topic.

Et voila. Finally, reviewed and tweaked the topic. See https://github.com/Slicer/Slicer/compare/master...jcfr:jbvimort-update-s4extGeneration

After testing it works as expected from the packaged Slicer, I will integrate.

jcfr:
Integrated in [r25014](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25014) to [r25016](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25016)

PR #388, opened 2015-10-28
BUG: Fix scaling of 3D window snapshots

naucoin:
Slicer svn revision 24668 incorporates a fix for a bug in vtkRenderLargeImage
so it now works with gradient backgrounds. Use this class to do off screen
magnified rendering of the 3D window when capturing screen shots.

A bug still remains for the 2D windows in that they get tiled by this
method instead of scaled, so continue using the Qt after capture
scaling for the full view and the slice views.

Issue #3885

### DIFF
--- a/Libs/MRML/Widgets/Resources/UI/qMRMLScreenShotDialog.ui
+++ b/Libs/MRML/Widgets/Resources/UI/qMRMLScreenShotDialog.ui
@@ -141,6 +141,9 @@
    </item>
    <item row="2" column="1">
     <widget class="ctkDoubleSpinBox" name="scaleFactorSpinBox">
+     <property name="toolTip">
+      <string>Adjust the Magnification factor. Note that only the 3D view is rendered in high resolution, the other views are rescaled after the window is captured.</string>
+     </property>
      <property name="sizePolicy">
       <sizepolicy hsizetype="Maximum" vsizetype="Fixed">
        <horstretch>0</horstretch>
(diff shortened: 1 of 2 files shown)


jcfr:
Nitpicks: `BUG: reenable scaling of 3D window snapshots ` -> `ENH: Implement scaling of 3D window snapshots`

naucoin:
Updated the branch with the new commit message, tool tip.

jcfr:
Great. 

Few more comment:

* seem the concept of screenshot and snapshot are used. The dialog is named `Annotations Screenshot`, the tooltip talk about screenshot. I propose to change the title of the commit message to `Use highres for 3D view included in magnified screenshots`.

After addressing this comment, it looks good and can be integrated.

Now here few questions that are for later:

* In the annotation module, `List of snapshots` is used. Should it be changed to `List of screenshots` ? I guess since we are using snapshot nodes, it may not be so easy.

* regarding the usability:
 * in the screenshot dialog. It would be great to add a text like:

```
What is an Annotation Screenshot?  Each time the user create an annotation screenshot, a traditional 
screenshot image of the viewers is created. Then, along side with a name and it a description, it is stored 
in the  with the current scene. 

The Annotations module allows to manage the list of screenshots.
```

  * Would it make sense to ask the user where to save the file ? The selected location would then be associated with the snapshot storage node.

jcfr:
To capture the usability remarks, I created issue [#4069](http://www.na-mic.org/Bug/view.php?id=4069)

jcfr:
Integrated in [r24670](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24670)

PR #393, opened 2015-10-30
BUG: SliceViewAnnotations: self.sliceViewNames.append() failed

che85:
fixes issue https://github.com/fedorov/PCampReview/issues/91

### DIFF
--- a/Modules/Scripted/DataProbe/DataProbeLib/SliceViewAnnotations.py
+++ b/Modules/Scripted/DataProbe/DataProbeLib/SliceViewAnnotations.py
@@ -471,7 +471,7 @@ class SliceAnnotations(VTKObservationMixin):
 
   def createCornerAnnotations(self):
     self.createGlobalVariables()
-    self.sliceViewNames = self.layoutManager.sliceViewNames()
+    self.sliceViewNames = list(self.layoutManager.sliceViewNames())
     for sliceViewName in self.sliceViewNames:
       self.addSliceViewObserver(sliceViewName)
       self.createActors(sliceViewName)


jcfr:
@che85 Thanks for the fix.

Would it be possible to provide more details regarding the error that is been fixed ?

After doing a quick test in Slicer python interactor, there are no issue iterating over a `tuple`:

```
>>> for viewName in slicer.app.layoutManager().sliceViewNames():
...   print("viewName: %s" % viewName)
... 
viewName: Green
viewName: Yellow
viewName: Red
>>> type(slicer.app.layoutManager().sliceViewNames())
<type 'tuple'>
```

It also confirmed by the fact a `tuple` is of sequence type: https://docs.python.org/2/library/functions.html#tuple

che85:
@jcfr see the following line

self.sliceViewNames is assigned as a tuple:
https://github.com/Slicer/Slicer/blob/master/Modules/Scripted/DataProbe/DataProbeLib/SliceViewAnnotations.py#L474

Appending to a tuple: 
https://github.com/Slicer/Slicer/blob/master/Modules/Scripted/DataProbe/DataProbeLib/SliceViewAnnotations.py#L582

You cannot append to a tuple like this.

```
sliceViewNames = slicer.app.layoutManager().sliceViewNames()
sliceViewNames.append('5')
AttributeError: 'tuple' object has no attribute 'append'
```

jcfr:
Thanks for clarifying :+1: 

After tweaking the commit message, patch has been integrated in [r24684](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24684)

PR #392, opened 2015-10-30
BUG: Fix vtkSlicerDoubleArraysLogicAddFileTest

jcfr:
**WIP** DO NOT INTEGRATE

This commit fixes test vtkSlicerDoubleArraysLogicAddFileTest to account
for commit introduced in r24675 (BUG: Fix support for missing labels in
vtkMRMLDoubleArrayStorage reader)

At time of integration, only the test vtkMRMLDoubleArrayStorageNodeTest1
updated in r24676 (ENH: Add test for vtkMRMLDoubleArrayStorageNode
reader/writer) has been executed to confirm it was working.

Failure reported on the dashboard: 
* Linux: http://slicer.cdash.org/testDetails.php?test=6885663&build=700523
* MacOSX: http://slicer.cdash.org/testDetails.php?test=6885265&build=700522
* Windows: http://slicer.cdash.org/testDetails.php?test=6886406&build=700521

PR #394, opened 2015-10-30
Fix DoubleArray tests

agirault:
Corrects typo returning wrong value at the end of `vtkMRMLDoubleArrayStorageNode::ReadDataInternal`.

It is now returning 0 if the DoubleArray is empty. This is not a usual behavior as we should allow reading an empty node, but we are doing this for backward compatibility [1].

[1] https://github.com/jcfr/Slicer/commit/96afb52#diff-8f3fa951eccab26a6d68e04cb1488da0L200

This fixes those two failing tests:
*  `vtkSlicerDoubleArraysLogicAddFileTest` : the error was shown by improving the test in commit d36d5ba
*  `vtkMRMLDoubleArrayNodeTest1` : error discovered by @jcfr in #392

### DIFF
--- a/Libs/MRML/Core/Testing/vtkMRMLDoubleArrayNodeTest1.cxx
+++ b/Libs/MRML/Core/Testing/vtkMRMLDoubleArrayNodeTest1.cxx
@@ -311,13 +311,12 @@ bool TestReadFileWithLabels(std::string filepath)
 
   vtkNew<vtkMRMLDoubleArrayStorageNode> doubleArrayRead;
   doubleArrayRead->SetFileName(filepath.c_str());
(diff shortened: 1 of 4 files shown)


jcfr:
Thanks :+1: 

Tested and integrated in [r24680](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24680) to [r24683](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24683)

jcfr:
@agirault Make surer that your editor do not add tab and also has the clean space for edited line feature enabled.

qtcreator-cleanup-spaces

PR #396, opened 2015-11-03
BUG: Fixed transform splitting

lassoan:
Fixes https://www.na-mic.org/Mantis/view.php?id=4073

Problem was that when a composite transform contained an inverse linear transform, the inverse transform was used as transformToParent. When transformToParent was later modified by the transform sliders on the GUI, transformToParent was updated, but transformFromParent was unchanged.

Script to reproduce the issue:

```
t1=vtk.vtkTransform()
m1=vtk.vtkMatrix4x4()
m1.SetElement(0,3,15)
t1.SetMatrix(m1)
t2=t1.GetInverse()
print t1.GetMatrix()
print t1.GetInverseFlag()
print t2.GetMatrix() # this is required, otherwise the behavior will not be inconsistent
print t2.GetInverseFlag()
t2.Inverse() # editing the inverse transform
print t1.GetMatrix() # returns ... 15 ...
print t1.GetInverseFlag() # returns 0
print t2.GetMatrix() # returns ... 15 ...
print t2.GetInverseFlag() # returns 0
# t1 and t2 should be inverse of each other, but instead InverseFlag is 0 on both and the matrix values are the same.
```
It seems that if a computed inverse transform is modified then the transform pipeline becomes inconsistent. It may be the expected behavior in VTK, one should simply never modify a computed transform. I'll confirm this with David Gobbi.

The solution was to always set the forward transform in the node so that when it is modified, it updates the inverse transform as well.


jcfr:
Great. I am testing on my side. Rebuild in progress. :construction:

jcfr:
```
51: Test command: /home/jcfr/Projects/Slicer-Debug/Slicer-build/Slicer "--launcher-no-splash" "--launch" "/home/jcfr/Projects/Slicer-Debug/Slicer-build/bin/MRMLCoreCxxTests" "vtkMRMLNonlinearTransformNodeTest1" "/home/jcfr/Projects/Slicer/Libs/MRML/Core/Testing/NonLinearTransformScene.mrml"
51: Test timeout computed to be: 1800
51: Success
1/1 Test #51: vtkMRMLNonlinearTransformNodeTest1 ...   Passed    1.36 sec
```

:+1:  LGTM 

I also confirm that both glyph and volume are updated after following the step to reproduce in https://www.na-mic.org/Mantis/view.php?id=4073

@lassoan Could you integrated ?

jcfr:
Integrated in [r24688](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24688)

PR #398, opened 2015-11-04
3462 model maker breaks slice view display

jcfr:
BUG: Avoid confusion when importing scene model hierarchy. Fixes #3462

*** WIP ***

Considering that:

1. the slice view representation in the 3D view is done using a `vtkMRMLModelDisplayNode`

2. the IDs associated with these nodes are `vtkMRMLModelDisplayNode1`, `vtkMRMLModelDisplayNode2`, ...

3. node references are updated after nodes are added to the scene. Indeed, this is done in the `UpdateNodeReferences()` function called from  `vtkMRMLScene::Import()`

4. By default, `qMRMLSceneDisplayableModel` updates itself each time a node is added (`NodeAddedEvent`), it doesn't consider the Import state of the scene.

importing a scene containing hierarchy nodes referencing model display nodes also identified with `vtkMRMLModelDisplayNode1`, ... will cause the `qMRMLSceneDisplayableModel` to <TO BE CONTINUED>

This commit implements the solution consisting in always setting the `qMRMLSceneDisplayableModel` as LazyUpdate, this means that **To be continued**

  (a) the model will always ignore the NodeAdded/NodeRemoved events  when batch operation  (ie Import) are done on the scene

  and (b) will rescan the scene after the batch operation is complete.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLSelectionNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLSelectionNode.cxx
@@ -717,9 +717,9 @@ vtkMRMLSelectionNode::GetModelHierarchyDisplayNodeClassNames()const
 
 //----------------------------------------------------------------------------
 void vtkMRMLSelectionNode::AddModelHierarchyDisplayNodeClassName(const std::string& dispayableNodeClass,
-                                                                 const std::string& dispayNodeClass)
+                                                                 const std::string& displayNodeClass)
 {
-  this->ModelHierarchyDisplayNodeClassName[dispayableNodeClass] = dispayNodeClass;
+  this->ModelHierarchyDisplayNodeClassName[dispayableNodeClass] = displayNodeClass;
 }
 
 //----------------------------------------------------------------------------
--- a/Libs/MRML/Core/vtkMRMLSelectionNode.h
+++ b/Libs/MRML/Core/vtkMRMLSelectionNode.h
@@ -234,7 +234,7 @@ class VTK_MRML_EXPORT vtkMRMLSelectionNode : public vtkMRMLNode
   /// For example, parameters of "vtkMRMLFiberbundleNode" , "vtkMRMLFiberbundleTubeDisplayNode"
   /// will change tube visibility
   void AddModelHierarchyDisplayNodeClassName(const std::string& dispayableNodeClass,
-                                             const std::string& dispayNodeClass);
+                                             const std::string& displayNodeClass);
 
   /// Clear node display class names
   /// to which display node to apply properties in hierarchy display widgets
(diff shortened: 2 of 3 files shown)


jcfr:
Cc: @pieper @naucoin 

This is a work in progress, I wanted to share my current result while finalizing the topic.

In a nutshell, ensuring the `qMRMLSceneDisplayableModel` rescan the scene after batch operation avoids issue. 

I think this could be optimized by generating a new ID that does't conflict each time a new reference is seen at import time. Currently the new ID generation is done when a node is added and not when a reference is first seen.

naucoin:
This makes sense to me as it's a solution I've used with the annotations/markups before.

jcfr:
@naucoin Thanks for commenting. 

>  it's a solution I've used with the annotations/markups before

Do you mean that you removed the call to `setLazyUpdate` in the annotations/markups and implemented a solution consisting in being smarted about References ?

Because I can't find any call to  `setLazyUpdate` in either the `Annotations` or `Markups` module.

lassoan:
Could you please check if this solves the very similar issue in the apply "Transformed" tree of the Transforms module?

How to reproduce:
* Start Slicer
* Switch to Transforms module
* Load this scene (it contains a transform applied to the MRBrainTumor1 volume):
http://1drv.ms/1XPkXuy
=> ERROR: The MRBrainTumor1 volume should appear in the "Transformed" tree at the bottom right of the Transforms module GUI
(if the scene is loaded _before_ switching to Transforms module then the MRBrainTumor1 correctly appears in the tree)

naucoin:
I meant that I recalled using lazy update before, but didn't double check exactly where. :-) Looking back on my notes:
I had used it in the Scene Views model to fix a bug on importing scenes, but then I rewrote the whole module to not use the hierarchy nodes.

lassoan:
I see that lazyUpdate is optional and presented as a performance optimization option (disabled by default). Probably we should remove this option and always wait for scene loading/import to complete before trying to interpret the scene.

lassoan:
I've changed the default this->LazyUpdate = true; to false in qMRMLSceneModelPrivate::qMRMLSceneModelPrivate and it solved the Transforms module Transformed tree issue. It really seems that LazyUpdate (more accurately: not trying to interpret the scene when it is in a transient state) in the scene model is a must.

jcfr:
@lassoan  Thanks for the report.

> Probably we should remove this option and always wait for scene loading/import to complete before trying to interpret the scene

Until it is not truly optional, I will look into enabling the option. 

I will take care of updating the scene model.

pieper:
Great work nailing down the issue and solution!

jcfr:
I have now a good understanding of what was happening. I am adding a test named `TestImport()` to `vtkMRMLNodeTest1.cxx` to show the root issue.

Then, I will mark the test as expected to fail and most likely implement the workaround of setting the model as `LazyUpdate` ... stay tuned.

jcfr:
Workaround for issue [#3462](https://www.na-mic.org/Mantis/view.php?id=3462) integrated [r24723](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24723)

The root issue has been documented in [#4081](https://www.na-mic.org/Mantis/view.php?id=4081)

jcfr:
Note that it also address issue mentioned by @lassoan in https://github.com/Slicer/Slicer/pull/398#issuecomment-153761297 (and tracked in [#4080](https://www.na-mic.org/Mantis/view.php?id=4080)) along with issue  [#2849](https://www.na-mic.org/Mantis/view.php?id=2849)

PR #399, opened 2015-11-04
BUG: Disable fiducials in light box mode.

naucoin:
Fiducials are not being properly positioned when in light box mode, this
change disables them and gives the user a warning that they are being disabled
(new fiducials can still be placed, but will not be visible until you exit light box mode).
This change fixes bugs that were left from the attempt to use the point handle
widgets in 2D so that they can be properly hidden: fix the visibility setting
on the point handles in light box mode, disable the point handles when invisible
so that they don't become visible when the mouse is near it, updated a
visibility check comparision to match return type from method.
Also removed warning messages by casting the 3d handles before changing the
picking state.
Fixed a bug wherein if you had a mix of fiducials having been placed in lightbox
and not light box mode, the seed widget would have a mix of 2d and 3d handles.
This fix removes all the current handles so that they're properly recreated as
the necessary type.

Issue #1690

### DIFF
--- a/Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx
+++ b/Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx
@@ -24,12 +24,14 @@
 #include <QHBoxLayout>
 #include <QLabel>
 #include <QMenu>
+#include <QMessageBox>
 #include <QPushButton>
 #include <QSpinBox>
 #include <QWidgetAction>
 
 // CTK includes
 #include <ctkDoubleSlider.h>
+#include <ctkMessageBox.h>
 #include <ctkPopupWidget.h>
 #include <ctkSignalMapper.h>
 #include <ctkDoubleSpinBox.h>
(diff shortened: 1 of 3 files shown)


naucoin:
Note: I'm seeing a warning on the terminal after switching back to non light box mode, it's due to a 2d actor during the render overlay process. I checked how the vtkSeedWidget deals with removing seeds and it seems correct as far as I can tell.
Warning: In /Users/nicole/Slicer-git/Slicer-build/VTKv6/Common/ExecutionModel/vtkAlgorithm.cxx, line 1421
vtkOpenGLImageMapper (0x7fd8696b4a80): Attempt to get connection index 0 for input port 0, which has 0 connections.

jcfr:
Thanks @naucoin 

Integrated in [r24726](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24726)

Note that in [r24727](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24727), I added a check to disable the popup if testing is enabled.

Message about `vtkOpenGLImageMapper` is independent of that patch and should be address later.

PR #400, opened 2015-11-04
ENH: set temporary directory variable

prastawa:


jcfr:
Thanks for the patch. 

We will integrate after the release.

Associated discussion: http://slicer-devel.65872.n3.nabble.com/Temporary-directory-for-CLI-modules-td4035464.html

jcfr:
Having an other look at your patch, this will be problematic when the CLI is executed directly without Slicer. 

In that case, the variable won't be set.

I would prefer that we discuss how to improve SlicerExecutionModel by introducing the concept of temp directory.

PR #401, opened 2015-11-06
BUG: Fix slice bounds for volumes not RAS aligned

finetjul:
See [#4077](http://www.na-mic.org/Bug/view.php?id=4077)


lassoan:
I've checked and this has been integrated already.

PR #404, opened 2015-11-10
BUG: Fix crash on saving node named with a single number

naucoin:
A previous fix[1] to avoid crashes on windows when saving
volumes named "1:xxxx" leads to a crash when a node name
is just a single number (the first part of the check passes, but
crashes when trying to get the second character in the name string).
This fix adds a string length check so that valid node names like
"1" or "2" will not cause a crash on save.

[1] http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24294

Issue #3991


jcfr:
Looks good. 

@naucoin Please, commit to the trunk.

naucoin:
@jcfr I used the github merge button, but it doesn't look like it showed up. I'll give it a few minutes then try on the command line.

cpinter:
I did the same thing last week and basically broke the github repo until Jc restored it. Apparently the github mirror is only one-way, so the merge needs to be done in the SVN trunk.

naucoin:
oh no! @jcfr  my apologies!

jcfr:
No problem :smile: 

That is a good way to test with the sync script can now account for this. And it look like they did. :v:

cpinter:
Nice :)

naucoin:
Phew!

PR #405, opened 2015-11-10
BUG: 2466, MRMLIDImageIO crashes with DWMRI volume

ihnorton:
http://www.na-mic.org/Bug/view.php?id=2466

Restore handling of b-values and gradients, and make code consistent
with DWIConvert: use the `sqrt(b/b_max)` scaling factor as documented
on the NA-MIC wiki page:

http://wiki.na-mic.org/Wiki/index.php/NAMIC_Wiki:DTI:Nrrd_format


jcfr:
Thanks :+1: 

Integrated in [r24724](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24724)

PR #395, opened 2015-11-02
Fiber Bundle Label Selection

zhangfanmark:
Hi @jcfr @pieper ,

Could you check the updates of the FiberBundleLabelSelection module? Please let us know if there is anything needed to be further updated. 
Thanks a lot!

Regards,
Fan

PR #406, opened 2015-11-11
EHN: Adding tests for vtp inputs/outputs for FiberBundleLabelSelect 

zhangfanmark:
Adding tests for vtp inputs and outputs, based on the last integration http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24720


jcfr:
Thanks. I will review shortly and integrate

jcfr:
@zhangfanmark I noticed that the email associated with this patch is also the "veggie" one. I will make sure to use the correct email when integrating. In the mean time, if not already done, make sure to fix your local git configuration as discussed in https://github.com/Slicer/Slicer/pull/395#issuecomment-155883544

jcfr:
Thanks again for the patch.

Integrated in [r24755](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24755)

PR #407, opened 2015-11-11
Util extension with displaying dialogs and progress

che85:
I think those methods can be quite useful since there is similar code in several projects duplicated and duplicated again.

### DIFF
--- a/Applications/SlicerApp/Testing/Python/FiducialLayoutSwitchBug1914.py
+++ b/Applications/SlicerApp/Testing/Python/FiducialLayoutSwitchBug1914.py
@@ -148,8 +148,8 @@ class FiducialLayoutSwitchBug1914Widget:
     except Exception, e:
       import traceback
       traceback.print_exc()
-      qt.QMessageBox.warning(slicer.util.mainWindow(),
-          "Reload and Test", 'Exception!\n\n' + str(e) + "\n\nSee Python Console for Stack Trace")
+      slicer.util.warningDisplay("Reload and Test", 'Exception!\n\n' + str(e) +
+                                 "\n\nSee Python Console for Stack Trace")
 
 
 #
--- a/Applications/SlicerApp/Testing/Python/JRC2013Vis.py
+++ b/Applications/SlicerApp/Testing/Python/JRC2013Vis.py
@@ -135,11 +135,7 @@ class JRC2013VisWidget:
       configFilePath = dicomFilesDirectory + '/Dcmtk-db/dcmqrscp.cfg'
       processCurrentPath = dicomFilesDirectory + '/Dcmtk-db/'
 
-      msgBox = qt.QMessageBox()
-      msgBox.setText('Do you want to choose local DCMTK database folder?')
-      msgBox.setStandardButtons(qt.QMessageBox.Yes | qt.QMessageBox.No)
-      val = msgBox.exec_()
-      if(val == qt.QMessageBox.Yes):
+      if slicer.util.confirmYesNoDisplay('Do you want to choose local DCMTK database folder?'):
         print 'Yes'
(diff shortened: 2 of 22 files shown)


lassoan:
Thank you, this will be useful.

We cannot merge this as is, as despite all the good intentions, it would just increase code duplication (because people would still copy-paste code from existing modules). Please remove all methods from existing modules that the new utility functions can replace and change the modules to use the new utility functions instead.

Ideally, scripted modules in all extensions should be checked, but that's a lot of work (I've done it once for the labelmap node related changes and it took several hours), so I'm OK if only scripted modules (and extension templates) in the core are updated to use the new utility functions.

fedorov:
@lassoan this is different from the labelmap node, since labelmap node change would otherwise break existing functionality. These helpers are style choices. Not adopting them is a choice or lack of proper information. Updating extension template is a great idea, but beyond that I think it is unfair to ask Christian to edit the existing code and introduce a chance of breaking existing functionality. If you don't think this can be merged, let's just close this PR, or keep it open until there's more time.

lassoan:
I'm not asking to update extensions, just modules in the Slicer core. As he already checked that these methods exist in modules it should be very quick to actually go through those modules and change them. It would also ensure that the new methods can actually replace the existing methods and they are not just additional variants that just increase duplication.

pieper:
Just to add one more opinion to this topic: as you all know I almost always
prefer python to C++, but this is an example where it would make sense to
put these into someplace where they could be used consistently in either
language.

fedorov:
@lassoan I don't have a good feeling of the effort involved. If Christian thinks this can be done in couple hours, please go ahead.

@pieper this is specifically a python helper util functions file. In principle (although, maybe not in this particular case), I completely agree with you, but 1) Christian's contribution is consistent with the choices made when this util.py was introduced, and 2) as always in Slicer development, we have to prioritize tasks based on the objectives of the projects funding Christian. Again, if this is something that can be achieved in couple hours, I have no problem if Christian moves those methods into C++ and exposes in python...

pieper:
@fedorov reasonable points.

che85:
Since I never used Slicer in C++ development , I don't know, where to implement all those methods. 

If you tell me where to implement such methods, I would think about that and do it besides my actual task. I think we and the rest of the world reimplements everything over and over again and I would like to support stopping it here.

pieper:
If you don't already have a C++ environment set up (and have experience with it) then it'll take more than the few hours that @fedorov suggested allocating to this.  I think the time would be better spent on the other project @lassoan suggested of replacing other MessageBox uses in slicer core scripted modules.  When looking at the code you may find other common patterns that could be migrated into the slicer.util.

fedorov:
@pieper :+1:

che85:
I think, that should do it. I replaced all occurrences of manual creation of progress and notification/confirmation dialogs

pieper:
Looks reasonable to me.  After 4.5 of course. ; )

fedorov:
Nice! Thank you Christian!

lassoan:
Very nice, thank you. The code got simpler at many places.

jcfr:
:+1:  This was very much needed. It will also allow to avoid blocking the interface when running automatic testing.

jcfr:
Thanks again for the patch. It will be integrated as soon as the dashboard settle down and there are no test failure.

jcfr:
And note that the following topic where changes have been re-organized will be integrated. See https://github.com/Slicer/Slicer/compare/master...jcfr:407-utilExtension-jcfr

jcfr:
@che85  Thanks again for your contribution :+1: 

It has been integrated in [r24897](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24897) to [r24901](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24901)

Keep your :eyes: open on Wednesday morning to confirm  the dashboard is green on all platforms.

PR #408, opened 2015-11-13
ENH: Add terminology information to color frame in Editor

naucoin:
In order to expand the semantic meanings of colors in Slicer, add medical terminology support that was originally developed in the Reporting extension. This change adds terminology for the Generic Anatomy Colors color node, shows it in the editor and on mouse over in the Colors module, and allows DICOM Segementation objects to build up new terminologies.

Details:
Added terminology files, read them on start up, associate with the generic anatomy colors node.
Add accessors to the color logic as well as basic methods to allow creating new terminologies (used by the Reporting extension).
Created Python callable utility methods to allow scripted access[1].
Add read only widgets to the Editor, shown if the selected color node has terminology associated with it (using node attributes).
Added a test for the terminology functionality to vtkSlicerColorLogicTest (rather than in vtkMRMLColorLogicTest1 since the Colors module logic adds the terminology files).
Pass the color logic to the color model so that it can look up terminology information to set the tool tips on the color table entries.

[1] Utility accessor functions:
- GetTerminologyFromLabel takes in the strings defining the color label categorization and the standard term elements and returns the specific term.
- Get the colon concatenated full standard term values via
GetSegmentedPropertyCategory, GetSegmentedPropertyType, GetSegmentedPropertyTypeModifier,
GetAnatomicRegion, GetAnatomicRegionModifier

Issue #4047


naucoin:
@fedorov @pieper  please review

naucoin:
colorterminology

fedorov:
Thank you Nicole! I will plan to finish the review tomorrow.

Quick comment - probably makes sense to add an acknowledgment that the terminology xml files are from the AIM on ClearCanvas project, and add a reference to the original repository: https://github.com/NCIP/annotation-and-image-markup

Base/Logic/Resources/ColorFiles/Terminology/GenericAnatomyColors-SNOMED.csv
```diff
@@ -0,0 +1,291 @@
+# Integer Label,Text Label,Segmented Property Category -- CID 7150++,Segmented Property Type,Segmented Property Type Modifier,Color
+# Version 2
+SlicerLUT=GenericAnatomyColors
+1,tissue,(T-D0050;SRT;Tissue),(T-D0050;SRT;Tissue),,rgb(128;174;128)
```
fedorov: Need to add to the commit message that this commit adds a number of SNOMED-CT (https://www.nlm.nih.gov/research/umls/Snomed/snomed_main.html) codes to Slicer source code repository. All of those codes we are adding are also included in the DICOM standard, and therefore are subject to the SNOMED-DICOM reuse agreement [1].

[1] ftp://medical.nema.org/medical/dicom/Geninfo/Policies/code_policy_20000921.pdf

Base/Logic/Resources/CMakeLists.txt
```diff
@@ -18,3 +18,26 @@ install(
   DESTINATION ${Slicer_INSTALL_SHARE_DIR}/ColorFiles
   COMPONENT Runtime
   )
+
+# Related terminology files
+file(GLOB TERMINOLOGYLUTFILES RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}" "ColorFiles/Terminology/*.csv")
+file(GLOB REGIONANDCATEGORYFILES RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}" "ColorFiles/Terminology/*.xml")
```
jcfr: I would recommend not to glob files. If you do so, adding new files will not cause a reconfiguration and the file will not be collected into the variable `REGIONANDCATEGORYFILES` and `TERMINOLOGYLUTFILES`.

See https://cmake.org/cmake/help/git-next/command/file.html?highlight=glob 

Instead you could so something like this:

```
set(Slicer_TERMINOLOGY_DIR "ColorFiles/Terminology")

set(terminology_filenames

  # LUT
  GenericAnatomyColors-SNOMED.csv

  # Region and Category
  AnatomicRegionAndModifier.xml
  SegmentationCategoryTypeModifier.xml

  )
foreach(filename ${terminology_filenames})
  message(STATUS "Configuring terminology file: ${filename}")
  set(terminology_filepath ${CMAKE_CURRENT_SOURCE_DIR}/${Slicer_TERMINOLOGY_DIR}/${filename})
  configure_file(
    ${terminology_filepath}
    ${CMAKE_BINARY_DIR}/${Slicer_SHARE_DIR}/${Slicer_TERMINOLOGY_DIR}/${filename}
    COPYONLY
    )
  install(
    FILES ${terminology_filepath}
    DESTINATION ${Slicer_INSTALL_SHARE_DIR}/ColorFiles/${Slicer_TERMINOLOGY_DIR}
    COMPONENT Runtime
    )
endforeach()
```

Libs/MRML/Logic/vtkMRMLColorLogic.h
```diff
@@ -39,6 +39,57 @@ class vtkMRMLColorTableNode;
 class VTK_MRML_LOGIC_EXPORT vtkMRMLColorLogic : public vtkMRMLAbstractLogic
 {
 public:
+  typedef struct
+    {
+      std::string CodeValue;
+      std::string CodingSchemeDesignator;
+      std::string CodeMeaning;
+
+      void StandardTerm() {
```
jcfr: Simply declare the struct contructor and rely on `std::string` being initialized to an empty string.
```
struct StandardTerm
{
  StandardTerm(){}

  StandardTerm(const std::string& value, const std::string& designator, const std::string& meaning)
    : CodeValue(value), 
      CodingSchemeDesignator(designator),
      CodeMeaning(meaning)
  {
  }
  [...]
};
```

No need for `typedef` and `void`
See http://www.cplusplus.com/doc/tutorial/structures/
See

Libs/MRML/Logic/vtkMRMLColorLogic.h
```diff
@@ -39,6 +39,57 @@ class vtkMRMLColorTableNode;
...
+        CodingSchemeDesignator = "";
+        CodeMeaning = "";
+      }
+      void StandardTerm(std::string value, std::string designator, std::string meaning) {
+        CodeValue = value;
+        CodingSchemeDesignator = designator;
+        CodeMeaning = meaning;
+      }
+
+      void PrintSelf(ostream &os) {
```
jcfr: std::osteam

jcfr: May be you could use `vtkIndent` ? and also implement the `<<` operator for these struct.

Libs/MRML/Logic/vtkMRMLColorLogic.h
```diff
@@ -39,6 +39,57 @@ class vtkMRMLColorTableNode;
 class VTK_MRML_LOGIC_EXPORT vtkMRMLColorLogic : public vtkMRMLAbstractLogic
 {
 public:
+  typedef struct
```
jcfr: Would it make sense to add Doxygen pointing to the standard ?

Libs/MRML/Logic/vtkMRMLColorLogic.h
```diff
@@ -125,6 +176,118 @@ class VTK_MRML_LOGIC_EXPORT vtkMRMLColorLogic : public vtkMRMLAbstractLogic
   /// procedural node with a blank color transfer function.
   static vtkMRMLProceduralColorNode* CopyProceduralNode(vtkMRMLColorNode* colorNode, const char* copyName);
 
+  /// Return the label's terminology in this color table.
+  bool LookupCategorizationFromLabel(int label, ColorLabelCategorization&, const char *lutName = "GenericAnatomyColors");
+  /// Returns a string with the terminology term requested, or an empty string if not found.
```
jcfr: For new doxygen. Consider adding `\brief` for the title.

```
/// \brief Returns a string with the terminology term requested, or an empty string if not found.
///
/// Valid values for categorization are SegmentedPropertyCategory, SegmentedPropertyType,
/// SegmentedPropertyTypeModifier, AnatomicRegion, AnatomicRegionModifier.
/// Valid values for standard term are CodeValue, CodeMeaning, CodingSchemeDesignator
///
/// \sa LookupCategorizationFromLabel
```

Libs/MRML/Logic/vtkMRMLColorLogic.h
```diff
@@ -125,6 +176,118 @@ class VTK_MRML_LOGIC_EXPORT vtkMRMLColorLogic : public vtkMRMLAbstractLogic
...
+  /// Valid values for categorization are SegmentedPropertyCategory, SegmentedPropertyType,
+  /// SegmentedPropertyTypeModifier, AnatomicRegion, AnatomicRegionModifier.
+  /// Valid values for standard term are CodeValue, CodeMeaning, CodingSchemeDesignator
+  /// \sa LookupCategorizationFromLabel
+  std::string GetTerminologyFromLabel(std::string categorization, std::string standardTerm, int label,
+                                      const char *lutName = "GenericAnatomyColors");
```
jcfr: In new API, I suggest we stick with `const std::string&`, in this case it would be:

```
std::string GetTerminologyFromLabel(const std::string& categorization,
                                    const std::string& standardTerm, int label,
                                    const std::string& lutName = "GenericAnatomyColors");
```

Libs/MRML/Logic/vtkMRMLColorLogic.h
```diff
@@ -93,7 +144,7 @@ class VTK_MRML_LOGIC_EXPORT vtkMRMLColorLogic : public vtkMRMLAbstractLogic
   /// Return a default color node id for a chart
   virtual const char * GetDefaultChartColorNodeID();
 
-  /// Add a file to the input list list, checking first for null, duplicates
+  /// Add a file to the input list Files, checking first for null, duplicates
   void AddColorFile(const char *fileName, std::vector<std::string> *Files);
```
jcfr: Why a pointer ? That means you will have to manage the memory ... would that work:

```
void AddColorFile(const std::string& fileName, const std::vector<std::string>& Files);
```

Libs/MRML/Logic/vtkMRMLColorLogic.h
```diff
@@ -125,6 +176,118 @@ class VTK_MRML_LOGIC_EXPORT vtkMRMLColorLogic : public vtkMRMLAbstractLogic
...
+  bool LookupCategorizationFromLabel(int label, ColorLabelCategorization&, const char *lutName = "GenericAnatomyColors");
+  /// Returns a string with the terminology term requested, or an empty string if not found.
+  /// Valid values for categorization are SegmentedPropertyCategory, SegmentedPropertyType,
+  /// SegmentedPropertyTypeModifier, AnatomicRegion, AnatomicRegionModifier.
+  /// Valid values for standard term are CodeValue, CodeMeaning, CodingSchemeDesignator
+  /// \sa LookupCategorizationFromLabel
+  std::string GetTerminologyFromLabel(std::string categorization, std::string standardTerm, int label,
+                                      const char *lutName = "GenericAnatomyColors");
+  /// From a given terminology, find the label value that corresponds to it for this color table.
+  bool LookupLabelFromCategorization(ColorLabelCategorization&, int&, const char *lutName = "GenericAnatomyColors");
```
jcfr: `ColorLabelCategorization&,` -> `const ColorLabelCategorization&,` 
That clearly shows it is an input.

jcfr: I would do this:

```
int LookupLabelFromCategorization(ColorLabelCategorization&, const char *lutName = "GenericAnatomyColors");
```

Then negative label means error ...

Libs/MRML/Logic/vtkMRMLColorLogic.h
```diff
@@ -125,6 +176,118 @@ class VTK_MRML_LOGIC_EXPORT vtkMRMLColorLogic : public vtkMRMLAbstractLogic
...
+  /// Print the terminology for this label in the color table to standard output.
+  bool PrintCategorizationFromLabel(int label, const char *lutName = "GenericAnatomyColors");
```
jcfr: Since you use `"GenericAnatomyColors"` everywhere as a default, may be you could add a 

`const char* DEFAULT_LUT_NAME` to your class and use that instead of a string.

Less chance of doing a typo

Libs/MRML/Logic/vtkMRMLColorLogic.h
```diff
@@ -125,6 +176,118 @@ class VTK_MRML_LOGIC_EXPORT vtkMRMLColorLogic : public vtkMRMLAbstractLogic
...
+
+  /// Utility methods to look up the terminology for the SegmentedPropertyCategory
+  /// for a specific label in a color node.
+  /// Returns an empty string if not found or defined.
+  std::string GetSegmentedPropertyCategoryCodeValue(int label, const char *lutName = "GenericAnatomyColors");
```
jcfr: By changing to something like:

```
std::string GetSegmentedPropertyCategoryProperty(int label, int propertyIdentifier, const char *lutName = DEFAULT_LUT_NAME);

Where `propertyIdentifier` would be any value from an enum like:

```
enum SegmentedPropertyCategoryPropertyIdentifiers
{
  CodeValue = 0,
  CodeMeaning,
  CodingSchemeDesignator
}
```

You would simplify you API.

Adding a `bool IsValidPropertyIdentifier(int)` would then be used to check the input ..

File: Libs/MRML/Logic/vtkMRMLColorLogic.h [hunk 12]
```diff
@@ -189,7 +390,17 @@ class VTK_MRML_LOGIC_EXPORT vtkMRMLColorLogic : public vtkMRMLAbstractLogic
   /// vtkMRMLApplication::GetColorFilePaths
   char *UserColorFilePaths;
 
+  // mappings used for terminology color look ups
+  typedef std::map<int,ColorLabelCategorization> ColorCategorizationMapType;
+  std::map<std::string, ColorCategorizationMapType> colorCategorizationMaps;
```
jcfr: `ColorCategorizationMaps`

Libs/MRML/Logic/vtkMRMLColorLogic.h
```diff
@@ -189,7 +390,17 @@ class VTK_MRML_LOGIC_EXPORT vtkMRMLColorLogic : public vtkMRMLAbstractLogic
...
+
   static std::string TempColorNodeID;
+
+ private:
+
+  std::string RemoveLeadAndTrailSpaces(std::string);
+  bool ParseTerm(std::string, StandardTerm&);
```
jcfr: Consider moving this as protected. That way in you test, you can derive `vtkMRMLColorLogic` into `vtkMRMLTestColorLogic` and test the method. For example, see https://github.com/jcfr/Slicer/commit/f2aa068e24277227931ca25049a970ce25429cba#diff-7aa2fc5db0ae46d14ee8f13646159833

Libs/MRML/Logic/vtkMRMLColorLogic.h
```diff
@@ -125,6 +176,118 @@ class VTK_MRML_LOGIC_EXPORT vtkMRMLColorLogic : public vtkMRMLAbstractLogic
...
+  /// terminology with a color node
+  /// \sa AssociateTerminologyWithColorNode
+  bool CreateNewTerminology(std::string lutName);
+  /// Return true if a terminology mapping already exists for this lutName,
+  // false if empty name or none found.
+  bool TerminologyExists(std::string lutName);
+  /// Link the color node with the terminology
+  /// Returns true on success, false if empty lut name, unable to find the color node
+  /// \sa CreateNewTerminology
+  bool AssociateTerminologyWithColorNode(std::string lutName);
```
jcfr: Can method `AssociateTerminologyWithColorNode` be protected ? 

This is an implementation details of `CreateNewTerminology`

Libs/MRML/Logic/vtkMRMLColorLogic.cxx
```diff
@@ -1076,3 +1094,699 @@ vtkMRMLProceduralColorNode* vtkMRMLColorLogic::CopyProceduralNode(vtkMRMLColorNo
...
+    }
+}
+
+//------------------------------------------------------------------------------
+bool vtkMRMLColorLogic::CreateNewTerminology(std::string lutName)
+{
+  if (lutName.length() == 0)
+    {
+    return false;
+    }
```
jcfr: Use `TerminologyExists` to check for existence.

Libs/MRML/Logic/vtkMRMLColorLogic.h
jcfr: I would suggest you add instead:

```
ColorLabelCategorization LookupCategorizationFromLabel(int label, const char *lutName = "GenericAnatomyColors");
```

Then you would be able to get any of the property of the returned `ColorLabelCategorization`.

jcfr: And to match what is done in MRMLScene: May be the name `GetCategorizationByLabel` ?

Libs/MRML/Logic/vtkMRMLColorLogic.h
```diff
@@ -125,6 +176,118 @@ class VTK_MRML_LOGIC_EXPORT vtkMRMLColorLogic : public vtkMRMLAbstractLogic
   /// procedural node with a blank color transfer function.
   static vtkMRMLProceduralColorNode* CopyProceduralNode(vtkMRMLColorNode* colorNode, const char* copyName);
 
+  /// Return the label's terminology in this color table.
+  bool LookupCategorizationFromLabel(int label, ColorLabelCategorization&, const char *lutName = "GenericAnatomyColors");
```
jcfr: I would suggest you add instead:

```
ColorLabelCategorization LookupCategorizationFromLabel(int label, const char *lutName = "GenericAnatomyColors");
```

Then you would be able to get any of the property of the returned `ColorLabelCategorization`.

Libs/MRML/Logic/vtkMRMLColorLogic.h
```diff
@@ -39,6 +39,57 @@ class vtkMRMLColorTableNode;
...
+
+  struct ColorLabelCategorization
+    {
+    unsigned LabelValue;
+    StandardTerm SegmentedPropertyCategory;
+    StandardTerm SegmentedPropertyType;
+    StandardTerm SegmentedPropertyTypeModifier;
+    StandardTerm AnatomicRegion;
+    StandardTerm AnatomicRegionModifier;
+
```
jcfr: What about adding a method:

```
std::string GetStandardTermProperty(int standardTermId, int propertyId)
```

where 

`standardTermId` 

would be any of 

```
enum StandardTermIdentifiers
{
  SegmentedPropertyCategory = 0,
  SegmentedPropertyType,
  SegmentedPropertyTypeModifier,
  AnatomicRegion,
  AnatomicRegionModifier
}
```

and 

`propertyId`

would be any of 

```
enum StandardTermIdentifiers
{
  CodeValue = 0,
  CodingSchemeDesignator,
  CodeMeaning
}
```


This approach would then allow you to remove these methods:

```
GetSegmentedPropertyCategoryCodeValue
GetSegmentedPropertyCategoryCodeMeaning
GetSegmentedPropertyCategory

GetSegmentedPropertyTypeCodeValue
GetSegmentedPropertyTypeCodeMeaning
GetSegmentedPropertyTypeCodingSchemeDesignator

GetSegmentedPropertyTypeModifierCodeValue
GetSegmentedPropertyTypeModifierCodeMeaning
GetSegmentedPropertyTypeModifierCodingSchemeDesignator

[...]
```

and simply have a method like this one:

```
std::string vtkMRMLColorLogic::GetStandardTermPropertyByLabel(int standardTermIdentifier, int standardTermProperty, int label, const char *lutName)
{

  ColorLabelCategorization labelCat = 
    GetCategorizationByLabel(int label, const char *lutName = "GenericAnatomyColors");

  return labelCat.GetStandardTermProperty(standardTermIdentifier, standardTermProperty);
}
```

And since enum are wrapped in python, that would also work nicely from python.

Testing would also be easier since you can loop through idenitifier and property.



Similarly, you could add:

```
std::string GetStandardTerm(int standardTermIdentifier,  int label, const char *lutName)
std::string GetStandardTermModifier(int standardTermIdentifier,  int label, const char *lutName)
```

to respectively remove method like:

```
GetSegmentedPropertyType()
GetAnatomicRegion()
[...]
```
 and

```
GetSegmentedPropertyTypeModifier()
GetAnatomicRegionModifier()
[...]
```

Overall it means that the concept would be accessed using the enum , stored using struct (these could even be moved into the cpp file as an implementation details into a `vtkInternal` class.

Your API would then be fully wrapped from python (thanks to enum)

Modules/Scripted/EditorLib/EditColor.py
```diff
@@ -50,27 +51,85 @@ def cleanup(self, QObject=None):
   def create(self):
```
jcfr: Could you consider moving the ColorBox and new terminology into a python widget like it is done for `LabelStructureListWidget` ? . See https://github.com/Slicer/Slicer/blob/master/Modules/Scripted/EditorLib/LabelStructureListWidget.py

Base/Logic/Resources/ColorFiles/Terminology/GenericAnatomyColors-SNOMED.csv
jcfr: An idea: may be you could do a first commit adding the functionality to manage terminology, and then a second one adding the data file where you would explicitly credit them ?

fedorov: Yet another idea - maybe we should just add this statement to the comment section of the .csv file?

fedorov:
@jcfr thank you for the thorough review. This is very helpful. I really appreciate this.

As I am going over the code, I also have quite a few suggestions for improving the structure and API, but the situation is that this functionality is in the critical path for our RSNA activities and we need some time to do real testing and share DICOM segmentation object support with our collaborators in advance of the conference. I would rather prefer have functionality that we can access and test, even though the underlying API is imperfect. Another consideration is that a lot of this will probably change after RSNA anyway, since we want to coordinate this development with the SlicerRT Segmentations project, which should also move into the core. That effort will likely overhaul most of colors handling anyway.

@jcfr @pieper @naucoin what do you think is the best way forward? @naucoin how much time do you need to address JC's comments? In my review, I will only add critical suggestions, but will hold off API improvements for now.

In QIICR we are committed to work on refining this, and Nicole will continue after RSNA, but I am just afraid if we don't release early, we may run into unexpected problems and may not have the functionality at RSNA. Releasing at the last moment is dangerous.

Libs/MRML/Logic/vtkMRMLColorLogic.cxx
```diff
@@ -1076,3 +1094,699 @@ vtkMRMLProceduralColorNode* vtkMRMLColorLogic::CopyProceduralNode(vtkMRMLColorNo
...
+  std::string ret = in;
+  ret.erase(ret.begin(), std::find_if(ret.begin(),ret.end(),
+    std::not1(std::ptr_fun<int,int>(std::isspace))));
+  ret.erase(std::find_if(ret.rbegin(),ret.rend(),
+    std::not1(std::ptr_fun<int,int>(std::isspace))).base(), ret.end());
+  return ret;
+}
+
+//---------------------------------------------------------------------------
+bool vtkMRMLColorLogic::ParseTerm(std::string str, StandardTerm& term)
```
fedorov: This function should probably be made more robust to the file content. As is, it may crash if the content of the file is not what is expected.

Libs/MRML/Logic/vtkMRMLColorLogic.cxx
```diff
@@ -1076,3 +1094,699 @@ vtkMRMLProceduralColorNode* vtkMRMLColorLogic::CopyProceduralNode(vtkMRMLColorNo
...
+      labelCat = this->colorCategorizationMaps[lutName][label];
+      success = true;
+      }
+    }
+  return success;
+}
+
+//---------------------------------------------------------------------------
+std::string vtkMRMLColorLogic::
+  GetTerminologyFromLabel(std::string categorization, std::string standardTerm, int label, const char *lutName)
```
fedorov: Along the comments JC did below, it makes sense to use enums instead of strings here.

Libs/MRML/Logic/vtkMRMLColorLogic.cxx
```diff
@@ -1076,3 +1094,699 @@ vtkMRMLProceduralColorNode* vtkMRMLColorLogic::CopyProceduralNode(vtkMRMLColorNo
...
+      {
+      returnString = term.CodingSchemeDesignator;
+      }
+    }
+  return returnString;
+}
+
+//---------------------------------------------------------------------------
+bool vtkMRMLColorLogic::
+LookupLabelFromCategorization(ColorLabelCategorization& labelCat, int& label, const char *lutName)
```
fedorov: I am not sure we need this function. It was present in Reporting logic since we did not create color tables on the fly. Currently it is used only by the test.

As a side-note, we should probably mention in the commit that some of the methods introduced here were migrated from Reporting logic.

Libs/MRML/Logic/vtkMRMLColorLogic.cxx
```diff
@@ -1076,3 +1094,699 @@ vtkMRMLProceduralColorNode* vtkMRMLColorLogic::CopyProceduralNode(vtkMRMLColorNo
...
+
+  if (labelFound != -1)
+    {
+    label = labelFound;
+    }
+  return (labelFound==-1) ? false: true;
+}
+
+//---------------------------------------------------------------------------
+bool vtkMRMLColorLogic::PrintCategorizationFromLabel(int label, const char *lutName)
```
fedorov: This can be used for the << operator JC suggested.

Libs/MRML/Widgets/qMRMLColorModel.cxx
```diff
@@ -325,6 +342,49 @@ void qMRMLColorModel::updateItemFromColor(QStandardItem* item, int color, int co
...
+        std::string category = this->mrmlColorLogic()->GetSegmentedPropertyCategoryCodeMeaning(color, lutName);
+        std::string type = this->mrmlColorLogic()->GetSegmentedPropertyTypeCodeMeaning(color, lutName);
+        std::string typeMod = this->mrmlColorLogic()->GetSegmentedPropertyTypeModifierCodeMeaning(color, lutName);
+        std::string region = this->mrmlColorLogic()->GetAnatomicRegionCodeMeaning(color, lutName);
+        std::string regionMod = this->mrmlColorLogic()->GetAnatomicRegionModifierCodeMeaning(color, lutName);
+        QString terminology = QString("Terminology:");
+        // only show the not empty terminology terms
+        if (!category.empty())
+          {
+          terminology = terminology + QString("\nCategory: %1").arg(category.c_str());
```
fedorov: I would use here full attribute names exactly as they are defined in DICOM: SegmentedPropertyCategory, SegmentedPropertyType + Modifier, AnatomicRegion + Modifier

fedorov:
As a general comment related to API: when this becomes part of Segmentations (http://wiki.slicer.org/slicerWiki/index.php/Documentation/Labs/Segmentations), there will be a single MRML node for each segment, and each segment will represent a single label. As a consequence, we should be able to simply keep various terms as node attributes pointing to the terminology triples. This will also be in line with how we are currently handling other DICOM attributes both in the core (UIDs) and in the subject hierarchy.

fedorov:
Trying to compile this branch from scratch, I have CTK build errors. I reported this before (http://slicer-devel.65872.n3.nabble.com/CTK-build-error-while-building-Slicer-td4034419.html), but then it went away, it seemed so at least, now I cannot build Slicer agian.

```
[ 64%] Building CXX object Libs/DICOM/Core/CMakeFiles/CTKDICOMCore.dir/ctkDICOMUtil.cpp.o
/Users/fedorov/local/builds/Slicer4-Debug/CTK/Libs/DICOM/Core/ctkDICOMUtil.cpp:36:3: error: use of undeclared identifier
      'log4cplus'; did you mean 'dcmtk::log4cplus'?
  log4cplus::Logger log = log4cplus::Logger::getRoot();
  ^~~~~~~~~
  dcmtk::log4cplus
/Users/fedorov/local/builds/Slicer4-Debug/DCMTK/oflog/include/dcmtk/oflog/logmacro.h:60:11: note: 'dcmtk::log4cplus'
      declared here
namespace log4cplus
          ^
/Users/fedorov/local/builds/Slicer4-Debug/CTK/Libs/DICOM/Core/ctkDICOMUtil.cpp:36:27: error: use of undeclared
      identifier 'log4cplus'; did you mean 'dcmtk::log4cplus'?
  log4cplus::Logger log = log4cplus::Logger::getRoot();
                          ^~~~~~~~~
                          dcmtk::log4cplus
/Users/fedorov/local/builds/Slicer4-Debug/DCMTK/oflog/include/dcmtk/oflog/logmacro.h:60:11: note: 'dcmtk::log4cplus'
      declared here
namespace log4cplus
          ^
/Users/fedorov/local/builds/Slicer4-Debug/CTK/Libs/DICOM/Core/ctkDICOMUtil.cpp:59:3: error: use of undeclared identifier
      'log4cplus'; did you mean 'dcmtk::log4cplus'?
  log4cplus::Logger log = log4cplus::Logger::getRoot();
  ^~~~~~~~~
  dcmtk::log4cplus
/Users/fedorov/local/builds/Slicer4-Debug/DCMTK/oflog/include/dcmtk/oflog/logmacro.h:60:11: note: 'dcmtk::log4cplus'
      declared here
namespace log4cplus
          ^
/Users/fedorov/local/builds/Slicer4-Debug/CTK/Libs/DICOM/Core/ctkDICOMUtil.cpp:59:27: error: use of undeclared
      identifier 'log4cplus'; did you mean 'dcmtk::log4cplus'?
  log4cplus::Logger log = log4cplus::Logger::getRoot();
                          ^~~~~~~~~
                          dcmtk::log4cplus
/Users/fedorov/local/builds/Slicer4-Debug/DCMTK/oflog/include/dcmtk/oflog/logmacro.h:60:11: note: 'dcmtk::log4cplus'
      declared here
namespace log4cplus
          ^
4 errors generated.
```

fedorov:
I didn't fix the wrong qmake path, rebuilding now, hopefully this will fix it. Sorry for the noise - I will delete this and the previous comment.

naucoin:
@fedorov I was working through the fixes from top to bottom on this page, doing the simpler changes as I went and evaluating the others. The major API reworking will take me this afternoon, but I agree that most of it is polishing that can happen after we're fairly sure that we need these API calls ( @jcfr  I was working from the starting point of Python accessible API calls that are needed in the scripts in the Reporting module to load DICOM Seg objects). 
My take: I can implement the enums that JC suggested as they should work from Python, but not do major modifications on the calls that I'm already using from Reporting. I'll do my best on the comments that address code that I dropped in from the Reporting logic, I honestly have been treating them like black boxes.

fedorov:
My testing of this branch didn't find any problems, things work as expected.

naucoin:
@fedorov @jcfr @pieper 
I've made a lot of the changes in this commit:
https://github.com/naucoin/Slicer/commit/15fae5ecc921521080b9d2d64e903371529445d2
I want to hold off on the complete re-work of the API using enums so that I can make sure to do it correctly for all the needed use cases and not break anything. I'll revisit it early next week and incorporate any further feedback.

pieper:
Wow!  Lots of activity here.  I didn't read through in detail but maybe
Andrey, Nicole and I should touch base on Monday.

Libs/MRML/Logic/vtkMRMLColorLogic.cxx
naucoin: @fedorov It seems like this could use the << operator rather than be used by it.

Libs/MRML/Widgets/qMRMLColorModel.cxx
naucoin: Strings updated to use the  full attribute names as you suggested.

naucoin:
@fedorov @pieper @jcfr just went through and commented specifically on all the line notes, apologies for not getting to that level of detail on Friday. From our meeting today, this branch is ready to integrate into the nightly, pending any changes suggested by Steve as he wraps up his changes in the Reporting module.

fedorov:
As far as I am concerned, please integrate once you get ok from Steve.

pieper:
Looks good to me - we should merge this today so it can be tested in tomorrow's nightly builds.

@naucoin , will you do the commit to svn?

naucoin:
Okay, will squash it down to two commits and get it into svn ASAP.

naucoin:
Done:
http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24742
https://github.com/Slicer/Slicer/commit/1f0a534f1979e008a371d6873438633be58ae227

jcfr:
As a follow-up, review of the code allowed to spot some issue with installation of terminology files .. @naucoin  is working on a fix.

naucoin:
@jcfr: installation fix pushed
https://github.com/Slicer/Slicer/commit/fe937b8ff1502ffcee00a740894a1ceff5aef1ae
http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24743

jcfr:
:+1:

fedorov:
:+1: thank you all!

Libs/MRML/Logic/vtkMRMLColorLogic.cxx
```diff
@@ -35,6 +35,7 @@
 
 // STD includes
```
fedorov:
```
include <algorithm>
```

fedorov:
@naucoin linux and windows builds are broken. One issue is missing algorithm include. The windows error I don't get:

```
vtkMRMLColorLogic.h(73) : error C2679: binary '<<' : no operator found which takes 
a right-hand operand of type 'std::string' (or there is no acceptable conversion)
```

jcfr:
@naucoin @fedorov FYI. Issue has been fixed last night.

jcfr:
See http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24745

fedorov:
Thank you JC, so that is the missing include. But I still don't understand what the error is for vtkMRMLColorLogic.h line 73. Did you check that adding algorithm also fixed that problem?

jcfr:
I added the include the the cpp file to fix the Linux error.

The windows error would have to be investigated and fixed on his own. I am assuming the `iostream` header should be included.

As a side note, the implementation of the method should also be moved to the cpp file.

naucoin:
@jcfr @fedorov thanks, looking into the windows error!

lassoan:
Windows error: probably you need to add .c_str() after the std::string type variables (e.g., CodeValue.c_str()).

naucoin:
@lassoan that's what I was testing first...

lassoan:
By the way, the method implementations should be in the .cxx file and not in the header. Maybe in the cxx there are already some additional headers included that can manage streaming of std::string. In that case adding .c_str() might not be needed.

naucoin:
Moved the implementation and added the .c_str() just in case:
http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24746

fedorov:
@naucoin down to just windows build this time, this time it looks like it's missing cctype include.

Do you think you could use the windows box Steve gave us access, to confirm no build errors on Windows today?

lassoan:
I can also run tests on Windows, just tell me what branch to test.

fedorov:
Andras, this is master. If you could add cctype include in the master branch vtkMRMLColorLogic (I think this is what breaks it now), and let us know if there are any other issues on Windows, it would be highly appreciated.

pieper:
Thanks Andras. I can test on windows too this afternoon if needed.

lassoan:
Adding #include <ctype.h> does not help. Instead, I had to change std::isspace() to isspace() and the build completes successfully. Let me know if you need to test a new version (probably after you check that removal of std:: does not cause any issue on Linux and Mac and commit the change to the trunk).

Background: std::isspace is declared in <local>, but this <local> library is not found in VS2008 STL. isspace() is declared in ctype.h.

fedorov:
Testing it on mac now

jcfr:
`ctype.h` is included in `vtkXMLParser.cxx` so that `isspace` is available. That shouldn't be an issue.
See https://github.com/Kitware/VTK/blob/1c54b47d1de7bb33359bb1cf24a7fe8826e927ca/IO/XMLParser/vtkXMLParser.cxx#L18

Let's just make sure these headers are included in the cpp file.

fedorov:
Confirmed - builds fine on mac 10.10.5

lassoan:
Unfortunately, there is also a linking issue:
2>vtkSlicerColorLogicTest1.obj : error LNK2019: unresolved external symbol "public: void __cdecl vtkMRMLColorLogic::ColorLabelCategorization::Print(class std::basic_ostream<char,struct std::char_traits<char> > &)" (?Print@ColorLabelCategorization@vtkMRMLColorLogic@@QEAAXAEAV?$basic_ostream@DU?$char_traits@D@std@@@std@@@Z) referenced in function "bool __cdecl `anonymous namespace'::TestTerminology(void)" (?TestTerminology@?A0x9379fd3b@@YA_NXZ)

fedorov:
It looks like in other places in slicer print takes ostream& os

@lassoan can you change the definition of the Print function to take ostream instead of std::ostream and see if it solves the issue?

jcfr:
I just pused a PR that integrate @lassoan change and should also fix the link error.

See #412

fedorov:
Wonderful, thank you @jcfr and @lassoan !

lassoan:
Neither Jc's changes nor replacing all std::ostream instances by ostream does NOT solve the link error.

As a workaround, we can move the print functions inline into the header file. Then the linking problem does not occur anymore.

fedorov:
Thank you Andras. The workaround is fine by me. If ok with JC, would you prefer @naucoin to create a pull request, or will you do it, since I assume you have already modified the code?

jcfr:
@fedorov Here is a topic including:
* Fix for isspace
* Fix for VS2013

See https://github.com/jcfr/Slicer/tree/fix-color-logic-build-link-error-ok

I will integrate it now. Then the last things left will be the link error @lassoan is working one.

jcfr:
Integrated in [r24750](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24750) and [r24751](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24751)

fedorov:
Thanks @jcfr - I tested it on mac, and it builds fine

fedorov:
:+1: 

image

PR #409, opened 2015-11-13
ENH: Simplified vtkSlicerApplicationLogic

lassoan:
This work is done as part of refactoring to prepare introduction of a new node (vtkMRMLTableNode).

Problem:
vtkSlicerApplicationLogic had node-specific logic for creating storage and display nodes.
This made it necessary to modify vtkSlicerApplicationLogic whenever a new node type is added.
CreateDefaultStorageNode and CreateDefaultDisplayNodes methods are introduced some time ago
to encapsulate all node-specific storage and display node creation. These methods were partly
used by vtkSlicerApplicationLogic, but not for all cases; also, CreateDefaultStorageNode and
CreateDefaultDisplayNodes methods were not implemented for all nodes.

Solution:
Moved node-specific storage and display node creation code to CreateDefaultStorageNode and
CreateDefaultDisplayNodes methods of the corresponding nodes.
Replaced node-specific logic from vtkSlicerApplicationLogic by CreateDefaultStorageNode and
CreateDefaultDisplayNodes method calls.

### DIFF
--- a/Base/Logic/vtkSlicerApplicationLogic.cxx
+++ b/Base/Logic/vtkSlicerApplicationLogic.cxx
@@ -15,33 +15,17 @@
 // MRML includes
 #include <vtkCacheManager.h>
 #include <vtkDataIOManagerLogic.h>
-#include <vtkMRMLColorTableStorageNode.h>
 #ifdef Slicer_BUILD_CLI_SUPPORT
 # include <vtkMRMLCommandLineModuleNode.h>
 #endif
-#include <vtkMRMLDiffusionTensorVolumeDisplayNode.h>
-#include <vtkMRMLDiffusionTensorVolumeNode.h>
-#include <vtkMRMLDiffusionTensorVolumeSliceDisplayNode.h>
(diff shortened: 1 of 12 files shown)


naucoin:
I am in favour of this refactoring! I'll take a closer look early next week.

pieper:
Agreed - this looks like a great step.

naucoin:
I built and ran the tests I wanted to be sure worked, all fine. I did manage to trigger a crash when I worked on the GUI though, not sure if it's due to something else, while closing the scene with model maker open:
load a greyscale and a label map and a model
go to model maker and make a model from the label map
save to mrml
file - close scene
crash

I'll check if this is also an issue on the trunk.

naucoin:
Not getting the crash on the trunk.

jcfr:
@naucoin I think the crash is also related to http://www.na-mic.org/Bug/view.php?id=4081 ... I will also have a look.

lassoan:
Integrated in rev 24780

PR #410, opened 2015-11-13
ENH: Added output checking to module CXX tests

lassoan:
Added a new option (WITH_VTK_ERROR_OUTPUT_CHECK) to slicerMacroConfigureModuleCxxTestDriver
that enables checking of error and warning messages. If the option is enabled then the test
fails if unexpected error or warning is logged.

Enabled the option for all tests that had debug leaks check option (WITH_VTK_DEBUG_LEAKS_CHECK)
already enabled. Fixed failing tests by fixing the test, the tested code, or (most frequently) marking
expected errors/warning (using TESTING_OUTPUT_ASSERT_ERRORS_BEGIN() / TESTING_OUTPUT_ASSERT_ERRORS_END()
and similar macros).

### DIFF
--- a/CMake/SlicerMacroConfigureModuleCxxTestDriver.cmake
+++ b/CMake/SlicerMacroConfigureModuleCxxTestDriver.cmake
@@ -21,6 +21,7 @@
 macro(SlicerMacroConfigureModuleCxxTestDriver)
   set(options
     WITH_VTK_DEBUG_LEAKS_CHECK
+    WITH_VTK_ERROR_OUTPUT_CHECK
     )
   set(oneValueArgs
     NAME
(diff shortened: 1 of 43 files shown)


pieper:
I didn't test this but it sounds like a great idea!

lassoan:
Yes, it's very useful, as it makes the automatic tests more sensitive. Every warning or error in any VTK class shows up in the dashboard (if the error is expected then it can be marked as such in the test). The same log messages are logged in production code to help diagnosing issues on user computers.

The mechanism will be able to detect more errors if developers put error or warning logs into failure branches (never just silently return when an operation fails - before that write in the log what was the problem).

The output checking is already in place in MRML widgets testing (https://github.com/Slicer/Slicer/tree/master/Libs/MRML/Widgets/Testing) and should be added to more and more places (MRML core tests, etc).

lassoan:
If there are no objections, I'll merge this tomorrow.

pieper:
Hi Andras - on yesterday's hangout we discussed that in the next few days we want to stabilize the nightlies in preparation for RSNA so we can get the segmentation object support in a form that people can download it.  Can this merge wait until after the middle of next week?

fedorov:
Yes, if this could wait until we have a functional nightly with all features we need for RSNA activities, we would really appreciate it Andras.

lassoan:
Ok. Please let me know as soon as the trunk is open for new stuff. Maybe let other developers know on the devel list that the trunk is still somewhat closed for modifications.

fedorov:
Fair enough!

fedorov:
fair enough!

jcfr:
To be precise, we discussed that this topic was good for integration. 
The one on hold are #409 and #401 
See http://www.slicer.org/slicerWiki/index.php/Developer_Meetings/20151117#Conclusion

jcfr:
I guess .. since it modifies the API for `RegisterNodeClass();`, I would like to help on this one too.

lassoan:
An optional parameter is added to RegisterNodeClass, for existing code (that does not specify the optional parameter) the behavior remains the same. Anyway, it can wait till next week.

jcfr:
Closing this PR. I pushed a new one (See #416) where there is no need to change the API of `vtkMRMLScene::RegisterNodeClass()`. 

The warnings in `vtkSlicerAnnotationModuleLogicTest1` were legitimate because `RegisterNodes` is implicitly called by `SetMRMLScene()`. The solution was to remove the extra call to `RegisterNodes()`.

lassoan:
The silent node registration is still needed as somerimes you want to copy all the registered node types from the main scene to an empty scene but then during copy you get a warning for re-registering all the standard node types. But I can make this API enhancement in a separate topic.

jcfr:
> can make this API enhancement in a separate topic.

Let's do it in a separate topic.

It means the implementation of this function would have to be updated. See vtkMRMLScene::CopyRegisteredNodesToScene

jcfr:
@lassoan Let me know if you like me to integrate #416 or if you will take care of it.

lassoan:
Please take care of it. Thank you!

PR #414, opened 2015-11-19
ENH: Added CLI support for Subject hierarchy

cpinter:
1. Moved vtkMRMLSubjectHierarchyNode and constants to MRML core. Moved transform related functions to SH logic from the node
2. Added reference argument support for Volume (image) and Model (geometry) nodes when outputs of a CLI module
3. Updated most CLI modules to support references for subject hierarchies


cpinter:
Please review and let me know if I should change anything.
Integration can of course wait until after RSNA.

fedorov:
This is a great feature! I don't have time at the moment to try this out or review, but I can check it out after RSNA (if it is not integrated by then). 

As far as RSNA Slicer features needed by the QIICR project, we are not done yet, Steve is working on the DICOM export integration.

jcfr:
@cpinter When rebasing the associated topic. Could you consider creating two commits: 1 adding the feature and 1 updating the clis ? Thanks

cpinter:
Integrated in commits
http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24798
and
http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24799

PR #417, opened 2015-11-20
BUG: Fixed volume rendering at incorrect position

lassoan:
Modification of any transforms applied on any volume may alter location where volume rendering appears.

How to reproduce:
* Load MRHead sample (MRHead)
* Load MRHead sample (MRHead_1)
* Apply a new linear transform (LinearTransform_3) to MRHead
* Go to Volume Rendering module
* Switch to MRHead_1, show volume rendering
=> volume rendering of MRHead_1 is displayed
* Go to transforms module
* Modify LinearTransform_3
=> ERROR: volume rendering of MRHead_1 moves on the image (while the transform is only applied to MRHead)
* Slightly rotate the 3D view
=> volume rendering of MRHead_1 jumps (back to the location where it should be)

Problem root cause: vtkMRMLVolumeRenderingDisplayableManager observes all TransformModifiedEvent changes and updates the position of the current volume based on that (even if not the current volume's transform is changed)

Solution: ignore transform modified changes if not the current volume's transform is modified

### DIFF
--- a/Modules/Loadable/VolumeRendering/MRMLDM/vtkMRMLVolumeRenderingDisplayableManager.cxx
+++ b/Modules/Loadable/VolumeRendering/MRMLDM/vtkMRMLVolumeRenderingDisplayableManager.cxx
@@ -709,8 +709,9 @@ void vtkMRMLVolumeRenderingDisplayableManager::UpdateClipping(
 //---------------------------------------------------------------------------
 void vtkMRMLVolumeRenderingDisplayableManager::TransformModified(vtkMRMLVolumeRenderingDisplayNode* vspNode)
 {
-  if (vspNode == NULL)
+  if (vspNode == NULL || vspNode != this->DisplayedNode)
     {
+    // only the current DisplayedNode is visible, therefore we can ignore all other transform updates
     return;
     }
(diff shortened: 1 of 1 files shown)


lassoan:
This is a quite visible bug and the impact of the fix seems to be minimal, so I would recommend to include it in the RSNA patch release.

naucoin:
Looks good to me!

fedorov:
:+1: !

pieper:
+1!

jcfr:
Thanks for the fix.

That said, I wonder why the `TransformModifiedEvent` is triggered in the first place. If the volume is not transformed, the event should not be triggered.

lassoan:
Event management in volume rendering displayable manager is a mess, which has to be cleaned up. I just did not want to make a big change now that would require extensive testing.

jcfr:
Fair enough. :+1:  then. 

And I suggest we change the comment to:
```
// XXX Only the current DisplayedNode is visible, therefore we can ignore all other transform updates.
//         Ideally event management in the Volume Rendering displayable manager should be refactored.
```

Let's just add a  a note in the commit message about it.

lassoan:
Merged into rev 24760

PR #418, opened 2015-11-20
ENH: Added labelmap conversion to Volumes module

cpinter:
1. Changed LabelMap checkbox to Volume type label showing the node tag without the Volume postfix (as scalar volume has the tag "Volume", "Scalar" is shown instead of empty string)
2. Add extra section into Volumes module, because
a. Label volume conversion cannot be added in the volume info widget due to scope issues (volumes logic not accessible from MRML widgets)
b. Conversion operation doesn't fit into the idea of "volume info"
3. Reogranized labelmap creation functions in volumes logic. Moved and corrected the body of FillLabelVolumeFromTemplate to CreateLabelVolumeFromVolume. Call the new function from CreateAndAddLabelVolume and FillLabelVolumeFromTemplate. Correction in CreateLabelVolumeFromVolume includes deep copy of the image data instead of keeping a floating threshold filter and setting connection (besides it seems to be unstable, it also did not work, dimensions of the volume were zeroes)
4. Removed margin from volumes info widget to make volumes module look more uniform

### DIFF
--- a/Libs/MRML/Widgets/Resources/UI/qMRMLVolumeInfoWidget.ui
+++ b/Libs/MRML/Widgets/Resources/UI/qMRMLVolumeInfoWidget.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>660</width>
-    <height>424</height>
+    <height>426</height>
    </rect>
   </property>
   <property name="sizePolicy">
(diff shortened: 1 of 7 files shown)


cpinter:
Function vtkSlicerVolumesLogic::FillLabelVolumeFromTemplate is not used anywhere in the core and it was broken (fixed now). Also its name is strange. I understand what 'template' means here but still. Can we remove it?

(This commit can also wait until after RSNA)

pieper:
LGTM - thanks!

jcfr:
Beside of few nitpicks. Looks good.

cpinter:
Addressed final comments and integrated in http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24797

PR #419, opened 2015-11-20
ENH: Fix scene save with default parameters

lassoan:
Novice Slicer users often complain that scene saving fails. The problem is that Slicer offers scene saving in a read-only directory by default (into the current working directory, which is the read-only program binary directory by default).

Fixes applied:
1. Changed the default scene root directory to be the user's documents folder (determined the same way as the default DICOM database directory).
2. Made default scene directory configurable in the application settings GUI (selection is stored in application settings under "DefaultScenePath" key).

### DIFF
--- a/Base/QTCore/qSlicerCoreApplication.cxx
+++ b/Base/QTCore/qSlicerCoreApplication.cxx
@@ -23,6 +23,7 @@
 
 // Qt includes
 #include <QDebug>
+#include <QDesktopServices>
 #include <QDir>
 #include <QLocale>
 #include <QMessageBox>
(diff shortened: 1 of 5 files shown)


lassoan:
This may be helpful for the RSNA tutorial and the scope of changes is quite limited, so it may be a good candidate for integration into the trunk now.

fedorov:
Looks fine to me. But I think that tutorial use the stable release, not nightly, so it will not affect the tutorial.

pieper:
I like this change - agreed, it's a common issue for users (young and old)

naucoin:
+1 and it looks good to me.

jcfr:
:+1:

jcfr:
Integrated in [r24771](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24771)

PR #420, opened 2015-11-23
Integrate editor seg export

pieper:


pieper:
committed to slicer svn at r24769

PR #421, opened 2015-11-24
ENH: Added DICOMUtils convenience function library

cpinter:
Usage:
import DICOMLib
DICOMLib.DICOMUtils.loadPatientByName('SamplePatient')

### DIFF
--- a/Modules/Scripted/DICOMLib/DICOMUtils.py
+++ b/Modules/Scripted/DICOMLib/DICOMUtils.py
@@ -1,0 +1,148 @@
+import os
+import vtk, qt, ctk, slicer
+import logging
+
+#########################################################
+#
+#
+comment = """
+
+  DICOMUtils is a collection of static-like DICOM
+  utility functions facilitating convenient scripted
+  use of the DICOM database
+
+"""
+#
+######################################################### 
+
+#------------------------------------------------------------------------------
+# Get patient UID by patient name for easy loading of a patient
+def getPatientUIDByPatientName(name):
+  if not hasattr(slicer,'dicomDatabase'):
+    logging.error('DICOM database cannot be accessed!')
+    return None
+  patients = slicer.dicomDatabase.patients()
+  for patientUID in patients:
+    currentName = slicer.dicomDatabase.nameForPatient(patientUID)
+    if currentName == name:
+      return patientUID
+  return None
+
+#------------------------------------------------------------------------------
+# Load patient by patient UID from DICOM database
+def loadPatientByUID(patientUID):
+  if not hasattr(slicer,'dicomDatabase') or not hasattr(slicer.modules,'dicom'):
+    logging.error('DICOM module or database cannot be accessed!')
+    return
+  dicomWidget = slicer.modules.dicom.widgetRepresentation().self()
+
+  # Select all series in selected patient
+  studies = slicer.dicomDatabase.studiesForPatient(patientUID)
+  series = [slicer.dicomDatabase.seriesForStudy(study) for study in studies]
+  seriesUIDs = [uid for uidList in series for uid in uidList]
+  dicomWidget.detailsPopup.offerLoadables(seriesUIDs, 'SeriesUIDList')
+  dicomWidget.detailsPopup.examineForLoading()
+
+  # Load selected data
+  dicomWidget.detailsPopup.loadCheckedLoadables()
+
+#------------------------------------------------------------------------------
+# Load patient by patient name from DICOM database
+def loadPatientByName(patientName):
+  loadPatientByUID(getPatientUIDByPatientName(patientName))
+
+#------------------------------------------------------------------------------
+# Open specified DICOM database
+def openDatabase(databaseDir):
+  if not hasattr(slicer,'dicomDatabase') or not hasattr(slicer.modules,'dicom'):
+    logging.error('DICOM module or database cannot be accessed!')
+    return False
+  if not os.access(databaseDir, os.F_OK):
+    logging.error('Specified database directory ' + repr(databaseDir) + ' cannot be found!')
+    return False
+  dicomWidget = slicer.modules.dicom.widgetRepresentation().self()
+  dicomWidget.onDatabaseDirectoryChanged(databaseDir)
+  if not slicer.dicomDatabase.isOpen:
+    logging.error('Unable to open DICOM database ' + databaseDir)
+    return False
+  return True
+
+#------------------------------------------------------------------------------
+# Open temporary DICOM database, return current database directory
+def openTemporaryDatabase(directory=None, absolutePath=False):
+  if not hasattr(slicer,'dicomDatabase') or not hasattr(slicer.modules,'dicom'):
+    logging.error('DICOM module or database cannot be accessed!')
+    return None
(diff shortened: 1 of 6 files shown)


Modules/Scripted/DICOMLib/DICOMUtils.py
```diff
@@ -0,0 +1,91 @@
...
+
+"""
+#
+######################################################### 
+
+#------------------------------------------------------------------------------
+# Get patient UID by patient name for easy loading of a patient
+def patientUIDByPatientName(name):
+  if not hasattr(slicer,'dicomDatabase'):
+    return None
```
jcfr: Should an error or warning be logged ?

cpinter:
I added a DICOM utility library to make it easier to do DICOM DB operations from python.
- patientUIDByPatientName, loadPatientByUID, and loadPatientByName facilitate easy loading of DICOM patients with one-liner python calls.
- openTemporaryDatabase and openDatabase can be used by self tests that download and import their own test data to easily switch to a temporary database and use that for the test, after which it can restore the original database (returned by openTemporaryDatabase). I haven't yet updated the core tests, but some of them can use these, such as JRC2013Vis, RSNAVisTutorial and SubjectHierarchyGenericSelfTest

Modules/Scripted/DICOMLib/DICOMUtils.py
cpinter: Sure, one sec.

Modules/Scripted/DICOMLib/DICOMUtils.py
```diff
@@ -0,0 +1,96 @@
...
+# Load patient by patient name from DICOM database
+def loadPatientByName(patientName):
+  loadPatientByUID(patientUIDByPatientName(patientName))
+
+#------------------------------------------------------------------------------
+# Open specified DICOM database
+def openDatabase(databaseDir):
+  if not hasattr(slicer,'dicomDatabase') or not hasattr(slicer.modules,'dicom'):
+    logging.error('DICOM module or database cannot be accessed!')
+    return
```
lassoan: It would be useful to return with True or False depending on if the database could be successfully opened or not

cpinter: Thanks! I fixed it before saw your comment :)

cpinter: Oh I see. So it should check if opening was successful?

Modules/Scripted/DICOMLib/DICOMUtils.py
```diff
@@ -0,0 +1,96 @@
...
+  utility functions facilitating convenient scripted
+  use of the DICOM database
+
+"""
+#
+######################################################### 
+
+#------------------------------------------------------------------------------
+# Get patient UID by patient name for easy loading of a patient
+def patientUIDByPatientName(name):
```
lassoan: This method name is not very clear for me. getPatientUIDByPatientName or patientUIDFromPatientName would be more clear.

lassoan: A small improvement possibility (probably not important for tests but could be useful for general use): In DICOM usually the patient name and patient ID are used together to identify a patient. It may be useful to add an optional patientID parameter that would allow to search for a specific patient name AND id.

Modules/Scripted/DICOMLib/DICOMUtils.py
```diff
@@ -0,0 +1,96 @@
...
+    logging.error('Specified database directory ' + repr(databaseDir) + ' cannot be found!')
+    return
+  dicomWidget = slicer.modules.dicom.widgetRepresentation().self()
+  dicomWidget.onDatabaseDirectoryChanged(databaseDir)
+  if not slicer.dicomDatabase.isOpen:
+    logging.error('Unable to open DICOM database ' + databaseDir)
+
+#------------------------------------------------------------------------------
+# Open temporary DICOM database, return current database directory
+def openTemporaryDatabase(folderName=None):
```
lassoan: What happens to the temporary database after it is no longer needed?
It could be nice to have a closeTemporaryDatabase(originalDatabaseFolderName, cleanup=True) method that can delete the temporary database and restore the original database.
originalDatabaseFolder = openTemporaryDatabase()
...
closeTemporaryDatabase(originalDatabaseFolder)

Modules/Scripted/DICOMLib/DICOMUtils.py
lassoan: A different return value would give a chance to the caller to detect the error and not try continue whatever it wanted to do. Small thing, it's just easier to change it now than add it later.

lassoan:
Thanks Csaba, this will be useful. Last time when somebody wanted to merge utility python functions I insisted that the utilities should be used by the Slicer core modules wherever it is applicable to serve as an example and to not duplicate code (see https://github.com/Slicer/Slicer/pull/407). So, I'm afraid I would need to ask you the same: please update the JRC2013Vis, RSNAVisTutorial and SubjectHierarchyGenericSelfTest tests as well.

Modules/Scripted/DICOMLib/DICOMUtils.py
cpinter: I will rename the function.

This file will be open to additional methods for all developers if the need arises.

cpinter:
I didn't do that in the hope of quick merge, as I don't think touching the RSNA tutorial is wise right now.

Modules/Scripted/DICOMLib/DICOMUtils.py
fedorov: Also, "patientUID" should be replaced with "patientID". There is no such thing as "patient UID" (unfortunately!).

lassoan:
RSNA demo will use stable and nothing will be merged until layer this week, when a new release will be created for additional RSNA activities. So, there will be no impact on RSNA.

Using the utility functions in the tests will also serves as tests for these new utility functions (in addition to serve as examples and reduce code duplication).

Modules/Scripted/DICOMLib/DICOMUtils.py
jcfr: I would suggest to great a context manager. That would allow to write code like this:
```
with openTemporaryDatabase() as d:
    d.doSomething()
```

The database would then be closed when exiting the context.

Example here: https://github.com/Slicer/Slicer/blob/7cfc9d6fc9ec2480810582750908ac05922f3a27/Base/Python/slicer/util.py#L491-L500

cpinter:
I changed the python tests using temporary databases to call the new utility functions

Modules/Scripted/DICOMLib/DICOMUtils.py
cpinter: Good idea, I’ll do it today.

cpinter:
I added the context manager for temporary databases.

jcfr:
@cpinter After return the database in the context manager. LGTM :+1:

cpinter:
Did the requested change.
Integrated in http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision
Thanks!

PR #424, opened 2015-12-01
COMP: qMRMLTableView: Fix unix build error using function instead of …

jcfr:
…macro

This commit fixes error introduced in r24783 (ENH: Added support for Tables)
similar to the following one:

```
/path/to/Slicer/Libs/MRML/Widgets/qMRMLTableView.cxx:217:44: error: macro "VERIFY_TABLE_MODEL_AND_NODE" requires 2 arguments, but only 1 given
   VERIFY_TABLE_MODEL_AND_NODE(setTransposed);
                                            ^
```

### DIFF
--- a/Libs/MRML/Widgets/qMRMLTableView.cxx
+++ b/Libs/MRML/Widgets/qMRMLTableView.cxx
@@ -43,30 +43,6 @@
 #include <vtkMRMLTableNode.h>
 #include <vtkMRMLTableViewNode.h>
 
-#define VERIFY_TABLE_MODEL_AND_NODE(methodName) \
-  if (!this->tableModel()) \
-    { \
-    qWarning() << "qMRMLTableView:: " << #methodName << " failed: invalid model"; \
-    return; \
-    } \
-  if (!this->mrmlTableNode()) \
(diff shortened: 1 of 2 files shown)


jcfr:
@naucoin @lassoan I was working on a fix and testing .. while you committed the other fix.

Instead of using a macro, I would suggest to use a function as suggested in this patch.

lassoan:
I think for trivial things like this macros are nicer because they allow more compact code, less noise. If we don't use macros then we could just as well have the actual checks everywhere (it would maybe one line more than calling a check method and return in case of error). So, I would not change the current code but I don't mind if you change it if you are sure that it is a nicer solution.

jcfr:
Generally speaking I think we should avoid macro, as mentioned in the [google style](https://google.github.io/styleguide/cppguide.html#Preprocessor_Macros), the code you see is not the code that is compiled.

In that particular case, the advantage of either the macro or the function is to avoid repeating the string in the error message and facilitate maintenance/refactoring when new check are added.

jcfr:
Closing. Superseded by #427

PR #426, opened 2015-12-01
BUG: Fix issues in sending PolyData. See Bug Tracker ID 0004095.

tokjun:


jcfr:
Thanks for the patch :+1: 

Integrated in [r24802](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24802)

PR #427, opened 2015-12-02
Generic macro verify table model and node

jcfr:


jcfr:
@lassoan Instead of having specific macros like `VERIFY_TABLE_MODEL_AND_NODE`, I would suggest having more generic macros like `CTK_CHECK_AND_RETURN_IF_FAIL` and `CTK_CHECK_AND_RETURN_FALSE_IF_FAIL` (See  2101bde for an example).

Ideally these macros would be added to CTK so that we can use across all Qt codes. 

Similar macros could be added to [VTKAddon](https://github.com/Slicer/Slicer/tree/master/Libs/vtkAddon).

That said, I am still thorn  ... (see https://google.github.io/styleguide/cppguide.html#Preprocessor_Macros)

Cc: @Slicer/community

Libs/MRML/Widgets/qMRMLTableView.cxx
```diff
@@ -219,14 +224,16 @@ vtkMRMLTableNode* qMRMLTableView::mrmlTableNode()const
 //------------------------------------------------------------------------------
 bool qMRMLTableView::transposed()const
 {
-  VERIFY_TABLE_MODEL_AND_NODE_RETURN_VALUE(transposed, false);
+  Q_D(const qMRMLTableView);
+  CTK_CHECK_AND_RETURN_FALSE_IF_FAIL(verifyTableModelAndNode)
```
lassoan: Thanks Jc, these changes look good!

The only thing that I don't like is that d-> is hidden in the macro. I think that instead of

CTK_CHECK_AND_RETURN_FALSE_IF_FAIL(verifyTableModelAndNode) 

this would be more clear:

CTK_CHECK_AND_RETURN_FALSE_IF_FAIL(d->verifyTableModelAndNode)

For example, without d-> it's not obvious why you need the Q_D macro.

lassoan:
Macros are useful and should be used when there are no equivalent C++ solutions. For example, getting line number, function name is not possible at all using C++. I think the solution that you implemented for these checks is also a good tradeoff between simple, easy-to-debug, and compact code.

jcfr:
Integrated in [r24858](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24858), [r24859](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24859) and [r24860](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24860).

PR #428, opened 2015-12-02
BUG: Fix hide from editor flag on volume rendering nodes

naucoin:
The volume rendering nodes are appearing in the Data tree
after the change to set the default of the hide from editors
flag to false on the MRML node superclass.
This change sets them hidden by default and sets up the
display node combo box to show them in the Volume
Rendering Inputs panel.

Issue #2906

### DIFF
--- a/Modules/Loadable/VolumeRendering/MRML/vtkMRMLCPURayCastVolumeRenderingDisplayNode.cxx
+++ b/Modules/Loadable/VolumeRendering/MRML/vtkMRMLCPURayCastVolumeRenderingDisplayNode.cxx
@@ -34,6 +34,8 @@ vtkMRMLNodeNewMacro(vtkMRMLCPURayCastVolumeRenderingDisplayNode);
 vtkMRMLCPURayCastVolumeRenderingDisplayNode::vtkMRMLCPURayCastVolumeRenderingDisplayNode()
 {
   this->RaycastTechnique = vtkMRMLCPURayCastVolumeRenderingDisplayNode::Composite;
+
+  this->SetHideFromEditors(1);
 }
 
 //----------------------------------------------------------------------------
--- a/Modules/Loadable/VolumeRendering/MRML/vtkMRMLGPURayCastVolumeRenderingDisplayNode.cxx
+++ b/Modules/Loadable/VolumeRendering/MRML/vtkMRMLGPURayCastVolumeRenderingDisplayNode.cxx
@@ -34,6 +34,8 @@ vtkMRMLNodeNewMacro(vtkMRMLGPURayCastVolumeRenderingDisplayNode);
 vtkMRMLGPURayCastVolumeRenderingDisplayNode::vtkMRMLGPURayCastVolumeRenderingDisplayNode()
 {
   this->RaycastTechnique = vtkMRMLGPURayCastVolumeRenderingDisplayNode::Composite;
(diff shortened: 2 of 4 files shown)


pieper:
LGTM

naucoin:
Rebased to the current trunk, okay to merge it now?

jcfr:
LGTM :+1:

naucoin:
Done:
http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24995
Github says it should be 6b2539c1f5b51d2fced8d3fa836f5065b58ed037 but it hasn't synched yet.

lassoan:
What was the issue with VolumeRendering (vtkMRMLGPURayCastVolumeRenderingDisplayNode) shown in Data module? It is very convenient that you can show/hide volume rendering from Data module (when showing Displayable hierarchy).

VolumeProperty is still displayed. Why it's not changed as well? (actually, I know, because hidden nodes are not saved if the node uses storage node; and VolumeProperty node uses storage node)

The actual problem is that in the Data widget: listed nodes should be restricted to transformable nodes when the Transform scene model is used.

I think http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24995 should be reverted and Data module's node tree widget should be fixed.

naucoin:
@lassoan I hid them due to the discussion on the bug report here:
http://www.na-mic.org/Bug/view.php?id=2906
After the default was changed to have hide from editors be set to 0 by default, the nodes that appeared in the Data module in the default Transforms view changed. This pull request was to go back to the original state.

lassoan:
Volume rendering display node was set to be shown in editors (in vtkMRMLVolumeRenderingDisplayNode). I don't really mind if it is changed but then don't add one more flipping of the hideFromEditor flag at the lowest class hierarchy levels but simply remove this->SetHideFromEditors(false) from vtkMRMLVolumeRenderingDisplayNode constructor.

Currently setting of this flag is completely inconsistent across the class hierarchy:
* vtkMRMLCPURayCastVolumeRenderingDisplayNode ->  this->SetHideFromEditors(1);
* vtkMRMLVolumeRenderingDisplayNode -> this->SetHideFromEditors(false);
* vtkMRMLDisplayNode -> this->HideFromEditors = 1;
* vtkMRMLNode -> this->HideFromEditors = 0;

lassoan:
I've made the changes that I described above (http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24997). There is no functional change compared to what Nicole committed, just removing the unnecessary SetHideFromEditors(false) in the base class and unnecessary SetHideFromEditors(1) in the children classes.

Note that rev 24995 did not actually restore the previous status of HideFromEditors, as HideFromEditors was explicitly specified in vtkMRMLVolumeRenderingDisplayNode, so changing default value did not have any effect. It is not clear why HideFromEditors was disabled for vtkMRMLVolumeRenderingDisplayNode previously and all other display nodes are hidden by default, so the current behavior is different than it was before, but more consistent

naucoin:
Thanks, I was looking at the wrong base class for your previous comment, this is more clear.

jcfr:
@naucoin @lassoan Thanks for moving this forward. :smile:

jcfr:
Now closing this PR.

PR #429, opened 2015-12-05
BUG: Update Git Tag for OpenIGTLinkIF to fix issue 0004101.

tokjun:
See http://www.na-mic.org/Bug/view.php?id=4101 for detail.


lassoan:
Merged in rev 24808

PR #430, opened 2015-12-07
BUG: Fixed crash in vtkMRMLTransformNode::GetMatrixTransformToNode

lassoan:
How to reproduce the crash:

```
# Create transform hierarchy t1-t2-t3
t1=slicer.vtkMRMLTransformNode()
t2=slicer.vtkMRMLTransformNode()
t3=slicer.vtkMRMLTransformNode()
slicer.mrmlScene.AddNode(t1)
slicer.mrmlScene.AddNode(t2)
slicer.mrmlScene.AddNode(t3)
t2.SetAndObserveTransformNodeID(t1.GetID())
t3.SetAndObserveTransformNodeID(t2.GetID())
# Get t3 to t1 transform
m = vtk.vtkMatrix4x4()
t3.GetMatrixTransformToNode(t1, m)
```

The problem was caused by infinite recursion. There were no tests for GetMatrixTransformToNode and GetTransformToNode methods and they did not work well.
Also, the transform node heavily used recursion in several methods when only a simple iteration on the transform tree was needed, which made computation time magnitudes higher (although it was still very quick, so it was not noticeable) and made the code harder to read and debug.

Solution:
Replaced all the unnecessary recursive method calls by simple for loops.
Added tests for Get(Matrix)TransformToNode methods.

### DIFF
--- a/Libs/MRML/Core/Testing/vtkMRMLTransformNodeTest1.cxx
+++ b/Libs/MRML/Core/Testing/vtkMRMLTransformNodeTest1.cxx
@@ -10,16 +10,187 @@
 
 =========================================================================auto=*/
 
+#include "vtkMRMLBSplineTransformNode.h"
 #include "vtkMRMLCoreTestingMacros.h"
+#include "vtkMRMLScene.h"
 #include "vtkMRMLTransformNode.h"
 
 #include <vtkGeneralTransform.h>
+#include <vtkMatrix4x4.h>
+#include <vtkNew.h>
+#include <vtkTransform.h>
+
+bool Matrix4x4AreEqual(vtkMatrix4x4 *m1, vtkMatrix4x4 *m2)
+{
+  const double tol = 1e-3;
+  for (int i = 0; i < 4; i++)
+    {
+    for (int j = 0; j < 4; j++)
+      {
+      if ( fabs(m1->GetElement(i, j) - m2->GetElement(i, j)) > tol )
+        {
+        return false;
+        }
+      }
+    }
+    return true;
+}
+
+vtkMatrix4x4* CreateTransformMatrix(double translateX, double translateY, double translateZ, double rotateX, double rotateY, double rotateZ)
+{
+  vtkNew<vtkTransform> tr;
+  tr->Translate(translateX, translateY, translateZ);
+  tr->RotateX(rotateX);
+  tr->RotateY(rotateY);
+  tr->RotateZ(rotateZ);
+  vtkMatrix4x4* matrix = vtkMatrix4x4::New();
+  tr->GetMatrix(matrix);
+  return matrix;
+}
 
 int vtkMRMLTransformNodeTest1(int , char * [] )
 {
   vtkNew<vtkMRMLTransformNode> node1;
   EXERCISE_BASIC_OBJECT_METHODS(node1.GetPointer());
   EXERCISE_BASIC_TRANSFORM_MRML_METHODS(vtkMRMLTransformNode, node1.GetPointer());
+
+  /// Test matrix get/set functions
+
+  // Create transform nodes
+  vtkNew<vtkMRMLTransformNode> bTransform; // == b_to_w
+  vtkNew<vtkMRMLTransformNode> cTransform; // == c_to_b
+  vtkNew<vtkMRMLTransformNode> dTransform; // == d_to_c
+  vtkNew<vtkMRMLTransformNode> eTransform; // == e_to_d
+  vtkNew<vtkMRMLTransformNode> qTransform; // == q_to_b
+  vtkNew<vtkMRMLTransformNode> rTransform; // == r_to_q
+  vtkSmartPointer<vtkMatrix4x4> w_from_b_mx = vtkSmartPointer<vtkMatrix4x4>::Take(CreateTransformMatrix( 34,  23, -12,  44,  12,  78));
+  vtkSmartPointer<vtkMatrix4x4> b_from_c_mx = vtkSmartPointer<vtkMatrix4x4>::Take(CreateTransformMatrix(-34,  11,  12, -22, 128,  18));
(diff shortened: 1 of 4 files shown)


Libs/MRML/Core/Testing/vtkMRMLTransformNodeTest1.cxx
```diff
@@ -10,10 +10,43 @@
...
 #include "vtkMRMLCoreTestingMacros.h"
+#include "vtkMRMLScene.h"
 #include "vtkMRMLTransformNode.h"
 
 #include <vtkGeneralTransform.h>
+#include <vtkMatrix4x4.h>
+#include <vtkNew.h>
+#include <vtkTransform.h>
+
+bool Matrix4x4AreEqual(vtkMatrix4x4 *m1, vtkMatrix4x4 *m2)
```
jcfr: Would it make sense to re-use `vtkMRMLSliceNode::Matrix4x4AreEqual()` ?
See https://github.com/Slicer/Slicer/blob/482657adbb55333e4f97ccb4eed525e16e6a3d06/Libs/MRML/Core/vtkMRMLSliceNode.h#L231

pieper:
LGTM - :+1:

Libs/MRML/Core/vtkMRMLTransformNode.cxx
```diff
@@ -350,22 +370,15 @@ vtkAbstractTransform* vtkMRMLTransformNode::GetTransformFromParent()
 //----------------------------------------------------------------------------
 int  vtkMRMLTransformNode::IsTransformToWorldLinear()
 {
-  if (!this->IsLinear())
-    {
-    return 0;
-    }
-  else
+  for (vtkMRMLTransformNode* current = this; current!=NULL; current = current->GetParentTransformNode())
```
jcfr: `current!=NULL;` -> `current != NULL;`

jcfr:
Beside of few nitpicks, :+1:

Libs/MRML/Core/Testing/vtkMRMLTransformNodeTest1.cxx
lassoan: That one does not have a tolerance in the comparison, so I'm not sure how well it would work. It would make sense to add some tolerance in vtkMRMLSliceNode::Matrix4x4AreEqual(), too, but I would hesitate to do that in this commit.

pieper: We should probably collect all the utilities we find useful and put them in
vtkAddon with the idea that they eventually make their way to vtk proper.
Fuzzy matrix comparison is definitely one of the candidates, but I'm sure
there are others.

lassoan: That's a good idea, I'll do that. I'll try to make it equivalent with ITK's fuzzy image to physical transform compare.

jcfr: > collect all the utilities we find useful and put them in
vtkAddon
:+1: 

> but I would hesitate to do that in this commit

Good point. This can definitively be done in an other commit.

Simply adding a note in this commit message would be good.

lassoan:
Thanks for the comments. I've committed as rev 24809.

I will send another pull request with the fuzzy compare added to vtkAddon. I'm not sure where to add it in vtkAddon though. Should I create a vtkAddonMath class and make it a static method of that?

cpinter:
When adding utilities to vtkAddon, it would be nice if you could check https://github.com/SlicerRt/SlicerRT/blob/master/SlicerRtCommon/SlicerRtCommon.h, there are many similar utility functions there that can be useful.

lassoan:
That's a good point. We can move common utility functions (that only depend on VTK) from SlicerRtCommon to vtkAddon. Csaba, if you have a list of candidate methods then I can add them in one batch (and then you can remove them from SlicerRT).

ihnorton:
+∞. (replacing inline transform handling with `GetTransformBetweenTransformables` could probably cut 5% of lines from Slicer base)

pieper:
: )

pieper:
How about adopting a naming convention like <vtkClassName>Addons so we would have vtkMathAddons, vtkMatrix4x4Addons, etc.  That would make it clear where the code is meant to be slotted in with respect to the VTK architecture.  At the least it would give us a way to keep track.

lassoan:
OK, I'll add a vtkAddonMatrix4x4 class, derived from vtkMatrix4x4 (similarly to how we did it for vtkOrientedGridTransform and vtkOrientedBSplineTransform)

jcfr:
> add a `vtkAddonMatrix4x4` class

In that particular case, may be we could contribute the method upstream ? In the mean time, we could update the `vtkMatrix4x4` in the Slicer/VTK fork ?

cpinter:
I'm writing a take-home exam right now, but here's a quick list of functions I can think of being integrated:
GetTransformBetweenTransformables
DoVolumeLatticesMatch
AreExtentsEqual
IsEqual(const vtkMatrix4x4& lhs, const vtkMatrix4x4& rhs)
Maybe the one with the templated vectors too.
Also there are a few about colors, but I don't think they'd belong in vtkAddon.
While we're at it, there are a few similar transforms in
https://github.com/SlicerRt/SlicerRT/blob/master/SegmentationCore/vtkOrientedImageDataResample.h
like
TransformExtent
AreEqualWithTolerance
IsTransformLinear

jcfr:
Integrated in [r24809](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24809)

I created an issue to capture the update of `vtkAddon`. See [#4103: Move general methods to vtkAddon ](http://www.na-mic.org/Bug/view.php?id=4103)

PR #431, opened 2015-12-08
BUG: Fixed vtkTransform::ApplyTransformMatrix crash

lassoan:
How to reproduce:

c=getNode('FullRainbow')
m=vtk.vtkMatrix4x4()
c.ApplyTransformMatrix(m)

Problem:
vtkTransform::ApplyTransform and vtkTransform::ApplyTransformMatrix are calling each other. Most of the time at least one is overridden, so no crash occurs.

Solution:
Removed calling of vtkTransform::ApplyTransformMatrix from vtkTransform::ApplyTransform. If a class has a more efficient version of transform application for linear transforms then this call can be added to vtkTransform::ApplyTransform in that specific class.
Removed vtkTransform::ApplyTransformMatrix where it is not necessary (when it's just a more complicated and possibly slower version of ApplyTransform).

### DIFF
--- a/Libs/MRML/Core/vtkMRMLFiducialListNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLFiducialListNode.cxx
@@ -1635,48 +1635,6 @@ bool vtkMRMLFiducialListNode::CanApplyNonLinearTransforms()const
 }
 
 //---------------------------------------------------------------------------
-void vtkMRMLFiducialListNode::ApplyTransformMatrix(vtkMatrix4x4* transformMatrix)
-{
-  int numPoints = this->GetNumberOfFiducials();
-  double (*matrix)[4] = transformMatrix->Element;
-  float xyzIn[3];
-  float xyzOut[3];
-  float orientationIn[4], quaternionIn[4];
-  float orientationMatrix3x3[3][3];
-  vtkNew<vtkMatrix4x4> orientationMatrix;
-  vtkNew<vtkMatrix4x4> newOrientationMatrix;
-  for (int n=0; n<numPoints; n++)
-    {
-    vtkMRMLFiducial *node = this->GetNthFiducial(n);
-
-    node->GetXYZ(xyzIn);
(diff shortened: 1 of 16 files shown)


jcfr:
If test passes. 

:+1:

lassoan:
All tests passed on Windows.

jcfr:
I will compile on Ubuntu and report back.

lassoan:
@jcfr were you able to test it on linux? I don't expect problems on non-Windows platforms, as everything was basic VTK classes (mostly deleting methods), but of course it's good if it's tested more.

jcfr:
Build is in process ... I will report back shortly so that you can integrate before the nightly starts.

jcfr:
Copying warning as they appear .. build is still in progress:

```
/home/jcfr/Projects/Slicer/Libs/MRML/Core/vtkMRMLTransformableNode.cxx:159:6: warning: unused parameter ‘transform’ [-Wunused-parameter]
 void vtkMRMLTransformableNode::ApplyTransform(vtkAbstractTransform* transform)
      ^

```

jcfr:
@lassoan Looks good.  :+1:

naucoin:
Testing on a color node caught my eye, I tend to forget that anything storable is transformable. There is a CanApplyNonLinearTransforms flag, but it looks like if a node is not actually transformable that it just didn't implement the virtual methods supporting that, leading to the crash. Would it make more sense to add a CanApplyLinearTransforms flag that defaulted to false and was checked in the superclass ApplyTransform? 
I haven't dug deeply into this change, so disregard this if the hardening of transforms still works for all nodes with this change.

lassoan:
We have to flip Storable and Transformable node in the hierarchy. The wrong order causes inconveniences in many places. I've checked and there is very limited impact of fixing it (by making Storable a parent of Transformable). I'll send a pull request soon.

lassoan:
Thanks for testing. I've fixed the build warning.

Integrated in rev 24813.

PR #432, opened 2015-12-10
ENH: VolumeRenderingModuleWidget: Extend API adding two new signals:

Punzo:
* newCurrentDisplayNode
* newCurrentMRMLVolumeNode

**** WIP ****

Why ?

signals to synch the current MRMLVolumeRendeingDisplayNode from the VolumeRenderingModuleWidget to another ModuleWidgets. 

if the signal approach is wrong, what about one of these two approaches (using MRML):
1) storing the current VolumeRenderingDisplayNodeID in the vtkMRMLVolumeRenderingScenarioNode.
or
2) storing the current VolumeRenderingDisplayNodeID in the SelectionNode.

Davide.

### DIFF
--- a/Modules/Loadable/VolumeRendering/Widgets/qSlicerVolumeRenderingModuleWidget.cxx
+++ b/Modules/Loadable/VolumeRendering/Widgets/qSlicerVolumeRenderingModuleWidget.cxx
@@ -354,6 +354,8 @@ void qSlicerVolumeRenderingModuleWidget::onCurrentMRMLVolumeNodeChanged(vtkMRMLN
     }
 
   this->setMRMLDisplayNode(dnode);
+
+  emit newCurrentMRMLVolumeNode(node);
 }
 
 // --------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


pieper:
Wouldn't the Qt naming convention be something more like currentDisplayNodeChanged ?

lassoan:
The signal name has to be changed but otherwise the signaling when current node selection changes is useful and I don't see any problems with that.

In addition to this, using the SelectionNode is necessary if you want to make sure that the same node is selected in the GUI after restarting Slicer and reloading the scene. For this there is no need to modify the SelectionNode but we can just add a node reference:

    selectionNode->SetNodeReferenceID("VolumeRendering.ActiveDisplayNode",displayNode.GetID());

Earlier each node reference to Active***Node was added manually because the node references infrastructure did not exist in vtkMRMLNode.

Punzo:
thanks for the feedback.

I'll change the name of the signals in:

1) currentVolumeRenderingDisplayNodeChanged
2) currentVolumeNodeChanged

and I'll use the setNodeReferenceID (which is a Set and Observe, isn't it?).
I'll do a new PR because I think I can't do commits on  jcfr:add-VR-module-widget-signal .

PR #434, opened 2015-12-14
BUG: Fix Slicer crash due to wrong value return in AddModel

agirault:
In `vtkSlicerModelsLogic::AddModel`, if the call to the storageNode
`ReadData()` failed, an error message was triggered and the node
removed from the scene, but `AddModel()` was still returning a pointer
to the `modelNode`, resulting in Slicer crashing.

The bug was encountered when loading a VTK Volume (DATASET set to
STRUCTURED_POINTS) and not selecting the 'Volume' description, since
the default description for the legacy VTK files is 'Model' in Slicer.


jcfr:
Thanks :+1: 

Integrated in [r24815](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24815)

PR #436, opened 2015-12-29
ENH: Removed unused vtkMRMLGlyphableVolumeDisplayPropertiesNode

lassoan:
The node was not used anywhere.

Isaiah Norton also confirmed that the node is not needed.


pieper:
Looks good to me - thanks!

jcfr:
Thanks for the patch. Looks good.

Could you specify in the commit message that the class was originally and inadvertently introduced in [r7369 ](https://github.com/Slicer/Slicer/commit/3cc318ca8c45a7be0c97ac7b8a1f2282b2268715#diff-599a7f242efcb6917dca3f14be0c2540) `ENH: merged demianVolumeGlyph-Simple branch` as `vtkMRMLGlyphVolumeDisplayPropertiesNode`

jcfr:
Will be integrating now. I updated the patch to fix the following error:

```
CMake Error at Libs/MRML/Core/Testing/CMakeLists.txt:112 (add_executable):
  Cannot find source file:

    vtkMRMLGlyphVolumeDisplayPropertiesNodeTest1.cxx

  Tried extensions .c .C .c++ .cc .cpp .cxx .m .M .mm .h .hh .h++ .hm .hpp
  .hxx .in .txx

```

jcfr:
Integrated in [r24831](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24831)

PR #437, opened 2016-01-04
ENH: Moved up vtkMRMLStorableNode is the MRML node hierarchy

lassoan:
vtkMRMLStorableNode is not a children of vtkMRMLTransformable node anymore, but directly a children of vtkMRMLNode.

This allows making a node storable without requiring it to be also transformable. It is important for several node types (color maps, tables, etc), which require separate storage node but not transformable.


pieper:
Sounds like a great idea.

jcfr:
Thanks again for the patch. It will be integrated as soon as the dashboard settle down and there are no test failure.

jcfr:
Integrated in [r24891](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24891)

PR #438, opened 2016-01-05
Cli infrastucture

vovythevov:
This pull request aims to solve issues when running CLIs from python.
The two cases that this solves are:
 - Running CLI A and on CLI A completion, run CLI A another time.
 - Running CLI A and CLI A in parallel.

This was previously addressed partially in pull request #358.

@pieper, @lassoan, @jcfr You looked at #358, would you mind looking at this one ?

### DIFF
--- a/Base/QTCLI/vtkSlicerCLIModuleLogic.h
+++ b/Base/QTCLI/vtkSlicerCLIModuleLogic.h
@@ -104,6 +104,9 @@ public:
 //   void LazyEvaluateModuleTarget(vtkMRMLCommandLineModuleNode* node)
 //     { this->LazyEvaluateModuleTarget(node->GetModuleDescription()); }
 
+  /// Set the application logic
+  virtual void SetMRMLApplicationLogic(vtkMRMLApplicationLogic* logic);
+
 protected:
   /// Reimplemented to observe NodeAddedEvent.
   virtual void SetMRMLSceneInternal(vtkMRMLScene * newScene);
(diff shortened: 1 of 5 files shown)


Base/QTCLI/vtkSlicerCLIModuleLogic.h
```diff
@@ -104,6 +104,9 @@ class Q_SLICER_BASE_QTCLI_EXPORT vtkSlicerCLIModuleLogic :
 //   void LazyEvaluateModuleTarget(vtkMRMLCommandLineModuleNode* node)
 //     { this->LazyEvaluateModuleTarget(node->GetModuleDescription()); }
 
+  /// Set the application logic
+  void SetMRMLApplicationLogic(vtkMRMLApplicationLogic* logic);
```
jcfr: explicitly add `virtual`

See https://google.github.io/styleguide/cppguide.html#Inheritance

___For clarity, use exactly one of override, final, or virtual when declaring an override. Rationale: A function or destructor marked override or final that is not an override of a base class virtual function will not compile, and this helps catch common errors. The specifiers serve as documentation; if no specifier is present, the reader has to check all ancestors of the class in question to determine if the function or destructor is virtual or not.___

jcfr:
Thanks for the patches :+1: 

Beside of few nitpicks, looks good.

vovythevov:
Thanks for the feedback @jcfr, that was a good catch !

pieper:
LGTM

vovythevov:
If none of you have any objection, I'll push this tonight.

jcfr:
:+1: Make sure to push before 10pm. 

Integrating now would also work.

vovythevov:
I can't manage to push to svn.
@jcfr Would you mind pushing this change ?

jcfr:
Sounds good. Currently testing the build. Will push afterward.

jcfr:
Out of curiosity, what is the error when you try to push to svn ?

jcfr:
> Out of curiosity, what is the error when you try to push to svn ?

Problem has been sorted out offline.

jcfr:
Integrated in [r24834](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24834)

agirault:
Thanks a bunch @vovythevov

PR #439, opened 2016-01-06
BUG: Fix display node copy in CloneVolume

agirault:
**Motivation**
If a volume is rendered through the Volume Rendering volume module then
cropped through the Crop Volume module, it is not possible to display the
volume rendering of the cropped volume: the original volume would be
displayed instead.

**Reason**
In the Crop Volume module, the method `vtkSlicerVolumesLogic::CloneVolume`
was called.  The volume and its first display node were cloned using
`vtkMRMLVolumeNode::CopyWithScene`, but the remaining display nodes were
ignored. This was resulting in having a cloned volume with references to
the original volume display nodes. In the case above, the display node of
type `vtkMRMLVolumeRenderingDisplayNode` was the same for the cropped and
original volumes, referencing the original volume for both of them.

**Fix**
In this commit only the first display node is cloned, and the other display
nodes are removed. Cloning a `vtkMRMLVolumeRenderingDisplayNode` would not be
as straightforward as copying a `vtkMRMLVolumeDisplayNode` since the display
node itself has a `VolumeNodeID`, a `ROINodeID` and a `VolumePropertyNodeID` that
would need to be created/updated.

This fixes the issue 4109: http://na-mic.org/Mantis/view.php?id=4109

### DIFF
--- a/Modules/Loadable/Volumes/Testing/Cxx/vtkSlicerVolumesLogicTest1.cxx
+++ b/Modules/Loadable/Volumes/Testing/Cxx/vtkSlicerVolumesLogicTest1.cxx
@@ -20,11 +20,15 @@
 
 // Volumes logic
 #include "vtkSlicerVolumesLogic.h"
+#include "vtkMRMLCoreTestingMacros.h"
 
 // MRML includes
 #include <vtkMRMLLabelMapVolumeNode.h>
 #include <vtkMRMLScalarVolumeNode.h>
 #include <vtkMRMLScene.h>
+#include <vtkMRMLScalarVolumeDisplayNode.h>
+#include <vtkMRMLLabelMapVolumeDisplayNode.h>
+#include <vtkMRMLVolumeArchetypeStorageNode.h>
 
 // VTK includes
 #include <vtkAlgorithm.h>
@@ -46,9 +50,12 @@ vtkMRMLScalarVolumeNode * TestScalarVolumeLoading( const char* volumeName,
                                                    vtkSlicerVolumesLogic* logic );
 vtkMRMLLabelMapVolumeNode * TestLabelMapVolumeLoading( const char* volumeName,
                                                        vtkSlicerVolumesLogic* logic );
-bool TestCheckForLabelVolumeValidity( vtkMRMLScalarVolumeNode* scalarVolume,
-                                      vtkMRMLLabelMapVolumeNode* labelMapVolume,
-                                      vtkSlicerVolumesLogic* logic );
+int TestCheckForLabelVolumeValidity( vtkMRMLScalarVolumeNode* scalarVolume,
+                                     vtkMRMLLabelMapVolumeNode* labelMapVolume,
+                                     vtkSlicerVolumesLogic* logic );
+int TestCloneVolume( vtkMRMLScalarVolumeNode* scalarVolume,
+                     vtkMRMLScene* scene,
+                     vtkSlicerVolumesLogic *logic);
 
 //-----------------------------------------------------------------------------
 int vtkSlicerVolumesLogicTest1( int argc, char * argv[] )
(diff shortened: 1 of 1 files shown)


Modules/Loadable/Volumes/Logic/vtkSlicerVolumesLogic.cxx
```diff
@@ -1216,15 +1204,38 @@ CloneVolume (vtkMRMLScene *scene, vtkMRMLVolumeNode *volumeNode, const char *nam
...
   clonedVolumeNode->CopyWithScene(volumeNode);
+
+  // remove storage nodes
   clonedVolumeNode->SetAndObserveStorageNodeID(NULL);
-  std::string uname = scene->GetUniqueNameByString(name);
-  clonedVolumeNode->SetName(uname.c_str());
-  if ( clonedDisplayNode )
+
+  // remove display nodes (but not the first one)
+  for ( unsigned int i = 1; i < clonedVolumeNode->GetNumberOfDisplayNodes(); ++i )
```
jcfr: Nitpicks: This form may be simpler:

```
while (clonedVolumeNode->GetNumberOfDisplayNodes() > 1)
  {
  clonedVolumeNode->RemoveNthDisplayNodeID(1); // always remove at index 1 since they will shift
  }
```

jcfr:
Thanks for the patch. Looks good. :+1: 

Could you also add/update test of the Volume logic ?

fedorov:
Thank you for working on this!

pieper:
Agreed - thanks for the fix!

agirault:
I'll work on the test for the Volume logic now.

agirault:
I refactored the `vtkSlicerVolumesLogicTest1` file and added new tests for `vtkSlicerVolumesLogic::CloneVolume` which check :
1. That a cloned volume has no storage node.
2. That a cloned volume has display nodes not referencing the original volume (fixed in commit 6ed6752 by copying only the first display node and removing the others).
3. That a cloned volume has the same imageData than the original volume.

Testing the copy of the display nodes involved including `vtkSlicerVolumeRenderingModuleMRML` in the CMakeLists.txt to be able to test the `vtkMRMLVolumeRenderingDisplayNode` classes. I don't think the way I did it is correct so please let me know how to make this cleaner.

Thank you.

jcfr:
> I don't think the way I did it is correct so please let me know

Looks good I will integrate asap.

jcfr:
@agirault Would it be possible to base the last two commits on top of the current master. There are merge conflicts.

Note that tests and associated functions should now all return EXIT_FAILURE, EXIT_SUCCESS 

Thanks

agirault:
Rebase made.

As for the return values, isn't it confusing since `true` will be equivalent to `EXIT_FAILURE` (1) and `false` equivalent to `EXIT_SUCCESS` (0) ?

jcfr:
Thanks.

That is the new convention. Since function returns int and use the EXIT_.
It is okay.

jcfr:
First two commits integrated as [r24861](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24861) and [r24862](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24862)

agirault:
@jcfr you integrated the second commit which tests returning `bool` instead of `int`. I am supposed to integrate the testing macros in the latest commit with the testing of my changes?

jcfr:
> integrated the second commit which tests returning bool instead of int

Indeed. This is my mistake.

> I am supposed to integrate the testing macros in the latest commit with the testing of my changes

Would be great if you could update things to use `EXIT_*` and `int` in your next commit.

Thanks

agirault:
Sorry for the delay, I had been busy on another project. I'll do this now. Since you integrated the two first commits what would you say would be the best way to update the last commit @jcfr ? Should I just cherry-pick this last commit from master, correct the conflicts and update the return values, then push force to only do the PR for this commit?

agirault:
Sorry for the delay, this should be good now @jcfr.

jcfr:
Thanks for following up with your contribution :+1: 

Integrated in [r24890](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24890)

PR #440, opened 2016-01-07
ENH: update MultiVolumeImporter

fedorov:
This update adds support for parsing multivolumes from Siemens DSC MRI. The
functionality was tested on the examples provided by the QIICR MGH team.

PR #443, opened 2016-01-08
ENH: update MultiVolumeImporting

fedorov:
This update should resolve the regression introduced by #440 and improves
robustness by not returning multivolumes that have single frames when parsed
based on ImagePositionPatient+AcquisitionTime. The frame identifying attribute for ImagePositionPatient+AcquisitionTime strategy is now set to AcquisitionTime.


fedorov:
integrated in r24840 = b12cc45049d45eb9b07ffe519781e0bb96fbfdd0

PR #444, opened 2016-01-08
Ihnorton jcfr remove dmri

jcfr:


jcfr:
Integrated in [r24855](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24855)

PR #445, opened 2016-01-13
ENH: update DCMTK hash

fedorov:
The DCMTK update fixes incorrect bit packing order, which would render
invalid any segmentation object created with its API.


fedorov:
committed in

r24867 = 7a44ad25fe4cae1ddabe51f7f2d8274834abc1eb

PR #447, opened 2016-01-19
BUG: Fixed the argument given to FileExists

ayamada0614:
The argument for FileExists function was converted from std::string to char type since FileExists requires const char* argument. I failed to build Slicer application on OS X 10.10.5 by this part.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLTableStorageNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLTableStorageNode.cxx
@@ -77,7 +77,7 @@ int vtkMRMLTableStorageNode::ReadDataInternal(vtkMRMLNode *refNode)
     }
 
   // check that the file exists
-  if (vtksys::SystemTools::FileExists(fullName) == false)
+  if (vtksys::SystemTools::FileExists(fullName.c_str()) == false)
     {
     vtkErrorMacro("ReadData: table file '" << fullName << "' not found.");
     return 0;


jcfr:
Thanks for the patch.

Before moving forward with the integration, would you mind copy/pasting here the build error you got ? 

Looking at the source code of `kwsys`, it seems to accept both `char*` and `std::string` as parameter.

See https://github.com/Slicer/VTK/blob/fe92273888219edca422f3a308761ddcd2882e2b/Utilities/KWSys/vtksys/SystemTools.hxx.in#L295-L298

ayamada0614:
Hi Jcfr,

Thank you very much for pointing out that SystemTools.hxx has 4 types of FileExists.
It was strange but my SystemTools.hxx downloaded by install script yesterday had only 2 types of FileExists as follows:

----------------
  /**
   * Return true if a file exists in the current directory.
   * If isFile = true, then make sure the file is a file and 
   * not a directory.  If isFile = false, then return true
   * if it is a file or a directory.
   */
  static bool FileExists(const char* filename, bool isFile);
  static bool FileExists(const char* filename);

  /**
   * Converts Cygwin path to Win32 path. Uses dictionary container for
   * caching and calls to cygwin_conv_to_win32_path from Cygwin dll
   * for actual translation.  Returns true on success, else false.
   */
---------------- 

SystemTools.cxx also had only 2 types of FileExists functions. So, I might received the error but today I tried to build Slicer again and it was succeeded without any modification I pointed out yesterday.
VTKv6 yesterday I used might be old version. So, please forget this issue though I could not understand why the script downloaded the old version.

Regards,

Atsushi

jcfr:
@ayamada0614  Thanks for following up on the problem.

And thanks again for submitting a pull request, it is always good to receive contribution and work on improving the framework. 

Let me know if you have any other questions, 

In the mean time, I am closing this issue.

jcfr:
And it is also entirely possible you had an older checkout, updated your operating system and then saw the problem.

PR #449, opened 2016-01-27
ENH: Refactored view ruler and orientation marker

lassoan:
Ruler and orientation marker: available in both 2D and 3D viewers, easier to discover (available in viewer controller widgets and application settings), more customizable (size, visibility, custom ruler scales, orientation marker models), implemented with cleaner design (using displayable managers for rendering and regular node attributes for data storage).

New features:
- Ruler and orientation marker is configurable per viewer in the slice/3D view controller widgets (same way as all view options)
- Ruler and orientation marker can be shown in 3D view
- Default ruler orientation marker settings (and additional 3D view settings, such as orthographic projection, show 3D cube, axis labels, depth peeling) can be configured in the application settings (saved in .ini file)
- Allow using any model in the scene as orientation marker (for example, set a custom orientation marker in the 3D viewer: getNode('View1').SetOrientationMarkerHumanModelNodeID(getNode('ToriaClipped03').GetID() )
- All ruler and orientation marker settings (including custom orientation marker) are saved in the scene
- Added "thick" ruler mode: there is an option now to show the ruler with a thicker line and larger font so that it can be seen from a larger distance (e.g., interventional use)

Fixes bugs in the old implementation:
- ruler showed incorrect value when zoomed in or zoomed out very much (instead of hiding the ruler, it was kept displayed at an incorrect scale)
- Ruler was unusable in lightboxview: ruler line was clipped in half, label was not visible
- Some observers were not removed

Design changes compared to the old implementation:
- Ruler and orientation marker are rendered by displayable managers. This allows much simpler and more robust resource management and rendering update (there was previously an infinite loop in the orientation marker update causing Slicer crash by stack overflow).
- Orientation marker is loaded from a single file, color is stored in scalars. This makes the implementation simpler and displaying of multi-color custom models feasible.

Slicer core changes:
- Allow modifying default node content: Scene can store "default" nodes. If a default node is specified for a class then the content of that node will be used to initialize new nodes and reset nodes. Currently used only for slice and 3D views but can be used for other purposes, too (e.g., disable interpolation by default for volume nodes).
- Removed roll/pitch/yaw buttons from 3D view: to reduce the additional complexity caused by adding 2 new buttons (for show/hide ruler and orientation marker)
- Added common helper functions from accessing file resources from MRML nodes. Previously a Slicer-specific environment variable was hardcoded in vtkMRMLFreeSurferProceduralColorNode. Now the environment variable name is configurable in CMake and GetApplicationHomeDirectory and GetApplicationShareDirectory methods in vtkMRMLLogic can be used to determine resource file paths.

Slice view corner annotations and scalar bar were left unchanged in the DataProbe module, as they were not obvious how to clean up and generalize. Slice corner annotations have to be made more generic, allowing displaying of custom text provided by various modules (see for example CornerAnnotations module). Scalar bar will need to be harmonized with the Color module's scalar bar display (that is currently only available in the 3D view).


fedorov:
Andras, thank you for working on this, the description looks really great! I am going to build this branch on mac and test, and report how it worked for me.

On a related note, I wanted to bring in @che85 into discussion, since he spent some time working on implementing text annotation in slice in the context of prostate IGT applications. For example, we need to indicate to the operator that certain image is pre-procedural, or registered. Not sure if this fits into the same topic, but Christian was thinking, and I agree with him, that it would make sense to put this as a generally accessible functionality somewhere in Slicer, and perhaps this module is the right place. Andras, if this is relevant, let us know, and we can provide more details.

che85:
Oh yeah I totally forgot, that I wanted to mention that in the Slicer community. 

I implemented a class, that handles such annotations (including resizing when slice view changes)

See https://github.com/SlicerProstate/SliceTracker/blob/master/Utils/helpers.py

lassoan:
Text display in views is not related to this topic and the design is not trivial (it has to be generic enough to handle all needs but still simple and practical to use). Let's move this discussion on the slicer-devel mailing list.

fedorov:
Makes sense. Christian, can you please initiate the discussion on slicer-devel?

I confirm it bulds and works fine on mac. It is very convenient to use with the link view!

One small thing that happened to me is that somehow the sizes of the orientation marker got out of sync and it is of a smaller size in the Red slice. I first enabled in Red only, then in 3D, then linked the views, and re-enabled, see the result below. Turning orientation marker on/off does not fix this.

image

lassoan:
It's a feature. You can set different stlye and size ruler and orientation marker in different views. If you link the slice views then any change (but only that change) will be applied to all slice views. if you change the orientation marker type for all slice views, the orientation marker size is unaffected.

fedorov:
Ok, let me rephrase the question. How can I make orientation maker have the same size in all slice views from the user interface? I can't figure that out.

lassoan:
Link slice views and select a size.

fedorov:
I see - makes sense. Thanks!

jcfr:
Closing this PR. Superseded by #450

PR #450, opened 2016-01-28
Refactored view ruler and orientation marker

jcfr:
Ruler and orientation marker:
-----------------------------

- available in both 2D and 3D viewers

- easier to discover (available in viewer controller widgets and
application settings)

- more customizable (size, visibility, custom ruler scales, orientation
marker models)

- implemented with cleaner design (using displayable managers for
rendering and regular node attributes for data storage).

New features:
-------------

- Ruler and orientation marker is configurable per viewer in the slice/3D
view controller widgets (same way as all view options)

- Ruler and orientation marker can be shown in 3D view

- Default ruler orientation marker settings (and additional 3D view settings,
such as orthographic projection, show 3D cube, axis labels, depth peeling)
can be configured in the application settings (saved in .ini file)

- Allow using any model in the scene as orientation marker (for example, set a
custom orientation marker in the 3D viewer:

```
getNode('View1').SetOrientationMarkerHumanModelNodeID(getNode('ToriaClipped03').GetID() )
```

- All ruler and orientation marker settings (including custom orientation marker)
are saved in the scene

- Added "thick" ruler mode: there is an option now to show the ruler with a
thicker line and larger font so that it can be seen from a larger distance
(e.g., interventional use)

Fixes bugs in the old implementation:
-------------------------------------

- ruler showed incorrect value when zoomed in or zoomed out very much (instead
of hiding the ruler, it was kept displayed at an incorrect scale)

- Ruler was unusable in lightboxview: ruler line was clipped in half, label was
not visible


jcfr:
@Slicer/slicer-core  @lassoan Beside of few nitpicks. Looks great :+1:

Punzo:
Hi,

I'd like to have a feedback how to change my code for this patch:
at the moment I override the updateRuler and makeRuler in this way
https://github.com/Punzo/SlicerAstro/blob/master/AstroDataProbe/AstroDataProbe.py#L450-L454

The nice new method in the API  AddRulerScalePreset, will not help me because it provides a ruler with ticks with a constant distance. In my case, see here:
https://raw.githubusercontent.com/Punzo/SlicerAstroWikiImages/master/wiki3.png
the position of each ticks both in vertical and horizontal has to be calculated with the astronomical non linear WCS:
https://github.com/Punzo/SlicerAstro/blob/master/AstroDataProbe/AstroDataProbe.py#L236-L278 

Is it possible at the moment in Slicer to do a kind of factory for DisplayableManager classes?
I don't know if here is the right place to ask, let me know if I have to shift the discussion elsewhere.

jcfr:
Thanks for reviewing. 

Since we would like the feature to be general so that it can be extended, it would be great to generalize a bit more to allow `SlicerAstro` to leverage this feature.

Do not hesitate to push a topic on your fork with additional commits generalizing further.

> I don't know if here is the right place to ask

Here is the place.

lassoan:
@Punzo The coordinate system axes in SlicerAstro look very nice. Note that I've added a ruler and not coordinate system axes, which looks somewhat similar, but actually serves a very different purpose.

Ruler is location-independent, we don't show absolute position values, just the total length. The current ruler is flexible enough to be used with non-linear coordinate systems, you just need to add all those values that you want to allow as ruler lengths and specify the corresponding unit name that will be displayed.

Coordinate system axes: We have never had this in Slicer (probably because in medical imaging we rarely have an absolute coordinate system), you can just keep your working implementation as is. If several Slicer users request it then we may move it into the Slicer core. I would recommend to implement coordinate system axes similarly to the new ruler: using displayable managers and the vtkAxisActor2D. You can implement displayable managers in either C++ or  Python.

pieper:
@lassoan can you describe or point to the definition of the ruler as an overlay on the 3D view?  It's not clear to me how it is used.  It seems it would make more sense to add tick marks to the cube natively in 3D.

jcfr:
> more sense to add tick marks to the cube natively

If I can chime in. The ruler is nice to give a sense of the scale when doing screenshot with only the 3D viewer is visible.

On the other hand, the 3D cube being the bounding box, it won't always be visible.

That said, it would be possible to improve the displayable manager associated with the 3D cube so that tick are displayed ...

lassoan:
We show the ruler in the 3D navigation scene to give surgeons an idea about the size of the objects they see. Of course this only works in parallel projection mode therefore the ruler is disabled in perspective projection mode. The ruler has to be always visible and should be in the view normal plane, so a 3D widget would not work.

For navigation the cube and the labels on the side are not helpful and often they are confusing (sometimes a big letter appears in the view and surgeons have no idea what it is), so we turn them off. In general, the cube and labels is only useful for a birds-eye view and not working well if you zoom in to see details.

I can imagine that showing coordinate axes on the cube might be useful for some people and it should not be difficult to change the cube's displayable manager to show ticks and labels.

Punzo:
@jcfr @lassoan thanks for the feedback.
My point was that if is possible to use a factory and registrate my axis displayable manager on the top of the ruler one. This for two reasons:
1)I could take advantage of this patch (I could use the ruler buttons in the 2-D, 3-D interface for display my axes).
2) this will automatically hide the ruler which is something I don't want to show. 

However, It seems to much work for doing this. The best for me, probably, is  to write my displayable manager for axes and connect it just to the interface of my module AstroDataProbe.

lassoan:
You should able unregister the ruler displayable manager from the slice view displayable manager factory singleton using UnRegisterDisplayableManager and register your displayable manager(s).

The button in the view controllers update the view nodes. Your displayable manager can show the coordinate system axes when the ruler is enabled in the node.

Punzo:
ok, thanks @lassoan! that was very helpful.

jcfr:
@Slicer/slicer-core  Now the backward compat and integration issues have been addressed and discussed. If there are no other comments, I will be integrated later this afternoon.

fedorov:
:+1:

jcfr:
@lassoan Note that your latest commit https://github.com/lassoan/Slicer/commit/19facbe0af8274e65f4c9cc381b11703f26f262d has already been integrated in this topic.

jcfr:
Integrated in [r24905](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24905) to [r24908](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24908)

PR #452, opened 2016-02-02
ENH: Added modified event for SetNthMarkupAssociatedNodeID

sumedhasingla:


pieper:
Makes sense to me - @naucoin can you have a look?

naucoin:
@sumedhasingla what's your motiviation for adding it? The commit messages should include the why of the change.
The code itself looks fine.

sumedhasingla:
Each time a markup is placed in 3d, we want to identify the object it is placed on. This information is stored as associated Node id. When a new mark up is placed, associated node id is set to blank (" ") by default ,  MarkupAddedEvent is triggered and then associated node id is set to appropriate node id. So there is no way of identifying the associated node id of the newly added node using events.

naucoin:
Perfect, thanks. If you can amend the commit message to include that we can integrate the request.

jcfr:
Thanks for the patch :+1: 

@naucoin @pieper Thanks for the review

Commit message updated, commit tntegrated in [r24921](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24921)

PR #453, opened 2016-02-04
ENH: Update Git tags for OpenIGTLink library and OpenIGTLink module to activate connection error log.

tokjun:
See: 

http://www.na-mic.org/Bug/view.php?id=4147

https://github.com/openigtlink/OpenIGTLinkIF/pull/63


jcfr:
Thanks for the patch :+1: 

Integrated in [r24920](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24920)

PR #454, opened 2016-02-05
COMP: Do not force VTK repository nor tag

vovythevov:
For external projects it is extremely beneficial to be able to prescribe
the VTK repository and tag so they can include project specific fixes of
VTK.


vovythevov:
After talking with @jcfr, this doesn't really solve the problem. Indeed, this will interfers with continous builds (that won't update themselves) and developer machines (that also won't update) when the hash is changed.

vovythevov:
Note: An issue describing the problem was submitted here: http://www.na-mic.org/Bug/view.php?id=4148

PR #456, opened 2016-02-11
BUG: Handle exception if string encoding fails

cpinter:
Sometimes there are tricky characters in the strings that need to be encoded (in my case 0xe9), throwing a UnicodeDecodeError exception, which, if not handled properly, results in a long list of SliceViewAnnotations error messages, furthermore stoppage of the running script

### DIFF
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -587,8 +587,14 @@ def toVTKString(str):
   """Convert unicode string into 8-bit encoded ascii string.
   Unicode characters without ascii equivalent will be stripped out.
   """
-  return str.encode('latin1', 'ignore')
-
+  vtkStr = ""
+  for c in str:
+    try:
+      cc = c.encode("latin1", "ignore")
+    except (UnicodeDecodeError):
+      cc = "?"
+    vtkStr = vtkStr + cc
+  return vtkStr
 
 #
 # File Utlities


cpinter:
Not sure if this is the best solution, but it is one nevertheless.

pieper:
Interesting, I would have thought the python unicode support would be more robust.

How about if the exception occurs we do something like:

newStr = ""
for c in str:
  try:
   cc = c.encode("latin1", "ignore")
  except (UnicodeDecodeError):
   cc = "?"
 newStr = newStr + cc

So that we get what we can out of the string?

jcfr:
From https://docs.python.org/2/howto/unicode.html#encodings

```
Python’s default encoding is the ‘ascii’ encoding. The rules for converting a Unicode string into the ASCII encoding are simple; for each code point:

    If the code point is < 128, each byte is the same as the value of the code point.
    If the code point is 128 or greater, the Unicode string can’t be represented in this encoding. (Python raises a UnicodeEncodeError exception in this case.)
```

This basically explained the behaviour observed by @cpinter 

May be you could use an approach similar to what is done in  `slicer.util.unicodeify`. See [Base/Python/slicer/util.py#L616-L621](https://github.com/Slicer/Slicer/blob/b7dad19b23c18b1f48aaf86ca432c7c299f74c0c/Base/Python/slicer/util.py#L616-L621)

pieper:
It would be great if someone could go through all the toolkits and determine who strings are best handled for each of them and how to convert to something the toolkit can handle.  Is the core issue that VTK cannot handle unicode text display?

lassoan:
VTK can display unicode text (see for example http://vtk.1045678.n5.nabble.com/Unicode-support-td5724528.html). Most (or all?) VTK filters and IO classes support unicode strings. Unicode filenames are a bit more tricky, but should be usable with some limitations (http://markmail.org/message/m34q4sh3srypb4a4).

cpinter:
I like the solution Steve suggested, but we can use unicodeify if it works for this case.
Another strange thing is that I cannot reproduce the issue with my PC at home, so I can only try it from the lab.

pieper:
It would be great to sort this out across all of slicer. We have had
problems with file paths, DICOM names, etc. But we need to be very careful
about reading and writing floating point numbers it text data files.

cpinter:
I finally went with Steve's suggestion, see updated commit.

pieper:
Looks good to me - thanks Csaba.

jcfr:
Thanks :+1: 

Integrated in [r24938](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24938)

cpinter:
Thanks!

PR #457, opened 2016-02-15
BUG: change findChildren function behavior from "OR" filter to "AND" filter

jonieva:
Modify behavior of the slicer.util.findChildren function


lassoan:
Thank you. Could you please convert this to a proper if/then/else than using eval? Eval use should be minimized to special operations that cannot be performed in any other way. For example, I cannot do step-by-step debugging of the evaluated expression.

PR #458, opened 2016-02-15
BUG: change "slicer.util.findChildren" behavior from "or" to "and" filter

jonieva:


Base/Python/slicer/util.py
```diff
@@ -140,29 +140,20 @@ def findChildren(widget=None,name="",text="",title="",className=""):
...
-    elif className:
-      try:
-        p.className()
-        if fnmatch.fnmatchcase(p.className(), className):
-          children.append(p)
-      except AttributeError:
-        pass
+    condition = True
+    if name:
+      condition &= fnmatch.fnmatchcase(p.name, name)
```
lassoan: If p doesn't have a name attribute then an exception is raised and all further children are ignored. To fix this, you would need to change
fnmatch.fnmatchcase(p.name, name) 
to something like
hasattr(p, 'name') and fnmatch.fnmatchcase(p.name, name)
for all checks (name, text, title, className)

Base/Python/slicer/util.py
```diff
@@ -140,29 +140,20 @@ def findChildren(widget=None,name="",text="",title="",className=""):
...
-      except AttributeError:
-        pass
-    elif className:
-      try:
-        p.className()
-        if fnmatch.fnmatchcase(p.className(), className):
-          children.append(p)
-      except AttributeError:
-        pass
+    condition = True
```
lassoan: Rename this variable to something more specific, which makes it clear what True and False mean. For example matched_filter_criteria would be specific enough.

Base/Python/slicer/util.py
```diff
@@ -140,29 +140,20 @@ def findChildren(widget=None,name="",text="",title="",className=""):
...
+    if name:
+      condition &= fnmatch.fnmatchcase(p.name, name)
+    if text:
+      condition &= fnmatch.fnmatchcase(p.text, text)
+    if title:
+      condition &= fnmatch.fnmatchcase(p.title, title)
+    if className:
+      condition &= fnmatch.fnmatchcase(p.className(), className)
+    try:
```
lassoan: If you use hasattr() then this try/catch should not be necessary.

jonieva:
I'm sorry I closed the request by mistake. I hope everything is fine now

lassoan:
thank you, it looks good to me

fedorov:
Small warning here, this function is probably used broadly in Slicer core and extensions. It might make sense to at least announce this change on slicer-devel, or maybe even look how it is used to get an idea if it will break any functionality.

jonieva:
I agree, that's why I sent it to the list first. But no one seemed to say
anything against Andras suggestion to make the change... We can maybe wait
a few days more if you think it's necessary.


Jorge Onieva

fedorov:
Jorge, you did the right thing, I am not complaining! I would just follow up on the devel list that this moves ahead and reference this PR. Often issues are discussed on the list, but then nothing happens.

lassoan:
I agree, a note on the mailing list and maybe an additional sentence in the function documentation that explains the behavior of the filtering criteria (logical AND operator) would be useful.

jonieva:
Andrey, I know, no offense taken. It's just this small change is becoming
harder than I expected. But you know, as in so many other situations in
life, I guess the first time is never easy :-)

I sent a follow-up to the list and I added a detailed documentation in the
function itself. I hope we are ready to go!

Honestly I think that people even didn't realize in the first term that
this was the behavior of this function, as I guess 'name' is generally used
and then the first result is typically chosen, so I don't think that the
impact is going to be very high.

Tx,

Jorge Onieva

cpinter:
Thanks Jorge for taking action! I'm pretty sure nobody expects "or" operation for this function. Hopefully nobody did, and the change will not affect anything functionally. Anyway, at least the previously mentioned announcement and running the Slicer python tests locally would make sense.

I did a quick search in the extensions (not the latest, but a three months old snapshot), and it seems that nobody uses findChildren with multiple criteria, except for SlicerRT (where I assumed AND), and SlicerIGT (it will be needed to check).

lassoan:
I've checked SlicerIGT. AND operation is assumed there, too.

cpinter:
@jonieva FYI in the extensions "text" was the most common argument of the function.

fedorov:
We extensively use findChildren in https://github.com/SlicerProstate/mpReview, but only the name argument.

@jonieva thanks for working on this! I hope you don't get discouraged by all the back and forth. Indeed a small change can take a lot more than initially expected, but I think this is all for better, and of course "first cut is the deepest" ...

lassoan:
By the way, not in this pull request, but I would also add a findChild (similar to http://doc.qt.io/qt-5/qobject.html#findChild), as it is quite ugly and inefficient that we most of the time use this method as findChildren(...)[0].

jonieva:
I agree. I actually thought about it in the first term, but I tried to keep
this first pull request as simple as possible, as I guessed it wouldn't be
as easy as I thought :-).

I'm happy to do it in this pull request though, I think it would be a nice
feature and wouldn't break any code. And well, it's still the same scope,
so I don't see why not to do it in this pull request.

Jorge Onieva

pieper:
Thanks for cleaning this up - it's definitely better than the old version.
 +1

jcfr:
@jonieva Thanks for working on this patch.

After reviewing, it shouldn't be needed to modify the `DataProbe`, your changes should be fully backward compatible.

Please, consider the following improved implementation: https://github.com/jcfr/Slicer/commit/8cc6c563a840df3a63ba5fb640ca5d7e90b1e2e5

jcfr:
Integrated in [r24957](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24957)

@jonieva Could you consider submitting an other PR adding `findChild`.

Instead of using the implementation originally suggested (see below).  The idea would be to:

* extend `findChildren` to support an additional parameter named (for example) `limit` (defaulting to `None`). Adding the `limit` parameter will allow to bail early from the `while` loop and be more efficient.

* add `findChild(widget=None, name="", text="", title="", className="", strict=False)`. 

* extend `findChildren` by a `strict` parameter. The meaning of `strict=True` + `limit != None` can be understood as setting `expected_children_count`.

Remarks:
* `strict` parameter doesn't have to be added. I just specified it here to also account for the case where you were throwing an `Exception` in your original implementation.

Suggested implementation:

```
def findChild(widget=None, name="", text="", title="", className="", strict=False):
        """ Return a single child widget that meet all the given criteria.
        :param widget: parent widget where the widgets will be searched
        :param name: name attribute of the widget
        :param text: text attribute of the widget
        :param title: title attribute of the widget
        :param className: className attribute of the widget
        :param strict: If ``True`` and if there is more than one widget matching the
        conditions, raise an :exc:`~exceptions.Exception`.
        :return: widget that meet all the given criteria (or ``None`` if no widget was found)
        """
        try:
          return findChildren(widget, name, text, title, className, limit=1, strict=strict)[0]
        except IndexError:
          return None
```

Original proposal for `findChild`:

```
def findChild(widget=None, name="", text="", title="", className=""):
        """ Return a single child widget that meet all the given criteria.
        If there is more than one widget that matches the conditions, an Exception is raised.
        If there is no widget that matches the conditions, the method returns None.
        :param widget: parent widget where the widgets will be searched
        :param name: name attribute of the widget
        :param text: text attribute of the widget
        :param title: title attribute of the widget
        :param className: className() attribute of the widget
        :return: single widget that meet all the given criteria (or None otherwise)
        """
        results = SlicerUtil.findChildren(widget, name, text, title, className)
        if len(results) == 0:
            return None
        if len(results) > 1:
            raise Exception("There is more than one widget that matches the given conditions")
        return results[0]
```

lassoan:
QObject has findChild (http://doc.qt.io/qt-4.8/qobject.html#findChild), probably the best would be to implement that behaviour. (as I remember, @jonieva's implementation does just that)

jcfr:
>  @jonieva's implementation does just that

It didn't. But after a quick check .. it works as expected if you specify the type of object.

```
>>> findChildren(name="LoadDICOMAction")[0]
QAction (QAction at: 0x42d0a70)
>>> mainWindow().findChild(qt.QAction, "LoadDICOMAction")
QAction (QAction at: 0x42d0a70)
```

I think it is preferable to have a fast lookup only given the object name with the option of specifying the className. The implementation would then use either (`slicer.util.findChildren` + `limit=1`) or `parentWidget.findChild` if a class was specified.

That said, the signature would then be:

```
def findChild(name, parentWidget=None, className=None):
        """ Return a single child widget that meet all the given criteria.
        :param name: name attribute of the widget
        :param parentWidget: parent widget where the widgets will be searched
        :param className: className attribute of the widget
        :return: widget that meet all the given criteria (or ``None`` if no widget was found)
        """
[...]
```

jonieva:
Following a comment from Csaba (http://slicer-devel.65872.n3.nabble.com/Changing-the-behavior-of-slicer-util-findChildren-td4036266.html) I found out there's an error in the current findChildren function.
Lines 158-159 should be replaced by:
if hasattr(p, attribute):
        attr_name = getattr(p, attribute)
        if attribute == 'className':
          # className is a method, not a direct attribute. Invoke the method
          attr_name = attr_name()
        if fnmatch.fnmatchcase(attr_name, kwargs[attribute]):
          matched_filter_criteria = matched_filter_criteria + 1

PR #463, opened 2016-02-25
ENH: update SlicerDMRI remote module to a82fde48

ihnorton:
```
$ git shortlog --no-merges d712d9522a..a82fde4870

Isaiah Norton (3):
      DWMasking: set b-value select slider max to 20k
      DWMasking: COMP: fix test that was using old parameter
      COMP: fix windows build error due to outdated CLI entrypoint
```


lassoan:
I've merged the change in git accidentally because I thought that it's an ExtensionIndex update but now I see that it's not. Hopefully it won't break git/svn sync...

lassoan:
Merged into SVN rev 24949

ihnorton:
Thanks.

jcfr:
>  it won't break git/svn sync...

System is robust and account for this.

I just need to resume the cron job that stopped working with the update for the factory to "El Capitan"

lassoan:
Very good then, thanks for the information

PR #467, opened 2016-03-07
STYLE: Rename Build Dir in PreventInSourceBuilds

agirault:
Make the name used as the build directory consistent with line 22.


jcfr:
Thanks for the patch :+1: 

Integrated in [r24962](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24962)

PR #468, opened 2016-03-09
BUG: Update BRAINSTools to BRAINSFit fix for centerOfROI initializer

jcfr:
Tested-by: Hans Johnson <hans-johnson@uiowa.edu>

```
$ git shortlog e257e62...29d203c --no-merges
Alexander Leinoff (5):
      COMP: BCD should only request VTK when building GUI
      COMP: CompareImageFilter should only use VTK when building GUI
      BUG: Replaces all baselines for BCDTests
      BUG: Fixes typo in previous commit
      BUG: Fixes version check in top level CMAKELists.txt

Ali Ghayoor (13):
      ENH: Upgrade vnl_powell to itkPowellOptimizerv4 in BCD
      ENH: Write the debug image based on requested debug level
      ENH: BCD can process input images with different dynamic range
      ENH: Change tolerance and step size for Powell optimizer
      BUG: Fixed a bug in BABC when a low debugLevel requested
      ENH: New parameters for multimodal segmentation in BABC
      ENH: Updated purePlugsThreshold value for AutoWorkup
      BUG: fixed a bug in BABC denoise filtering
      ENH: Initialize a matrix variable
      ENH: remove an intermediate variable to avoid extra memory usage
      BUG: defined an orderedmap class for BABC
      STYLE: Cleanup unnecessary if statement
      ENH: Prevent holes in created brain mask by ROIAutoFilter

Andrey Fedorov (2):
      BUG: fix centerOfROI initializer
      ENH: adjust restriction on the required registration phases

Hans J. Johnson (40):
      ENH: Update ITK, ANTS, and teem
      STYLE: Remove MinSizeRel option
      COMP: Resolve clang static analysis warnings
      BUG:  Compilation error fixed due to typo.
      ENH: Update external tools support.
      COMP: Improve coverage by using removeIntensityOutliers flag.
      COMP: Force C++11 compliance testing
      COMP: Added compatibility testing with apple and CXX11
      BUG: Nipype now reqires invert_initial_moving_transform to be set
      ENH: Selection of denoising filter improved.
      ENH: Changed options to reflect current processing
      ENH: Updated ANTs for improved DenoiseImage
      BUG: The Rician noise model fails for DenoiseImage
      ENH: Added key files for new denoised gaussian images.
      ENH: Use denoised images for atlas reference
      ENH: Add atlas images to reference atlas.
      ENH: Fine-tune denoising options in BAW.
      BUG: Need to work around nipype bug
      ENH: Use denoised t1 from atlas for registration
[...]

PR #469, opened 2016-03-10
BUG: Update BRAINSTools to fix mask checks

fedorov:
This should resolve the regression introduced in r24963 that caused failing
EMSegment test. The problem was identified and solution proposed by @jcfr

$ git shortlog 29d203c...16526a5 --no-merges
Andrey Fedorov (1):
      BUG: fix mask checks for centerOfROI initializer

cc: @hjmjohnson

also see https://github.com/Slicer/Slicer/pull/468


jcfr:
Thanks :+1: 

Integrated in [r24964](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24964)

PR #471, opened 2016-03-11
ENH: Exposing jqplot options to select marker size and line width in …

fbudin69500:
…qMRMLChartView

Exposing two new options for jqplot: 'size' allows to select the size of the marker, and 'lineWidth' allows to select the width of a line. This allows to change these options when using qMRMLChartView. Both properties should be set with a string containing an integer (greater than 0).


pieper:
I like it : )

+1

jcfr:
:+1: Thanks for the patch. As soon as my local rebuild complete .. I will integrate.

jcfr:
Integrated in [r24968](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24968)

PR #472, opened 2016-03-14
ENH: Modernizing vcl_ to std::

hjmjohnson:
In all supported compilers, the need for vcl_ specialized
functions has been removed.  there is no longer a need
to have these overrides, and code is easier to read
and easier to maintain without these specializations.

The vcl_* definitions can now be greatly simplified.
After removing specializations for early non-conformant
c++ compilers the end result was that only the
std:: version of the functions could ever be
used by the compiler.

ITK_SCRIPT=ITK/Utilities/Maintenance/VCL_ModernizeNaming.py
SRC_BASE_DIR=$(pwd)
for ext in ".h" ".cxx" ".cpp" ".hxx" ".hpp" ".txx"; do
  find ${SRC_BASE_DIR} -type f -name "*${ext}" \         -exec python ${ITK_SCRIPT} {} \;
done

### DIFF
--- a/Libs/vtkITK/itkGrowCutSegmentationImageFilter.h
+++ b/Libs/vtkITK/itkGrowCutSegmentationImageFilter.h
@@ -9,9 +9,11 @@
 
 //#include "itkGrowCutSegmentationUpdateFilter.h"
 
-#include "vcl_list.h"
+#include <vcl_compiler.h>
+#include <iostream>
+#include "list"
 
-#include <vcl_vector.h>
+#include <vector>
 
 
 #ifndef PixelState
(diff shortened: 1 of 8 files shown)


hjmjohnson:
@pieper ,@jcfr This is backwards compatible for a long time with all supported compilers.  Recent cleanup in vxl indicated that vcl_* now is just an alias to std::* for all of these cases.

pieper:
Sounds like a great idea to me.

jcfr:
Thanks for the patch :+1: 

Integrated in [r24974](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24974)

PR #466, opened 2016-03-04
3032 improve packaging error reporting

jcfr:
This topic updates the build system:

 * address [#3032](http://na-mic.org/Mantis/view.php?id=3032) ensuring Slicer application packaging errors are reported on the dashboard
 * update dashboard driver scripts to provide better feedback regarding the build context (it displays variable in friendly way)

I would like to also get your feedback regarding [f18100da5da10b9](ENH: Remove support for building an ExtensionsIndex directly from Slicer), was this (undocumented) way of building extensions used ?

For reference, documentation to build an extensions index is [here](http://wiki.slicer.org/slicerWiki/index.php/Documentation/Nightly/Developers/Build_ExtensionsIndex)

PR #473, opened 2016-03-18
BUG: Slicer_EXTENSION_DESCRIPTION_DIR is a PATH, not a STRING

fbudin69500:
Slicer_EXTENSION_DESCRIPTION_DIR was declared as a STRING. cmake-gui does not offer the possibility to open a file browser and select a directory from the user interface if a variable is not defined as a PATH or FILEPATH.


jcfr:
Thanks for the PR.

I added your commit on top of the PR modernizing the extension build system. See #466

PR #475, opened 2016-03-23
ENH: DICOMListener with optional indexing, specifying port and incoming data directory

che85:
- added possibility for:
-- defining directory for incoming data
-- incoming port (default: 11112)
-- enabling/disabling indexing by providing dicomDatabase (True) or not (False)
- improved error handling and feedback for user
- when no dicomDatabase specified, incomingDir is required, otherwise an UserWarning is raised

for reference see: http://slicer-devel.65872.n3.nabble.com/DICOMListener-with-optional-Indexing-td4036422.html


che85:
After discusstons on http://slicer-devel.65872.n3.nabble.com/DICOMListener-with-optional-Indexing-td4036422.html modifying the code again

PR #476, opened 2016-03-24
ENH: New class DICOMStoreSCPProcess inherited from DICOMProcess:

che85:
- DICOMListener now inherited from DICOMStoreSCPProcess since DICOMListener is depends on storescp
- improved error handling and feedback for user

for reference see: http://slicer-devel.65872.n3.nabble.com/DICOMListener-with-optional-Indexing-td4036422.html

### DIFF
--- a/Modules/Scripted/DICOMLib/DICOMProcesses.py
+++ b/Modules/Scripted/DICOMLib/DICOMProcesses.py
@@ -64,7 +64,6 @@ class DICOMProcess(object):
     print ("Starting %s with " % cmd, args)
     self.process.start(cmd, args)
 
-
   def onStateChanged(self, newState):
     print("process %s now in state %s" % (self.cmd, self.QProcessState[newState]))
     if newState == 0 and self.process:
@@ -73,6 +72,8 @@ class DICOMProcess(object):
       print('error code is: %d' % self.process.error())
       print('standard out is: %s' % stdout)
       print('standard error is: %s' % stderr)
+      return stdout, stderr
+    return None, None
 
   def stop(self):
     if hasattr(self,'process'):
@@ -80,6 +81,7 @@ class DICOMProcess(object):
         print("stopping DICOM process")
         self.process.kill()
         self.process = None
+
 
 class DICOMCommand(DICOMProcess):
   """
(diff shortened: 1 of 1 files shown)


pieper:
Looks good to me.

I know these are hard to thoroughly test because of the dependencies on other systems, but were you able to confirm that you can use storescu to put files into the database with this changed version of the code?

che85:
Yes, I tested sending to the port and it was indexed into the dicom database

jcfr:
Thanks for the patch.

If there is consensus. LGTM

Nitpicks: to facilitate review, you could even split 5533337 in two commits.

lassoan:
Looks good in general. The class should be used carefully, as when you request to stop a store process, actually all of them are killed. It would be nice to store the pid when the process is started and kill the process corresponding to that pid (not needed in this pull request).

che85:
Well, what I was thinking is, what if the DICOMListener has to get started and there is a other storescp process already running (maybe on another port) then the user is prompted with the question if running processes should be termintated. Clicking yes results in killing all storescp processes. 

It should better be checked if a storescp process is running that uses the same port as the DICOMListener uses and then specifically kills this process. 

Storescp processes that are listening to other ports should not stop DICOMListener from getting started.

lassoan:
We shouldkill those processes that we started. In addition to that it could be nice to search for processes that listening on the port that we want to use to take care of processes that are hanging around due to any reason (netstat can tell which process uses a particular port; or you can parse the command-line of storescp processes).

jcfr:
Related to netstat ... 

* A program with interesting part we could potentially re-use: http://qt-apps.org/content/show.php/QNetStatView?content=157088  Issue is that it is GPL, may be possible to contact the original author to ask to re-license. 

* An other one in python that is license under BSD: https://pypi.python.org/pypi/pynetstat

che85:
Any news about when you are going to merge this PR?

jcfr:
@che85 Thanks again for the patches

Commit squashed and integrated in [r25002](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25002)

PR #477, opened 2016-03-25
BUG: When using slicer.util.createProgressDialog() QProgressDialog di…

che85:
…dn't show up

- setting modal=True as default fixed this bug

### DIFF
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -724,6 +724,7 @@ def createProgressDialog(parent=None, value=0, maximum=100, labelText="", window
   """
   import qt
   progressIndicator = qt.QProgressDialog(parent if parent else mainWindow())
+  progressIndicator.modal = True
   progressIndicator.minimumDuration = 0
   progressIndicator.maximum = maximum
   progressIndicator.value = value
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -714,7 +714,7 @@ def messageBox(text, parent=None, **kwargs):
       setattr(mbox, key, value)
   return mbox.exec_()
 
-def createProgressDialog(parent=None, value=0, maximum=100, labelText="", windowTitle="Processing...", **kwargs):
+def createProgressDialog(**kwargs):
   """Display a modal QProgressDialog. Go to QProgressDialog documentation
   http://pyqt.sourceforge.net/Docs/PyQt4/qprogressdialog.html for more keyword arguments, that could be used.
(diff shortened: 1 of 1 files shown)


pieper:
That method is used a few places in Modules/Scripted/DICOMLib/DICOMWidgets.py  Did you see that the impact of the change is on that code?

che85:
Maybe we can do it like when there is no parent given, that it should be modal?

pieper:
I realize my earlier post was not well worded.  I meant: did you try using the parts of the dicom module where createProgressDialog is used and confirm that they behave the as expected when the modal flag is turned on?  That code was developed years ago and I don't recall right now why the flags were set as they were but I do recall that it was tricking getting the progress dialog to behave as expected with the cancel button working etc.

In general yes, I think it's better that all options be exposed rather than hard coded, especially if we are changing behavior.

che85:
I tried the DICOM module and it seems like there is not really any difference....

Let me take a look at the code and I will try to make it as flexible as possible

jcfr:
Thanks for streamlining the code :+1: 

In 763ccb1 , you mention  `fixed this bug`. Which bug are you referring too ?

che85:
@pieper What do you think about the last commit with just using kwargs. With using kwargs.pop we are able to define defaults if it has not been defined. If there is no parent defined then it's automatically modal  (except the user added modal=True, what would not make sense to me)

che85:
@jcfr Sorry for the confusion, I am not sure if it is really a bug. In one of my modules progress dialogs simply don't appear since I am not defining a parent and it seems that mainWindow() somehow makes problems regarding this. When I explicitly say qt.QProgressDialog() then the progress dialog appears like it is expected.

lassoan:
Modal dialogs are interfering with event processing. They should be avoided unless you really know what you are doing. Have you called QApplication.processEvents() after creating the window? If not then it's normal that the window does not appear.

che85:
@lassoan Ok but then I don't get why it works when I don't define a parent. Can you explain to me why? 

When I use qt.QProgressDialog() instead of qt.QProgressDialog(parent if parent else mainWindow())  it works just fine. Or are there any ongoing events in mainWindow and that's why it never appears?

lassoan:
Some underlying code anywhere may call processEvents, but you cannot rely on that, so if you need processEvents then you have to call it.

Does calling processEvents fix the problem?

che85:
Yes using processEvents works ...

che85:
I guess we can close this PR since processEvents fixes the problem?

jcfr:
Per discussion, closing this PR.

PR #480, opened 2016-04-07
BUG: Fixed edit properties option in node selectors

lassoan:
Fixes http://www.na-mic.org/Bug/view.php?id=2109

Now any module can associate a MRML node type with a module (instead of having a hardcoded list of modules
in qSlicerApplication) and when switching to a module, the chosen node is selected as active node.
Associations can be added by calling qSlicerApplication::registerNodeModule().
Multiple modules can be associated with the same MRML node type.

The best module for a specific node instance is determined run-time, by calling
qSlicerAbstractModuleWidget::nodeEditable() for each candidate and choosing
the one that has the highest confidence in handling the node.
This mechanism is used for selecting the CLI module that corresponds to a specific CLI module parameter node.
This mechanism is also used for switching to volume rendering module if a volume rendering clipping ROI node
is selected for editing (instead of switching to the Annotations module).
The most suitable module's qSlicerAbstractModuleWidget::setEditedNode() method is called to select the currently edited node.
qSlicerAbstractModuleWidget::nodeEditable() and qSlicerAbstractModuleWidget::setEditedNode() methods can be overridden
in scripted loadable modules as well.

As edited node selection is now possible using a clean, unified method call, it is no longer necessary
to access GUI widgets directly. Therefore, almost all editProperties() method calls in subject hierarchy plugins were
removed (replaced by a common implementation in the base class). This new, simplified method fixes robustness issues
of using node name filters for selecting a node in model hierarchy, inability to choose annotation hierarchy nodes, etc.

qSlicerAbstractModuleWidget::setEditedNode(vtkMRMLNode* node, QString role, QString context) has two extra (optional)
arguments. Role can be used to specify additional nodes for a module (e.g., if there are multiple inputs for a module
then role argument can select between them). Context can be used for specifying a sub-selection within a node (for
example, Segment Editor module can be activated for a selected segment - specified by 'context' - of a specific
segmentation node - specified by 'node').

### DIFF
--- a/Base/QTCore/qSlicerAbstractModuleRepresentation.h
+++ b/Base/QTCore/qSlicerAbstractModuleRepresentation.h
@@ -21,6 +21,9 @@
 #ifndef __qSlicerAbstractModuleRepresentation_h
 #define __qSlicerAbstractModuleRepresentation_h
 
+// Qt includes
+#include <QString>
+
 // CTK includes
 #include <ctkPimpl.h>
 
@@ -31,6 +34,7 @@
 
 /// class vtkSlicerApplicationLogic;
 class vtkMRMLAbstractLogic;
+class vtkMRMLNode;
 class qSlicerAbstractCoreModule;
 class QAction;
 class qSlicerAbstractModuleRepresentationPrivate;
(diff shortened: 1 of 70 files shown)


cpinter:
Nice work, thanks! The UI widget poking in SH was indeed a bit messy, and it's always good to get rid of hard-coded "server-side" handlers (like qSlicerApplication::nodeModule). :thumbsup:

pieper:
Very nice!  Thanks!

Base/QTCore/qSlicerAbstractModuleRepresentation.h
```diff
@@ -50,6 +54,24 @@ class Q_SLICER_BASE_QTCORE_EXPORT qSlicerAbstractModuleRepresentation : virtual
   /// The module is set right before setup() is called.
   qSlicerAbstractCoreModule* module()const;
 
+  /// Allows other modules to selecting input and output nodes in the module's GUI.
```
jcfr: `selecting` -> `select`

Base/QTGUI/Testing/Cxx/qSlicerScriptedLoadableModuleWidgetTest.cxx
```diff
@@ -22,9 +22,13 @@
 #include <ctkTest.h>
 #include <ctkUtils.h>
 
+// VTK includes
+#include "vtkNew.h"
```
jcfr: `"vtkNew.h"` -> `<vtkNew.h>`

jcfr:
Very nice :smile:

Nitpick: Instead of having the module registering itself with the application, would it makes sense to have the module describing itself and then making use of the information in module manager when the module is setup. It would facilitate documentation.

lassoan:
The register method allows an arbitrary module allows any module to associate a node with any other module (not with only itself). For example, if a module adds a new node type then it can associate it with the Data module or Subject hierarchy module or some other existing generic module. This flexibility might not be necessary, so if you think that a member function similar to contributors() etc would be nicer then we can change this.

lassoan:
Thanks for the feedbacks. I've fixed code review findings and committed to r25003.

jcfr:
Thanks Andras.

PR #482, opened 2016-04-13
BUG: 4169-two bugs are fixed in the OpenIGTLink library, SetSendBlock and IGTLUnitAppend

leochan2009:


pieper:
Tested that it built on mac and committed as r25010.

PR #484, opened 2016-04-15
ENH: Added SQLite MRML storage node.

ayarmarkovich:
I have added support for SQLite database including:
- new vtkMRMLTableSQLiteStorageNode
- test for vtkMRMLTableSQLiteStorageNode
- support for vtkMRMLTableSQLiteStorageNode in the Tables module including:
  - drag and drop of .db files will read tables in the database into vtkMRMLTableNnode's
  - saving of vtkMRMLTableNnode's into .db files

### DIFF
--- a/Libs/MRML/Core/vtkMRMLTableSQLiteStorageNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLTableSQLiteStorageNode.cxx
@@ -36,6 +36,7 @@
 #include <vtkSQLDatabase.h>
 #include <vtkSQLiteDatabase.h>
 #include <vtkSQLiteQuery.h>
+#include <vtkSmartPointer.h>
 
 #include <vtksys/SystemTools.hxx>
 
@@ -71,7 +72,7 @@ int vtkMRMLTableSQLiteStorageNode::ReadDataInternal(vtkMRMLNode *refNode)
 {
   std::string fullName = this->GetFullNameFromFileName();
 
-  if (fullName == std::string(""))
+  if (fullName.empty())
     {
     vtkErrorMacro("ReadData: File name not specified");
     return 0;
@@ -91,15 +92,17 @@ int vtkMRMLTableSQLiteStorageNode::ReadDataInternal(vtkMRMLNode *refNode)
     }
 
   std::string dbname = std::string("sqlite://") + fullName;
-  vtkSQLiteDatabase *database = vtkSQLiteDatabase::SafeDownCast( vtkSQLiteDatabase::CreateFromURL(dbname.c_str()));
-  if (!database->Open(this->GetPassword(), vtkSQLiteDatabase::USE_EXISTING))
+  vtkSmartPointer<vtkSQLiteDatabase> database = vtkSmartPointer<vtkSQLiteDatabase>::Take(
+                   vtkSQLiteDatabase::SafeDownCast( vtkSQLiteDatabase::CreateFromURL(dbname.c_str())));
+
+  if (!database.GetPointer() || !database->Open(this->GetPassword(), vtkSQLiteDatabase::USE_EXISTING))
     {
     vtkErrorMacro("ReadData: database file '" << fullName << "cannot be openned");
-    database->Delete();
     return 0;
     }
 
-  vtkSQLiteQuery *query = vtkSQLiteQuery::SafeDownCast( database->GetQueryInstance());
+  vtkSmartPointer<vtkSQLiteQuery> query = vtkSmartPointer<vtkSQLiteQuery>::Take(
+                   vtkSQLiteQuery::SafeDownCast( database->GetQueryInstance()));
   std::string queryString("select * from ");
   queryString += std::string(this->TableName);
   query->SetQuery(queryString.c_str());
@@ -115,9 +118,6 @@ int vtkMRMLTableSQLiteStorageNode::ReadDataInternal(vtkMRMLNode *refNode)
 
   vtkDebugMacro("ReadData: successfully read table from file: " << fullName);
 
-  //database->Close();
-  query->Delete();
-  database->Delete();
   return 1;
 }
 
@@ -132,7 +132,7 @@ int vtkMRMLTableSQLiteStorageNode::WriteDataInternal(vtkMRMLNode *refNode)
   std::string fullName = this->GetFullNameFromFileName();
   if (fullName.empty())
     {
-      vtkErrorMacro("WriteData: file name not specified");
+    vtkErrorMacro("WriteData: file name not specified");
     return 0;
     }
 
@@ -142,22 +142,38 @@ int vtkMRMLTableSQLiteStorageNode::WriteDataInternal(vtkMRMLNode *refNode)
     vtkErrorMacro("WriteData: unable to cast input node " << refNode->GetID() << " to a valid table node");
     return 0;
     }
-  if(this->TableName == "")
+  if(!this->TableName || std::string(this->TableName).empty())
     {
     vtkErrorMacro(<<"No table name specified!");
     return 0;
     }
 
   std::string dbname = std::string("sqlite://") + fullName;
+
+  //vtkSmartPointer<vtkSQLiteDatabase> database = vtkSmartPointer<vtkSQLiteDatabase>::Take(vtkSQLiteDatabase::SafeDownCast( vtkSQLiteDatabase::CreateFromURL(dbname.c_str())));
   vtkSQLiteDatabase *database = vtkSQLiteDatabase::SafeDownCast( vtkSQLiteDatabase::CreateFromURL(dbname.c_str()));
-  if (!database->Open(this->GetPassword(), vtkSQLiteDatabase::USE_EXISTING_OR_CREATE))
+
+  if (!database || !database->Open(this->GetPassword(), vtkSQLiteDatabase::USE_EXISTING_OR_CREATE))
     {
     vtkErrorMacro("ReadData: database file '" << fullName << "cannot be openned");
-    database->Delete();
     return 0;
     }
 
   vtkTable *table = tableNode->GetTable();
+  if (!table)
+    {
+    vtkErrorMacro("ReadData: no table to write for the node '" << std::string(tableNode->GetName()));
+    return 0;
+    }
+
+  if (!this->TableName || std::string(this->TableName).empty())
+    {
+    vtkErrorMacro("ReadData: no table name for the node '" << std::string(tableNode->GetName()));
+    return 0;
+    }
+
+  // first try to drop the table
+  this->DropTable(this->TableName, database);
 
   //converting this table to SQLite will require two queries: one to create
   //the table, and another to populate its rows with data.
@@ -250,20 +266,60 @@ int vtkMRMLTableSQLiteStorageNode::WriteDataInternal(vtkMRMLNode *refNode)
   return 1;
 }
 
+int vtkMRMLTableSQLiteStorageNode::DropTable(char *tableName, vtkSQLiteDatabase* database)
+{
+  if(!tableName || std::string(tableName).empty())
+    {
+    std::cerr <<"No table name specified!";
+    return 0;
+    }
+
+  if (!database)
+    {
+    std::cerr << "No databse specified!";
+    return 0;
+    }
+
+  vtkStringArray *tables = database->GetTables();
+  vtkSmartPointer<vtkSQLiteQuery> query = vtkSmartPointer<vtkSQLiteQuery>::Take(
+                       vtkSQLiteQuery::SafeDownCast( database->GetQueryInstance()));
+
+  for (int i=0; i<tables->GetNumberOfValues(); i++)
+    {
+    if (!tables->GetValue(i).compare(tableName))
+      {
+      std::string dropTableQuery = "DROP TABLE ";
+      dropTableQuery += tableName;
+      query->SetQuery(dropTableQuery.c_str());
+      query->Execute();
+      break;
+      }
+    }
+
+  //database->Close();
+  return 1;
+}
+
 //----------------------------------------------------------------------------
 void vtkMRMLTableSQLiteStorageNode::InitializeSupportedReadFileTypes()
 {
   this->SupportedReadFileTypes->InsertNextValue("SQLight database (.db)");
+  this->SupportedReadFileTypes->InsertNextValue("SQLight database (.db3)");
+  this->SupportedReadFileTypes->InsertNextValue("SQLight database (.sqlite)");
+  this->SupportedReadFileTypes->InsertNextValue("SQLight database (.sqlite3)");
 }
 
 //----------------------------------------------------------------------------
 void vtkMRMLTableSQLiteStorageNode::InitializeSupportedWriteFileTypes()
 {
   this->SupportedWriteFileTypes->InsertNextValue("SQLight database (.db)");
+  this->SupportedWriteFileTypes->InsertNextValue("SQLight database (.db3)");
+  this->SupportedWriteFileTypes->InsertNextValue("SQLight database (.sqlite)");
+  this->SupportedWriteFileTypes->InsertNextValue("SQLight database (.sqlite3)");
 }
 
 //----------------------------------------------------------------------------
 const char* vtkMRMLTableSQLiteStorageNode::GetDefaultWriteFileExtension()
 {
-  return "db";
+  return "sqlite3";
 }
--- a/Libs/MRML/Core/vtkMRMLTableSQLiteStorageNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLTableSQLiteStorageNode.cxx
@@ -1,0 +1,269 @@
+/*==============================================================================
+
+  Program: 3D Slicer
+
+  Portions (c) Copyright 2015 Brigham and Women's Hospital (BWH) All Rights Reserved.
+
+  See COPYRIGHT.txt
+  or http://www.slicer.org/copyright/copyright.txt for details.
+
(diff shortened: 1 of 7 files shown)


pieper:
Hi Alex -

I had been thinking this would be tested first as part of an extension, but it seems small enough so maybe putting it straight into the core is safe.  

@jcfr @lassoan @fedorov would you mind having a look and commenting?

-Steve

ayarmarkovich:
Hi Steve,
The functionality is general enough, and seems not large enough for it's own extension. Plus it depends on the new Table MRML and module functionality which is only part of the slicer trunk for now.
Alex

jcfr:
I didn't have to review.

That said, I think it make sense to integrate the sqlite implementation into the core. (as a reference one)

Before doing so we need to make sure the integration is done like it would be done in an extension, meaning we should avoid hardcoding sqlite function calls into table logic.

lassoan:
This is a nice new IO class and I think having it in the Slicer core is useful. I would suggest to improve a few things before integration:

1. Figure out a strategy something for saving data. Currently if an existing file is chosen to save the data into then user gets a warning popup that file will be overwritten and then all the existing records are kept and the entire table contents is appended (so all existing records are duplicated in the table). I would have expected the entire table contents replaced. I can imagine that an append mode can be useful for some cases (especially when modifying the database programmatically) but it should be a mode that is disabled by default.

2. It's not nice that all the tables in the database are loaded when the user opens a database file. The user should be able to select tables(s) and enter password. Please add a widget to the AddData dialog box (similarly to for example volumes can be loaded centered, etc).

I'll add additional smaller comments to the code.

Libs/MRML/Core/vtkMRMLTableSQLiteStorageNode.cxx
```diff
@@ -0,0 +1,269 @@
...
+{
+  return refNode->IsA("vtkMRMLTableNode");
+}
+
+//----------------------------------------------------------------------------
+int vtkMRMLTableSQLiteStorageNode::ReadDataInternal(vtkMRMLNode *refNode)
+{
+  std::string fullName = this->GetFullNameFromFileName();
+
+  if (fullName == std::string(""))
```
lassoan: use fullName.empty()

Libs/MRML/Core/vtkMRMLTableSQLiteStorageNode.cxx
```diff
@@ -0,0 +1,269 @@
...
+    }
+  vtkMRMLTableNode *tableNode = vtkMRMLTableNode::SafeDownCast(refNode);
+  if (tableNode == NULL)
+    {
+    vtkErrorMacro("ReadData: unable to cast input node " << refNode->GetID() << " to a table node");
+    return 0;
+    }
+
+  // Check that the file exists
+  if (vtksys::SystemTools::FileExists(fullName) == false)
```
lassoan: very small thing, but
(!vtksys::...)
would be more compact and just as readable

Libs/MRML/Core/vtkMRMLTableSQLiteStorageNode.cxx
```diff
@@ -0,0 +1,269 @@
...
+    {
+    vtkErrorMacro("ReadData: table file '" << fullName << "' not found.");
+    return 0;
+    }
+
+  std::string dbname = std::string("sqlite://") + fullName;
+  vtkSQLiteDatabase *database = vtkSQLiteDatabase::SafeDownCast( vtkSQLiteDatabase::CreateFromURL(dbname.c_str()));
+  if (!database->Open(this->GetPassword(), vtkSQLiteDatabase::USE_EXISTING))
```
lassoan: add a database==NULL check

Libs/MRML/Core/vtkMRMLTableSQLiteStorageNode.cxx
```diff
@@ -0,0 +1,269 @@
...
+
+  // Check that the file exists
+  if (vtksys::SystemTools::FileExists(fullName) == false)
+    {
+    vtkErrorMacro("ReadData: table file '" << fullName << "' not found.");
+    return 0;
+    }
+
+  std::string dbname = std::string("sqlite://") + fullName;
+  vtkSQLiteDatabase *database = vtkSQLiteDatabase::SafeDownCast( vtkSQLiteDatabase::CreateFromURL(dbname.c_str()));
```
lassoan: Use vtkSmartPointer::Take to avoid littering the code with Delete() statements late, similar to this:

vtkSmartPointer<vtkAbstractTransform> transformCopy=vtkSmartPointer<vtkAbstractTransform>::Take(transform->MakeTransform());

Libs/MRML/Core/vtkMRMLTableSQLiteStorageNode.cxx
```diff
@@ -0,0 +1,269 @@
...
+  if (!database->Open(this->GetPassword(), vtkSQLiteDatabase::USE_EXISTING))
+    {
+    vtkErrorMacro("ReadData: database file '" << fullName << "cannot be openned");
+    database->Delete();
+    return 0;
+    }
+
+  vtkSQLiteQuery *query = vtkSQLiteQuery::SafeDownCast( database->GetQueryInstance());
```
lassoan: use SmartPointer::Take and check for NULL

Libs/MRML/Core/vtkMRMLTableSQLiteStorageNode.cxx
```diff
@@ -0,0 +1,269 @@
...
+{
+  if (this->GetFileName() == NULL)
+    {
+    vtkErrorMacro("WriteData: file name is not set");
+    return 0;
+    }
+  std::string fullName = this->GetFullNameFromFileName();
+  if (fullName.empty())
+    {
+      vtkErrorMacro("WriteData: file name not specified");
```
lassoan: remove extra spaces

Libs/MRML/Core/vtkMRMLTableSQLiteStorageNode.cxx
```diff
@@ -0,0 +1,269 @@
...
+    return 0;
+    }
+
+  vtkMRMLTableNode *tableNode = vtkMRMLTableNode::SafeDownCast(refNode);
+  if (tableNode == NULL)
+    {
+    vtkErrorMacro("WriteData: unable to cast input node " << refNode->GetID() << " to a valid table node");
+    return 0;
+    }
+  if(this->TableName == "")
```
lassoan: This will probably not always work (even if this->TableName points to an empty string, its address will be probably different from the constant one).

TableName is const char, so first we need to check if it's NULL and if not then check if it's empty (by strcmp or by casting it to std::string and calling .empty).

Libs/MRML/Core/vtkMRMLTableSQLiteStorageNode.cxx
```diff
@@ -0,0 +1,269 @@
...
+    {
+    vtkErrorMacro(<<"No table name specified!");
+    return 0;
+    }
+
+  std::string dbname = std::string("sqlite://") + fullName;
+  vtkSQLiteDatabase *database = vtkSQLiteDatabase::SafeDownCast( vtkSQLiteDatabase::CreateFromURL(dbname.c_str()));
+  if (!database->Open(this->GetPassword(), vtkSQLiteDatabase::USE_EXISTING_OR_CREATE))
```
lassoan: same as above, SmartPointer and NULL check needed

Libs/MRML/Core/vtkMRMLTableSQLiteStorageNode.cxx
```diff
@@ -0,0 +1,269 @@
...
+    {
+    vtkErrorMacro("ReadData: database file '" << fullName << "cannot be openned");
+    database->Delete();
+    return 0;
+    }
+
+  vtkTable *table = tableNode->GetTable();
```
lassoan: NULL-pointer check needed

Libs/MRML/Core/vtkMRMLTableSQLiteStorageNode.cxx
```diff
@@ -0,0 +1,269 @@
...
+      }
+    else
+      {
+      createTableQuery += ", ";
+      insertPreamble += ", ";
+      }
+    }
+
+  //perform the create table query
+  vtkSQLiteQuery *query =
```
lassoan: smartpointer

Libs/MRML/Core/vtkMRMLTableSQLiteStorageNode.cxx
```diff
@@ -0,0 +1,269 @@
...
+  database->Delete();
+
+  vtkDebugMacro("WriteData: successfully wrote table to database: " << fullName);
+  return 1;
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLTableSQLiteStorageNode::InitializeSupportedReadFileTypes()
+{
+  this->SupportedReadFileTypes->InsertNextValue("SQLight database (.db)");
```
lassoan: add sqlite and (sqlite2 or sqlite3)

Libs/MRML/Core/vtkMRMLTableSQLiteStorageNode.cxx
```diff
@@ -0,0 +1,269 @@
...
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLTableSQLiteStorageNode::InitializeSupportedWriteFileTypes()
+{
+  this->SupportedWriteFileTypes->InsertNextValue("SQLight database (.db)");
```
lassoan: add sqlite and (sqlite2 or sqlite3)

Modules/Loadable/Tables/qSlicerTablesReader.cxx
```diff
@@ -98,6 +102,7 @@ QStringList qSlicerTablesReader::extensions()const
     << "Table (*.tsv)"
     << "Table (*.csv)"
     << "Table (*.txt)"
+    << "Table (*.db)"
```
lassoan: add sqlite, ...

Modules/Loadable/Tables/Logic/vtkSlicerTablesLogic.cxx
```diff
@@ -68,27 +72,75 @@ ::AddTable(const char* fileName, const char* name)
...
-    {
-    tableNode->SetName(name);
-    }
-  this->GetMRMLScene()->AddNode(tableNode.GetPointer());
+  vtkMRMLTableNode *tableNode;
 
-  // Read
-  int res = tableStorageNode->ReadData(tableNode.GetPointer());
-  if (res == 0) // failed to read
+  if (std::string(fileName).find(".db") != std::string::npos)
```
lassoan: Determining file extension is not trivial, use the method that is added for this purpose: vtkMRMLStorageNode::GetLowercaseExtensionFromFileName

Example:
   // determine file type
  std::string extension = vtkMRMLStorageNode::GetLowercaseExtensionFromFileName(filename);
  if( extension.empty() )
    {
    vtkErrorMacro("ReadData: no file extension specified: " << filename.c_str());
    return false;
    }

  if ( !extension.compare(".nrrd")
      || !extension.compare(".nhdr")
      || !extension.compare(".mha")
      || !extension.compare(".mhd")
      || !extension.compare(".nii") )
    {
...

Modules/Loadable/Tables/Logic/vtkSlicerTablesLogic.cxx
```diff
@@ -68,27 +72,75 @@ ::AddTable(const char* fileName, const char* name)
...
     {
-    vtkErrorMacro("vtkSlicerTablesLogic::AddTable failed: failed to read data from file: "<<fileName);
-    this->GetMRMLScene()->RemoveNode(tableStorageNode.GetPointer());
-    this->GetMRMLScene()->RemoveNode(tableNode.GetPointer());
-    return 0;
+    // SQLite
+    std::string dbname = std::string("sqlite://") + std::string(fileName);
```
lassoan: Add a static method to the storage node that returns a list of tables from a filename to remove repeating very low-level details, such as creating dbname from filename, opening database, etc.

Modules/Loadable/Tables/Logic/vtkSlicerTablesLogic.cxx
```diff
@@ -68,27 +72,75 @@ ::AddTable(const char* fileName, const char* name)
...
+    vtkSQLiteDatabase *database = vtkSQLiteDatabase::SafeDownCast( vtkSQLiteDatabase::CreateFromURL(dbname.c_str()));
+    if (!database->Open(name?name:"", vtkSQLiteDatabase::USE_EXISTING))
+      {
+      vtkErrorMacro("Failed to read tables from " << fileName);
+      database->Delete();
+      return 0;
+      }
+    vtkStringArray *tables = database->GetTables();
```
lassoan: user shouldbe able to specify a regexp (or some other filter) to choose which tables to read (and also specify a password)

Modules/Loadable/Tables/qSlicerTablesReader.cxx
```diff
@@ -114,6 +119,30 @@ bool qSlicerTablesReader::load(const IOProperties& properties)
     name = properties["name"].toString();
     }
   std::string uname = this->mrmlScene()->GetUniqueNameByString(name.toLatin1());
+
+  if (fileName.contains(".db"))
+    {
+   uname = "";
```
lassoan: missing spaces

Modules/Loadable/Tables/qSlicerTablesReader.cxx
```diff
@@ -114,6 +119,30 @@ bool qSlicerTablesReader::load(const IOProperties& properties)
     name = properties["name"].toString();
     }
   std::string uname = this->mrmlScene()->GetUniqueNameByString(name.toLatin1());
+
+  if (fileName.contains(".db"))
```
lassoan: use storeage node's extension method to get extension and compare - see above

Modules/Loadable/Tables/qSlicerTablesReader.cxx
```diff
@@ -114,6 +119,30 @@ bool qSlicerTablesReader::load(const IOProperties& properties)
...
+    {
+   uname = "";
+    std::string dbname = std::string("sqlite://") + fileName.toStdString();
```
lassoan: use toLatin1 for consistency (used most places in Slicer source code instead of toStdString)

Modules/Loadable/Tables/qSlicerTablesReader.cxx
```diff
@@ -114,6 +119,30 @@ bool qSlicerTablesReader::load(const IOProperties& properties)
...
+    vtkSQLiteDatabase *database = vtkSQLiteDatabase::SafeDownCast( vtkSQLiteDatabase::CreateFromURL(dbname.c_str()));
+    if (!database->Open("", vtkSQLiteDatabase::USE_EXISTING))
+      {
+      bool ok;
+      QString text = QInputDialog::getText(0, tr("QInputDialog::getText()"),
+                                           tr("Database Password:"), QLineEdit::Normal,
+                                           "", &ok);
+      if (ok && !text.isEmpty())
+        {
+        uname = text.toStdString();
```
lassoan: use toLatin1 for consistency (used most places in Slicer source code instead of toStdString)

Libs/MRML/Core/vtkMRMLTableSQLiteStorageNode.cxx
lassoan: Due to some reason, if you create a new table, saving as .db is not offered as an option.
Probably because only one storage node provides all the file extensions, and it's the vtkMRMLTableStorageNode.

lassoan:
Comments not for immedate fix, just to be recorded for future enhancements (added to the bugtracker, along with other improvement possibilities):
* Database columns with VARCHAR type that don't contain any values are loaded as a "vtkDoubleArray" - see for example Studies/PerformingPhysician column in the Slicer DICOM databaes. Probably it's VTK's column type detector heuristics. Would be nice to fix.
* Ability to specify absolute path (instead of relative to MRML file location) may be useful
* Renaming the file name to the node name is again very annoying in this use case, too. It's essentially impossible to update a database because the database file name is changed to the table's name (I don't propose this to be fixed immediately, because first it has to be decided how to handle updates - e.g., rewrite entire database table with table in Slicer?)

jcfr:
@ayarmarkovich Thanks for integrating into the trunk [r25034](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25034).

Having a closer look,  there are difference between `r25034` and this PR ...

For example, these two files are different:

 trunk:  https://github.com/Slicer/Slicer/blob/master/Libs/MRML/Core/vtkMRMLTableSQLiteStorageNode.cxx

 PR#484: https://github.com/Slicer/Slicer/pull/484/commits/cda4add6b9acff406bd5cd74f669b299fd76a467#diff-32c9db00492c53d5cd32e5a3ea069321

The headers are not alphabetically ordered and are missing "<", ">" in the trunk version.

Could you have a look ?

Thanks
Jc

jcfr:
@ayarmarkovich @pieper  What do you think of the previous comment ?

ayarmarkovich:
I don't see any changes on my machine. If the file changed, the changes you
mentioned do not affect anything.

Alex

jcfr:
> see any changes on my machine. Thanks for checking.

Great. 

Then what do you think explain the difference between the work you pushed in the topic and what was committed in svn ? 

Then, I will do a diff later one to make sure nothing was missed.

jcfr:
Closing. After double checking, everything is ok.

PR #487, opened 2016-04-29
Add builds of pull requests via CircleCI

thewtex:
Pull requests are built with CircleCI. Slicer itself is built against a Docker image with Slicer's dependencies to reduce build time. If a change to dependency ExternalProject's are made, the build intentionally fails. The resulting package is made available as a CircleCI artifact.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -5,9 +5,14 @@ machine:
     - docker
 
 dependencies:
+  cache_directories:
+        - "~/docker"
+
   override:
     - docker info
+    - if [[ -e ~/docker/image.tar ]]; then docker load -i ~/docker/image.tar; fi
     - docker pull slicer/slicer-build-deps
+    - mkdir -p ~/docker; docker save slicer/slicer-build-deps > ~/docker/image.tar
 
 test:
   override:
--- a/circle.yml
+++ b/circle.yml
@@ -17,4 +17,5 @@ dependencies:
 test:
   override:
     - "git diff --name-only master | grep -q SuperBuild > /dev/null || echo 'Notice: TravisCI does not build changes to Slicer dependencies'"
-    - docker run -e "BUILD_TOOL_FLAGS=-j5" --rm -v ~/Slicer:/usr/src/Slicer slicer/slicer-build-deps
+    - docker run -e "BUILD_TOOL_FLAGS=-j5" --name slicer -v ~/Slicer:/usr/src/Slicer slicer/slicer-build-deps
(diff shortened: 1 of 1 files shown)


jcfr:
Thanks, that is great !

jcfr:
@thewtex I just added CircleCi to the Slicer/Slicer project.
See https://circleci.com/gh/Slicer/Slicer

jcfr:
Integrated in [r25031](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25031) to [r25033](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25033)

jcfr:
screenshot from 2016-04-29 00-45-24

The build was triggered after integrating the commits into master. :8ball:

pieper:
Looks awesome!

fedorov:
Very nice! 👍 

Is it possible to make the circleci dashboard publicly visible?

Did you think about a process when the dependencies docker can be updated when Slicer Superbuild directory content is changed?

thewtex:
Thanks for the integration!

The CircleCI build results should be publicly visible. The resulting Slicer package artifact download can only be downloaded by user's with GitHub push permissions to Slicer/Slicer according to the CircleCI policies:

  https://circleci.com/docs/build-artifacts/

While there is still more work to do in thewtex/SlicerDocker, it may be time to transfer ownership to Slicer/SlicerDocker. To update the dependency Docker image, the revision there can be bumped, the dependency image built locally, and the dependency image pushed to DockerHub.

fedorov:
> The CircleCI build results should be publicly visible. 

Ah, right - I missed the url, thanks!

https://circleci.com/gh/Slicer/Slicer

circle.yml
```diff
@@ -0,0 +1,15 @@
...
+    - docker
+
+dependencies:
+  override:
+    - docker info
+    - docker pull slicer/slicer-build-deps
+
+test:
+  override:
+    - "git diff --name-only master | grep -q SuperBuild > /dev/null || echo 'Notice: TravisCI does not build changes to Slicer dependencies'"
```
msmolens: Is this line right? It looks like `grep` exits with a zero status code on a match, so the warning won't be printed. Also, does this line need to explicitly call `exit` with a non-zero status code to cause the build to fail? And--nitpick--it should say `CircleCI` instead of `TravisCI`.

thewtex: Yes, it was supposed to have a non-zero status code and cause the build to fail -- please check this follow-up.

  https://github.com/Slicer/Slicer/pull/488

I had also tried TravisCI and missed the name update. TravisCI would time out at its fixed 40 minute limit.

PR #489, opened 2016-05-02
PERF: use hashmap lookup instead of linear search.

ihnorton:
@jcfr @lassoan this provides significant performance improvement in large scene loading task
(1200 models), from >20 minutes on master to 5:30.

The comment regarding `find` was incorrect: pointer comparison issue
only applies if a non-const pointer is used *as the key*.

(cc @ljod)

### DIFF
--- a/Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.cxx
+++ b/Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.cxx
@@ -1849,16 +1849,13 @@ vtkProp3D * vtkMRMLModelDisplayableManager::GetActorByID(const char *id)
     return (0);
     }
 
-  std::map<std::string, vtkProp3D *>::iterator iter;
-  // search for matching string (can't use find, since it would look for
-  // matching pointer not matching content)
-  for(iter=this->Internal->DisplayedActors.begin(); iter != this->Internal->DisplayedActors.end(); iter++)
+  std::map<std::string, vtkProp3D *>::iterator iter =
+    this->Internal->DisplayedActors.find(std::string(id));
+  if (iter != this->Internal->DisplayedActors.end())
     {
-    if ( iter->first.c_str() && !strcmp( iter->first.c_str(), id ) )
-      {
-      return (iter->second);
-      }
+    return iter->second;
     }
+
   return (0);
 }
 


jcfr:
:+1: Good catch. Makes complete sense. 

Will integrate as soon as CI completes.

Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.cxx
```diff
@@ -1850,15 +1850,10 @@ vtkProp3D * vtkMRMLModelDisplayableManager::GetActorByID(const char *id)
...
-  // search for matching string (can't use find, since it would look for
-  // matching pointer not matching content)
-  for(iter=this->Internal->DisplayedActors.begin(); iter != this->Internal->DisplayedActors.end(); iter++)
-    {
-    if ( iter->first.c_str() && !strcmp( iter->first.c_str(), id ) )
-      {
-      return (iter->second);
-      }
-    }
+  iter = this->Internal->DisplayedActors.find(std::string(id));
```
jcfr: Nitpick, add bracket:

```
+  if (iter != this->Internal->DisplayedActors.end())
+  {
+    return iter->second;
+  }
``

Comment (ihnorton) [inline Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.cxx]: done

**Comment by lassoan** [comment]
Nice. find is used many places, so I don't know why it was not used here.

Very small thing: it would be nicer to merge these two lines
std::map<std::string, vtkProp3D *>::iterator iter;
iter = this->Internal->DisplayedActors.find(std::string(id));
to
std::map<std::string, vtkProp3D *>::iterator iter = this->Internal->DisplayedActors.find(std::string(id));

**Comment by jcfr** [comment]
@lassoan I will do the tweak you suggested at integration time.

**Comment by ihnorton** [comment]
@lassoan that construct is not supported by clang (at least).

**Comment by lassoan** [comment]
Interesting. A slightly different syntax is used at many places in the file:

std::map<std::string, vtkProp3D *>::iterator iter = this->Internal->DisplayedActors.find(dnode->GetID());

https://github.com/ihnorton/Slicer/blob/perf1/Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.cxx#L1201

**Comment by ihnorton** [comment]
That line still specifies the types for the map...

Maybe I misunderstood: if you meant
```
std::map<std::string, vtkProp3D *>::iterator iter = this->Internal->DisplayedActors.find(dnode->GetID());
```
-- sure, that looks good. The only reason I didn't do that is line-length (could be split on two lines).

lassoan:
I just usually prefer to avoid leaving any variables uninitialized, as having uninitilized variables may cause different behavior in debug/release builds, may cause problems when code is added or moved around and you accidentally use the uninitialized value, etc.

You can still break it into two lines if you prefer shorter lines:
std::map<std::string, vtkProp3D *>::iterator iter =
  this->Internal->DisplayedActors.find(std::string(id));

ihnorton:
Ok, understood. Updated!

(sorry, I was reading your comment too literally as `std::map::iterator` with no types specified)

jcfr:
Merci :+1: 

Integrated in [r25043](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25043)

ljod:
Fabulous thank you Isaiah, JC, and Andras. I look forward to the speed improvements for atlas viewing.

PR #490, opened 2016-05-05
More large scene optimizations

ihnorton:
Follow-up to https://github.com/Slicer/Slicer/pull/489
For same large scene (1200 models):
- Reduces scene-loading time to ~4 minutes.
- Reduces delay when switching to Models module from 3 min to ~5s.
- Reduces delay when enabling "Include Fibers" option from 3 min to <5s.
- Reduces delay when changing visibility of all models from "very slow" to "usable" (few seconds)

All the tests pass (locally) except for: `py_BRAINSFitRigidRegistrationCrashIssue4139`, `vtkMRMLNonlinearTransformNodeTest1`, and `qSlicerSslTest`.

The SSL one is known on buildbots, and I believe the other two are ITK linking issues specific to this build toolchain ([mailing list](http://slicer-devel.65872.n3.nabble.com/Empty-deformable-transforms-with-Mac-build-td4036248.html)).

cc @ljod

### DIFF
--- a/Libs/MRML/Core/Testing/vol_and_cube.mrml
+++ b/Libs/MRML/Core/Testing/vol_and_cube.mrml
@@ -1,12 +1,15 @@
 <MRML>
 
 <Selection
-  id="vtkMRMLSelectionNodeSingleton" name="vtkMRMLSelectionNode1" hideFromEditors="true"
-  activeVolumeID="vtkMRMLScalarVolumeNode1" activeLayoutID="vtkMRMLLayoutNode1">
+  id="vtkMRMLSelectionNodeSingleton" name="Selection" hideFromEditors="true"
+  activeVolumeID="vtkMRMLScalarVolumeNode1" activeLayoutID="vtkMRMLLayoutNode1"
+  singletonTag="Singleton"
+  >
 </Selection>
 <Interaction
-  id="vtkMRMLInteractionNodeSingleton" name="vtkMRMLInteractionNode1" hideFromEditors="true"
-  currentInteractionMode="ViewTransform" lastInteractionMode="ViewTransform">
+  id="vtkMRMLInteractionNodeSingleton" name="Interaction" hideFromEditors="true"
+  currentInteractionMode="ViewTransform" lastInteractionMode="ViewTransform"
+  singletonTag="Singleton">
 </Interaction>
 
 <Slice
--- a/Libs/MRML/Widgets/qMRMLSceneModel.cxx
+++ b/Libs/MRML/Widgets/qMRMLSceneModel.cxx
@@ -1153,6 +1153,14 @@ void qMRMLSceneModel::onMRMLSceneEvent(vtkObject* vtk_obj, unsigned long event,
   vtkMRMLNode* node = reinterpret_cast<vtkMRMLNode*>(call_data);
   Q_ASSERT(scene);
   Q_ASSERT(sceneModel);
+
+  // we don't care about importing events. we will
+  // repopulate entire scenemodel at EndImportEvent.
+  if (event != vtkMRMLScene::EndImportEvent && scene->IsImporting())
+    {
+    return;
+    }
+
   switch(event)
     {
     case vtkMRMLScene::NodeAboutToBeAddedEvent:
@@ -1499,7 +1507,6 @@ bool qMRMLSceneModel::isANode(const QStandardItem * item)const
 void qMRMLSceneModel::onMRMLSceneAboutToBeImported(vtkMRMLScene* scene)
 {
   Q_UNUSED(scene);
-  //this->beginResetModel();
 }
 
 //------------------------------------------------------------------------------
@@ -1507,11 +1514,8 @@ void qMRMLSceneModel::onMRMLSceneImported(vtkMRMLScene* scene)
 {
   Q_D(qMRMLSceneModel);
   Q_UNUSED(scene);
-  if (d->LazyUpdate)
-    {
-    this->updateScene();
-    }
-  //this->endResetModel();
+  this->updateScene();
+  emit sceneUpdated();
 }
 
 //------------------------------------------------------------------------------
--- a/Libs/MRML/Widgets/qMRMLSceneModel.h
+++ b/Libs/MRML/Widgets/qMRMLSceneModel.h
@@ -124,6 +124,8 @@ public:
   Q_INVOKABLE virtual void setMRMLScene(vtkMRMLScene* scene);
   Q_INVOKABLE vtkMRMLScene* mrmlScene()const;
 
+  virtual void updateScene();
+
   /// 0 until a valid scene is set
   QStandardItem* mrmlSceneItem()const;
 
@@ -266,7 +268,6 @@ protected:
 
   qMRMLSceneModel(qMRMLSceneModelPrivate* pimpl, QObject *parent=0);
 
-  virtual void updateScene();
   virtual void populateScene();
   virtual QStandardItem* insertNode(vtkMRMLNode* node);
   virtual QStandardItem* insertNode(vtkMRMLNode* node, QStandardItem* parent, int row = -1);
--- a/Modules/Loadable/Models/qSlicerModelsModuleWidget.cxx
+++ b/Modules/Loadable/Models/qSlicerModelsModuleWidget.cxx
@@ -547,14 +547,9 @@ void qSlicerModelsModuleWidget::updateWidgetFromSelectionNode()
 
   if (include)
     {
-    // force update mrml widgets
-    std::vector<vtkMRMLNode *> nodes;
-    vtkMRMLScene *scene = this->mrmlScene();
-    scene->GetNodesByClass(displayNodeClass.c_str(), nodes);
-    for (unsigned int i = 0; i < nodes.size(); i++)
-      {
-      nodes[i]->InvokeEvent(vtkCommand::ModifiedEvent);
-      }
+    // invoke batch scenemodel update. triggering modified on each model
+    // node is extremely expensive with larger scenes.
+    d->ModelHierarchyTreeView->sceneModel()->updateScene();
     }
   d->ModelDisplayWidget->setMRMLModelOrHierarchyNode(d->ModelDisplayWidget->mrmlDisplayableNode());
 }


pieper:
I'm not very familiar with the scene model but these look like very reasonable changes to me.  And if they work for you on the large datasets I'm confident they aren't likely to break anything else.

Nice work!

jcfr:
Thanks for the perf improvements :+1: 

I will review more carefully tonight.

@naucoin  Could the caching the of the selection node cause issue with scene view ? 

@ihnorton What do you think of associating the MRMLApplicationLogic to the scene model, and have the code calling `this->selectionNode()` .. that way we wouldn't have to keep a pointer that could potentially be invalidated.

jcfr:
>  STYLE: clarify cast

Integrated as [r25046](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25046)

jcfr:
>  	PERF: track and reuse index during populateScene

Integrated as [r25048](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25048)
:+1:  Roughly a 8% decrease in load time of "NAC Brain Atlas 2015" including ~1600 nodes. I added a commits to include comment & fix warning `-Woverloaded-virtual` warnings like the ones reported below.



Note that since  [r25047](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25047), 
the loading time of any file is now displayed.

```
In file included from /home/jcfr/Projects/Slicer-Release/Slicer-build/Modules/Remote/OpenIGTLinkIF/GUI/../../../../../OpenIGTLinkIF/GUI/qMRMLIGTLIOModel.h:24:0,
                 from /home/jcfr/Projects/Slicer-Release/Slicer-build/Modules/Remote/OpenIGTLinkIF/GUI/moc_qMRMLIGTLIOModel.cxx:9:
/home/jcfr/Projects/Slicer/Libs/MRML/Widgets/qMRMLSceneModel.h:272:26: warning: ‘virtual QStandardItem* qMRMLSceneModel::insertNode(vtkMRMLNode*, int)’ was hidden [-Woverloaded-virtual]
   virtual QStandardItem* insertNode(vtkMRMLNode* node, int index);
                          ^
In file included from /home/jcfr/Projects/Slicer-Release/Slicer-build/Modules/Remote/OpenIGTLinkIF/GUI/moc_qMRMLIGTLIOModel.cxx:9:0:
/home/jcfr/Projects/Slicer-Release/Slicer-build/Modules/Remote/OpenIGTLinkIF/GUI/../../../../../OpenIGTLinkIF/GUI/qMRMLIGTLIOModel.h:72:26: warning:   by ‘qMRMLIGTLIOModel::insertNode’ [-Woverloaded-virtual]
   virtual QStandardItem* insertNode(vtkMRMLNode* node);
                          ^

                 from /home/jcfr/Projects/Slicer-Release/OpenIGTLinkIF/GUI/qMRMLIGTLIOModel.cxx:26:
/home/jcfr/Projects/Slicer/Libs/MRML/Widgets/qMRMLSceneModel.h:272:26: warning: ‘virtual QStandardItem* qMRMLSceneModel::insertNode(vtkMRMLNode*, int)’ was hidden [-Woverloaded-virtual]
   virtual QStandardItem* insertNode(vtkMRMLNode* node, int index);
                          ^
In file included from /home/jcfr/Projects/Slicer-Release/OpenIGTLinkIF/GUI/qMRMLIGTLIOModel.cxx:26:0:
/home/jcfr/Projects/Slicer-Release/OpenIGTLinkIF/GUI/qMRMLIGTLIOModel.h:72:26: warning:   by ‘qMRMLIGTLIOModel::insertNode’ [-Woverloaded-virtual]
   virtual QStandardItem* insertNode(vtkMRMLNode* node);
                          ^
[ 77%] Built target vtkSlicerTractographyInteractiveSeedingModuleLogicPythonD
[ 77%] Linking CXX executable ../../../../bin/qSlicerTractographyInteractiveSeedingModuleCxxTests
In file included from /home/jcfr/Projects/Slicer-Release/OpenIGTLinkIF/GUI/qMRMLIGTLIOTreeView.cxx:32:0:
/home/jcfr/Projects/Slicer/Libs/MRML/Widgets/qMRMLSceneModel.h:272:26: warning: ‘virtual QStandardItem* qMRMLSceneModel::insertNode(vtkMRMLNode*, int)’ was hidden [-Woverloaded-virtual]
   virtual QStandardItem* insertNode(vtkMRMLNode* node, int index);
                          ^
In file included from /home/jcfr/Projects/Slicer-Release/OpenIGTLinkIF/GUI/qMRMLIGTLIOTreeView.cxx:36:0:
/home/jcfr/Projects/Slicer-Release/OpenIGTLinkIF/GUI/qMRMLIGTLIOModel.h:72:26: warning:   by ‘qMRMLIGTLIOModel::insertNode’ [-Woverloaded-virtual]
   virtual QStandardItem* insertNode(vtkMRMLNode* node);
                          ^
```

ihnorton:
> What do you think of associating the MRMLApplicationLogic to the scene model, and have the code calling this->selectionNode()

Makes sense. It looks like that is what SubjectHierarchy does. I'll make the change later today and update this PR.

jcfr:
> ll make the change later today

Great. Thanks

jcfr:
> re: PERF: factor out constant display node lookup.

Integrated in [r25050](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25050)

Note  that I skipped the part about `qMRMLSceneHierarchyModel.cxx`, I didn't see how it was reducing the number of lookup. Existing function `qMRMLSceneHierarchyModel::parentNode` is lightweight and allow derived class to overwrite the behaviour.

jcfr:
> re:  	COMP: fix vtkObserverManagerTest1 fail in debug

Integrated in [r25051](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25051)

jcfr:
@ihnorton Will review and integrate the remaining later.

ihnorton:
> Note that I skipped the part about qMRMLSceneHierarchyModel.cxx, I didn't see how it was reducing the number of lookup. Existing function qMRMLSceneHierarchyModel::parentNode is lightweight and allow derived class to overwrite the behaviour.

IMHO that is the more important part, it avoids the need to call `vtkMRMLHierarchyNode::GetAssociatedHierarchyNode` twice, which was showing up as a hot spot (I don't remember what %, but it was not trivial).

jcfr:
re: `vtkMRMLHierarchyNode::GetAssociatedHierarchyNode` twice

Got it. Will integrate and describe this in the commit msg.

jcfr:
> re: vtkMRMLHierarchyNode::GetAssociatedHierarchyNode twice

@ihnorton I came up with a simpler version that avoid the lookup and is still calling the virtual method `parentNode`.
See https://github.com/jcfr/Slicer/commit/839297805f461d39cc6a6d08f486226ab241a9b4 

Let me know what you think

ihnorton:
> See jcfr@8392978

I'm not sure if your change means the same thing, see comment inline.

ihnorton:
> What do you think of associating the MRMLApplicationLogic to the scene model, and have the code calling this->selectionNode()

After looking at this for a while, I disagree. The scenemodel is set up by qMRMLTreeView, which has no association with the logic. It seems much simpler to use the initial proposed solution -- which is the same as used by the DisplayableManager.

(I tried the `qSlicerCoreApplication::application()->applicationLogic()->GetSelectionNode();` version used by SubjectHierarchy, but that breaks a lot of tests that don't do a full app setup)

Since all the code uses an implicit contract that there is only one  `vtkMRMLSelectionNode*` per scene, it seems like we could just add a `vtkMRMLSelectionNode*` as a member of the MRMLScene and delete a lot of this redundant code.

lassoan:
> we could just add a  vtkMRMLSelectionNode*  as a member of the MRMLScene

There is no need for that. vtkMRMLSelectionNode is a singleton node, so if it is in the scene there is always only one and it can be retrieved very quickly using

    scene->GetNodeByID("vtkMRMLSelectionNodeSingleton")

 It's a single map lookup, so in most cases caching is not worth the trouble.

jcfr:
> re: vtkMRMLHierarchyNode::GetAssociatedHierarchyNode twice

@ihnorton  Thanks for the explanation. Make complete sense. I integrated your change.

Integrated in [r25053](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25053)

jcfr:
> re: vtkMRMLSelectionNode

@ihnorton @lassoan Make complete sense. Being singleton, it make sense to do a direct look up. I will integrate the changes accordingly.

jcfr:
> re PERF: cache SelectionNode, avoid expensive lookup
> re PERF: use cached SelectionNode for few more %

Following the suggestion of Andras, here is a commit updating all places where the selection node was retrieved using `GetNodesByClass`. It now uses `GetNodeByID("vtkMRMLSelectionNodeSingleton")`. See https://github.com/jcfr/Slicer/commit/c5c60c93ffff27f772a92aa1080c85ca7f059b50

Cc: @ihnorton @lassoan

jcfr:
Integrated in [r25054](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25054)

ihnorton:
Thanks for the merges; PR updated to remove those commits.

The changes using the `SelectionNodeSingleton` are great (net code deletion 👍 ).

jcfr:
Thanks @ihnorton 

Since I am travelling until Monday of next week, I suggest we hold on integrating the last two commits.

It would also be great to have a look at the failing tests `qMRMLSliceWidgetTest1_vol_and_cube.mrml` and `qMRMLVolumeThresholdWidgetTest2_vol_and_cube.mrml`. Looks like warning messages are now displayed.

ihnorton:
I pushed a commit that fixes those tests locally. CircleCI still says failure, but I don't see any test results in the command line logs on CircleCI, only compilation results. The only visible error is this:

```
Notice: CircleCI does not build changes to Slicer dependencies
```

jcfr:
Thanks. Since ITK has been updated, we need to publish a new docker image.

ihnorton:
Is CircleCI supposed to show test results?

jcfr:
Not yet. This will be addressed later.
Jc

thewtex:
An updated version of the Slicer Docker image, https://github.com/thewtex/SlicerDocker/commit/23960c7d368b8d2409be165de474e30ea699157c has been pushed to DockerHub.

jcfr:
Thanks Matt.

ihnorton:
The remaining test failure here is fixed by #517. Rebased, but the CI build died at the end for some unspecified reason.

ihnorton:
@jcfr if you have a chance to look at this again, the last two commits should be low-risk. The first one (a06ee38) is somewhat related to #398 and so possibly risky. Maybe it makes sense to discard in favor of the lazy loading approach when that is resurrected.

lassoan:
The changes look OK to me.

By the way, what is in your "large scene"? Is it a large model hierarchy? We could also speed up things by replacing many simple nodes by one or a few more complex node. For example, a single Segmentation node can contain a large number of segments (each segment may be displayed in 2D and 3D the same way as a model). Segmentation may need some optimization to manage many nodes quickly (we have never really checked performance bottlenecks with hundreds of segments), but that optimization is much simpler and has a potential of achieving much better performance improvement.

fedorov:
> we have never really checked performance bottlenecks with hundreds of segments

you might want to test with SPL atlases https://www.spl.harvard.edu/pages/Software

ihnorton:
> By the way, what is in your "large scene"? Is it a large model hierarchy?

1200 models in scene. (For comparison, loading those in Python+VTK takes 3 or 4 seconds)

> We could also speed up things by replacing many simple nodes by one or a few more complex node.

Yes, we've talked about doing some of that. But these MRMLScene algorithmic complexity issues, especially because of `GetNodesByClass`, come up everywhere.

> you might want to test with SPL atlases

On Slicer 4.5 it takes ~5 seconds to change visibility of all of the models in the brain atlas scene (XTK atlas viewer can do it ~instantly: http://www.na-mic.org/Bug/view.php?id=2810). On current master it's down to maybe 2 seconds, which is a big improvement, although still hangs the whole window for the duration.

lassoan:
It's great that you are improving scene performance because it improves the application's speed in many use cases. But if you want to improve performance of managing thousands of models and you have some more time to invest then it may be better to address two specific bottlenecks:

1. Scene model: A complete scene model is created in many MRML widgets (e.g., in each MRML node selector and node tree), with lots of observers on many nodes. Having a single scene model that contains all necessary information about all nodes in the scene (and a couple of filters&proxies) used by all widgets would probably make things scalable much better.

2. Model hierarchies: Slicer doesn't store low-level each low-level item (such as a fiducial point, a label, or segment) in a separate node for good reasons - mainly for performance for simplicity for module developers. For example, at some point fiducials were attempted to be stored as "one node per fiducial" but it was complex to use and it was terribly slow. Changing fiducial management to have multiple markups in a single node solved the problems. I think model nodes should work the same way: they should be able to manage many meshes internally (probably not just in a flat list but organized using tags or parent/children relationships), we should not require a separate node for each.

By investing a couple of weeks into either 1 and/or 2, you might achieve 10-100x performance improvement in managing large number of models (vs an optimistic 2-10x performance improvement by further general optimization of scene management).

pieper:
+1 for optimizing the scene model.  It would help speed up a lot of
operations.

cpinter:
I also second the idea of a common scene model for the widgets.

One thing that occurs me worth considering is that tree models have type-specific logic in the model itself: parenting (parentNode, canBeParent, etc.) and order (nodeIndex) related. As we have relatively few of those (less than ten I'd say) in Slicer compared to simple list-type scene models (hundreds of node comboboxes), most of the performance improvement could be achieved by having a common scene model for the comboboxes, and - as Andras suggests - using different filter-proxy models for each type that sit on the same scene model.

ihnorton:
@lassoan thanks for the advice. (I've noticed and wondered-about number 1 in particular when looking at profile traces).

jcfr:
@ihnorton  If you would like to integrate some of the remaining tweaks. Would be great if you could submit an new PR. In the mean time, I will close this one. 

For future reference, discussion has been moved to this topic on the forum: https://discourse.slicer.org/t/improving-performance-to-manage-thousands-of-models/813

PR #492, opened 2016-05-11
BUG: Enable use of labelValue for output label map

naucoin:
A bug was introduced in svn 22211 [1] when the labelValue parameter in the ModelToLabelMap CLI was exposed to the interface.
The erosion and dilation filters used in the code are binary ones and by default use the maximum of the data type of the label map, in this case hard coded to 255 for unsigned char. In order to do the full pipeline with the user selected label value, the code would need to be updated to use:
erodeFilter->SetErodeValue(labelValue);
dilateFilter->SetDilateValue(labelValue);
A quicker fix is to adjust the output label map in the last step to set the desired label value, implemented in this patch.

Investigating this issue, the hard coded kernel size was set into a variable, in anticipation of exposing it via the XML (quick tests showed no visible difference).

Added a test and output using the label value 128. When run from the GUI, the model and volume were offset, so a transformed model was generated that overlaps the input volume and was used to generate the new baseline comparison volume.
Removed unused baseline image.

[1] http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=22211

Issue #3233


pieper:
Makes sense to me.  @fedorov or @naucoin can you ask Christian to test this?  (I don't see his gitub name popping up or I'd include him directly)

naucoin:
@pieper I emailed him the link to this pull request when I made it. I also tested this with lots of different labels before commit (I was trying a few different ways to make the fix), and included a new test.

che85:
I am going to test it tomorrow. 
@naucoin Thanks for the quick fix 👍

naucoin:
@che85 Thanks for testing it! Looks like the CircleCI testing just covers the configure and build stages, it doesn't run the tests.

che85:
@naucoin Actually I wanted to test it today, but todays nightly build is from 05/08/2016

naucoin:
@che85 the change didn't get merged yet so it's not in the nightly build pipeline yet - the last change was made on May 8th, so the nightly is up to date with the repository.
@pieper - the patch built via CirlceCI, can I merge it?

pieper:
Yes, merge.  Thanks for the fix!

naucoin:
Done:
http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25062

naucoin:
@che85 The patch will be in the next nightly build.

PR #493, opened 2016-05-11
ENH: Addition of ctk_cli

fbudin69500:
ctk_cli [1] is a command line parsing python package that parses
arguments the same way SlicerExecutionModel does.
Using ctk_cli allows in Python scripts allows to have the same
user interface in Python or in C++.

[1] https://github.com/commontk/ctk-cli


pieper:
Sounds cool - can you add a short example?

fbudin69500:
@pieper Sure. Do you have any preference where the example should go?

fbudin69500:
@pieper : I added an example in the extension examples. Let me know if that seems reasonable.

pieper:
Yes, that makes sense - thanks!

jcfr:
Thanks for the patch.

Nitpick:  In 80a5a9b, instead of modifying the files in Extensions/Testing. Could you:
* Add a type of modules in https://github.com/Slicer/Slicer/tree/master/Utilities/Templates/Modules
* Update the module [SlicerGenerateExtensionTemplates](https://github.com/Slicer/Slicer/blob/90211eb32fd8f6c2cf5d6c1104760216869e417e/CMake/SlicerExtensionTemplatesGenerator.cmake#L84)
* Build the target `SlicerGenerateExtensionTemplates`

Thanks
Jc

jcfr:
Closing. Superseded by #497

PR #494, opened 2016-05-13
Threed view jump slice

lassoan:
Added "jump to slice" by shift+mouse move in 3D view the same way as it has been already available in slice views.

Also removed remnants of some old picking mechanism (unused, dysfunctional code).


pieper:
Love it!  Thanks!

naucoin:
@lassoan regarding removing the old picking mechanisms: I thought the code looked familiar but turns out that it was a change by Alex Yarmarkovich at svn 13978:
http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=13978
The OnExpose method makes me think that it was added to try and resolve models not showing up when scene views were changed, with your patch does that still work the same?

lassoan:
I've committed the slice jump feature (r25063). I skip the cleanup.

PR #496, opened 2016-05-16
BUG: Attribute does not exist if _haveGit False

fbudin69500:
If Python is not built with SSL support, the variable _haveGit
is set to False. In that case, some command line arguments are
not defined. These arguments should be tested only if _haveGit
is True.


vovythevov:
This is (closely) related to this issue: http://www.na-mic.org/Bug/view.php?id=4182

jcfr:
Closing. Superseded by https://github.com/Slicer/Slicer/pull/763

PR #497, opened 2016-05-16
Add ctk cli2

fbudin69500:
@jcfr This is the new branch I created to address your comments from  #493


jcfr:
Thanks for the PR.

I integrated the changes related to `BUG: Template argument name was modified in template declaration … 	`

I will post comment regarding the PythonCLI template

Utilities/Templates/Modules/PythonCLI/PythonCLIModuleTemplate.py
```diff
@@ -0,0 +1,36 @@
+import itk
```
jcfr: See https://pypi.python.org/pypi/flake8-import-order

```
import logging
import sys

from ctk_cli import CLIArgumentParser

import itk
```

Useful resrouces: https://pypi.python.org/pypi/flake8-import-order and

jcfr: Additionally, would it be possible to use `SimpleITK`, wrapped ITK is not yet enabled by default. 

Thanks for your help

fedorov:
@jcfr what is the status of this PR?

jcfr:
After updating the example so that it does not fail when wrapped itk is not available, it should be good.

fedorov:
Thanks JC!

@mehrtash perhaps it makes sense to build this branch and use it to wrap the deep learning models, as we discussed today? This way you can get started, and at the same time provide feedback to this PR. @jcfr perhaps if you are interested, we could discuss this project at a project week planning call? I will be traveling next Tue, but perhaps we could do it after that. Or should we arrange a separate meeting?

jcfr:
This branch will be integrated as soon as we complete the SlicerITK extension.

Cc: @fbudin69500 @thewtex

jcfr:
Update `PyCLIModule4Test.py` introduced in https://github.com/Slicer/Slicer/pull/894/files to use ctk-cli

PR #498, opened 2016-05-17
Upgrade to scipy 0.14.1

fbudin69500:


fbudin69500:
@jcfr I have removed the special case of Windows from your original branch, and updated the version of scipy (as well as added the dependency to cython and CLAPACK).

jcfr:
Thanks for working on this :+1: 

The reason why we originally removed the CLAPACK external project is that it couldn't be used with NUMPY. See [r20539](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=20539) for the details.

I would prefer we use the actively maintained OpenBLAS project.

fbudin69500:
@jcfr scipy and numpy tests are currently failing.

jcfr:
Renamed PR since numpy version has already been updated to the latested

jcfr:
Closing. An updated PR integrating the latest scipy and an updated numpy will be pushed in the next few days.

PR #499, opened 2016-05-17
ENH : factorize SliceController + new method for displayStringOrientation

Punzo:
@pieper @jcfr 
as discussed this morning for the first point.

The mapping and the methods are static so an estension can do this in its setup method

```
  vtkMRMLSliceNode::SetAxialOrientationDisplayString("XZ");
  vtkMRMLSliceNode::SetSagittalOrientationDisplayString("ZY");
  vtkMRMLSliceNode::SetCoronalOrientationDisplayString("XY");
```

and set up the mapping for all the SliceNodes. 

To get an idea in which situations  this mapping will be useful:

https://github.com/Punzo/SlicerAstro/blob/master/AstroDataProbe/AstroDataProbe.py#L75

https://github.com/Punzo/SlicerAstro/blob/master/AstroVolume/Widgets/qMRMLSliceAstroControllerWidget.cxx#L111

### DIFF
--- a/Libs/MRML/Core/vtkMRMLSliceNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLSliceNode.cxx
@@ -361,6 +361,45 @@ void vtkMRMLSliceNode::SetOrientation(const char* orientation)
 }
 
 //----------------------------------------------------------------------------
+void vtkMRMLSliceNode::SetAxialOrientationDisplayString(const char *displayOrientation)
+{
+  DisplayOrientationStringsMap["Axial"] = displayOrientation;
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLSliceNode::SetSagittalOrientationDisplayString(const char *displayOrientation)
+{
+  DisplayOrientationStringsMap["Sagittal"] = displayOrientation;
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLSliceNode::SetCoronalOrientationDisplayString(const char *displayOrientation)
+{
+  DisplayOrientationStringsMap["Coronal"] = displayOrientation;
+}
+
+//----------------------------------------------------------------------------
+vtkMRMLSliceNode::displayOrientationStringsMap vtkMRMLSliceNode::createdisplayOrientationStringsMap()
+{
+    displayOrientationStringsMap map;
+    map["Axial"] = "Axial";
+    map["Coronal"] = "Coronal";
+    map["Sagittal"] = "Sagittal";
+    return map;
+}
+
+//----------------------------------------------------------------------------
+vtkMRMLSliceNode::displayOrientationStringsMap vtkMRMLSliceNode::DisplayOrientationStringsMap =
+    vtkMRMLSliceNode::createdisplayOrientationStringsMap();
+
+//----------------------------------------------------------------------------
+std::string vtkMRMLSliceNode::GetOrientationDisplayString()
+{
+  std::string orientation = this->GetOrientationString();
+  return DisplayOrientationStringsMap[orientation];
+}
+
+//----------------------------------------------------------------------------
 void vtkMRMLSliceNode::SetOrientationToReformat()
 {
     // Don't need to do anything.  Leave the matrices where they were
--- a/Libs/MRML/Core/vtkMRMLSliceNode.h
+++ b/Libs/MRML/Core/vtkMRMLSliceNode.h
@@ -30,6 +30,9 @@ class vtkMatrix4x4;
 /// \li FieldOfView tells the size of  slice plane
 class VTK_MRML_EXPORT vtkMRMLSliceNode : public vtkMRMLAbstractViewNode
 {
+
+  typedef std::map<std::string, std::string> displayOrientationStringsMap ;
+
   public:
   static vtkMRMLSliceNode *New();
   vtkTypeMacro(vtkMRMLSliceNode,vtkMRMLAbstractViewNode);
@@ -132,6 +135,29 @@ class VTK_MRML_EXPORT vtkMRMLSliceNode : public vtkMRMLAbstractViewNode
   /// SetOrientationToSagittal(), SetOrientationToCoronal() or
   /// SetOrientationToReformat() depending on the value of the string
   void SetOrientation(const char* orientation);
+
+  ///
+  /// map the Axial plane string with a custom displayOrientation string
+  static void SetAxialOrientationDisplayString(const char* displayOrientation);
+
+  ///
+  /// map the Sagittal plane string with a custom displayOrientation string
+  static void SetSagittalOrientationDisplayString(const char* displayOrientation);
+
+  ///
+  /// map the Coronal plane string with a custom displayOrientation string
+  static void SetCoronalOrientationDisplayString(const char* displayOrientation);
+
+  ///
+  /// A description of the current orientation,
+  /// the method return customized names
+  /// for the planes Axial, Sagittal and Coronal.
+  /// The mapping is set using the static map:
+  /// displayOrientationStringsMap.
+  /// SetAxialOrientationDisplayString
+  /// SetSagittalOrientationDisplayString
+  /// SetCoronalOrientationDisplayString
+  std::string GetOrientationDisplayString();
 
   /// Description
   /// A description of the current orientation
@@ -457,6 +483,9 @@ protected:
   int IsUpdatingMatrices;
 
   std::vector< std::string > ThreeDViewIDs;
+
+  static displayOrientationStringsMap createdisplayOrientationStringsMap();
+  static displayOrientationStringsMap DisplayOrientationStringsMap;
 };
 
 //----------------------------------------------------------------------------
--- a/Libs/MRML/Widgets/qMRMLSliceControllerWidget.h
+++ b/Libs/MRML/Widgets/qMRMLSliceControllerWidget.h
@@ -158,7 +158,7 @@ public slots:
 
   /// Set slice orientation.
   /// \note Orientation could be either "Axial, "Sagittal", "Coronal" or "Reformat".
-  void setSliceOrientation(const QString& orientation);
+  virtual void setSliceOrientation(const QString& orientation);
 
   /// Set slice \a offset. Used to set a single value.
   void setSliceOffsetValue(double offset);
--- a/Libs/MRML/Widgets/qMRMLSliceControllerWidget_p.h
+++ b/Libs/MRML/Widgets/qMRMLSliceControllerWidget_p.h
@@ -109,7 +109,7 @@ public slots:
   void updateFromMRMLScene();
 
   /// Update widget state using the associated MRML slice node
-  void updateWidgetFromMRMLSliceNode();
+  virtual void updateWidgetFromMRMLSliceNode();
 
   /// Update widget state using the associated MRML slice composite node
   void updateWidgetFromMRMLSliceCompositeNode();


Libs/MRML/Core/vtkMRMLSliceNode.h
```diff
@@ -30,6 +30,9 @@ class vtkMatrix4x4;
 /// \li FieldOfView tells the size of  slice plane
 class VTK_MRML_EXPORT vtkMRMLSliceNode : public vtkMRMLAbstractViewNode
 {
+
+  typedef std::map<std::string, std::string> displayOrientationStringsMap ;
+
   public:
   static vtkMRMLSliceNode *New();
```
jcfr: Add the following 

```
typedef vtkMRMLSliceNode Self;
```

above the `vtkTypeMacro`

Libs/MRML/Core/vtkMRMLSliceNode.h
```diff
@@ -133,6 +136,29 @@ class VTK_MRML_EXPORT vtkMRMLSliceNode : public vtkMRMLAbstractViewNode
   /// SetOrientationToReformat() depending on the value of the string
   void SetOrientation(const char* orientation);
 
+  ///
+  /// map the Axial plane string with a custom displayOrientation string
```
jcfr: 'map' -> 'Map'

Libs/MRML/Core/vtkMRMLSliceNode.h
```diff
@@ -133,6 +136,29 @@ class VTK_MRML_EXPORT vtkMRMLSliceNode : public vtkMRMLAbstractViewNode
...
+  /// map the Coronal plane string with a custom displayOrientation string
+  static void SetCoronalOrientationDisplayString(const char* displayOrientation);
+
+  ///
+  /// A description of the current orientation,
+  /// the method return customized names
+  /// for the planes Axial, Sagittal and Coronal.
+  /// The mapping is set using the static map:
+  /// displayOrientationStringsMap.
+  /// SetAxialOrientationDisplayString
```
jcfr:
```
\sa SetAxialOrientationDisplayString(const char* displayOrientation)
\sa SetSagittalOrientationDisplayString(const char* displayOrientation)
\sa SetCoronalOrientationDisplayString(const char* displayOrientation)
```

Libs/MRML/Core/vtkMRMLSliceNode.h
```diff
@@ -30,6 +30,9 @@ class vtkMatrix4x4;
 /// \li FieldOfView tells the size of  slice plane
 class VTK_MRML_EXPORT vtkMRMLSliceNode : public vtkMRMLAbstractViewNode
 {
+
+  typedef std::map<std::string, std::string> displayOrientationStringsMap ;
```
jcfr: `displayOrientationStringsMap` -> `DisplayOrientationStringsMapType`

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -361,6 +361,45 @@ void vtkMRMLSliceNode::SetOrientation(const char* orientation)
...
+
+//----------------------------------------------------------------------------
+vtkMRMLSliceNode::displayOrientationStringsMap vtkMRMLSliceNode::DisplayOrientationStringsMap =
+    vtkMRMLSliceNode::createdisplayOrientationStringsMap();
+
+//----------------------------------------------------------------------------
+std::string vtkMRMLSliceNode::GetOrientationDisplayString()
+{
+  std::string orientation = this->GetOrientationString();
+  return DisplayOrientationStringsMap[orientation];
```
jcfr:
```
return Self::DisplayOrientationStringsMap[orientation];
```

Libs/MRML/Widgets/qMRMLSliceControllerWidget.h
```diff
@@ -158,7 +158,7 @@ public slots:
 
   /// Set slice orientation.
   /// \note Orientation could be either "Axial, "Sagittal", "Coronal" or "Reformat".
-  void setSliceOrientation(const QString& orientation);
+  virtual void setSliceOrientation(const QString& orientation);
```
jcfr: The addition of the `virtual` keyword should be in its own commit

Libs/MRML/Widgets/qMRMLSliceControllerWidget_p.h
```diff
@@ -109,7 +109,7 @@ public slots:
   void updateFromMRMLScene();
 
   /// Update widget state using the associated MRML slice node
-  void updateWidgetFromMRMLSliceNode();
+  virtual void updateWidgetFromMRMLSliceNode();
```
jcfr: The addition of the `virtual` keyword should be in its own commit

Libs/MRML/Core/vtkMRMLSliceNode.h
```diff
@@ -457,6 +483,9 @@ class VTK_MRML_EXPORT vtkMRMLSliceNode : public vtkMRMLAbstractViewNode
   int IsUpdatingMatrices;
 
   std::vector< std::string > ThreeDViewIDs;
+
+  static displayOrientationStringsMap createdisplayOrientationStringsMap();
+  static displayOrientationStringsMap DisplayOrientationStringsMap;
```
jcfr:
```
static DisplayOrientationStringsMapType DisplayOrientationStringsMap;
```

jcfr:
Nice. Thanks for the contribution :+1: 

Thinking more about it, instead of using a map. Let's simply have static ivars like:
```
std::string AxialOrientationDisplayString;
std::string CoronalOrientationDisplayString;
std::string SagitalOrientationDisplayString;
```

then, you could remove the static method `createdisplayOrientationStringsMap()` and simply statically initialize the value to their default.

Punzo:
Ok, thanks for the feedback. Tomorrow I'll change it.

lassoan:
@Punzo Thanks for your pull request. I agree that orientation string management is very rigid and unsatisfactory as it is implemented now. The most limiting factor is that only certain hardcoded orientations can have a name: orientations and corresponding names cannot be changed.

While your patch introduces minimal changes, it actually takes us farther from the actual solution of the problem, as it just adds a layer, trying to hide the problem from the user. This layer would just make implementation of a proper solution more complex.

I would recommend to actually solve the root problem: make the list of named orientations configurable. The proper solution requires only a few changes in the Slicer core (most of it makes the code nicer, by removing hardcoded constants and method names):

1. Add public methods to define named orientations in vtkMRMLSliceNode. Note that the scene can store a default slice node, so an extension only need to define named orientations only once, for this default slice node (as it is done for default orientation marker settings).

2. Make a number of small changes in vtkMRMLSliceLogic, qMRMLSliceControllerWidget, and qMRMLSliceInformationWidget to use orientations list in the slice node instead of expecting the current 3 named orientations.

In your extension you can create custom layouts with custom orientations (this is already implemented, extensions can define their own layouts; you should be also able to change any existing layout).

Punzo:
Hi @lassoan, thanks for your feedback. Yestreday, I talked exctaly about this with Jc and Steve. 
Our first idea was indeed to add a vtkStringVector to the SliceNode. Of course that solution, as you said, means to change SlicerCore in vtkMRMLSliceLogic, qMRMLSliceControllerWidget, qMRMLSliceInformationWidget and, I think also, qSlicerReformatModuleWidget. A part the layouts, there is also plenty of code in python that uses the hard-coded orientation string, example:
https://github.com/pieper/CompareVolumes/blob/master/CompareVolumes.py#L472
 
For me the proposed solution seemed accetable avoiding any break in SlicerCore and python modules (and extensions). However, I understand completly your view. @jcfr, @pieper, do you have any thought about this?

lassoan:
I searched for the axial, sagittal, coronal strings in the entire slicer source code and it pops up in very few classes. Changes in vtkMRMLSlicerSliceLogic, controller widget, info widget are all completely trivial. No impact on layouts, no backward compatibility issue, any module can still use axial, sagittal, coronal orientations, as we would keep these default slice orientations in the list by default.

Punzo:
you are completly right, what I meant is simply that I implemented what we decided yesterday. Of course I can change it (:

pieper:
If you do generalize we could certainly change the CompareVolumes code to
query for the orientation strings to initialize the variable.  It would be
a good improvement.

jcfr:
> as we would keep these default slice orientations in the list by default.

Good point.

> make the list of named orientations configurable

:+1: 

> changes in the Slicer core
> public methods to define named orientations in vtkMRMLSliceNode

Are you thinking about something like this:

```
/// Add orientation \a name associated with \a sliceToRAS transformation.
bool vtkMRMLSliceNode::AddNamedOrientation(const std::string& name, vtkMatrix4x4* sliceToRAS);

/// Insert orientation \a name associated with \a sliceToRAS transformation matrix \a afterOrientationID
///
/// \warning orientation can NOT be inserted for orientationID <= OrientationID::REFORMAT
bool vtkMRMLSliceNode::InsertNamedOrientation(const std::string& name, vtkMatrix4x4* sliceToRAS, int afterOrientationID);

/// Remove an orientation given its \a name
///
/// \warning orientationID <= OrientationID::REFORMAT can NOT be removed.
bool vtkMRMLSliceNode::RemoveNamedOrientation(const std::string& name);

/// Remove an orientation given its \a orientationID
///
/// \warning orientationID <= OrientationID::REFORMAT can NOT be removed.
bool vtkMRMLSliceNode::RemoveNamedOrientation(int orientationID);

/// Rename an orientation to \a updatedName given its \a name
bool vtkMRMLSliceNode::RenameNamedOrientation(const std::string& name, const std::string& updatedName);

/// Rename an orientation to \a updatedName given its \a orientationID
bool vtkMRMLSliceNode::RenameNamedOrientation(int orientationID, const std::string& updatedName);

/// Return list of named orientations. This corresponds to the available orientation in the Slice controllers.
void vtkMRMLSliceNode::GetNamedOrientations(vtkCollection* namedOrientations)const;

/// Return ID of the current orientation
int vtkMRMLSliceNode::GetOrientationID() const; // This is basically the position in the named orientation array

/// Convenience enum to retrieve current orientation independently of its name
enum OrientationID
{
  AXIAL = 0,
  SAGITAL = 1,
  CORONAL = 2,
  REFORMAT = 3  // This a special orientation
}
```

Then existing code doing:

```
if (slide->GetOrientationString() == "Axial")
```

would be changed into 

```
if (slide->GetOrientationID() == vtkMRMLSliceNode::AXIAL)
```

> the scene can store a default slice node

Good point. I forgot about this. 

See `vtkMRMLScene::AddDefaultNode` - https://www.slicer.org/doc/html/classvtkMRMLScene.html#ae302c5ed4aabb2910bc35dcc9aa2513f




For reference, here are the occurrences of `GetOrientationString()` 

```
$ ack "GetOrientationString\("
Modules/Loadable/Reformat/qSlicerReformatModuleWidget.cxx
264:      QString::fromStdString(this->MRMLSliceNode->GetOrientationString()));

Modules/Scripted/DataProbe/DataProbe.py
299:              orient=sliceNode.GetOrientationString(),

Libs/MRML/Widgets/qMRMLSliceInformationWidget.cxx
100:      QString::fromStdString(this->MRMLSliceNode->GetOrientationString()));

Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx
786:      QString::fromStdString(this->MRMLSliceNode->GetOrientationString()));
796:          QString::fromStdString(this->MRMLSliceNode->GetOrientationString())];

Libs/MRML/Core/vtkMRMLSliceNode.cxx
1145:  this->SetOrientationString(node->GetOrientationString());
1186:  std::string orientation = this->GetOrientationString();

Libs/MRML/Core/vtkMRMLSliceNode.h
53:  /// \sa GetOrientationString()

Applications/SlicerApp/Testing/Python/SliceLinkLogic.py
340:    self.assertEqual(compareNode2.GetOrientationString(), compareNode.GetOrientationString())
341:    self.assertEqual(compareNode3.GetOrientationString(), compareNode.GetOrientationString())
```

lassoan:
Yes, Get/Set methods like that would work.

I would allow changing/deleting the axial/sagittal/coronal names, as they are completely arbitrary (e.g., I may want to override the "axial" orientation to make patient's right point to left or right on screen). We should keep only one special string, which would be used when there is no matching named orientation for the current slice orientation (currently: "Reformat").

I don't think we need an orientation ID. There should be no need for distinguished orientations anywhere.

lassoan:
The API could be simplified to something like this:

```
/// Add orientation preset associated with \a sliceToRAS transformation.
bool vtkMRMLSliceNode::AddSliceOrientationPreset(const std::string& presetName, vtkMatrix4x4* sliceToRAS);

/// Remove an orientation given its \a name
bool vtkMRMLSliceNode::RemoveSliceOrientationPreset(const std::string& presetName);

/// Get preset name from \a sliceToRAS transformation.
vtkMatrix4x4* sliceToRAS vtkMRMLSliceNode::GetSliceOrientationPreset(const std::string& name)const;

/// Get preset name from \a sliceToRAS transformation.
std::string vtkMRMLSliceNode::GetSliceOrientationPresetName(vtkMatrix4x4* sliceToRAS) const;

/// Return list of named orientations. This corresponds to the available orientation in the Slice controllers.
void vtkMRMLSliceNode::GetSliceOrientationPresetNames(vtkStringArray* namedOrientations)const;

```

jcfr:
I like it.

To move forward, here are a possible first steps:

* [ ] Consolidate:
 * `vtkMRMLSliceNode::Matrix4x4AreEqual`
 * `vtkMRMLSliceLayerLogic.AreMatricesEqual` 
 * `vtkMRMLTransformNodeTest1.Matrix4x4AreEqual` 
 adding one method `Matrix4x4AreEqual(vtkMatrix4x4 *m1, vtkMatrix4x4 *m2, double tolerance = 1e-3)` to `vtkAddon/vtkAddonMathUtilities.h`

* [ ]  Add function `static vtkMatrix4x4* vtkMRMLSliceNode::GetSliceOrientationPreset(std::string& name) const` factoring out code like [this one](https://github.com/Slicer/Slicer/blob/bec0bb893028a8ce0d179dc31df5443373637d7f/Libs/MRML/Core/vtkMRMLSliceNode.cxx#L375-L386).

* [ ] Add function `static std::string vtkMRMLSliceNode::GetSliceOrientationPresetName(vtkMatrix4x4* sliceToRAS) const` factoring out [this code](https://github.com/Slicer/Slicer/blob/bec0bb893028a8ce0d179dc31df5443373637d7f/Libs/MRML/Core/vtkMRMLSliceNode.cxx#L650-L687) and using the `GetSliceOrientationPreset` and `vtkMRMLSliceNode::Matrix4x4AreEqual` methods

Questions:

* @lassoan For [this code](https://github.com/Slicer/Slicer/blob/bec0bb893028a8ce0d179dc31df5443373637d7f/Libs/MRML/Core/vtkMRMLSliceNode.cxx#L2028-L2068) in `vtkMRMLSliceNode::RotateToVolumePlane`, I wonder what is the best approach to get the alignedRAS indices given the named preset associated with the reference orientation ?  Is it sufficient to rely on the fallback case in case name of orientation is different from  `Axial`, `Sagittal`, ...

pieper:
The RotateToVolumePlane code is special.  It might be possible to
generalize the concepts there but it should work as-is as long as the
orientation strings are maintained for backwards compatibility.  The idea
of the code is to keep the patient from flipping over or on the side when
snapping to acquisition planes and that may not matter in non-medical
contexts.

lassoan:
Using the orientation string to decide which axis the view should be snapped to was just a temporary solution that Steve quickly implemented when we needed that feature. The correct solution would be to snap to closest axis, I've just never had the time to implement that logic.

jcfr:
Great. Thanks for the comments.

Looks like we a good first draft for the API.

Now looking at `qMRMLSliceControllerWidget`, the list of orientation listed by the controller should be automatic.

* `GetSliceOrientationPresetNames` will return all the named orientation that were added using `AddSliceOrientationPreset`. `Reformat` will be considered as a special case and always the last item of the list. With this, we can easily get all available orientations and populate the combox box.

* Next, there is a `Prefix` and a `Tooltip` associated with each orientation. This could be done adding a virtual methods.

To summarize, these changes would then be required:

 * [ ] Adding `virtual qMRMLOrientation qMRMLSliceControllerWidgetPrivate::mrmlOrientation(const QString& name)` will allow to customize the text in custom layout.

* [ ] Update `qMRMLSliceControllerWidgetPrivate::qMRMLSliceControllerWidgetPrivate` to use `vtkMRMLSlicerNode::GetSliceOrientationPresetNames` and `qMRMLSliceControllerWidgetPrivate::mrmlOrientation`

 * [ ] Remove assert in `qMRMLSliceInformationWidget::setSliceOrientation`, `qMRMLSliceControllerWidget::setSliceOrientation` and `qSlicerReformatModuleWidget::
onSliceOrientationChanged`

jcfr:
@Punzo I think they are now in good shape regarding the design, it that looks reasonable to you, you could start incrementally adding commit for each one of the check boxes.

Punzo:
Thanks to everyone for the help and feedbacks. I'll close this pull request and I'll push new ones. I'll follow the Jc's design. Of course, any further feedback on the following up pull requests will be very welcome.

jcfr:
Great. Thanks

PR #504, opened 2016-05-20
ENH: #4146 add advisor for dicom extensions

pieper:
After an import to the dicom database the contents
are checked against a pre-configured file that suggests
possible extensions.

See:

http://www.na-mic.org/Bug/view.php?id=4146

https://github.com/QIICR/Reporting/pull/71#

### DIFF
--- a/Modules/Scripted/DICOMLib/DICOMWidgets.py
+++ b/Modules/Scripted/DICOMLib/DICOMWidgets.py
@@ -76,6 +76,8 @@ class DICOMDetailsPopup(VTKObservationMixin):
     setDatabasePrecacheTags(self.dicomBrowser)
 
     self.dicomBrowser.connect('databaseDirectoryChanged(QString)', self.onDatabaseDirectoryChanged)
+    self.extensionCheckPending = False
+    self.dicomBrowser.connect('directoryImported()', self.onDirectoryImported)
 
     # Update visibility
     for name in [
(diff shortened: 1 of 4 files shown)


pieper:
@fedorov you can test this one and let me know what you think.  I found a way to avoid changing CTK.

@jcfr I looked at how to pre-load the extension manager with these extensions selected but it looked like a non-trivial change for not a lot of benefit.  Instead I give the user a dialog with instructions on how to perform the installation and I think that's enough for now.

image

Modules/Scripted/DICOMLib/DICOMWidgets.py
```diff
@@ -386,6 +388,92 @@ def onDatabaseDirectoryChanged(self,databaseDirectory):
...
+    signals during the same operation, so we collapse them
+    into a single check for compatible extensions."""
+    if not self.extensionCheckPending:
+      self.extensionCheckPending = True
+      def timerCallback():
+        self.promptForExtensions()
+        self.extensionCheckPending = False
+      qt.QTimer.singleShot(0, timerCallback)
+
+  def promptForExtensions(self):
```
jcfr: -> `def promptForExtensions(self, extensionsToInstall):`

Modules/Scripted/DICOMLib/DICOMWidgets.py
```diff
@@ -386,6 +388,92 @@ def onDatabaseDirectoryChanged(self,databaseDirectory):
...
+        pluralOrNot = " is"
+      else:
+        pluralOrNot = "s are"
+      message = "The following extension%s not installed, but may help you work with DICOM data in your database:\n\n" % pluralOrNot
+      for extension in extensionsToOffer:
+        message += '  ' + extension + '\n'
+      message += "\n\nYou can install extensions using the Extension Manager option from the View menu."
+      slicer.util.infoDisplay(message, windowTitle='DICOM')
+
+  def checkForExtensions(self):
```
jcfr: -> `checkForExtensions(dicomDatabase, extensionDescriptorPath)`

The idea is to make the function independent of the internal state and potentially re-usable if different `database` and `descriptorfile`

Modules/Scripted/DICOMLib/DICOMWidgets.py
```diff
@@ -386,6 +388,92 @@ def onDatabaseDirectoryChanged(self,databaseDirectory):
...
+    then return matches
+
+    See
+    http://www.na-mic.org/Bug/view.php?id=4146
+    """
+
+    # 1 - load json
+    import logging, os, json
+    logging.info('Imported a DICOM directory, checking for extensions')
+    modulePath = os.path.dirname(slicer.modules.dicom.path)
```
jcfr: Add a method `lookupExtensionDescriptorPath()`

Modules/Scripted/DICOMLib/DICOMWidgets.py
```diff
@@ -386,6 +388,92 @@ def onDatabaseDirectoryChanged(self,databaseDirectory):
...
       slicer.app.setDICOMDatabase(slicer.dicomDatabase)
 
+  def onDirectoryImported(self):
+    """The dicom browser will emit multiple directoryImported
+    signals during the same operation, so we collapse them
+    into a single check for compatible extensions."""
+    if not self.extensionCheckPending:
+      self.extensionCheckPending = True
+      def timerCallback():
+        self.promptForExtensions()
```
jcfr: -> `promptForExtensions(self.checkForExtensions())`

jcfr:
Nicee.

> pre-load the extension manager with these extensions selected

Will see what it would take to give you an api for it ..

fedorov:
I will test and report back tomorrow. Thank you Steve!

fedorov:
Sorry for the delay in getting to test this.

Here are few suggestions:

 - how about adding explicit check any time it is not possible to parse an instance without warnings? 

 - I would add a human-readable description of the object type, and at the time of import (or explicit check), and formulate the message to the user as "The data you are trying to parse contains DICOM Segmentation Object(s), which cannot be parsed properly. You might want to install Reporting extension to help with this data type"

pieper:
New update includes Andrey's @fedorov suggestion for human readable object description.

I didn't add the check on loads with warnings because I felt this might be too frequent and users could get annoyed.

I didn't implement suggestions from @jcfr because I felt this it is premature to generalize the code for use cases that don't currently exist.

jcfr:
> premature to generalize the code 

Agree that adding `lookupExtensionDescriptorPath()` may be overkill at that stage, that said I would think that having two functions like:

```
self.checkForExtensions()
def promptForExtensions(self, extensionsToInstall)
```

instead of current `promptForExtensions()` would facilitate testing and re-use and aligned with the general idea of splitting UI and logic.

It is quite common to integrate part of the DICOM interface in Slicelet or other applications.

That said, It is perfectly fine to squash commits and keep things as it is. It always be possible to refactor later ...

pieper:
Okay, yes, agreed.  Let's let Andrey have a look.

fedorov:
@pieper @jcfr I did not realize this one is not merged yet! Does it make sense to integrate and refine later?

pieper:
I squashed and committed.

r25567 = 51220abe619be7225377d50494dc158083cfdbc9

Looked again at generalizing the promptForExtensions method but I think that if there's a generalized version it probably belongs in the ExtensionManager itself and not in DICOM.  Also I don't have a good way to test any generalized logic so I don't want to add dead code.

fedorov:
@pieper I've just experienced this new feature, and I realized that it probably makes sense to revisit it.

The problem is that it adds yet another popup to the interface:

1. From Slicer - open DICOM Browser -- popup 1
2. From DICOM Browser - import data -- file open dialog -- popup 2
3. Once import is done - notification of how many patients are added -- popup 3
4. Advisor message about the available extensions -- popup 4

I have to say I was overwhelmed, since it is not clear which popup I should "ok" and what this all means!

Does it make sense to at least consolidate popup 3 and popup 4 into one? It is quite messy otherwise, as you might be able to see below:

image

fedorov:
I am also wondering if it is technically possible to close the file open dialog before opening the import notification popup to further reduce the total?

pieper:
Sure, if you have a design you think would work better let's discuss.  I
can imagine combining the dicom extension advice in with the import
complete dialog.  But even better would be a way to offer to auto-download
the required extensions, which requires some extra features for the
extension manager.

fedorov:
> if you have a design you think would work better let's discuss

yes, I suggest 1) close file open dialog before any completion-related popups show up; 2) have one popup for both reporting the number of imported series/patients, and suggesting useful extensions. What do you think?

pieper:
I used this with some data and I didn't find the extra dialog box to be
objectionable.

I tend to this that if the info is tacked at the end of another dialog that
people are used to ignoring it might get missed.

There's also the small complication that the 'new
patient/study/series/instance count' dialog is in CTK, while the plugin
advisor is in slicer.  that's solvable but means adding some extra
abstraction and generalization features to CTK that may not IMO be worth
it...

About closing the file dialog first, I guess we could but since this result
is generated inside the slot triggered by the Ok button it's 'logically'
part of that operation and so seeing the layers of responses helps
reinforce to the user that these feedback dialogs are a part of the load
operation.  Unless I'm mistaken that's consistent with the way something
like Microsoft Word would work.  If the dialog came up after the dialog is
gone it might seem unrelated.

fedorov:
@pieper ok, I guess there can be a logical explanation to anything, but this didn't change my opinion about the current interface. This is not the interface I would deliver to the users I would want to be happy, but we should be able to customize as needed in the extensions. I guess the question is whether users who can't customize UI will be happy, and I will leave it to them to decide whether to push this agenda forward. Too bad CTK appears to be more of an obstacle than a helper when it comes to customizing UI.

pieper:
If we are going to put effort into making things easier for the user we
should just bundle the dicom interpretation code with slicer core and not
make people install them.  That's the main reason this is clunky.

lassoan:
There are dependencies to extension code - for example you need special RT nodes to read all info from DICOM-RT files. Also, having DICOM plugins in extensions have several advantages:
* it makes DICOM loading much faster that less plugins parse the data
* there is a better chance that the best loadable is chosen (by installing an extension the user indicates that he is interested in a particular domain, so plugins in extensions can use higher loadable confidence values)
* DICOM plugin developers don't have to worry about breaking the Slicer core
* DICOM plugin developers can update their plugins more easily

pieper:
Yes, those are good points Andras.

I guess what I keep coming back to is that I feel we need a more
comprehensive overhaul of the user experience for DICOM loading and
interpretation.

jcfr:
@pieper What is the status on this ? Could we close the PR ?

pieper:
yes, this has been integrated (still could use improvements as discussed above) so I'll close this.

jcfr:
Thanks @pieper . For future reference, feel free to move some of the discussion point to the Slicer forum

PR #506, opened 2016-05-25
ENH: enable customization of SliceNode and SliceController

Punzo:
Hi @jcfr . 

I implemented all the API and infrastructure as designed in https://github.com/Slicer/Slicer/pull/499.
I apologize that it is all in one PR. I wanted to test it so I preferred to implement everything. 

Tests:
1) I loaded MRHead, I navigated the datacube using the SliceController and I used several modules.
2) I loaded my extension (which customize the 2-D View with the new API), I navigated the datacube using the SliceController and I used several modules.

@lassoan @pieper may you also kindly have a look?

Thanks,

Davide.

### DIFF
--- a/Base/Logic/vtkSlicerApplicationLogic.cxx
+++ b/Base/Logic/vtkSlicerApplicationLogic.cxx
@@ -26,10 +26,12 @@
 #include <vtkMRMLModelHierarchyNode.h>
 #include <vtkMRMLScene.h>
 #include <vtkMRMLSelectionNode.h>
+#include <vtkMRMLSliceNode.h>
 #include <vtkMRMLTableNode.h>
 #include <vtkMRMLRemoteIOLogic.h>
 
 // VTK includes
+#include <vtkMatrix4x4.h>
 #include <vtkNew.h>
 #include <vtkObjectFactory.h>
 #include <vtkPointData.h>
@@ -295,6 +297,60 @@ void vtkSlicerApplicationLogic::SetMRMLSceneDataIO(vtkMRMLScene* newMRMLScene,
       {
       remoteIOLogic->AddDataIOToScene();
       }
+    }
+
+  // Setting Orientation Matrices presets
+  vtkNew<vtkMatrix4x4> axialSliceToRAS;
+  axialSliceToRAS->SetElement(0, 0, -1.0);
+  axialSliceToRAS->SetElement(1, 0,  0.0);
+  axialSliceToRAS->SetElement(2, 0,  0.0);
+  axialSliceToRAS->SetElement(0, 1,  0.0);
+  axialSliceToRAS->SetElement(1, 1,  1.0);
+  axialSliceToRAS->SetElement(2, 1,  0.0);
+  axialSliceToRAS->SetElement(0, 2,  0.0);
+  axialSliceToRAS->SetElement(1, 2,  0.0);
+  axialSliceToRAS->SetElement(2, 2,  1.0);
+
+  vtkNew<vtkMatrix4x4> sagittalSliceToRAS;
+  sagittalSliceToRAS->SetElement(0, 0,  0.0);
+  sagittalSliceToRAS->SetElement(1, 0,  -1.0);
+  sagittalSliceToRAS->SetElement(2, 0,  0.0);
+  sagittalSliceToRAS->SetElement(0, 1,  0.0);
+  sagittalSliceToRAS->SetElement(1, 1,  0.0);
+  sagittalSliceToRAS->SetElement(2, 1,  1.0);
+  sagittalSliceToRAS->SetElement(0, 2,  1.0);
+  sagittalSliceToRAS->SetElement(1, 2,  0.0);
+  sagittalSliceToRAS->SetElement(2, 2,  0.0);
+
+  vtkNew<vtkMatrix4x4> coronalSliceToRAS;
+  coronalSliceToRAS->SetElement(0, 0, -1.0);
+  coronalSliceToRAS->SetElement(1, 0,  0.0);
+  coronalSliceToRAS->SetElement(2, 0,  0.0);
+  coronalSliceToRAS->SetElement(0, 1,  0.0);
+  coronalSliceToRAS->SetElement(1, 1,  0.0);
+  coronalSliceToRAS->SetElement(2, 1,  1.0);
+  coronalSliceToRAS->SetElement(0, 2,  0.0);
+  coronalSliceToRAS->SetElement(1, 2,  1.0);
+  coronalSliceToRAS->SetElement(2, 2,  0.0);
+
+  // Setting a Slice Default Node
+  vtkSmartPointer<vtkMRMLSliceNode> defaultNode = vtkMRMLSliceNode::SafeDownCast
+      (newMRMLScene->GetDefaultNodeByClass("vtkMRMLSliceNode"));
+  if (defaultNode)
+    {
+    defaultNode->AddSliceOrientationPreset("Axial", axialSliceToRAS.GetPointer());
+    defaultNode->AddSliceOrientationPreset("Sagittal", sagittalSliceToRAS.GetPointer());
+    defaultNode->AddSliceOrientationPreset("Coronal", coronalSliceToRAS.GetPointer());
+    }
+  else
+    {
+    defaultNode = vtkMRMLSliceNode::SafeDownCast
+        (newMRMLScene->CreateNodeByClass("vtkMRMLSliceNode"));
+    defaultNode->AddSliceOrientationPreset("Axial", axialSliceToRAS.GetPointer());
+    defaultNode->AddSliceOrientationPreset("Sagittal", sagittalSliceToRAS.GetPointer());
+    defaultNode->AddSliceOrientationPreset("Coronal", coronalSliceToRAS.GetPointer());
+    newMRMLScene->AddDefaultNode(defaultNode);
+    defaultNode->Delete(); // scene owns it now
     }
 }
 
--- a/Libs/MRML/Core/vtkMRMLSliceNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLSliceNode.cxx
@@ -19,12 +19,15 @@ Version:   $Revision: 1.2 $
 #include "vtkMRMLVolumeNode.h"
 
 // VTK includes
+#include <vtkAddonMathUtilities.h>
 #include <vtkMath.h>
 #include <vtkMatrix4x4.h>
 #include <vtkNew.h>
 #include <vtkObjectFactory.h>
+#include <vtkSmartPointer.h>
+#include <vtkStringArray.h>
 #include <vtkVector.h>
-#include <vtkAddonMathUtilities.h>
+
 
 // VNL includes
 #include <vnl/vnl_double_3.h>
@@ -43,7 +46,7 @@ vtkMRMLNodeNewMacro(vtkMRMLSliceNode);
 vtkMRMLSliceNode::vtkMRMLSliceNode()
 {
   // set by user
-  this->SliceToRAS = vtkMatrix4x4::New();
+  this->SliceToRAS = vtkSmartPointer<vtkMatrix4x4>::New();
   this->SliceToRAS->Identity();
 
   this->JumpMode = OffsetJumpSlice;
@@ -52,14 +55,14 @@ vtkMRMLSliceNode::vtkMRMLSliceNode()
   this->OrientationReference = NULL;
 
   // calculated by UpdateMatrices()
-  this->XYToSlice = vtkMatrix4x4::New();
-  this->XYToRAS = vtkMatrix4x4::New();
-  this->UVWToSlice = vtkMatrix4x4::New();
-  this->UVWToRAS = vtkMatrix4x4::New();
+  this->XYToSlice = vtkSmartPointer<vtkMatrix4x4>::New();
+  this->XYToRAS = vtkSmartPointer<vtkMatrix4x4>::New();
+  this->UVWToSlice = vtkSmartPointer<vtkMatrix4x4>::New();
+  this->UVWToRAS = vtkSmartPointer<vtkMatrix4x4>::New();
 
   // set the default field of view to a convenient size for looking
   // at slices through human heads (a 1 pixel thick slab 25x25 cm)
-  // TODO: how best to represent this as a slab rather than infinitessimal slice?
+  // TODO: how best to represent this as a slab rather than infinitesimal slice?
   this->FieldOfView[0] = 250.0;
   this->FieldOfView[1] = 250.0;
   this->FieldOfView[2] = 1.0;
@@ -113,7 +116,8 @@ vtkMRMLSliceNode::vtkMRMLSliceNode()
   this->LayoutColor[1] = vtkMRMLSliceNode::grayColor()[1];
   this->LayoutColor[2] = vtkMRMLSliceNode::grayColor()[2];
 
-  this->SetOrientationToAxial();
+  this->SetOrientationString("Reformat");
+  this->SetOrientationReference("Reformat");
   this->SetLayoutLabel("");
 
   this->OrientationMarkerEnabled = true;
@@ -123,26 +127,6 @@ vtkMRMLSliceNode::vtkMRMLSliceNode()
 //----------------------------------------------------------------------------
 vtkMRMLSliceNode::~vtkMRMLSliceNode()
 {
-  if ( this->SliceToRAS != NULL)
-    {
-    this->SliceToRAS->Delete();
-    }
-  if ( this->XYToSlice != NULL)
-    {
-    this->XYToSlice->Delete();
-    }
-  if ( this->XYToRAS != NULL)
-    {
-    this->XYToRAS->Delete();
-    }
-  if ( this->UVWToSlice != NULL)
-    {
-    this->UVWToSlice->Delete();
-    }
-  if ( this->UVWToRAS != NULL)
-    {
-    this->UVWToRAS->Delete();
-    }
   if ( this->OrientationString )
     {
     delete [] this->OrientationString;
@@ -333,32 +317,32 @@ bool vtkMRMLSliceNode
 }
 
 //----------------------------------------------------------------------------
-void vtkMRMLSliceNode::SetOrientation(const char* orientation)
+bool vtkMRMLSliceNode::SetOrientation(const char* orientation)
 {
   if (!orientation)
     {
-    return;
+    return false;
     }
- if (!strcmp(orientation, "Axial"))
+
+  vtkMatrix4x4 *orientationPreset;
+  orientationPreset = this->GetSliceOrientationPreset(orientation);
+
+  if (!strcmp(orientation, "Reformat"))
     {
-    this->SetOrientationToAxial();
+    this->SetOrientationString(orientation);
     }
-  else if (!strcmp(orientation, "Sagittal"))
+
+  if (orientationPreset == NULL)
     {
-    this->SetOrientationToSagittal();
+    return false;
     }
-  else if (!strcmp(orientation, "Coronal"))
-    {
-    this->SetOrientationToCoronal();
-    }
-  else if (!strcmp(orientation, "Reformat"))
-    {
-    this->SetOrientationToReformat();
-    }
-  else
-    {
-    vtkErrorMacro("SetOrientation: invalid orientation: " << orientation);
-    }
+
+  this->SliceToRAS->DeepCopy(orientationPreset);
+  this->SetOrientationString(orientation);
+  //SetOrientationReference() behaviour will be mantained for backward compatibility
+  this->SetOrientationReference(orientation);
+  this->UpdateMatrices();
+  return true;
 }
 
 //----------------------------------------------------------------------------
@@ -371,73 +355,190 @@ void vtkMRMLSliceNode::SetOrientationToReformat()
 }
 
 //----------------------------------------------------------------------------
-void vtkMRMLSliceNode::SetOrientationToAxial()
+// This method is matained for backward compatibility.
+bool vtkMRMLSliceNode::SetOrientationToAxial()
 {
-    // Px -> Patient Left
-    this->SliceToRAS->SetElement(0, 0, -1.0);
-    this->SliceToRAS->SetElement(1, 0,  0.0);
-    this->SliceToRAS->SetElement(2, 0,  0.0);
-    // Py -> Patient Anterior
-    this->SliceToRAS->SetElement(0, 1,  0.0);
-    this->SliceToRAS->SetElement(1, 1,  1.0);
-    this->SliceToRAS->SetElement(2, 1,  0.0);
-    // Pz -> Patient Inferior
-    this->SliceToRAS->SetElement(0, 2,  0.0);
-    this->SliceToRAS->SetElement(1, 2,  0.0);
-    this->SliceToRAS->SetElement(2, 2,  1.0);
-
-    this->SetOrientationString( "Axial" );
-    this->SetOrientationReference( "Axial" );
-    this->UpdateMatrices();
+  return this->SetOrientation("Axial");
 }
 
 //----------------------------------------------------------------------------
-void vtkMRMLSliceNode::SetOrientationToSagittal()
+// This method is matained for backward compatibility.
+bool vtkMRMLSliceNode::SetOrientationToSagittal()
 {
-    // Px -> Patient Posterior
-    this->SliceToRAS->SetElement(0, 0,  0.0);
-    this->SliceToRAS->SetElement(1, 0, -1.0);
-    this->SliceToRAS->SetElement(2, 0,  0.0);
-    // Py -> Patient Inferior
-    this->SliceToRAS->SetElement(0, 1,  0.0);
-    this->SliceToRAS->SetElement(1, 1,  0.0);
-    this->SliceToRAS->SetElement(2, 1,  1.0);
-    // Pz -> Patient Right
-    this->SliceToRAS->SetElement(0, 2,  1.0);
-    this->SliceToRAS->SetElement(1, 2,  0.0);
-    this->SliceToRAS->SetElement(2, 2,  0.0);
-
-    this->SetOrientationString( "Sagittal" );
-    this->SetOrientationReference( "Sagittal" );
-    this->UpdateMatrices();
+  return this->SetOrientation("Sagittal");
 }
 
 //----------------------------------------------------------------------------
-void vtkMRMLSliceNode::SetOrientationToCoronal()
+// This method is matained for backward compatibility.
+bool vtkMRMLSliceNode::SetOrientationToCoronal()
 {
-    // Px -> Patient Left
-    this->SliceToRAS->SetElement(0, 0, -1.0);
-    this->SliceToRAS->SetElement(1, 0,  0.0);
-    this->SliceToRAS->SetElement(2, 0,  0.0);
-    // Py -> Patient Inferior
-    this->SliceToRAS->SetElement(0, 1,  0.0);
-    this->SliceToRAS->SetElement(1, 1,  0.0);
-    this->SliceToRAS->SetElement(2, 1,  1.0);
-    // Pz -> Patient Anterior
-    this->SliceToRAS->SetElement(0, 2,  0.0);
-    this->SliceToRAS->SetElement(1, 2,  1.0);
-    this->SliceToRAS->SetElement(2, 2,  0.0);
-
-    this->SetOrientationString( "Coronal" );
-    this->SetOrientationReference( "Coronal" );
-    this->UpdateMatrices();
+  return this->SetOrientation("Coronal");
 }
 
 //----------------------------------------------------------------------------
 // Local helper to compare matrices -- TODO: is there a standard version of this?
 bool vtkMRMLSliceNode::Matrix4x4AreEqual(const vtkMatrix4x4 *m1, const vtkMatrix4x4 *m2)
 {
-  return vtkAddonMathUtilities::Matrix4x4AreEqual(m1, m2, /* tolerance= */ 0);
+  return vtkAddonMathUtilities::Matrix4x4AreEqual(m1, m2);
+}
+
+//----------------------------------------------------------------------------
+vtkMatrix4x4 *vtkMRMLSliceNode::GetSliceOrientationPreset(const std::string &name)
+{
+  if (!name.compare("Reformat"))
+    {
+    return NULL;
+    }
+
+  std::vector< std::pair <std::string, vtkSmartPointer<vtkMatrix4x4> > >::iterator it;
+  for (it = this->OrientationMatrices.begin(); it != this->OrientationMatrices.end(); ++it)
+    {
+    if (!it->first.compare(name))
+      {
+      return it->second;
+      }
+    }
+
+  vtkErrorMacro("GetSliceOrientationPreset: invalid Orientation Name: " << name);
+  return NULL;
+}
+
+//----------------------------------------------------------------------------
+std::string vtkMRMLSliceNode::GetSliceOrientationPresetName(vtkMatrix4x4 *sliceToRAS)
+{
+  if (sliceToRAS == NULL)
+    {
+    vtkErrorMacro("GetSliceOrientationPresetName: invalid input Matrix. ");
+    return "";
+    }
+
+  std::vector< std::pair <std::string, vtkSmartPointer<vtkMatrix4x4> > >::reverse_iterator it;
+  for (it = this->OrientationMatrices.rbegin(); it != this->OrientationMatrices.rend(); ++it)
+    {
+    std::string orientation = it->first;
+    vtkMatrix4x4* storedSliceToRAS = this->GetSliceOrientationPreset(orientation);
+    if (storedSliceToRAS == NULL)
+      {
+      continue;
+      }
+    if (vtkMRMLSliceNode::Matrix4x4AreEqual(storedSliceToRAS, sliceToRAS))
+      {
+      return orientation;
+      }
+    }
+  return "Reformat";
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLSliceNode::GetSliceOrientationPresetNames(vtkStringArray *namedOrientations)
+{
+  if (namedOrientations == NULL)
+    {
+    vtkErrorMacro("GetSliceOrientationPresetNames: invalid input vtkStringArray.");
+    return;
+    }
+
+  namedOrientations->SetNumberOfValues(this->OrientationMatrices.size());
+
+  std::vector< std::pair <std::string, vtkSmartPointer<vtkMatrix4x4> > >::iterator it;
+  int id = 0;
+  for (it = this->OrientationMatrices.begin(); it != this->OrientationMatrices.end(); ++it)
+    {
+    namedOrientations->SetValue(id, it->first);
+    id++;
+    }
+}
+
+//----------------------------------------------------------------------------
+bool vtkMRMLSliceNode::AddSliceOrientationPreset(const std::string &name, vtkMatrix4x4 *sliceToRAS)
+{
+  if (!name.compare("Reformat"))
+    {
+    vtkWarningMacro("AddSliceOrientationPreset: Reformat refer to any arbitrary orientation. "
+                    "Therefore, it can not have a preset.");
+    return false;
+    }
+
+  std::vector< std::pair <std::string, vtkSmartPointer<vtkMatrix4x4> > >::iterator it;
+  for (it = this->OrientationMatrices.begin(); it != this->OrientationMatrices.end(); ++it)
+    {
+    if (!it->first.compare(name))
+      {
+      vtkDebugMacro("AddSliceOrientationPreset: the orientation preset " << name << " is already stored.");
+      return false;
+      }
+    }
+
+  this->OrientationMatrices.push_back(std::pair<std::string, vtkSmartPointer<vtkMatrix4x4> >(name, sliceToRAS));
+  return true;
+}
+
+//----------------------------------------------------------------------------
+bool vtkMRMLSliceNode::RemoveSliceOrientationPreset(const std::string &name)
+{
+  std::vector< std::pair <std::string, vtkSmartPointer<vtkMatrix4x4> > >::iterator it;
+  for (it = this->OrientationMatrices.begin(); it != this->OrientationMatrices.end(); ++it)
+    {
+    if (!it->first.compare(name))
+      {
+      this->OrientationMatrices.erase(it);
+      return true;
+      }
+    }
+
+  vtkErrorMacro("RemoveSliceOrientationPreset: the orientation preset " << name << " is not stored.");
+  return false;
+}
+
+//----------------------------------------------------------------------------
+bool vtkMRMLSliceNode::RenameSliceOrientationPreset(const std::string &name, const std::string &updatedName)
+{
+  if (!name.compare(updatedName))
+    {
+    vtkErrorMacro("RenameSliceOrientationPreset: the orientation name " << name <<
+                  "and the update orientation name " << updatedName << "coincide.");
+    return false;
+    }
+
+  if (!name.compare(this->GetOrientationString()))
+    {
+    this->SetOrientationString(updatedName.c_str());
+    }
+
+  std::vector< std::pair <std::string, vtkSmartPointer<vtkMatrix4x4> > >::iterator it;
+  for (it = this->OrientationMatrices.begin(); it != this->OrientationMatrices.end(); ++it)
+    {
+    if (!it->first.compare(name))
+      {
+      it->first = updatedName;
+      return true;
+      }
+    }
+
+  vtkErrorMacro("RenameSliceOrientationPreset: the orientation preset " << name << " is not stored.");
+  return false;
+}
+
+//----------------------------------------------------------------------------
+bool vtkMRMLSliceNode::HasSliceOrientationPreset(const std::string &name)
+{
+  if (!name.compare("Reformat"))
+    {
+    vtkWarningMacro("HasSliceOrientationPreset: Reformat refer to any arbitrary orientation. "
+                    "Therefore, it does not have a preset.");
+    return false;
+    }
+
+  std::vector< std::pair <std::string, vtkSmartPointer<vtkMatrix4x4> > >::iterator it;
+  for (it = this->OrientationMatrices.begin(); it != this->OrientationMatrices.end(); ++it)
+    {
+    if (!it->first.compare(name))
+      {
+      return true;
+      }
+    }
+
+  return false;
 }
 
 //----------------------------------------------------------------------------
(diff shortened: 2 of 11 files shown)


Punzo:
Hi @jcfr I found a bug. it is related to the fact taht I allocate the matrices in the costructor. 

Then when an extension add its Deafult Slice Node, the matrices are not overrided: it will have still a copy ol the old ones in the map plus the new ones. 

This can be solved using a default Node also in the core with the AddSliceOrientationPreset as you proposed. However, it will not work in vtkMRMLSliceLogic::SetMRMLSceneInternal, because I think the layout is instantiated before that and the matrices are missing.

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -113,6 +116,42 @@ vtkMRMLSliceNode::vtkMRMLSliceNode()
   this->LayoutColor[1] = vtkMRMLSliceNode::grayColor()[1];
   this->LayoutColor[2] = vtkMRMLSliceNode::grayColor()[2];
 
+  this->axialSliceToRAS = vtkSmartPointer<vtkMatrix4x4>::New();
```
lassoan: use vtkNew<vtkMatrix4x4> instead

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -113,6 +116,42 @@ vtkMRMLSliceNode::vtkMRMLSliceNode()
...
+  this->axialSliceToRAS->SetElement(2, 0,  0.0);
+  this->axialSliceToRAS->SetElement(0, 1,  0.0);
+  this->axialSliceToRAS->SetElement(1, 1,  1.0);
+  this->axialSliceToRAS->SetElement(2, 1,  0.0);
+  this->axialSliceToRAS->SetElement(0, 2,  0.0);
+  this->axialSliceToRAS->SetElement(1, 2,  0.0);
+  this->axialSliceToRAS->SetElement(2, 2,  1.0);
+  this->AddSliceOrientationPreset("Axial", this->axialSliceToRAS);
+
+  this->sagittalSliceToRAS = vtkSmartPointer<vtkMatrix4x4>::New();
```
lassoan: vtkNew

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -113,6 +116,42 @@ vtkMRMLSliceNode::vtkMRMLSliceNode()
...
+  this->sagittalSliceToRAS->SetElement(2, 0,  0.0);
+  this->sagittalSliceToRAS->SetElement(0, 1,  0.0);
+  this->sagittalSliceToRAS->SetElement(1, 1,  0.0);
+  this->sagittalSliceToRAS->SetElement(2, 1,  1.0);
+  this->sagittalSliceToRAS->SetElement(0, 2,  1.0);
+  this->sagittalSliceToRAS->SetElement(1, 2,  0.0);
+  this->sagittalSliceToRAS->SetElement(2, 2,  0.0);
+  this->AddSliceOrientationPreset("Sagittal", this->sagittalSliceToRAS);
+
+  this->coronalSliceToRAS = vtkSmartPointer<vtkMatrix4x4>::New();
```
lassoan: vtkNew

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -143,6 +182,7 @@ vtkMRMLSliceNode::~vtkMRMLSliceNode()
     {
     this->UVWToRAS->Delete();
     }
+  this->OrientationMatrices.clear();
```
lassoan: no need for this

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -339,25 +379,22 @@ void vtkMRMLSliceNode::SetOrientation(const char* orientation)
...
-    {
-    this->SetOrientationToCoronal();
-    }
-  else if (!strcmp(orientation, "Reformat"))
-    {
-    this->SetOrientationToReformat();
-    }
-  else
+
+  if (!strcmp(orientation, "Reformat"))
```
lassoan: No need to check every time if an input is "Reformat". Just make sure "Reformat" string cannot be added as an orientation preset name and everywhere else you can rely on that.

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -339,25 +379,22 @@ void vtkMRMLSliceNode::SetOrientation(const char* orientation)
...
+   {
+   this->SetOrientationToReformat();
+   return;
+   }
+
+  this->SliceToRAS->DeepCopy(this->GetSliceOrientationPreset(orientation));
```
lassoan: this will probably crash if no orientation preset is not found by that name

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -339,25 +379,22 @@ void vtkMRMLSliceNode::SetOrientation(const char* orientation)
...
+
+  //SetOrientationString() and SetOrientationReference()
+  //behaviour will be mantained for backward compatibility
+  if(this->SliceToRAS)
```
lassoan: this can never happen (you've just called a method of this->SliceToRAS, so if it was NULL then the application has already crashed)

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -339,25 +379,22 @@ void vtkMRMLSliceNode::SetOrientation(const char* orientation)
...
     {
-    vtkErrorMacro("SetOrientation: invalid orientation: " << orientation);
+    this->SetOrientationString(orientation);
+    this->SetOrientationReference(orientation);
```
lassoan: Get rid of this orientation reference member variable. Rotate to volume plane should snap to the closest volume axis.

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -371,73 +408,194 @@ void vtkMRMLSliceNode::SetOrientationToReformat()
...
-    this->SliceToRAS->SetElement(2, 1,  0.0);
-    // Pz -> Patient Inferior
-    this->SliceToRAS->SetElement(0, 2,  0.0);
-    this->SliceToRAS->SetElement(1, 2,  0.0);
-    this->SliceToRAS->SetElement(2, 2,  1.0);
-
-    this->SetOrientationString( "Axial" );
-    this->SetOrientationReference( "Axial" );
-    this->UpdateMatrices();
+  this->SliceToRAS->DeepCopy(this->GetSliceOrientationPreset("Axial"));
```
lassoan: it will crash if Axial is deleted from the list of presets

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -371,73 +408,194 @@ void vtkMRMLSliceNode::SetOrientationToReformat()
...
+
+  //SetOrientationString() and SetOrientationReference()
+  //behaviour will be mantained for backward compatibility
+  this->SetOrientationString("Axial");
```
lassoan: why don't you just call vtkMRMLSliceNode::SetOrientation("Axial")?

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -371,73 +408,194 @@ void vtkMRMLSliceNode::SetOrientationToReformat()
...
+  if (sliceToRAS == NULL)
+    {
+    vtkErrorMacro("GetSliceOrientationPresetName: invalid input Matrix. ");
+    return "";
+    }
+
+  std::map<vtkStdString, vtkMatrix4x4*>::iterator it;
+  for (it = this->OrientationMatrices.begin(); it != this->OrientationMatrices.end(); ++it)
+    {
+    std::string orietation = it->first;
```
lassoan: orietation - typo

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -371,73 +408,194 @@ void vtkMRMLSliceNode::SetOrientationToReformat()
...
+    if (!orietation.compare("Reformat"))
```
lassoan: no need for this, we know there is no preset with the name "Reformat"

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -371,73 +408,194 @@ void vtkMRMLSliceNode::SetOrientationToReformat()
...
+      {
+      continue;
+      }
+    vtkMatrix4x4* storedSliceToRAS = this->GetSliceOrientationPreset(orietation);
+    if (storedSliceToRAS != NULL)
```
lassoan: don't have so many nested conditionals, just add
if (storedSliceToRas == NULL)
{
  log error
  continue;
}

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -371,73 +408,194 @@ void vtkMRMLSliceNode::SetOrientationToReformat()
...
+    }
+
+  vtkErrorMacro("GetSliceOrientationPresetName: orientation string not found.");
+  return "";
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLSliceNode::GetSliceOrientationPresetNames(vtkStringArray *namedOrientations)
+{
+  namedOrientations->SetNumberOfValues(this->OrientationMatrices.size() + 1);
```
lassoan: add namedOrientations NULL check

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -371,73 +408,194 @@ void vtkMRMLSliceNode::SetOrientationToReformat()
...
+
+  std::map<vtkStdString, vtkMatrix4x4*>::iterator it;
+  int id = 0;
+  for (it = this->OrientationMatrices.begin(); it != this->OrientationMatrices.end(); ++it)
+    {
+    namedOrientations->SetValue(id, it->first);
+    id++;
+    }
+  namedOrientations->SetValue(id, "Reformat");
```
lassoan: add a constant for "Reformat" and use that everywhere instead of the string itself

lassoan: I don't think it's a good idea to return the "Reformat" string along with the real orientation preset names. The "Reformat" option can be added at the GUI level (e.g., the widget asks for the orientation preset names and it may add the "Reformat" name if it wants to).

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -371,73 +408,194 @@ void vtkMRMLSliceNode::SetOrientationToReformat()
...
+}
+
+//----------------------------------------------------------------------------
+bool vtkMRMLSliceNode::AddSliceOrientationPreset(const std::string &name, vtkMatrix4x4 *sliceToRAS)
+{
+  if (!name.compare("Reformat"))
+    {
+    return false;
```
lassoan: add a warning message, it may not be trivial for developers why this fails

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -371,73 +408,194 @@ void vtkMRMLSliceNode::SetOrientationToReformat()
...
+    }
+
+  std::map<vtkStdString, vtkMatrix4x4*>::iterator it = this->OrientationMatrices.find(name);
+  if (it != this->OrientationMatrices.end())
+    {
+    vtkErrorMacro("AddSliceOrientationPreset: the orietation preset " << name << " is already stored.");
+    return false;
+    }
+
```
lassoan: We don't check if the matrix is the same as the matrix in any of the existing presets. This means that we may have multiple names with the same matrix and therefore the result of name lookup based on matrix is undeterministic (especially because we store presets in a map and maps are not sorted).
It would be better to somehow make the behavior more deterministic, for example use a vector instead of a map (and when we search for a preset name, always start from the most recently added preset name).
It's probably useful to be able to have user-defined presets that may have the same orientation as a default preset, but it would be nice if we could ensure that the name lookup would always return the user-defined orientation.

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -371,73 +408,194 @@ void vtkMRMLSliceNode::SetOrientationToReformat()
...
+  this->OrientationMatrices[name] = sliceToRAS;
```
lassoan: Serious issue: object ownership is not clear. We should deep-copy the matrix into the preset list.

Punzo:
@lassoan  thank you very much for your feedback.

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -371,73 +408,194 @@ void vtkMRMLSliceNode::SetOrientationToReformat()
...
+  return true;
+}
+
+//----------------------------------------------------------------------------
+bool vtkMRMLSliceNode::RemoveSliceOrientationPreset(const std::string &name)
+{
+  if (!name.compare("Reformat"))
```
lassoan: not needed, we never have a preset named "Reformat"

Libs/MRML/Core/vtkMRMLSliceNode.cxx
Punzo: ok I agree on that, I can be done at the Controller widget level. So I can remove all the if in the various methods

Libs/MRML/Core/vtkMRMLSliceNode.cxx
Punzo: yes

Libs/MRML/Core/vtkMRMLSliceNode.cxx
Punzo: yeah I also noted it. I was pushing a commit, than I got more problem relative the allocation of the matrices

Libs/MRML/Core/vtkMRMLSliceNode.cxx
Punzo: ok, So RotatePlane will work automatically, or I have to check and modify it (If I remove OrietationReference?)

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -371,73 +408,194 @@ void vtkMRMLSliceNode::SetOrientationToReformat()
...
+bool vtkMRMLSliceNode::RenameSliceOrientationPreset(const std::string &name, const std::string &updatedName)
+{
+  if (!name.compare("Reformat"))
+    {
+    vtkErrorMacro("RenameSliceOrientationPreset: Reformat has no orietation preset.");
+    return false;
+    }
+
+  std::map<vtkStdString, vtkMatrix4x4*>::iterator it = this->OrientationMatrices.find(name);
+  if (it != this->OrientationMatrices.end())
```
lassoan: Keep the main success scenario in the main branch of the code and exit if there is error.
In this case, if it == ...end() then exit, if not then execution continues

Libs/MRML/Core/vtkMRMLSliceNode.cxx
Punzo: yes, it is a problem. It is a pointer, if the matric get deleted, it will not work.

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -371,73 +408,194 @@ void vtkMRMLSliceNode::SetOrientationToReformat()
...
+    {
+    std::swap(this->OrientationMatrices[updatedName], it->second);
```
lassoan: if name == updatedName then return

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -371,73 +408,194 @@ void vtkMRMLSliceNode::SetOrientationToReformat()
...
+    this->OrientationMatrices.erase(it);
```
lassoan: I don't understand this swap and erase. Are you sure it's working? Is there a test that checks if it works?

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -371,73 +408,194 @@ void vtkMRMLSliceNode::SetOrientationToReformat()
...
+    }
+
+  vtkErrorMacro("RenameSliceOrientationPreset: the orietation preset " << name << " is not stored.");
+  return false;
+}
+
+//----------------------------------------------------------------------------
+bool vtkMRMLSliceNode::HasSliceOrientationPreset(const std::string &name)
+{
+  if (!name.compare("Reformat"))
```
lassoan: not needed, there is no such preset

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -636,47 +794,17 @@ void vtkMRMLSliceNode::UpdateMatrices()
...
-         this->SliceToRAS->GetElement(0, 2) ==  1.0 &&
-         this->SliceToRAS->GetElement(1, 2) ==  0.0 &&
-         this->SliceToRAS->GetElement(2, 2) ==  0.0 )
+    // It update the Orientation Matrices with the Position coordinates
+    // which are set only on the SliceToRAS matrix.
+    std::map<vtkStdString, vtkMatrix4x4*>::iterator it;
+    for (it = this->OrientationMatrices.begin(); it != this->OrientationMatrices.end(); ++it)
       {
-        orientationString = "Sagittal";
+      it->second->SetElement(0, 3, this->SliceToRAS->GetElement(0, 3));
```
lassoan: Why are we doing something completely different here compared to what was done here before?

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -751,6 +879,26 @@ void vtkMRMLSliceNode::WriteXML(ostream& of, int nIndent)
       }
     }
   of << indent << " sliceToRAS=\"" << ss.str().c_str() << "\"";
+
+  std::map<vtkStdString, vtkMatrix4x4*>::iterator it;
+  for (it = this->OrientationMatrices.begin(); it != this->OrientationMatrices.end(); ++it)
+    {
+    std::stringstream ss;
+    int j;
+    for (i=0; i<4; i++)
```
lassoan: if we only define orientation why we write a 4x4 matrix?

Libs/MRML/Core/vtkMRMLSliceNode.cxx
Punzo: yeah that true, and yes the vector can help. However, in principle if the matrix, instead of adding it, it should be used just the rename method.

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -1135,6 +1303,19 @@ void vtkMRMLSliceNode::Copy(vtkMRMLNode *anode)
   this->SetOrientationString(node->GetOrientationString());
   this->SetOrientationReference(node->GetOrientationReference());
 
+  vtkNew<vtkStringArray> namedOrientations;
+  node->GetSliceOrientationPresetNames(namedOrientations.GetPointer());
+
+  for (int i = 0; i < namedOrientations->GetNumberOfValues(); i++)
+    {
+    if (!namedOrientations->GetValue(i).compare("Reformat"))
```
lassoan: no need

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -1253,6 +1434,13 @@ void vtkMRMLSliceNode::PrintSelf(ostream& os, vtkIndent indent)
   os << indent << "SliceToRAS: \n";
   this->SliceToRAS->PrintSelf(os, indent.GetNextIndent());
 
+  std::map<vtkStdString, vtkMatrix4x4*>::iterator it;
+  for (it = this->OrientationMatrices.begin(); it != this->OrientationMatrices.end(); ++it)
+    {
+    os << indent << "OrientationMatrix"<< it->first <<": \n";
```
lassoan: If you want to serialize like this then you have to make sure to reject names that contain space or special characters. Instead, it would be better to serialize all names and matrices into a single element. Name could be encoded/decoded to handle special characters (URLEncodeString/URLDecodeString).

Libs/MRML/Core/vtkMRMLSliceNode.cxx
Punzo: I mean another thing that I can do instead to use the vector, it is to eliminate the old preset in the map if the new preset has the same matrix. Will it be acceptable?

Libs/MRML/Core/vtkMRMLSliceNode.h
```diff
@@ -418,6 +440,11 @@ class VTK_MRML_EXPORT vtkMRMLSliceNode : public vtkMRMLAbstractViewNode
   vtkMatrix4x4 *UVWToSlice;
   vtkMatrix4x4 *UVWToRAS;
 
+  std::map<vtkStdString, vtkMatrix4x4*> OrientationMatrices;
+
+  vtkSmartPointer<vtkMatrix4x4> axialSliceToRAS;
```
lassoan: There is no need for these member variables. You just need to create matrices temporarily and set it in a preset,

Libs/MRML/Core/vtkMRMLSliceNode.h
```diff
@@ -418,6 +440,11 @@ class VTK_MRML_EXPORT vtkMRMLSliceNode : public vtkMRMLAbstractViewNode
   vtkMatrix4x4 *UVWToSlice;
   vtkMatrix4x4 *UVWToRAS;
 
+  std::map<vtkStdString, vtkMatrix4x4*> OrientationMatrices;
```
lassoan: Very important!! vtkMatrix4x4 should be replaced by vtkSmartPointer<vtkMatrix4x4> otherwise you may leak memory and/or crash. We have to make a copy of the matrix that the caller provides and store that.

lassoan: if you use std::map then use a std::string instead of a vtkStdString (unless you have a good reason to do so)

Libs/MRML/Core/vtkMRMLSliceNode.cxx
lassoan: You (or I) need to modify the rotate to volume plane. It would be great if you could give it a go and let me know if you need help or don't have time.

Libs/MRML/Core/vtkMRMLSliceNode.cxx
lassoan: I think it's actually good to let multiple names for the same matrix. E.g., there could be user-defined orientations that may sometimes (even temporarily) coincide with existing standard orientations. We don't want to wipe the standard orientations and don't want our user-defined orientations to fail either. It would be just nice if we could prioritize them.

Vector instead of a map could work, as we only have a few presets. Actually, a vector may even be faster for 3-4 elements.

Punzo: ok

Libs/MRML/Core/vtkMRMLSliceNode.cxx
Punzo: ok

Libs/MRML/Core/vtkMRMLSliceNode.cxx
Punzo: Here I just add, to the presets also the "Origin" information stored in the SliceToRAS. 
This because, when SetOrietation is called, the SliceToRAS matrix is override by the present (all the 4x4 elements)

Libs/MRML/Core/vtkMRMLSliceNode.cxx
Punzo: see up. Of course the behaviour can be changed, and I can store MAtrix3x3 for the orietation and then make the copy to SliceToRAS manually element by element, instead to use DeepCopy

Libs/MRML/Core/vtkMRMLSliceNode.h
Punzo: ok

Libs/MRML/Core/vtkMRMLSliceNode.cxx
Punzo: yes it is working, because the values of the map are all pointers. However, we decide to change the map to vector, right?

Punzo:
@lassoan thank you again for all your comments. A question: if I change from the map to the vector to store the matrices, still I need to store somewhere the names. Shall I so two vectors?

Libs/MRML/Core/vtkMRMLSliceNode.cxx
lassoan: Yes probably a vector would work better

Libs/MRML/Core/vtkMRMLSliceNode.cxx
lassoan: Maybe we should store vtkMatrix3x3 to reduce chance of confusion (we don't actually use the fourth row and column).

lassoan:
Put a std::string and vtkSmartPointer<vtkMatrix3x3> into a struct or class. You can even create the matrix in the constructor to make things simple. The preset list will be a std::vector of this class.

Punzo:
ok, thanks, perfect.

Libs/MRML/Core/vtkMRMLSliceNode.cxx
Punzo: ok, no we don't use them. I'll make a method or something like thaht then to copy the Matrix3x3 of the preset into the SliceToRAS matrix4x4

Libs/MRML/Core/vtkMRMLSliceNode.cxx
lassoan: You can add a method to the name-matrix struct (or class) that sets the vtkMatrix3x3 content based on a vtkMatrix4x4 input.

Punzo:
@jcfr  setting the defult Node in vtkSlicerApplicationLogic::SetMRMLSceneDataIO works!!!!

Libs/MRML/Core/vtkMRMLSliceNode.cxx
jcfr: > (especially because we store presets in a map and maps are not sorted)

Looking at the documentation, map are ordered.
See http://www.cplusplus.com/reference/map/map/

Would it be enough to iterate from the end in `vtkMRMLSliceNode::GetSliceOrientationPresetName` ?

lassoan: Yes, they are usually kind of alphabetically sorted, but we don't want that. We need to be able to specify a sort order (for example, by the order of insertion). That would allow us to specify order like "Axial, Sagittal, Coronal" followed by custom orientations.

jcfr: Good point. Not sure what I was thinking.

>  specify a sort order

Indeed, that is critical.

a `std::vector` containing `std::pair` would work then.

For reference:
* http://www.cplusplus.com/reference/utility/make_pair/

Punzo: ok perfect, tomorrow I'll submit a commit with all the modification. Thanks.

Punzo:
Hi @jcf @lassoan,
I pushed the modification.
I couple of notes:
1) RotateVolumeToPlane has not been modified (there is still the OrientationReference attribute). 
   @jcfr proposed to do it in another PR. In pirinciple, should not be too difficult:
   we should add a method such NearstOrietationPreset(vtkMatrix4x4*),
   which check the input matrix with all the presets matrices
2) I leaft the orietation matrices presets as vtkMatrix4x4 (instead of Matrix3x3) because we always
    compare sliceToRAS with the presets which is a  vtkMatrix4x4. Of course some method to "take"
   the "orietation" matrix from sliceToRAS may be implemented, but to me seems only adding         
   complexity. 

Please let me know your feedback

Punzo:
P.S.: One way, to fix the RotateVolumeToPlane method may be the use of cosine similarity beetween vector

Base/Logic/vtkSlicerApplicationLogic.cxx
```diff
@@ -296,6 +298,60 @@ void vtkSlicerApplicationLogic::SetMRMLSceneDataIO(vtkMRMLScene* newMRMLScene,
...
+  coronalSliceToRAS->SetElement(2, 0,  0.0);
+  coronalSliceToRAS->SetElement(0, 1,  0.0);
+  coronalSliceToRAS->SetElement(1, 1,  0.0);
+  coronalSliceToRAS->SetElement(2, 1,  1.0);
+  coronalSliceToRAS->SetElement(0, 2,  0.0);
+  coronalSliceToRAS->SetElement(1, 2,  1.0);
+  coronalSliceToRAS->SetElement(2, 2,  0.0);
+
+  // Setting a Slice Default Node
+  vtkMRMLSliceNode* defaultNode = vtkMRMLSliceNode::SafeDownCast
```
jcfr:
```
vtkSmartPointer<vtkMRMLSliceNode*> defaultNode = vtkMRMLSliceNode::SafeDownCast( ...
```

Base/Logic/vtkSlicerApplicationLogic.cxx
```diff
@@ -296,6 +298,60 @@ void vtkSlicerApplicationLogic::SetMRMLSceneDataIO(vtkMRMLScene* newMRMLScene,
...
+  else
+    {
+    defaultNode = vtkMRMLSliceNode::SafeDownCast
+        (newMRMLScene->CreateNodeByClass("vtkMRMLSliceNode"));
+    defaultNode->AddSliceOrientationPreset("Axial", axialSliceToRAS.GetPointer());
+    defaultNode->AddSliceOrientationPreset("Sagittal", sagittalSliceToRAS.GetPointer());
+    defaultNode->AddSliceOrientationPreset("Coronal", coronalSliceToRAS.GetPointer());
+    newMRMLScene->AddDefaultNode(defaultNode);
+    defaultNode->Delete(); // scene owns it now
+    }
```
jcfr: Nitpick: Would suggest the following form:
```
vtkSmartPointer<vtkMRMLNode> defaultNode = newMRMLScene->GetDefaultNodeByClass("vtkMRMLSliceNode");
if (!defaultNode)
  {
  defaultNode.Take(newMRMLScene->CreateNodeByClass("vtkMRMLSliceNode"));
  newMRMLScene->AddDefaultNode(defaultNode);
  }
vtkMRMLsliceNode * defaultSliceNode = vtkMRMLSliceNode::SafeDownCast(defaultNode);
defaultNode->AddSliceOrientationPreset("Axial", axialSliceToRAS.GetPointer());
defaultNode->AddSliceOrientationPreset("Sagittal", sagittalSliceToRAS.GetPointer());
defaultNode->AddSliceOrientationPreset("Coronal", coronalSliceToRAS.GetPointer());
```

Punzo: ok, thanks

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -113,7 +116,17 @@ vtkMRMLSliceNode::vtkMRMLSliceNode()
...
-  this->SetOrientationToAxial();
+  this->SliceToRAS->SetElement(0, 0, -1.0);
+  this->SliceToRAS->SetElement(1, 0,  0.0);
+  this->SliceToRAS->SetElement(2, 0,  0.0);
+  this->SliceToRAS->SetElement(0, 1,  0.0);
+  this->SliceToRAS->SetElement(1, 1,  1.0);
+  this->SliceToRAS->SetElement(2, 1,  0.0);
+  this->SliceToRAS->SetElement(0, 2,  0.0);
+  this->SliceToRAS->SetElement(1, 2,  0.0);
+  this->SliceToRAS->SetElement(2, 2,  1.0);
```
jcfr: May we could initialize to `-1` and the the OrientationString to "Unknown" ?

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -333,111 +326,229 @@ ::IsDisplayableInThreeDView(const char* viewNodeID)const
...
     {
-    return;
+    return false;
     }
- if (!strcmp(orientation, "Axial"))
+
+  vtkMatrix4x4 *orientationPreset;
+  orientationPreset = this->GetSliceOrientationPreset(orientation);
+
+  if(!strcmp(orientation, "Reformat"))
```
jcfr: `if (`

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -333,111 +326,229 @@ ::IsDisplayableInThreeDView(const char* viewNodeID)const
...
     {
-    this->SetOrientationToAxial();
+    this->SetOrientationString(orientation);
     }
-  else if (!strcmp(orientation, "Sagittal"))
+
+  if(orientationPreset == NULL)
```
jcfr: `if (`

Libs/MRML/Core/vtkMRMLSliceNode.cxx
Punzo: do you mean leaving this->SliceToRAS == Identity?

jcfr: Yes. 

The issue is that having the Axial matrix set to in two places is not ideal. That is why setting to Identity here and relaying on the presets make more sense

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -333,111 +326,229 @@ ::IsDisplayableInThreeDView(const char* viewNodeID)const
...
+    {
+    return false;
+    }
+
+  //SetOrientationString() and SetOrientationReference()
+  //behaviour will be mantained for backward compatibility
```
jcfr: Do you mean that the method could be completely removed from the class ? Or do you mean calling these method below is only done to maintain backward compatibility ?

Base/Logic/vtkSlicerApplicationLogic.cxx
Punzo: changing in taht way, it gives this error in compilation. I don't know why

vtkMRMLScene (0x2c0f1e0): vtkMRMLScene::AddDefaultNode failed: invalid node

Libs/MRML/Core/vtkMRMLSliceNode.cxx
Punzo: yeah in fact my extesion that it is annoying I'll try

Libs/MRML/Core/vtkMRMLSliceNode.cxx
jcfr: What do you thnk of:

```
// Since the method `GetOrientation/SetOrientation` are sufficient to manage the orientation, the 
// following calls to `SetOrientationString` and `SetOrientationReference` are only done to
// maintain backward compatibility.
```

Punzo: sorry copy and paste typo. I meant only SetOrientationReference()

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -333,111 +326,229 @@ ::IsDisplayableInThreeDView(const char* viewNodeID)const
...
+  return true;
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLSliceNode::SetOrientationToReformat()
+{
+    // Don't need to do anything.  Leave the matrices where they were
+    // so the reformat starts where you were.
+
+    this->SetOrientationString( "Reformat" );
```
jcfr: If `SetOrientationString` is deprecated .. should we do something else here ?

Libs/MRML/Core/vtkMRMLSliceNode.cxx
Punzo: I think 

//SetOrientationReference() behaviour will be mantained for backward compatibility

is fine. SetOrientation is the method to set the orietation

Libs/MRML/Core/vtkMRMLSliceNode.cxx
Punzo: I think SetOrietationString is not really deprecated. It is a string that tell us which preset is using at the moment the matrix sliceToRAS

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -333,111 +326,229 @@ ::IsDisplayableInThreeDView(const char* viewNodeID)const
...
-    this->SliceToRAS->SetElement(0, 1,  0.0);
-    this->SliceToRAS->SetElement(1, 1,  1.0);
-    this->SliceToRAS->SetElement(2, 1,  0.0);
-    // Pz -> Patient Inferior
-    this->SliceToRAS->SetElement(0, 2,  0.0);
-    this->SliceToRAS->SetElement(1, 2,  0.0);
-    this->SliceToRAS->SetElement(2, 2,  1.0);
+  if (!name.compare("Reformat"))
+    {
+    vtkWarningMacro("AddSliceOrientationPreset: Reformat refer to any arbitrary orietation. "
```
jcfr: `orietation` -> `orientation`

Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -333,111 +326,229 @@ ::IsDisplayableInThreeDView(const char* viewNodeID)const
...
 
-    this->SetOrientationString( "Axial" );
-    this->SetOrientationReference( "Axial" );
-    this->UpdateMatrices();
+  std::vector< std::pair <std::string, vtkSmartPointer<vtkMatrix4x4> > >::iterator it;
+  for (it = this->OrientationMatrices.begin(); it != this->OrientationMatrices.end(); ++it)
+    {
+    if (!it->first.compare(name))
+      {
+      vtkDebugMacro("AddSliceOrientationPreset: the orietation preset " << name << " is already stored.");
```
jcfr: `orietation` -> `orientation`

Libs/MRML/Logic/vtkMRMLSliceLayerLogic.cxx
```diff
@@ -55,7 +55,7 @@ vtkStandardNewMacro(vtkMRMLSliceLayerLogic);
 
 bool AreMatricesEqual(const vtkMatrix4x4* first, const vtkMatrix4x4* second)
 {
-  return vtkAddonMathUtilities::Matrix4x4AreEqual(first, second, /* tolerance= */ 0);
+  return vtkAddonMathUtilities::Matrix4x4AreEqual(first, second);
```
jcfr: This is a change in behavior, it it okay to have a tolerance of 10e-3

Libs/MRML/Core/vtkMRMLSliceNode.cxx
Punzo: no you are right to change orietation SetOrientation must be called, but I guess it was the same also before, right?

Libs/MRML/Widgets/Resources/UI/qMRMLSliceControllerWidget.ui
```diff
@@ -4,17 +4,26 @@
  <widget class="ctkPopupWidget" name="qMRMLSliceControllerWidget">
```
jcfr: Look like the changes in this file could be discarded

Libs/MRML/Core/vtkMRMLSliceNode.cxx
Punzo: ok, sorry corrected in ALL the vtkError and Warning

Libs/MRML/Logic/vtkMRMLSliceLayerLogic.cxx
Punzo: It is needed, otherwise it'll not work. I tested both

Libs/MRML/Widgets/Resources/UI/qMRMLSliceControllerWidget.ui
Punzo: sorry I wanted only to remove Reformat from the ctkcombox

Libs/MRML/Core/vtkMRMLSliceNode.cxx
Punzo: SliceToRAS == Identity works and I set up the  OrientationString to "Reformat" (so sliceController will not complain

Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx
```diff
@@ -781,19 +782,59 @@ void qMRMLSliceControllerWidgetPrivate::updateWidgetFromMRMLSliceNode()
   // Update abbreviated slice view name
   this->ViewLabel->setText(this->MRMLSliceNode->GetLayoutLabel());
 
+  vtkNew<vtkStringArray> orientationNames;
+  this->MRMLSliceNode->GetSliceOrientationPresetNames(orientationNames.GetPointer());
+
+  wasBlocked = this->SliceOrientationSelector->blockSignals(true);
+  int count = this->SliceOrientationSelector->count();
+  if(this->SliceOrientationSelector->itemText(count - 1).compare("Reformat") &&
```
jcfr: `this->SliceOrientationSelector->itemText(count - 1).compare("Reformat")` -> `this->SliceOrientationSelector->itemText(count - 1) != "Reformat"`

Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx
```diff
@@ -781,19 +782,59 @@ void qMRMLSliceControllerWidgetPrivate::updateWidgetFromMRMLSliceNode()
   // Update abbreviated slice view name
   this->ViewLabel->setText(this->MRMLSliceNode->GetLayoutLabel());
 
+  vtkNew<vtkStringArray> orientationNames;
+  this->MRMLSliceNode->GetSliceOrientationPresetNames(orientationNames.GetPointer());
+
+  wasBlocked = this->SliceOrientationSelector->blockSignals(true);
+  int count = this->SliceOrientationSelector->count();
```
jcfr:
```
// If last item in the combox box is not "Reformat" and if the current orientation is not "Reformat", remove it ... 
```

Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx
```diff
@@ -781,19 +782,59 @@ void qMRMLSliceControllerWidgetPrivate::updateWidgetFromMRMLSliceNode()
...
+  if(this->SliceOrientationSelector->itemText(count - 1).compare("Reformat") &&
+     strcmp(this->MRMLSliceNode->GetOrientationString(), "Reformat"))
+    {
+    this->SliceOrientationSelector->removeItem(count - 1);
+    count--;
+    }
```
jcfr: It would be good to explain the role of these looks ..

Punzo: essentially "Reformat" is added as last orietation in the comboBox if the sliceToRAS matrix is not a preset

Libs/MRML/Widgets/Resources/UI/qMRMLSliceControllerWidget.ui
Punzo: However, It was not needed to remove the Reformat one in the ctkCombobox. It is automatically updated updateWidgetFromMRMLSliceNode. I will restore the original file

Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx
jcfr: Wouldn't make sense to always have `Reformat` as the last item independently of the current `sliceToRAS` ?

Punzo: the problem is that is you click on reformat, it will fire setOrientation("Reformat")
which will not change the sliceToRAS matrix, and therefore the ctkCombobox selection it will be automatically ressetted to the last orietationpreset

Libs/MRML/Logic/vtkMRMLSliceLayerLogic.cxx
jcfr: I wonder which change in this patch made this a requirement

Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx
Punzo: On the other hand, if the sliceToRAS is not one of the preset, now it will display Reformat on the ctkCombobox

Libs/MRML/Logic/vtkMRMLSliceLayerLogic.cxx
Punzo: Not sure, but before the float of the matrix where compared with -1.0, 0.0 and 1.0. I guess when comparing the elements of the two matricies there are some arrotondation errors of the order of the precision of the float.

Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx
jcfr: Adding comment explaining what is done will below will be sufficient then

Punzo: ok, I will

Punzo:
@jcfr corrections applied and pushed. The error taht we saw using the reformat module was not a bug due to this PR. When using the widget relative to normal it may happen that the matrix is not weel aligned. Example : 

a)I rotate the slice in 3-D randomely
b) I push normalZ
c) in the combobox it shows Reformat instead of axial.

However, the comboBox is right because the sliceToRAS matrix is this one:
   -0.998356 0.0573213 5.82867e-16 -2.79519 
    0.0573213 0.998356 9.99201e-16 6.4286 
    6.38378e-16 -1.22125e-15 1 -10.2143 
    0 0 0 1 

and instead should be 
    -1 0 0 -2.79519 
    0 1 0 6.4286 
    0 0 1 -10.2143 
    0 0 0 1 

rotating in 3-D more the slice and trying to rest with normal make the matrix even more different (at a certain point it is even visible by eye that it is not the axial orientation).

I tested also by python script and the infrastructure introduced in this PR worked.

Punzo:
@jcfr  squash done.

Base/Logic/vtkSlicerApplicationLogic.cxx
jcfr: FYI: `TakeReference` had to be used.

jcfr:
Closing. This PR is superseded by #522

PR #508, opened 2016-05-31
PERF: Improve startup time by ~25% delaying import of SimpleITK

jcfr:
Startup time reduced from  5.1s to 3.8s (cold cache)

Stats obtained after averaging values obtained after 5 runs clearing
the file cache between runs. The following command was used:

```
~/Projects/Slicer/Applications/SlicerApp/Testing/Python/MeasureStartupTimes.py --drop-cache -n5 --normal ./Slicer
```

With a warm cache, it improve the startup time by ~32% (4s -> 2.7s)

Results have been gathered on Ubuntu 15.10 on a workstation with the
following specs: 64GB / M.2 PCIe NVMe SSD / Quad Core 3.80GHz

PR #509, opened 2016-05-31
[PREVIEW-WIP] PERF: Improve startup time by 8% with lazy loading of wrapped libraries

jcfr:
*** WORK IN PROGRESS: For now, you have to make sure SlicerApp-real and
Slicer launcher are built to ensure the successful generation of the json
files ***

Startup time reduced from 3.8s to 3.5s with a "cold cache"
and from 2.7s to 2.38s with a "warm cache".

For each logic/mrml/dm/widgets python modules, a json files listing
the associated attributes is generated. Then, when the application is
initialized, the "slicer" module is created as a "lazy" module with
the attributes associated with logic/mrml/dm/widgets set as "not loaded".

Finally, as soon as an attribute not yet loaded is accessed, the specialized
__getattribute__ load the associated python module and update the module
dictionary.

The "lazy" module has been adapted from "itkLazy.py"

Results have been gathered on Ubuntu 15.10 on a workstation with the
following specs: 64GB / M.2 PCIe NVMe SSD / Quad Core 3.80GHz


jcfr:
Note: Timing performance obtained after applying the fixes associated with #508

pieper:
Today on the tcon it would help if we could discuss how to run these
performance tests on other platforms.  Slicer on macs has always been
terribly slow to start so I suspect the tune ups will have even bigger
impact there.  Also the Instruments tool on mac is good for finding hot
spots in the code.

PR #510, opened 2016-05-31
ENH: Add "--hide-application-information" option

jcfr:


jcfr:
Cc: @Slicer/slicer-core

naucoin:
I'd suggest using information singular in all cases.

jcfr:
> information rather than informations in all cases

Thanks for the review, I just updated the option to be `--hide-application-information`. Let me know if that is what you were thinking.

Later today, I will also add a test.

lassoan:
Operating system, 32/64 bit, RAM, etc. information is essential for any kind of problem investigation, so it should always be written to the log file. On the other hand, this information is probably doesn't need to be printed on the console at all.

So, I would change the behavior to always write to file, don't write to console by default, and maybe add a switch to enable writing it to console.

I think the switch is not needed: we may just disable writing of system info on the console permanently. Writing it to the console is not really needed because although it may be useful for automatic tests, system information may be obtained by adding a test that specifically prints this info. It would simplify things if we didn't add yet another startup switch.

naucoin:
@jcfr yes, that's what I meant, thanks.

lassoan:
OK. Then the name of the switch should be --show-application-information (as the option is disabled by default).

jcfr:
Thanks for the comments.

Updated behavior (logged to file but not to console), along with new option `--application-info` and tests have been integrated in [25149](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25149)

PR #511, opened 2016-05-31
BUG: Ensure extension python wrapped modules can be loaded

jcfr:
This commit fixes a regression introduced in r25128 (STYLE: Simplify
import of Logic/MRML/DisplayableManager python modules)

Reported-by: Andras Lasso <lasso@queensu.ca>


jcfr:
Cc: @lassoan

jcfr:
Since CircleCI is green and it fixes the issue after building and installing locally an extension contain a loadable module (generated using the wizard), I am integrating.

Integrated in [25137](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25137)

PR #516, opened 2016-06-01
ENH: New test to verify the Volumes module works well with NaN in Volumes

sumedhasingla:
This fix is to verify the bug #0003851.
http://www.na-mic.org/Bug/view.php?id=3851


jcfr:
Thanks for the PR :+1: 

I will re-organize the current test data and will then apply your patch.

jcfr:
Integrated in [r26195](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26195)

PR #517, opened 2016-06-01
BUG: Remove invalid assert while populating the scene

sankhesh:
The assert was checking whether the index of the current node
matches the row count. However, while inserting nodes, the
logic adds a parent node if there isn't one. This insertion of
parent node increments the RowCache size. Next time the assert
was encountered, it would fail.


jcfr:
Patch integrated in r25143
See http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25143

Test added in r25144
See http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25144

PR #520, opened 2016-06-02
STYLE: Improve documentation of GetFirstNode()

fbudin69500:
GetFirstNode third parameter "byHideFromEditor" was not explained
in the doxygen documentation. It did not seem intuitive to have
a parameter that is a pointer when one might expect to specify a
boolean. The new doxygen string explains that this is a 3-state
variable that is either disregarded if the pointer is set to 0,
or the function only returns a node that matches the given value
(hidden from editor or not).


jcfr:
Thanks for the patch.

Integrated in [r25151](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25151)

PR #521, opened 2016-06-03
COMP: virtualize SaveArchetypeVolume method in VolumeLogic

Punzo:
@jcfr


jcfr:
Note: I manually cancelled the build. Since the Docker image used in the CI is not up-to-date .. it would have fail.

Docker images is being rebuilt and will be pushed to docker up as soon as I have access to a fast connection.

For reference: https://github.com/thewtex/SlicerDocker

naucoin:
@Punzo could you please update the commit message to include why you're making this change?

Punzo:
@naucoin sorry I'll do

Punzo:
sorry Actually this PR was not needed. I just realized that after the change in my Extension suggested by @jcfr  
https://github.com/Punzo/SlicerAstro/commit/18be7a17b3e32dee89826d70a0ebc1ff6fd724d3
(making my AstroVolumeLogic child of VolmesLogic) I don't need anymore SaveArchetypeVolume method in my Logic. Thanks!

PR #523, opened 2016-06-03
ENH: Update VTK to handle default values for array parameters in python

fbudin69500:
Without this commit, one must specify parameters like (,int array[3]=0) when
the method is called from python.

$ git shortlog 1c30cb0..4b9957b --no-merges
David Gobbi (2):
      Fix wrapping pointer parameters with default value 0.
      Handle default values for array parameters in python.


jcfr:
Thanks for the PR.

jcfr:
Will test locally and integrate.

jcfr:
Integrated in [25160](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25160)

PR #525, opened 2016-06-08
ENH: Allow modules to detect when the user clicks on a markup fiducial

lassoan:
*** Do not integrate until https://github.com/lassoan/VTK/tree/seedwidget_point_select is integrated into Slicer's VTK ***

Many times a module need to know when the user selects an existing markup. Often the markup position has to be locked but we still want to be able to detect when the user clicked on it.

VTK allowed locking a seed but then no interaction is possible. VTK handle and seed widgets were improved (see https://github.com/lassoan/VTK/tree/seedwidget_point_select) to allow locking of position but still able to select.

Slicer markups module behavior was changed so that if a markup point is locked then it can still be selected (but not moved). If the entire markup node is locked then markup points cannot be selected or moved (same behavior as before).

How to test the new event:

---

@vtk.calldata_type(vtk.VTK_INT)
def markupCallback(caller, eventId, callData):
  print("PointStartInteractionEvent: {0}".format(callData))

markupsNode = getNode('F')
observerTag = markupsNode.AddObserver(slicer.vtkMRMLMarkupsNode.PointStartInteractionEvent, markupCallback)

---

Markups module may be improved in the future to use this new event, for example select the markup in the markup list if it is selected in a 2D or 3D view. Also, selection state of markups could be changed when the user clicks on a markup (there could be single-select and multi-select modes).

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.h
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.h
@@ -137,6 +137,7 @@ public:
     LockModifiedEvent = 19000,
     LabelFormatModifiedEvent,
     PointModifiedEvent,
+    PointStartInteractionEvent,
     PointEndInteractionEvent,
     NthMarkupModifiedEvent,
     MarkupAddedEvent,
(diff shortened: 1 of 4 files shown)


lassoan:
@naucoin

naucoin:
I left some comments on the change to Slicer VTK. I think we need to distinguish between the at least two types of interactions that trigger the start interaction event (mouse is near the seed, mouse is clicked and moving the seed).

jcfr:
Cc: @vovythevov @msmolens

naucoin:
@pieper  do the proposed changes work with Landmark Registration?

pieper:
I haven't tried but I don't see why they wouldn't work.

lassoan:
integrated:
https://github.com/Slicer/Slicer/commit/a2b893bc93c5221d5bb091c078f0c7aad8fcf8fc

PR #527, opened 2016-06-08
ENH: allow custom DICOM volume naming for Scalar Volumes

jonieva:
Allow to specify the template that will be used to name a DICOM scalar
volume, allowing to use any tag from the DICOM Browser


pieper:
I took a quick look and it seems good (I didn't have time to test)

jcfr:
Thanks for the patch.

Bonus: Could you add a basic SelfTest to exercise the new features ? 

I would like to start having more test checking the DICOM module work as expected.

jonieva:
Hi JC.

Do we have any example of using DICOM sample datasets to create the test?

Thank you.

Jorge Onieva

cpinter:
Hi Jorge,

I made a really small (~30KB) DICOM study specifically for testing.
http://slicer.kitware.com/midas3/download/folder/2822/TinyRtStudy.zip
It contains a dose and a structure set too, but you can remove those and make a new test set from only the CT if you want.

Downloading and unzipping files within self tests can be done, here’s an example
https://github.com/SlicerRt/SlicerRT/blob/master/DicomRtImportExport/Testing/Python/DicomRtImportSelfTest.py

I hope this helps,
csaba

jcfr:
@cpinter Thanks for the reference :+1: 

@jonieva Thanks for going the extra mile adding test :smile:

Modules/Scripted/DICOMLib/DICOMPlugin.py
```diff
@@ -140,19 +156,48 @@ def export(self,exportable):
...
-    if seriesNumber != "":
-      name = seriesNumber + ": " + name
+    on the series level data in the database.
+
+    Arguments
+
+      files: files that will be used to build the volume name using the DICOM tags in a list of DICOM files
+    """
+    if not files or len(files) == 0:
+      return "UNNAMED VOLUME"
```
cpinter: Is this change from camel case to all upper case intentional?

jonieva: Yap. The idea is to stress the cases in which the volume naming is not correct for any reason

fedorov:
I want to review and test this as well, but few urgent things I need to take care of today. I will get to this a bit later.

Modules/Scripted/DICOMLib/DICOMPlugin.py
cpinter: In sample data repositories (i.e. data we work with on a daily basis) it is a quite common scenario. that series description etc is empty.

Also I'm pretty sure changing the default name would break tests. Unless it is very important for you to make this change, I'd suggest leaving the default as it was.

jonieva: Makes sense.

Will do.

Tx!

jonieva:
Related to Csaba comment, I was taking a deeper look. The default name that was being used so far is "Unnamed Series". I don't have strong feelings about uppercase/lowercase, but with this new naming schema we shouldn't include the word "Series" anyway.
What happens right now is that any DICOM field that is not found is named as "UNKNOWN", and this is generic for any field. 
So I would argue kind of the other way around. If a test fails because of this "Unnamed Series" thing, I think it's actually ok that it fails, and it should be fixed to be adapted to the new schema.
If you think this is going to cause too much trouble, I could distinguish this special case when there is no template, and return the same "Unnamed Series", but honestly I don't think it's a very elegant thing to do.
Thoughts?

cpinter:
Agreed, without the word Series the tests will fail anyway, and they shouldn't be too hard to fix.

I'd put my vote against all uppercase though, I just don't find it pleasant when I'm being yelled at by the test data :) Not too big a deal though in any case.

lassoan:
Yes, I agree with Csaba. There should be no yelling (all-uppercase) in any text that is displayed to user.

lassoan:
Using a static template is way too much limiting. Determining a nice and informative name from DICOM tags usually involves a very complex logic (see the current simple rule of generating node name in the scalar volume plugin, or see a more complex logic in the CTK DICOM browser). The templated method cannot even handle fields in private modules or in sequences. It's so rigid that I don't think it would fulfill the needs of even a single user who works on data that comes from different scanners or hospitals.

We must allow plugins to define names using complex rules that take into account which tags are defined, what is the imaging modality, etc. The existing method of allowing plugins to define a logic of any complexity to generate node names correspond to this requirement.

The question is how to make this logic more customizable without dumbing it down.

What kind of customization would you need?
Could you give examples of how you would construct the node name for different DICOM objects?

jonieva:
I removed the "yelling" (using "Unknown" for incomplete fields).
I also added a self test, but I'm missing something because apparently the slicer.dicomDatabase object has not been created when the test is executed. Any help here would be appreciated.
Andras, regarding our needs, the proposed solution would fit our needs. Of course there is always room to improve, but I would argue that this solution is not "harming" and it's much much less limiting than the current scheme.

lassoan:
The solution in the pull request is an improvement in the sense that you can use any fields but a a step back in the sense that you can only use dumb rigid structure. Now if series name or number is not defined you end up with ugly "Unknown" strings and separators.

In general, DICOM standard is not exact enough so that you can blindly link any field to any meaning in general. A solution may be to keep a logic that extracts more abstract fields, such as "id" and "description" and then allow the user to use these fields instead of raw DICOM fields. You would still need a solution for the separator problem (if a field is missing you may either want to show a default value or don't show the value and corresponding separator at all).

lassoan:
Please give specific examples about how would you like to construct your node names.

Modules/Scripted/DICOMLib/Testing/DICOMBrowserTest.py
```diff
@@ -0,0 +1,120 @@
...
+class DICOMBrowserTest:
+  """
+  This class is the 'hook' for slicer to detect and recognize the test
+  as a loadable scripted module (with a hidden interface)
+  """
+  def __init__(self, parent):
+    parent.title = "DICOMBrowserTest"
+    parent.categories = ["Testing.TestCases"]
+    parent.contributors = ["Jorge Onieva (BWH)"]
+    parent.helpText = """Self tests for the DICOM Browser. So far, no GUI is tested"""
```
jcfr: Adding:
```
parent.dependencies = ['DICOM']
```
will ensure `slicer.dicomDatabase` is set.

See here for more details: https://github.com/Slicer/Slicer/blob/acc9915ce733f2ad8394d9566a73580ad3327cb7/Modules/Scripted/DICOM/DICOM.py#L42-L65

jcfr:
> `slicer.dicomDatabase` object has not been created when the test is executed.

Adding dependencies on `DICOM` module should help. See https://github.com/Slicer/Slicer/pull/527/files#r66526885

The test should also be updated to do something like this too: https://github.com/Slicer/Slicer/blob/acc9915ce733f2ad8394d9566a73580ad3327cb7/Applications/SlicerApp/Testing/Python/RSNAVisTutorial.py#L345-L356

This will ensure a temp database is created.

Existing limitation: Doing so will cause any existing database created by a prior test to be used. Moving forward, the idea is to have (by default) every DICOM tests create their own database, and (by default) remove it when test is successful. This part could be done later at a later time.

Modules/Scripted/DICOMLib/DICOMPlugin.py
```diff
@@ -140,19 +156,48 @@ def export(self,exportable):
...
+        fixedTag = dicomTagCode[0].lower() + dicomTagCode[1:]
+        # Remove the possible parenthesis
+        if fixedTag.startswith("(") and fixedTag.endswith(")"):
+          fixedTag = fixedTag[1:-1]
+        if self.tags.has_key(fixedTag):
+          # Instead of a code, the tag is one of the common well known DICOM tags (ex: seriesNumber)
+          fixedTag = self.tags[fixedTag]
+
+        # Search for the real value in the first image of the series
+        codeValue = slicer.dicomDatabase.fileValue(files[0], fixedTag)
```
fedorov: Note that values can contain special characters that can cause problems. There are 2 places where filtering for special characters is done: 

1. https://github.com/pieper/dicomsort/blob/master/dicomsort.py#L99-L108
2. (simpler) https://github.com/naucoin/CTK/blob/86837e6a68a0518f045be0ec93006af5500140f9/Libs/DICOM/Widgets/ctkDICOMBrowser.cpp#L957-L961

See related discussion in http://www.na-mic.org/Bug/view.php?id=3163

jonieva:
@jcfr
JC: thanks for the tip. Steve also pointed me to RSNAVisTutorial. Just adding the dependency and what it's done there, it seems to work.
However, I think there may be a bug in that test, as dicomWidget is not initialized? See
https://github.com/Slicer/Slicer/blob/acc9915ce733f2ad8394d9566a73580ad3327cb7/Applications/SlicerApp/Testing/Python/RSNAVisTutorial.py#L356

@lassoan 
Andras,

We have some collaborators that are going to do a study using our tools. I don't know exactly the tags that they will want to use, but I'm pretty confident that just this extra functionality will work for them (I already showed them the functionality and they are very pleased with it). 
This obviously won't cover any real clinical scenario, but it will cover a big part of clinical studies scenarios where the data are more controlled. I agree it can be annoying to end up with an "Unknown" label in some cases, but it's much worst to have 0 flexibility as it is the case now.

@fedorov 
Andrey,
Tx for your comment. I tested different scenarios and so far the only one that seems to be making trouble in the regular expression is the '@' symbol. I will replace that symbol with "_" as it's suggested, but I think that the rest of the symbols should work fine.

lassoan:
Also note some fields require pre-processing to convert them to human-readable representation. For example, patient name may have 5 components, any of them can be empty, and they are separated by ^ characters (http://dicom.nema.org/dicom/2013/output/chtml/part05/sect_6.2.html). If you want to include patient name in the node name then you have to decode it. It is doable in Python but using a dumb string replacement you would have "Morrison-Jones^Susan" instead of Susan Morrison-Jones or Morrison-Jones, Susan.
This is just one more example that dumb string pattern replacement does not work for DICOM and would be a step back compared to the intelligent construction of fields from tags.

lassoan:
I agree that the current method should be more customizable so that custom DICOM fields can be used for generating node names. However, we cannot afford to lose having a custom logic for constructing correct string values from the custom DICOM fields.
Ideally, the same method should be used for generating the fields that are shown in the DICOM browser, so it would be nice to find a solution that can be used there, too.

lassoan:
If you need a quick-and-dirty solution then feel free to add a custom scalar volume DICOM plugin that generates node names the way you want. If this plugin gives higher confidence value for scalar volumes than the default plugin then your plugin will be used by default. Inside your plugin you can use the default scalar volume DICOM plugin for doing everything else than the node name generation.

If you have time to implement a proper solution (allow customization but also allow constructing the name using custom logic, work safely with special characters, etc.) then it would be even better. We would be happy to help you with the design and implementation as needed and then integrate that into the Slicer core.

fedorov:
I would think that Andras' concern could partially be addressed by making sure that if the user is running the updated code without knowing anything about the new feature Jorge added, the behavior should be identical to the one we currently have. This should be the starting point for the implementation.

Having tested the code, (and I know Jorge you are probably enough discouraged already) but I have to say I am not that happy with the functionality either. I am not expecting a perfect solution, but I am struggling to use the proposed functionality.

* The way UI is structured right now is very un-intuitive. I would add a gear icon somewhere in the corner, and have a popup window with loadable name configuration being one of the options.

* There is no guidelines (tooltip, help button) of how the field should be configured. I tried to guess, and failed (see below)

image

 ... and if I remove all text from the blank field, and re-examine, even on a new series - it has memory!

image

* I don't see how it can be helpful to the user to have a completely unconstrained blank field to configure node name based on the DICOM tags, while most users have no clue what those possible choices are. To me, it would be a lot more logical to have a pre-defined set of tags that might make sense to use while constructing node name, and give an option for an advanced user to specify arbitrary tags.

* do you _really_ need to allow the use of arbitrary tags while forming the series name? how about having a hard-coded set of tags that are recognized, and amend that list as new use cases come up? This will also address Andras concerns about sequences and private attributes. Did you think what would happen if someone sets PixelData as the series name?

> There should be no yelling (all-uppercase) in any text that is displayed to user.

Except when using DICOM! :-D (sorry, could not resist!)

jonieva:
Thank you for your comments.

@fedorov 
Andrey: there is a tooltip in the textbox that gives you some help. This could be replaced with a "help" icon if needed, but I don't see any help at all in the rest of the DICOM plugins. If you remove all the text from the textbox, the plugin would have the "old" behaviour (Series name: series description). If this is not how it's working for you, then there may be an error in the code.

@lassoan 
Andras,
Just to make sure. If I refactor the proposed solution to make a new DICOMPlugin, would this be accepted as it is right now? Please note that I would still need the textbox for the mask in Advanced options, although I could configure the DICOM Browser to hide it when the plugin is not selected.

lassoan:
Yes, I think the simplest is to do it all in your own plugin. No need to change the DICOM dialog in the core, you can add a textbox in the GUI of your own module or add a widget to an existing frame in the DICOM module GUI.

jonieva:
Got it.

As it seems clear our proposed solution is not good enough, I'm closing this pull request. We will think about the plugin road. 

Thank you all for your time and your comments.

fedorov:
I think if you do it in a plugin, you should do it in an extension, not in the core.

fedorov:
@jonieva I am curious what happened with this? Let me guess - you forked and tweaked Slicer app and use a non-standard binary?

jonieva:
Well... I couldn't get the Slicer tests to work, so I assumed that the pull
request wouldn't be accepted. So yeah, basically our users used a tweaked
Slicer for their purposes. Maybe one day I can resume the work to see what
was going on with the test, because the manual tests that I did worked
fine.

Jorge Onieva

PR #528, opened 2016-06-09
BUG: Fix logging warning message when slicer.vtkPVScalarBarActor doesn't exist

msmolens:
This commit fixes a bug with logging a warning message. Because 'slicer.logging'
doesn't exist, the call to 'slicer.logging.warning()' failed with an error like:

      File "/path/to/Slicer-build/lib/Slicer-4.5/qt-scripted-modules/DataProbeLib/SliceViewAnnotations.py", line 25, in __init__
	slicer.logging.warning("SliceAnnotations: Disable features relying on vtkPVScalarBarActor")
    AttributeError: 'module' object has no attribute 'logging'

Now the code calls 'logging.warning()'.

This warning was added in r24058. As described in that commit, the warning can
be triggered by starting Slicer with '--disable-builtin-loadable-modules'.

### DIFF
--- a/Modules/Scripted/DataProbe/DataProbeLib/SliceViewAnnotations.py
+++ b/Modules/Scripted/DataProbe/DataProbeLib/SliceViewAnnotations.py
@@ -1,5 +1,6 @@
 from __future__ import division
 import os
+import logging
 import qt
 import vtk
 import slicer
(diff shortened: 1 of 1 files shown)


jcfr:
Thanks for the PR :+1: 

Integrated in [r25178](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25178)

PR #537, opened 2016-06-16
ENH: Update EMSegment to include Python wrapping fixes for VTK 7.1

msmolens:
This commit updates EMSegment to include Python wrapping fixes for the current
VTK master (upcoming VTK 7.1).


jcfr:
Integrated in [r25196](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25196)

PR #538, opened 2016-06-16
STYLE: remove dead code in vtkMRMLViewDisplayableManager mantis 2341

QuadmasterXLII:
http://na-mic.org/Mantis/view.php?id=2341


jcfr:
:+1: 

Will integrate after CI is green.

jcfr:
After checking it compiled locally, integrated in [r25194](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25194)

PR #500, opened 2016-05-18
ENH: customize ctkAxesWidget

Punzo:
This is a following up of https://github.com/Slicer/Slicer/pull/464

the Axes Labels stored in the ViewNode will be used to customize automatically also
the ctkAxisWidget.

It will work after the merging of this
https://github.com/commontk/CTK/pull/647

PR #541, opened 2016-06-23
BUG: the labels on the controller widget were swapped

Punzo:
The bug has reported during the project week. The label were swapped on the widget controller 
(R <-> L, S <-> I and A <-> P).

The bug was introduced by myself in https://github.com/Slicer/Slicer/pull/500

### DIFF
--- a/Libs/MRML/Widgets/qMRMLThreeDViewControllerWidget.cxx
+++ b/Libs/MRML/Widgets/qMRMLThreeDViewControllerWidget.cxx
@@ -334,12 +334,12 @@ void qMRMLThreeDViewControllerWidget::updateWidgetFromMRML()
     }
 
   QStringList axesLabels;
-  axesLabels <<  d->ViewNode->GetAxisLabel(0);
   axesLabels <<  d->ViewNode->GetAxisLabel(1);
-  axesLabels <<  d->ViewNode->GetAxisLabel(4);
+  axesLabels <<  d->ViewNode->GetAxisLabel(0);
   axesLabels <<  d->ViewNode->GetAxisLabel(5);
-  axesLabels <<  d->ViewNode->GetAxisLabel(2);
+  axesLabels <<  d->ViewNode->GetAxisLabel(4);
   axesLabels <<  d->ViewNode->GetAxisLabel(3);
+  axesLabels <<  d->ViewNode->GetAxisLabel(2);
   d->AxesWidget->setAxesLabels(axesLabels);
 
   d->actionSet3DAxisVisible->setChecked(d->ViewNode->GetBoxVisible());
--- a/Libs/MRML/Widgets/qMRMLThreeDViewControllerWidget.cxx
+++ b/Libs/MRML/Widgets/qMRMLThreeDViewControllerWidget.cxx
@@ -333,13 +333,15 @@ void qMRMLThreeDViewControllerWidget::updateWidgetFromMRML()
     return;
     }
 
+  // In the axes widget the order of labels is: +X, -X, +Z, -Z, +Y, -Y
+  // and in the view node axis labels order is: -X, +X, -Y, +Y, -Z, +Z.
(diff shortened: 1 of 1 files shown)


jcfr:
Good catch and thanks for the PR

While working on this, could you look into introducing an enum named for example `vtkMRMLAbstractViewNode::AxisLabelType` 

That way we wouldn't have to hard code arbitrary integer to retrieve the label names

References:
* https://github.com/Slicer/Slicer/blob/71357481556d85e33544f1b8a0445f8e2b1a1e00/Libs/MRML/Core/vtkMRMLAbstractViewNode.h#L203
* https://github.com/Slicer/Slicer/blob/71357481556d85e33544f1b8a0445f8e2b1a1e00/Libs/MRML/Core/vtkMRMLAbstractViewNode.cxx#L34

Libs/MRML/Widgets/qMRMLThreeDViewControllerWidget.cxx
```diff
@@ -334,12 +334,12 @@ void qMRMLThreeDViewControllerWidget::updateWidgetFromMRML()
     }
 
   QStringList axesLabels;
-  axesLabels <<  d->ViewNode->GetAxisLabel(0);
   axesLabels <<  d->ViewNode->GetAxisLabel(1);
```
lassoan: Add a comment here that: in the axes widget the order of labels is: +X, -X, +Z, -Z, +Y, -Y and in the view node axis labels order is -X, +X, -Y, +Y, -Z, +Z. And maybe after each line add a comment which label is that, for example:
axesLabels <<  d->ViewNode->GetAxisLabel(1); // +X

lassoan:
I agree that enums could be added to ViewNode, but it's really the axis widget that has an exotic order (I would expect LPS or RAS), so ctkAxesWidget should have enums. But the ctkAxesWidget API that we used to set the labels don't use any enums just labels are added to the list in a specific order.
So, either we change the ctkAxesWidget's set methods to make it more explicit what order is used or just manage this by comments in the code.

Punzo:
Hi, thanks for your feedback Jc and Andras. I agree that using the enum we will benift a lot.
The issue is that in the ctkAxes the labels are stored as a QStringList  https://github.com/commontk/CTK/blob/master/Libs/Widgets/ctkAxesWidget.cpp#L56

(and therefore the set method that we implemented in https://github.com/commontk/CTK/pull/653 takes in input a QStringList).

If for you it is ok, I can apply a commit with Andras comments with the labels order for a fast fix, so you may integrate the fix quickly. If you can wait, next week I'll see how to use enum coherently both in the Slicer and ctkAxes.

Punzo:
Hi @jcfr and @lassoan, I have a question regarding the enum:

in order to store the labels strings in an enum is required to use the enum class 
http://en.cppreference.com/w/cpp/language/enum
which requires C++11.  Does Slicer support C++11?

Another solution will be to couple the "normal" integer enum with string values in a map.
However, I don't know if this last solution will be really better respect what we have now.

jcfr:
> or just manage this by comments in the code.

Since @Punzo fixes a regression, I squashed the commits and updated the message. It is now integrated as [r25208](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25208)

We can still add enum an other time.

> Does Slicer support C++11?

Not yet.

Punzo:
Hi Jc, sorry I am currently still very busy with the revision of the paper. Of course, I can add the enum later on. However, (I think), it will be better and easier if I can use enum class with c++11 (in which the enum can be mapped also with string value, not only integer value).

Thanks for integrating.

jcfr:
>  I am currently still very busy with the revision of the paper

No problemi. Good luck wrapping up.

> use enum class with c++11

Agreed.  After transitioning to c++11, we should be able to simplify enums

PR #540, opened 2016-06-17
Update to VTK 7

msmolens:
This updates Slicer to use VTK 7.

See:
- https://www.slicer.org/slicerWiki/index.php/Documentation/Labs/VTK7.
- http://massmail.spl.harvard.edu/public-archives/slicer-devel/2016/040450.html

PR #542, opened 2016-06-24
ENH: Added vtkSegmentationCore and Segmentations MRML, module, widget…

cpinter:
…s, editor

Integration of the segmentations infrastructure that used to be in SlicerRT. It is a new paradigm of handling segmentations, see slides [1] to understand underlying concepts.

[1] https://www.slicer.org/slicerWiki/index.php/File:20160526_Segmentations.pptx


cpinter:
@jcfr @pieper: If you give the green light I'll commit this whole thing to SVN. Thanks!

jcfr:
Exciting !

I will review. If CI turn green ... we should also be in good shape.

Cc: @msmolens  Could you check if the PR is compatible with transition to VTK 7.1 (https://github.com/Slicer/Slicer/pull/540)

cpinter:
Issues I discovered since the pull request:
* qSlicerSegmentationsModuleWidgetGenericTest fails due to too wide module. I'll re-check/address this once I'm back to my regular computer (widget UI files behave very differently on the 3Kx2K display I'm currently using)
* py_SegmentationsModuleTest1 fails due to missing RT loading feature. This is an oversight on my side and I will fix this very soon.

cpinter:
There's a build error due to unaccessible vtkLoggingMacros.h. I'll amend the commit if I applied a fix (it compiled for me so cannot really validate I guess)

cpinter:
It seems there are other errors that didn't come out on my side. I'll fix them one by one. I'm going to visit a friend of mine in Germany, but I can still start new pull requests I'm sure.

cpinter:
The latest error is something I cannot really fix from here. vtkImageFillROI.h apparently cannot be found:
/usr/src/Slicer/Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorAbstractLabelEffect.cxx:50:29: fatal error: vtkImageFillROI.h: No such file or directory

However its directory is included in Modules\Loadable\Segmentations\EditorEffects\CMakeLists.txt, and if I print out the cmake variable vtkSlicerEditorLibModuleLogic_INCLUDE_DIRS then it seems to contain the right paths. I'm without ideas right now, and could use some tips.

lassoan:
vtkImageFillROI is only used in qSlicerSegmentEditorAbstractLabelEffect::createMaskImageFromPolyData, which creates mask from 2D polydata. Almost all effect now generate 3D polydata (to allow real-time preview in all 2D and 3D views, painting in arbitrary slice orientation, slice thickness, etc).

If you cannot resolve this vtkImageFillROI finding problem then let me know and I replace it with 3D painting at the remaining few places.

cpinter:
The include directories are fine, but the error remains. I think it's because Loadable modules are added before the Scripted ones. If this is true, then the only option to fix the include problem would be to move the vtkImageFillROI class to vtkAddon or someplace else (or to duplicate in segmentations which I wouldn't like to do).
I think ultimately it would be better to get rid of the code the uses it. @lassoan if you have time to fix it by Monday then I'd appreciate your help, otherwise I'll try to do it early next week (it's a long weekend here in Canada and we're planning to make use of it).

pieper:
vtkImageFillROI is very old code from slicer2 that could definitely be retired if we have a valid substitute.

When I ported it to slicer3 I looked for alternatives and I found that there were some special cases that other rasterizers didn't handle cleanly.  Specifically if you have a bow-tie like outline vtkImageFillROI usually handles it in an "expected" way but other algorithms didn't.  On the other hand I've noticed that sometimes the RectangleEffect can get an extra pixel artifact due to some rasterizing bug.

It sounds like moving the code to vtkAddon would be the simplest solution, but something more general would also be welcome.

lassoan:
Several effects use 3D rasterization already and we should move towards that in all effects. So, in the long term we won't need this filter. So, we only need a short-term solution (for a few weeks, until I get back from vacation and work on this for a few days).

jcfr:
> move the vtkImageFillROI class to vtkAddon or someplace else

It could be moved to `Base/Logic`.  

`vtkAddons` would make sense if you anticipate the class to be contributed upstream VTK

> Several effects use 3D rasterization already and we should move towards that in all effects

:+1:

cpinter:
I moved it to vtkAddon for now, as I started doing that before Jc's comment and at least we can see if there are further issues. I can still move it to Base/Logic if that's the concensus.

pieper:
Yes, Jc is correct, Base/Logic is better since this class isn't something
suitable for vtk proper.

cpinter:
OK! I'll wait until the test is done, then will amend the commit with the class moved to Base/Logic.

cpinter:
Amendment pushed. It should be good now.
There is one test that will fail, but I plan to fix it next week if that's alright.

cpinter:
Can I go ahead and commit to SVN?

jcfr:
@cpinter  I am doing a final check now. Will report shortly.

jcfr:
@cpinter You will find here an updated topic:
 
 https://github.com/jcfr/Slicer/commits/segmentations-integration

where:

* Relocation of `vtkImageFillROI` is in its own commit
* There are fixes for warnings

Would be great if you could start review the commit fixing warnings. There is one for each type of warnings (narrowing, sign-compare, unused, ...). The idea is to make sure the warnings were innocuous and not the symptoms of some  larger issue. 

I will follow up with few more commits.

cpinter:
Thanks, Jc! The warning fixes look good to me.

jcfr:
> The warning fixes look good to me.

Great. Thanks for reviewing. 

To facilitate review process ... I will create a PR from my topic. 
When we are happy with the fixes, I propose to squash them with the commit adding the segmentations support.

jcfr:
See #545

jcfr:
Closing this PR. It is superseded by #545

PR #546, opened 2016-07-07
ENH: Adding vtkMatrix4x4 to/from string methods to vtkAddonMathUtilities.

mholden8:
Please see background at: https://github.com/SlicerRt/Sequences/pull/16#discussion_r67286616.
Factoring out parallel implementation in the Transforms module's widget.

### DIFF
--- a/Libs/vtkAddon/vtkAddonMathUtilities.cxx
+++ b/Libs/vtkAddon/vtkAddonMathUtilities.cxx
@@ -162,9 +162,14 @@ bool vtkAddonMathUtilities::FromString(vtkMatrix4x4* mat, const std::string& str
   int numElements = 0;
   while(itr!=emptyItr)
     {
-    ss << *itr << " "; // Space is a delimiter that stringstream can handle
-    numElements++;
+    std::string valString( *itr );
     itr++;
+    if (valString.empty())
+      {
+      continue; // Ignore empty strings - it just means we have back-to-back delimiters
+      }
+    ss << valString << " "; // Space is a delimiter that stringstream can handle
+    numElements++;
     }
 
   // Ensure the matrix is 1x1, 2x2, 3x3, or 4x4
@@ -174,7 +179,6 @@ bool vtkAddonMathUtilities::FromString(vtkMatrix4x4* mat, const std::string& str
     return false;
     }
   int sizeInt = std::floor(size);
-
 
   // Put into matrix
   for (int i = 0; i < sizeInt; i++)
--- a/Libs/vtkAddon/vtkAddonMathUtilities.cxx
+++ b/Libs/vtkAddon/vtkAddonMathUtilities.cxx
@@ -150,47 +150,47 @@ bool vtkAddonMathUtilities::FromString(vtkMatrix4x4* mat, const std::string& str
     return false;
   }
 
-  // Convert expression into a stringstream to convert and parse into doubles for the matrix
-
   // Parse the string using the regular expression
   std::regex delimiterRegex(delimiterExp);
   std::sregex_token_iterator itr(str.begin(), str.end(), delimiterRegex, -1); // Use matches as delimiters to split the string
   std::sregex_token_iterator emptyItr;
 
-  // Put into stringstream
-  std::stringstream ss;
-  int numElements = 0;
+  // Convert each string token into a double and put into vector
+  char* end;
+  std::vector<double> elements;
   while(itr!=emptyItr)
     {
     std::string valString( *itr );
-    itr++;
-    if (valString.empty())
+    double val = std::strtod( valString.c_str(), &end );
+    if ( *end != 0 )
       {
-      continue; // Ignore empty strings - it just means we have back-to-back delimiters
+      return false; // Parsing failed due to non-numeric character
       }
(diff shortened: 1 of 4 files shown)


pieper:
Great idea - thanks!

Libs/vtkAddon/vtkAddonMathUtilities.cxx
```diff
@@ -116,3 +119,80 @@ void vtkAddonMathUtilities::GetOrientationMatrix(vtkMatrix4x4* source,
       }
     }
 }
+
+//----------------------------------------------------------------------------
+std::string vtkAddonMathUtilities::ToString(const vtkMatrix4x4* mat, const std::string delimiter, const std::string rowDelimiter)
```
lassoan: Could you please consolidate this with the to/from string conversion implemented in qSlicerTransformsModuleWidget::pasteTransform() and qSlicerTransformsModuleWidget::copyTransform()?

Probably the best is to have the string converters in vtkAddonMathUtilities and use that in qSlicerTransformsModuleWidget but we shouldn't lose any functionality. Conversion from string is not trivial, as we want to support 2x2, 3x3, 4x4 matrix with different separators (space, tab, comma, semicolon, etc) and ignore extra characters ( ()[] etc) to be convenient to both Python and Matlab users.

Modules/Loadable/Transforms/qSlicerTransformsModuleWidget.cxx
```diff
@@ -429,39 +420,12 @@ void qSlicerTransformsModuleWidget::pasteTransform()
...
-  QString text = QApplication::clipboard()->text();
-  QRegExp rx("(\\ |\\,|\\:|\\t|\\n|\\[|\\])");
-  QStringList entries = text.split(rx, QString::SkipEmptyParts);
-
-  if (entries.count() != 4
-      && entries.count() != 9
-      && entries.count() != 16)
+  std::string text = QApplication::clipboard()->text().toStdString();
+  bool success = vtkAddonMathUtilities::FromString(tempMatrix.GetPointer(), text);
+  if (success)
```
lassoan: In case of failure log a warning and in the message include the string that was not converted.

Libs/vtkAddon/vtkAddonMathUtilities.cxx
```diff
@@ -116,3 +119,80 @@ void vtkAddonMathUtilities::GetOrientationMatrix(vtkMatrix4x4* source,
...
+      {
+      continue; // Ignore empty strings - it just means we have back-to-back delimiters
+      }
+    ss << valString << " "; // Space is a delimiter that stringstream can handle
+    numElements++;
+    }
+
+  // Ensure the matrix is 1x1, 2x2, 3x3, or 4x4
+  double size = std::sqrt(numElements);
+  if (std::floor(size)!=size || size>4)
```
lassoan: Don't do floating point comparison without tolerance. It would be simpler to just check that numElements is 1, 4, 9, or 16.

Libs/vtkAddon/vtkAddonMathUtilities.cxx
```diff
@@ -116,3 +119,80 @@ void vtkAddonMathUtilities::GetOrientationMatrix(vtkMatrix4x4* source,
...
+  int numElements = 0;
+  while(itr!=emptyItr)
+    {
+    std::string valString( *itr );
+    itr++;
+    if (valString.empty())
+      {
+      continue; // Ignore empty strings - it just means we have back-to-back delimiters
+      }
+    ss << valString << " "; // Space is a delimiter that stringstream can handle
```
lassoan: It would be nicer to parse the value immediately and store the result in a std::vector instead.

Libs/vtkAddon/vtkAddonMathUtilities.cxx
```diff
@@ -116,3 +119,80 @@ void vtkAddonMathUtilities::GetOrientationMatrix(vtkMatrix4x4* source,
...
+
+  // Put into matrix
+  for (int i = 0; i < sizeInt; i++)
+    {
+    for (int j = 0; j < sizeInt; j++)
+      {
+        double val;
+        if (!(ss >> val))
+          {
+          return false;
```
lassoan: it's not nice that we return with failure but still make some changes in the output matrix (parsing the string earlier will probably solve this)

lassoan:
Thank you, it looks good to me.

mholden8:
@lassoan Thanks for the feedback. I've updated it with your suggestions.

lassoan:
@mholden8 Tests fail because regexp is only available for Cxx11 compliant compilers. Could you please try if you could rewrite the code to use vtksys::RegularExpression instead?

lassoan:
@mholden8 Thanks for the quick fix. It seems that this time automatic tests failed because the original pull request was submitted when Slicer still used VTK6 but since Slicer moved to VTK7. Would you mind submitting a new pull request that contains the same changes (preferably squashed into one commit) and rebased on the latest Slicer trunk version? Let me know if you need any help or clarification. Sorry for the lots of requests, I think now we are really close to complete this integration.

mholden8:
@lassoan Thanks for the input. Please see the new pull request rebased on the latest Slicer trunk version #554.

PR #544, opened 2016-07-03
Fix and test extension build system

jcfr:
This topic will fix errors happening when building, testing or packaging extensions with dependencies.

It checks for the following:
* Generated description files are correct
* Submission to CDash are done
* Upload to extension server are done
* CMakeCache.txt of built extensions contains expected variables (e.g _DIR variables, ...)

Remaining tasks:
* [x] Check list of requests done to mock server are the expected ones.
* [ ] Install packages extensions and check dependent one are also installed

Notes:
* To ensure the test run in reasonable time, extensions used in the tests contains only scripted module. 
* extension sources used in the test are currently checked in the source (Testing/TestExtA, ...). I will revisit this to use the extension wizard.

PR #547, opened 2016-07-08
ENH: Add test checking ExtensionBuildSystem works as expected

jcfr:
This PR is a follow-up of #544 where all improvements and fixes have been integrated.

This set of patches integrate the test infrastructure used to validate the improvements and fixes integrated in #544 .

It checks for the following:
* Generated description files are correct
* Submission to CDash are done
* Upload to extension server are done
* CMakeCache.txt of built extensions contains expected variables (e.g _DIR variables, ...)
* Check list of requests done to mock server are the expected ones.

Remaining tasks:
* [x] Do not directly add sources of test extension sources (Testing/TestExtA, ...), instead generate their source using the wizard.
* [ ] Install packaged extensions and check dependent one are also installed

Notes:
* To ensure the test run in reasonable time, extensions used in the tests contains only scripted module.


jcfr:
Integrated in [25296](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25296)

Check that packaged extensions can be installed will be added later.

PR #549, opened 2016-07-12
BUG: Fix OpenGL errors in py_RSNA2012ProstateDemo test

msmolens:
Debug builds showed OpenGL errors while running the py_RSNA2012ProstateDemo
test.

The problem is that vtkMRMLAnnotationDisplayableManager could directly call
vtkRenderer::Render() in some cases, but only vtkRenderWindow should call that
method. The errors occur when OpenGL functions are called before an OpenGL
context has been initialized. When the VTK OpenGL2 backend is enabled, running
the py_RSNA2012ProstateDemo test leads to a crash.

See the documentation for vtkRenderer::Render():
https://github.com/Kitware/VTK/blob/5f52c1a91a9e9fbfc1efe0d14b3c11b548c7acb1/Rendering/Core/vtkRenderer.h#L231-L236

<strike>This commit sets the widget representation's NeedToRender flag instead of
attempting to render directly.</strike>
This commit replaces the direct call to vtkRenderer::Render() with a call to the
displayable manager's RequestRender() method. This queues a render request to be
handled by the framework at an appropriate time.

### DIFF
--- a/Modules/Loadable/Annotations/MRMLDM/vtkMRMLAnnotationDisplayableManager.cxx
+++ b/Modules/Loadable/Annotations/MRMLDM/vtkMRMLAnnotationDisplayableManager.cxx
@@ -1357,10 +1357,7 @@ bool vtkMRMLAnnotationDisplayableManager::IsWidgetDisplayable(vtkMRMLSliceNode* 
             }
 
           // we need to render again
-          if (currentRenderer)
-            {
-            currentRenderer->Render();
-            }
+          this->RequestRender();
           }
 
         }


jcfr:
Make sense.

That said, _grepping_ through the VTKv7 [code base](https://github.com/Slicer/VTK/tree/slicer-v7.1.0-2016-07-06-b824d94), I could find only few locations where `NeedToRender` is used: `vtkContourRepresentation` and `vtkContourWidget`. For that reason, I suspect calling `widget->GetRepresentation()->NeedToRenderOn();` is a no-op

msmolens:
Thanks for the review. Indeed, `NeedToRender` is not generally used when rendering widget representations, even though it's used in several classes in the `vtkMRMLAnnotationDisplayableManager` class hierarchy.

Do any @Slicer/slicer-core developers know how to trigger the follow branch outside of the py_RSNA2012ProstateDemo test?
https://github.com/Slicer/Slicer/blob/700bcbe4973741c7054d953eebe1d76b175f7c33/Modules/Loadable/Annotations/MRMLDM/vtkMRMLAnnotationDisplayableManager.cxx#L1332

pieper:
@naucoin may have some ideas about that code but she's out sick today -- I'm afraid that the annotations module code and infrastructure was contributed by some developers who didn't have the needed level of experience with VTK; this is why the Fiducials got moved to Markups and the other annotations have been patch where possible.  Fixing the crashes is good but there will probably always be some corner cases that won't work right until things are pretty much revamped.

msmolens:
I updated this with an alternate implementation that also avoids the OpenGL errors/crash.

jcfr:
Nice

naucoin:
@pieper is correct, there was a lot of hacking done to get the vtk widgets to render after programatic updates. The RequestRender call is the right way to go.

msmolens:
Integrated in [r25237](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25237).

PR #551, opened 2016-07-14
ENH: Test for sitkUtils to make sure to push/pull data from Slicer 

MayeulChassagnard:
According to the request [ #1693](http://na-mic.org/Mantis/view.php?id=1693) on MantisBugTracker, I added a test for Push/Pull data from Slicer

A minor error linked with the call of PullFromSlicer (a function from sitkUtils) still exist but do not interfere with the test.

The detail of the error can be seen in the file attached [here](https://github.com/Slicer/Slicer/files/364637/error_PullFromSlicer.zip). Otherwise, when running the following terminal command : `ctest -V -R py_nomainwindow_test_sitkUtils`  in the directory `Slicer-SuperBuild/Slicer-build`

### DIFF
--- a/Applications/SlicerApp/Testing/Python/CMakeLists.txt
+++ b/Applications/SlicerApp/Testing/Python/CMakeLists.txt
@@ -225,11 +225,11 @@ set_tests_properties(py_nomainwindow_${testname} PROPERTIES
   )
 
 if(Slicer_USE_SimpleITK)
-slicer_add_python_unittest(
-  SCRIPT ${Slicer_SOURCE_DIR}/Base/Python/tests/test_sitkUtils.py
-  SLICER_ARGS --no-main-window
-  TESTNAME_PREFIX nomainwindow_
-  )
+  slicer_add_python_unittest(
+    SCRIPT ${Slicer_SOURCE_DIR}/Base/Python/tests/test_sitkUtils.py
+    SLICER_ARGS --no-main-window
+    TESTNAME_PREFIX nomainwindow_
+    )
 endif()
 
 slicer_add_python_unittest(
--- a/Applications/SlicerApp/Testing/Python/CMakeLists.txt
+++ b/Applications/SlicerApp/Testing/Python/CMakeLists.txt
@@ -224,6 +224,14 @@ set_tests_properties(py_nomainwindow_${testname} PROPERTIES
   WILL_FAIL TRUE
   )
 
+if(Slicer_USE_SimpleITK)
+slicer_add_python_unittest(
+  SCRIPT ${Slicer_SOURCE_DIR}/Base/Python/tests/test_sitkUtils.py
+  SLICER_ARGS --no-main-window
+  TESTNAME_PREFIX nomainwindow_
+  )
+endif()
+
 slicer_add_python_unittest(
   SCRIPT ${Slicer_SOURCE_DIR}/Base/Python/tests/test_saferef.py
   SLICER_ARGS --no-main-window --disable-modules
--- a/Base/Python/tests/test_sitkUtils.py
+++ b/Base/Python/tests/test_sitkUtils.py
@@ -32,9 +32,62 @@ class SitkUtilsTests(unittest.TestCase):
                          'Error Push Pull: Modify Time are not the same')
 
         """ To perform the test:
-        Try to modified the image with sitk tools (just few sanity checks)
-        Verify the GetMTime are different after a modification of sitkimage
+        DONE : Try to modified the image with sitk tools (just few sanity checks)
+        DONE : Verify the GetMTime are different after a modification of sitkimage
         Try with all the compositeView and overwrite (any possibility)
+        """
+
+        """ Few modification of the image : Direction, Origin """
+        sitkimage.SetDirection((-1.0, 1.0, 0.0, 0.0, -1.0, 1.0, 1.0, 0.0, 1.0))
+        sitkimage.SetOrigin((100.0, 100.0, 100.0))
+
+        """ Few pixel changed """
+        size = sitkimage.GetSize()
+        for x in xrange(0,size[0],(int)(size[0]/10)):
+            for y in xrange(0,size[1],(int)(size[1]/10)):
+                for z in xrange(0,size[2],(int)(size[2]/10)):
+                    sitkimage.SetPixel(x,y,z,0L)
+
+        su.PushToSlicer(sitkimage, 'ImageChanged', compositeView=0, overwrite=False)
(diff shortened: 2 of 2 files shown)


Base/Python/tests/test_sitkUtils.py
```diff
@@ -0,0 +1,49 @@
+import slicer
+#import SimpleITK as sitk
+import sitkUtils as su
+
+import unittest
+
+class SitkUtilsTests(unittest.TestCase):
+
+    def runTest(self):
```
jcfr: This function should not be needed here

Base/Python/tests/test_sitkUtils.py
```diff
@@ -0,0 +1,49 @@
...
+        self.assertEqual(self.myNode1.GetMTime(), myNode2.GetMTime(),
+                         'Error Push Pull: Modify Time are not the same')
+
+        """ To perform the test:
+        Try with a modified sitkimage and verify the GetMTime are diffrents
+        Try with all the compositeView and overwrite (all possiblities)
+        """
+
+    def tearDown(self):
+        del self.myNode1
```
jcfr: I suggest to call `slicer.mrmlScene.Close()` here

Base/Python/tests/test_sitkUtils.py
```diff
@@ -0,0 +1,49 @@
+import slicer
+#import SimpleITK as sitk
```
jcfr: Remove commented code

Base/Python/tests/test_sitkUtils.py
```diff
@@ -0,0 +1,49 @@
...
+    def setUp(self):
+        """ Download the MRHead node
+        """
+
+        from SampleData import SampleDataLogic
+        SampleDataLogic().downloadMRHead()
+        myNode1 = slicer.util.getNode('MRHead')
+
+        print("Node Name is %s" % myNode1.GetName())
+        self.assertIsNotNone(myNode1.GetName(),"ERROR: Node Name is None !")
```
jcfr: Instead uses:
```
self.assertEqual(myNode1.GetName(), "MRHead")
```

Base/Python/tests/test_sitkUtils.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def setUp(self):
+        """ Download the MRHead node
+        """
+
+        from SampleData import SampleDataLogic
+        SampleDataLogic().downloadMRHead()
+        myNode1 = slicer.util.getNode('MRHead')
+
+        print("Node Name is %s" % myNode1.GetName())
```
jcfr: Print statement not needed

Base/Python/tests/test_sitkUtils.py
```diff
@@ -0,0 +1,49 @@
...
+        self.test_SimpleITK_SlicerPushPull()
+        self.tearDown()
+
+    def setUp(self):
+        """ Download the MRHead node
+        """
+
+        from SampleData import SampleDataLogic
+        SampleDataLogic().downloadMRHead()
+        myNode1 = slicer.util.getNode('MRHead')
```
jcfr: `myNode1` -> `volumeNode`

jcfr:
Nice.  Thanks for the PR :+1: 

Look like the error message happen because of the hack done [here](https://github.com/Slicer/Slicer/blob/b09de614b5ebc9e6a8e7974113fe82c637c79a24/Base/Python/sitkUtils.py#L83-L90)

A nice improvement would be to have a more obvious way of registering the IO factories.

@blowekamp What do you think ?

Applications/SlicerApp/Testing/Python/CMakeLists.txt
```diff
@@ -224,6 +224,14 @@ set_tests_properties(py_nomainwindow_${testname} PROPERTIES
   WILL_FAIL TRUE
   )
 
+if(Slicer_USE_SimpleITK)
+slicer_add_python_unittest(
```
vovythevov: This method should be indented (as it's inside a if()/endif())

MayeulChassagnard: Thanks for the review folks !

Base/Python/tests/test_sitkUtils.py
```diff
@@ -0,0 +1,49 @@
...
+        self.assertIsNotNone(myNode1.GetName(),"ERROR: Node Name is None !")
+
+        self.myNode1 = myNode1
+
+    def test_SimpleITK_SlicerPushPull(self):
+        """ Try to pull a slicer MRML image, return the SimpleITK,
+        push it to Slicer, and then compare if the image are the same
+        """
+
+        sitkimage = su.PullFromSlicer(self.myNode1.GetName())
```
blowekamp: I would add some sanity checks for the SimpleITK image, size, pixel type, direction matrix, and at least a couple pixels.

blowekamp:
Perhaps we should ensure EnsureRegistration only gets called once in sitkUtils.py? Or there should be a wrapped C++ method which checks if the registration needs to occur?

Base/Python/tests/test_sitkUtils.py
MayeulChassagnard: So, I can erase runTest ? What about setUp and tearDown ? Must I delete them also ?

jcfr:
> Perhaps we should ensure EnsureRegistration only gets called once in sitkUtils.py?

Calling it once by capturing its stderr would indeed be an improvement

> Or there should be a wrapped C++ method which checks if the registration needs to occur?

Agreed. That would be better

Base/Python/tests/test_sitkUtils.py
jcfr: You need `setUp` and `tearDown`, they are called before and after running each test function.

Base/Python/tests/test_sitkUtils.py
MayeulChassagnard: This command doesn't exist. Do you mean `slicer.mrmlScene.Clear()` instead ?

jcfr: Ooops. Yess, `slicer.mrmlScene.Clear(0)` it is.

jcfr:
Could you rebase  and push force your topic. This would ensure the CI will success.

```
git svn rebase
git push origin name_of_topic --force
```

MayeulChassagnard:
rebase + push done

jcfr:
Closing. Tests have been added by @lassoan  in [r26230](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26230)  (f66714c)

PR #555, opened 2016-07-27
ENH: Made Python interactor a docking widget

lassoan:
Problem:
It was always necessary to rearrange windows after showing the Python interactor so that both the console and Slicer screen.
This has to be repeated every time Slicer is started, since the Python interactor window's position was not saved.

Implemented solution:

1. Made Python interactor a docking widget (if Slicer starts with main window visible).
For single-screen configuration this allows sharing of the screen between the Python interactor and other Slicer windows.
On multi-screen configurations the interactor can be dragged to the second screen.

2. The position (which screen, where), size, and state (docked/undocked, shown/hidden) of the Python interactor window is now also saved,
so there is no need for any repositioning of the windows between sessions.

3. Ctrl-3 shows/hides the interactor (not just shows as before) and when the interactor is shown it gets keyboard focus.
It allows easy opening of the interactor, typing some text, and closing it, without the need of taking the mouse to reposition
any windows.

Also cleaned up deletion of ctkPythonConsole. As the ctkPythonConsole widget can be owned by a widget (docking window), it is not
possible to create it on the stack anymore. Instead, qSlicerApplication creates the ctkPythonConsole (similarly to qSlicerPythonManager)
and destroys it if it is not owned by a widget.

### DIFF
--- a/Base/QTGUI/qSlicerApplication.cxx
+++ b/Base/QTGUI/qSlicerApplication.cxx
@@ -34,6 +34,9 @@
 #include <ctkErrorLogStreamMessageHandler.h>
 #include <ctkITKErrorLogMessageHandler.h>
 #include <ctkMessageBox.h>
+#ifdef Slicer_USE_PYTHONQT
+# include "ctkPythonConsole.h"
+#endif
 #include <ctkSettings.h>
 #ifdef Slicer_USE_QtTesting
 #include <ctkQtTestingUtility.h>
@@ -55,6 +58,7 @@
 #include "qSlicerModuleManager.h"
 #ifdef Slicer_USE_PYTHONQT
 # include "qSlicerPythonManager.h"
+# include "qSlicerSettingsPythonPanel.h"
 #endif
 #ifdef Slicer_BUILD_EXTENSIONMANAGER_SUPPORT
 # include "qSlicerExtensionsManagerDialog.h"
@@ -172,13 +176,31 @@ void qSlicerApplicationPrivate::init()
 #ifdef Slicer_USE_PYTHONQT
   if (!qSlicerCoreApplication::testAttribute(qSlicerCoreApplication::AA_DisablePython))
     {
-    // Note: qSlicerCoreApplication class takes ownership of the pythonManager and
+    // qSlicerCoreApplication class takes ownership of the pythonManager and
     // will be responsible to delete it
     q->setCorePythonManager(new qSlicerPythonManager());
+    // qSlicerCoreApplication does not take ownership of PythonConsole, therefore
+    // we have to delete it in the destructor if it is not deleted already
+    // and not owned by a widget (it is owned and deleted by a widget if it is added
+    // to the GUI)
+    q->setPythonConsole(new ctkPythonConsole());
     }
 #endif
 
   this->Superclass::init();
+
+#ifdef Slicer_USE_PYTHONQT
+  if (!qSlicerCoreApplication::testAttribute(qSlicerCoreApplication::AA_DisablePython))
+    {
+    q->pythonConsole()->initialize(q->pythonManager());
+    QStringList autocompletePreferenceList;
+    autocompletePreferenceList
+      << "slicer"
+      << "slicer.mrmlScene"
+      << "qt.QPushButton";
+    q->pythonConsole()->completer()->setAutocompletePreferenceList(autocompletePreferenceList);
+    }
+#endif
 
   this->initStyle();
 
@@ -244,6 +266,13 @@ void qSlicerApplicationPrivate::init()
   qSlicerSettingsInternationalizationPanel* qtInternationalizationPanel =
       new qSlicerSettingsInternationalizationPanel;
   this->SettingsDialog->addPanel("Internationalization", qtInternationalizationPanel);
+#endif
+
+#ifdef Slicer_USE_PYTHONQT
+  if (!qSlicerCoreApplication::testAttribute(qSlicerCoreApplication::AA_DisablePython))
+    {
+    q->settingsDialog()->addPanel("Python", new qSlicerSettingsPythonPanel);
+    }
 #endif
 
   qSlicerSettingsDeveloperPanel* developerPanel = new qSlicerSettingsDeveloperPanel;
@@ -325,6 +354,17 @@ qSlicerApplication::qSlicerApplication(int &_argc, char **_argv)
 //-----------------------------------------------------------------------------
 qSlicerApplication::~qSlicerApplication()
 {
+  // We have to delete PythonConsole if it is not deleted already
+  // and not owned by a widget (it is owned and deleted by a widget if it is added
+  // to the GUI).
+  ctkPythonConsole* pythonConsolePtr = this->pythonConsole();
+  if (pythonConsolePtr)
+  {
+    if (pythonConsolePtr->parent() == NULL)
+    {
+      delete pythonConsolePtr;
+    }
+  }
 }
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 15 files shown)


fedorov:
much needed feature, thanks!

pieper:
+1

cpinter:
Can't wait to try!

Base/QTGUI/qSlicerApplication.cxx
```diff
@@ -325,6 +354,17 @@ qSlicerApplication::qSlicerApplication(int &_argc, char **_argv)
 //-----------------------------------------------------------------------------
 qSlicerApplication::~qSlicerApplication()
 {
+  // We have to delete PythonConsole if it is not deleted already
+  // and not owned by a widget (it is owned and deleted by a widget if it is added
+  // to the GUI).
```
jcfr: * I think this should be protected by `#ifdef Slicer_USE_PYTHONQT ` and `qSlicerCoreApplication::testAttribute(qSlicerCoreApplication::AA_DisablePython)`

* Fix indent

jcfr:
Very nice.

Few nitpicks and it can be integrated

lassoan:
Thanks for the feedback.
I'll wait for circleci results, add #ifdef, fix indentation, and integrate.

lassoan:
Integrated into rev 25248

cpinter:
Is the disappearance of the Python interactor button from the toolbar related to this change? I always used that button to show the interactor, I found it convenient

lassoan:
Yes. It's a toggle now, it's not like the extension manager anymore. To use the Python console you need to use the keyboard anyway and you don't need to use the mouse anymore. So, any time you want to start with the Python console or finished what you wanted to do just press Ctrl-3. With mouse you can still see the state of the Python console visibility and show/hide it in the View menu.

Of course we can still add a toolbar button if there is a strong need.

cpinter:
To me it was a useful feature that was now arbitrarily discarded. There are other toggle buttons on the toolbar too, I don't understand the opposition. I think the "strong need" is not needed to restore it, but to hide it (=make the change).

fedorov:
I also don't see any need for this button to be removed. All developers expect to see it.

lassoan:
I've committed a change now to show the button on the toolbar and you'll see what's the problem. For me, the view menu look quite inconsistent after this change (as the modules pane show/hide action is shown as a checkmark). What do you think?
We could add a separate, non-toggle action to the toolbar for showing the Python console.
Or, we could add an icon for the module panel (and maybe to most other actions in the view menu have an icon anyway).
I'm also thinking about making the error log window a dockable widget as well (and allow toggling with ctrl-0), for convenience and consistency.

fedorov:
I see your concern. Why didn't you raise it at the time of this PR review?

I will follow up after I take a look and have time to think about it.

fedorov:
Algorithm name/type missing in output ...

fedorov:
sorry, wrong thread!

lassoan:
I thought I have a good solution. These kinds of UI tweaks are continuous, ongoing activities, we can keep iterating through them in nightly builds until we reach a consensus.

pieper:
I kind of like this behavior, but not enough to prefer it over the older
behavior unless we make some changes.  Here are my thoughts:

Pros:
* I like the console remembering it's size and position across runs

Cons:
* I don't imagine myself ever wanting to use the console in 'docked' mode.
* The docked widget is always 'on top' of the rest of slicer.
* I don't like that the console disappears when switching to another app
(at least this is what happens on mac).  This makes it harder to refer to
the output when looking at a web page for instance.  Also impossible to
watch the python console output when using a different app.

It's easy to try the non-docking version of this with the following snippet:

c = findChildren(name='Python*')[0]
c.setParent(None); c.show()

Toggling still works from the toolbar.  The only thing that would be needed
is to add the control-3 hotkey to the console widget and we would have the
nice behavior of Andras's change without the drawbacks.  (If you really
want it to be dockable then it would be possible to have a menu command
that puts it in a QDockWidget).

Extra comment:
What I'd really like to have is command line history in the python
console.  I've been meaning to do that for years but have never gotten
around to it because I expected we'd be able to use ipython someday.

lassoan:
> I don't imagine myself ever wanting to use the console in 'docked' mode.

Docked mode is important for using it on singles screen (especially laptops). It's optional anyway.

> The docked widget is always 'on top' of the rest of slicer.

If it's docked then it's actually not on top of anything (doesn't hide anything), it resizes other parts of the GUI to make space.

> I don't like that the console disappears when switching to another app

Would you like to show another application's window between the main window and an undocked window? I don't think it's possible with Qt dock widgets. If you need this then maybe docking/undocking should be an application settings.

pieper:
I don't mind docking as an option for other people, but I don't want to
have it myself, and I don't like the QDockWidget behavior.  For example, I
don't want the rest of the layout to change when I bring up the python
console.  It should be completely independent of the main window.

Yes, I would like to be able to have the slicer main window on top of the
python interactor.  I'd also like to be able to have another application on
top of my slicer main window and then have the python console on top of
that.

I agree the issue is the QDockWidget.  That's why I suggest not using it
unless the user specifically requests it.  If you paste in the python
statements I sent earlier you can see what the behavior is like without the
QDockWidget in the way (it's nicer!  : )

cpinter:
+1 for adding it it to Application settings

Also I don't insist on re-adding the toolbar button if removing it is fine with everyone. I just don't see a reason for it, as I can't imagine anyone being hindered by it in any way.

lassoan:
I've made dockable Python console optional. Can be set along with other Python console settings in Application Settings / Python. The option is enabled by default, as novice users are not likely to switch between many applications while in Slicer and this way there is no need for manual window management. As I wrote earlier, I also added back the toolbar button.

pieper:
You aren't saving the size and location state when it's not in docking mode
- is that on purpose?  that was the most useful feature!  I was suggesting
using same widget management but without the QDockWidget parent so that we
got the benefit of your updates without the drawbacks.

I'd like to hear what other developers as they use this.  If you have the
docking-style console undocked on a small screen you won't be able to place
it behind the main window.  I guess the idea is that you use the hotkey to
pop it up and down when you need it.

lassoan:
I just restored the old behavior of allowing non-dockable Python console. I haven't improved anything on that. Saving/restoring position, state, etc. are all managed automatically for QDockWidget parent, I agree that it would be useful for the separate window as well, but that would need to be implemented from scratch.

On a single screen I usually want to see the effect immediately as I enter multiple commands.
With separate window: ctrl-3, enter command, Ctrl-3, look at the screen. With docking it's simpler:enter command, look at screen.

The separate window style Python console does not capture Ctrl-3, so actually you need to use the mouse to hide it (old behavior kept). I've partially fixed this by making the toolbar button a toggle, but somehow a shortcut should be registered for the console, too. Maybe it could be implemented with an event filter or by putting the console widget into another window, but again, it was all automatically working for the docking window and I just did not have the time to improve the separate window as well.

jcfr:
> useful for the separate window

This could probably be done using something like https://github.com/commontk/CTK/blob/bacd3602caa8e522d081dd204e9feb3df473235d/Libs/Widgets/ctkSettings.cpp#L116-L119

> QDockWidget

I played with it and I like it. I used to have a one liner in my `.slicerrc.py` to do it and never got the time to upstream the functionality.

Thanks for working on improving the python interactor. Much appreciated. Coupled with the coming changes of @MayeulChassagnard (https://github.com/commontk/CTK/pull/659 and https://github.com/commontk/CTK/pull/662), the overall python experience in Slicer will be improved a lot :+1:

pieper:
Just setting the parent to be None and showing the console widget was the
only thing needed to get the 'best of both worlds' behavior.  Seems like
that would be easy enough to do when we get a dockLocationChanged signal
with a value of NoDockWidgetArea, but I guess there could be side effects
of that.

But if you don't want to do that and nobody else objects I suppose I can
try to get used to the new behavior or else change the undocked mode myself.

PR #556, opened 2016-07-28
ENH: Update CreateDICOMSeries.xml doc link

ihnorton:


jcfr:
Thanks for the patch :+1: 

Integrated in [r25251](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25251)

PR #557, opened 2016-07-29
ENH: Make it faster/easier to work with long fiducial lists

naucoin:
Put Start/EndModify calls around removing a list of markups from the current node (speed up is significant when deleting 100 fiducials from a list of 272 fiducials, down to ~4s from ~120s).

Automatically select the last loaded fiducial node when entering the Markups module if there are nodes loaded and none are selected.

Adjust the right click context menu in the table:
Use one line per markup, consolidating the label (use index if no label) and the coordinate (half the rows in the menu).
Put the operations first in the menu in case so many are selected that part of the menu ends up off screen.

Fixes Issue #4239


pieper:
LGTM - thanks!

Committed to svn as r25253

jcfr:
Very nice :+1

PR #559, opened 2016-08-02
COMP: bump DCMTK version to 3.6.1_20160630

ihnorton:
This has updates needed by us and @fedorov. There was only one patch in the CTK branch, and it is no longer necessary, so I switched back to upstream.


fedorov:
👍 thanks for testing and contributing the patch!

msmolens:
Thanks, integrated in [r25259](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25259).

PR #560, opened 2016-08-03
COMP: Support building in RelWithDebInfo configuration on Windows

msmolens:
This commit fixes build errors when building in the RelWithDebInfo configuration
on Windows. Previously, projects that use OpenSSL_DIR would fail to find the
OpenSSL headers because only 'Debug' and 'Release' directories--not
'RelWithDebInfo'--are provided by the precompiled OpenSSL binaries.

An example error is:

    10>C:\S\R\Python-2.7.11\Modules\_hashopenssl.c(39): fatal error C1083: Cannot open include file: 'openssl/evp.h': No such file or directory

Fixes http://www.na-mic.org/Mantis/view.php?id=4058.

### DIFF
--- a/SuperBuild/External_OpenSSL.cmake
+++ b/SuperBuild/External_OpenSSL.cmake
@@ -249,10 +249,24 @@ this version of visual studio [${MSVC_VERSION}]. You could either:
       )
 
     set(OpenSSL_DIR ${EP_SOURCE_DIR})
+    set(_openssl_base_dir ${OpenSSL_DIR})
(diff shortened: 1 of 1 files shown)


msmolens:
cc: @dzenanz

dzenanz:
Good job Max!

msmolens:
Integrated in [r25269](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25269).

PR #561, opened 2016-08-03
WIP: CI: set DISPLAY in CircleCI; check if image updated before re-saving

ihnorton:
This seems to work -- I can't shell in to a running build container itself because CircleCI doesn't support `docker exec`, but launching a new instance does allow to connect to the X server on the host.

For now this hard-codes the default IP address used for the `docker0` bridge and client. We can do a little more work with some awk'ing to get the values dynamically.

Also adds a check for whether `docker pull` actually updated anything, to avoid re-saving the image unnecessarily (saves 3-5 minutes).

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -3,6 +3,8 @@ machine:
     - curl -sSL https://s3.amazonaws.com/circle-downloads/install-circleci-docker.sh | bash -s -- 1.10.0
   services:
     - docker
+  environment:
+    DISPLAY: 172.17.0.1:99
 
 dependencies:
   cache_directories:
@@ -10,12 +12,16 @@ dependencies:
 
   override:
     - docker info
+    - mkdir -p ~/docker
     - if [[ -e ~/docker/image.tar ]]; then docker load -i ~/docker/image.tar; fi
+    - docker inspect --format='{{.Id}}' slicer/slicer-build-deps > ~/docker/image.id.last
     - docker pull slicer/slicer-build-deps
-    - mkdir -p ~/docker; docker save slicer/slicer-build-deps > ~/docker/image.tar
+    - docker inspect --format='{{.Id}}' slicer/slicer-build-deps > ~/docker/image.id
+    - if [[ `diff ~/docker/image.id ~/docker/image.id.last` ]] ; then docker save slicer/slicer-build-deps > ~/docker/image.tar ; fi
 
 test:
   override:
     - "echo 'Notice: CircleCI does not build changes to Slicer dependencies' && if git diff --name-only master | grep -q SuperBuild > /dev/null; then false; fi"
+    - xhost +172.17.0.2
     - docker run -e "BUILD_TOOL_FLAGS=-j5" --name slicer -v ~/Slicer:/usr/src/Slicer slicer/slicer-build-deps
     - docker cp slicer:$(docker cp slicer:/usr/src/Slicer-build/Slicer-build/PACKAGE_FILE.txt - | tar xO) $CIRCLE_ARTIFACTS


thewtex:
Cool! Look promising.

Here is another reference that may help with exposing the X environment:

  https://github.com/thewtex/docker-opengl-mesa/blob/master/run.sh

HTH,
Matt

jcfr:
@ihnorton We now successfully run test required X display using CircleCI for CTK project:
* https://github.com/commontk/CTK/tree/master/CMake/CircleCI
* https://github.com/commontk/CTK/blob/master/circle.yml

Let me know if you have time to update Slicer to do the same. If not, I will ask @MayeulChassagnard  to work on it.

MayeulChassagnard:
To clarify, I should make work on CircleCI tests which need display mode, am I right?
In this case, I should modify the docker image to looks like this [docker-opengl](https://github.com/thewtex/docker-opengl), right?

thewtex:
@MayeulChassagnard Best results may come from using a testing image like the docker-opengl-mesa image referenced above but `FROM ubuntu:14.04`, then setting the CircleCI build to run on their Ubuntu 14.04 image. This will ensure both mesa versions are the same, which can avoid issues.

VTK has issues rendering to the CPU rendering backend in the docker-opengl image.

ihnorton:
I was out last week and missed this thread. Sounds great! You folks are clearly most familiar with CircleCI.

(hopefully this will provide an incentive to make the tests faster -- last time I looked, there seemed to be many tests, particularly python workflow tests, that completed quickly and then hung inexplicably until some timeout was hit)

jcfr:
@thewtex Instead of re-using the mesa approach, I suggest @MayeulChassagnard looks into finalizing https://github.com/thewtex/SlicerDocker/pull/5 What do you think ?

for reference, we tried 

```
XSOCK=/tmp/.X11-unix
XAUTH=/tmp/.docker.xauth
touch $XAUTH
xauth nlist $DISPLAY | sed -e 's/^..../ffff/' | xauth -f $XAUTH nmerge -

docker run \
	-v $XSOCK:$XSOCK:rw \
	-v $XAUTH:$XAUTH:rw \
	--device=/dev/dri/card0:/dev/dri/card0 \
	-e DISPLAY=$DISPLAY \
        -e XAUTHORITY=$XAUTH \
        -ti slicer/slicer-build bash
```

then

```
cmake -DBUILD_TESTING:BOOL=ON
ninja qSlicerBaseQTGUICxxTests
ctest -R qSlicerWidgetTest2 -V
```

but it fails (as expected, since VTK is not built with mesa) with:

```
281: Test command: /usr/src/Slicer-build/Slicer-build/Slicer "--launcher-no-splash" "--launch" "/usr/src/Slicer-build/Slicer-build/bin/qSlicerBaseQTGUICxxTests" "qSlicerWidgetTest2"
281: Test timeout computed to be: 1800
281: X Error: BadValue (integer parameter out of range for operation) 2
281:   Extension:    154 (Uknown extension)
281:   Minor opcode: 3 (Unknown request)
281:   Resource id:  0x0
281: X Error: GLXBadContext 169
281:   Extension:    154 (Uknown extension)
281:   Minor opcode: 5 (Unknown request)
281:   Resource id:  0x5400011
281: X Error: GLXBadContext 169
281:   Extension:    154 (Uknown extension)
281:   Minor opcode: 5 (Unknown request)
281:   Resource id:  0x5400011
281: error: [/usr/src/Slicer-build/Slicer-build/bin/qSlicerBaseQTGUICxxTests] exit abnormally - Report the problem.
```

pieper:
+1 to working on the SlicerDocker

thewtex:
Finalizing thewtex/SlicerDocker#5 is worth a try, but some false positive test failures are anticipated.

In the end, a Docker image with a Slicer build on Ubuntu 14.04 intended for testing GitHub pull requests, separate from the current SlicerDocker image that is primarily intended for Linux packaging, may be required.

MayeulChassagnard:
I think that I am going to create a new image where we will build the hole project + the tests and run them based on the thewtex/docker-opengl image (in view to be submitted to CDash). 

Indeed, separate build and test wouldn't be useful I think, and so we need to do everything in the same place. 

Are you agree with me?

thewtex:
@MayeulChassagnard That is fine, but a known limitation then will be that the package binaries created will only be compatible with Linux distributions with the same or newer versions of glibc in the thewtex/docker-opengl image (i.e. the version of glibc in Debian Jessie).

PR #563, opened 2016-08-05
ENH: Fixed warning when slice intersection display enabled for empty polydata

lassoan:
When a new model node was created (or polydata was reset in an existing model node) and model slice intersection display was enabled then VTK logged a lot of warnings. It is a normal use case to have empty model node, so no warnings should be logged.

Added a check so that now if the polydata is NULL or does not contain any points then the model slice intersection actor visibility is turned off.

How to reproduce:

---
p = vtk.vtkPolyData()
mn = slicer.vtkMRMLModelNode()
mn.SetAndObservePolyData(p)
slicer.mrmlScene.AddNode(mn)
mn.CreateDefaultDisplayNodes()
mn.GetDisplayNode().SetSliceIntersectionVisibility(1)
---

Prints into the log:

No input data

No input data

No input data

No input data

No input data

No input data


pieper:
LGTM  - thanks!

lassoan:
Committed r25267

PR #564, opened 2016-08-15
Fix widget size and reference counting tests

lassoan:
Fix for two issues that show up on the dashboards:
1. Max allowed widget width hint for HiDPI screens (fix: compute dynamically from screen size)
2. Node reference counting test errors (fix: remove test that tries to check undefined behavior of double-delete)

See details in the two commit comments.


jcfr:
:+1: 

Integrated in [r25286](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25286) and [r25287](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25287)

PR #570, opened 2016-08-19
ENH: Fractional labelmap display in Segmentations

Sunderlandkyl:
This commit modifies the displayable manager in Segmentations to allow an image data containing a range of scalar values to be visualized/interpolated, allowing fractional labelmaps to be displayed correctly.

### DIFF
--- a/Modules/Loadable/Segmentations/MRMLDM/vtkMRMLSegmentationsDisplayableManager2D.cxx
+++ b/Modules/Loadable/Segmentations/MRMLDM/vtkMRMLSegmentationsDisplayableManager2D.cxx
@@ -65,6 +65,9 @@
 #include <vtkTriangleFilter.h>
 #include <vtkCleanPolyData.h>
 #include <vtkCellArray.h>
+#include <vtkDoubleArray.h>
+#include <vtkIntArray.h>
+#include <vtkImageThreshold.h>
 
 // STD includes
 #include <algorithm>
@@ -146,6 +149,7 @@ public:
     vtkSmartPointer<vtkImageLabelOutline> LabelOutline;
     vtkSmartPointer<vtkLookupTable> LookupTableOutline;
     vtkSmartPointer<vtkLookupTable> LookupTableFill;
+    vtkSmartPointer<vtkImageThreshold> ImageThreshold;
     };
 
   typedef std::map<std::string, const Pipeline*> PipelineMapType;
@@ -484,6 +488,7 @@ vtkMRMLSegmentationsDisplayableManager2D::vtkInternal::CreateSegmentPipeline(std
   pipeline->LabelOutline = vtkSmartPointer<vtkImageLabelOutline>::New();
   pipeline->LookupTableOutline = vtkSmartPointer<vtkLookupTable>::New();
   pipeline->LookupTableFill = vtkSmartPointer<vtkLookupTable>::New();
+  pipeline->ImageThreshold = vtkSmartPointer<vtkImageThreshold>::New();
 
   // Set up image pipeline
   pipeline->Reslice->SetBackgroundColor(0.0, 0.0, 0.0, 0.0);
@@ -506,6 +511,10 @@ vtkMRMLSegmentationsDisplayableManager2D::vtkInternal::CreateSegmentPipeline(std
   pipeline->LookupTableFill->SetTableRange(0, 1);
   pipeline->LookupTableFill->SetTableValue(0,  0, 0, 0,  0);
   pipeline->LookupTableFill->SetTableValue(1,  0, 0, 0,  0);
+
+  pipeline->ImageThreshold->SetInputConnection(pipeline->Reslice->GetOutputPort());
+  pipeline->ImageThreshold->SetOutValue(1);
+  pipeline->ImageThreshold->SetInValue(0);
 
   // Image outline
   pipeline->LabelOutline->SetInputConnection(pipeline->Reslice->GetOutputPort());
@@ -791,11 +800,32 @@ void vtkMRMLSegmentationsDisplayableManager2D::vtkInternal::UpdateDisplayNodePip
       pipeline->PolyDataOutlineActor->SetVisibility(false);
       pipeline->PolyDataFillActor->SetVisibility(false);
 
+      // Set the range of the scalars in the image data from the ScalarRange field if it exists
+      // Default to the scalar range of 0.0 to 1.0 otherwise
+      double minimumValue = 0.0;
+      double maximumValue = 1.0;
+      vtkDoubleArray* scalarRange = vtkDoubleArray::SafeDownCast(
+        imageData->GetFieldData()->GetAbstractArray(vtkMRMLSegmentationsDisplayableManager2D::GetScalarRangeFieldName()));
+      if (scalarRange && scalarRange->GetNumberOfValues() == 2)
+        {
+        minimumValue = scalarRange->GetValue(0);
+        maximumValue = scalarRange->GetValue(1);
+        }
+
       // Set segment color
       pipeline->LookupTableOutline->SetTableValue(1,
         properties.Color[0], properties.Color[1], properties.Color[2], properties.Opacity2DOutline * displayNode->GetOpacity());
-      pipeline->LookupTableFill->SetTableValue(1,
-        properties.Color[0], properties.Color[1], properties.Color[2], properties.Opacity2DFill * displayNode->GetOpacity());
+      pipeline->LookupTableFill->SetNumberOfTableValues(maximumValue - minimumValue + 1); //TODO: potentially specify the number of values in the table
+      pipeline->LookupTableFill->SetRampToLinear();
+      pipeline->LookupTableFill->SetTableRange(minimumValue, maximumValue);
+      double hsv[3] = {0,0,0};
+      vtkMath::RGBToHSV(properties.Color, hsv);
+      pipeline->LookupTableFill->SetHueRange(hsv[0], hsv[0]);
+      pipeline->LookupTableFill->SetSaturationRange(hsv[1], hsv[1]);
+      pipeline->LookupTableFill->SetValueRange(hsv[2], hsv[2]);
+      pipeline->LookupTableFill->SetAlphaRange(0.0, properties.Opacity2DFill * displayNode->GetOpacity());
+      pipeline->LookupTableFill->ForceBuild();
+      pipeline->Reslice->SetBackgroundLevel(minimumValue);
 
       // Calculate image IJK to world RAS transform
       pipeline->SliceToImageTransform->Identity();
@@ -825,8 +855,17 @@ void vtkMRMLSegmentationsDisplayableManager2D::vtkInternal::UpdateDisplayNodePip
         {
         pipeline->Reslice->SetResliceTransform(pipeline->SliceToImageTransform);
         }
-      //TODO: Interpolate fractional labelmaps
+
+      // Set the interpolation mode from the InterpolationType field if it exists
+      // Default to nearest neighbor interpolation otherwise
       pipeline->Reslice->SetInterpolationModeToNearestNeighbor();
+      vtkIntArray* interpolationType = vtkIntArray::SafeDownCast(
+        imageData->GetFieldData()->GetAbstractArray(vtkMRMLSegmentationsDisplayableManager2D::GetInterpolationTypeFieldName()));
+      if (interpolationType && interpolationType->GetNumberOfValues() == 1)
+        {
+        pipeline->Reslice->SetInterpolationMode(interpolationType->GetValue(0));
+        }
+
       pipeline->Reslice->SetInputData(identityImageData);
       int dimensions[3] = {0,0,0};
       this->SliceNode->GetDimensions(dimensions);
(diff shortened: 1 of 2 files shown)


cpinter:
Thanks! Looks good to me.

In C++ we usually define constants via constant strings, but since we rely on python wrapping a lot in Slicer, I think the Get functions in the header can stay.

Modules/Loadable/Segmentations/MRMLDM/vtkMRMLSegmentationsDisplayableManager2D.cxx
```diff
@@ -781,11 +793,37 @@ void vtkMRMLSegmentationsDisplayableManager2D::vtkInternal::UpdateDisplayNodePip
...
 
+      // Set the range of the scalars in the image data from the ScalarRange field if it exists
+      // Default to the scalar range of 0.0 to 1.0 otherwise
+      double minimumValue = 0.0;
+      double maximumValue = 0.0;
+      vtkSmartPointer<vtkDoubleArray> scalarRange = vtkDoubleArray::SafeDownCast(
+        imageData->GetFieldData()->GetAbstractArray(vtkMRMLSegmentationsDisplayableManager2D::GetScalarRangeFieldName()));
+      if (scalarRange)
+        {
+        minimumValue = scalarRange->GetValue(0);
```
lassoan: Check if there are actually 2 values in the array before trying to retrieve them.

Modules/Loadable/Segmentations/MRMLDM/vtkMRMLSegmentationsDisplayableManager2D.cxx
```diff
@@ -781,11 +793,37 @@ void vtkMRMLSegmentationsDisplayableManager2D::vtkInternal::UpdateDisplayNodePip
       pipeline->PolyDataOutlineActor->SetVisibility(false);
       pipeline->PolyDataFillActor->SetVisibility(false);
 
+      // Set the range of the scalars in the image data from the ScalarRange field if it exists
+      // Default to the scalar range of 0.0 to 1.0 otherwise
+      double minimumValue = 0.0;
+      double maximumValue = 0.0;
+      vtkSmartPointer<vtkDoubleArray> scalarRange = vtkDoubleArray::SafeDownCast(
```
lassoan: You can use a simple pointer here, as the data set owns the array and it won't release it while we are reading the values from it.

Modules/Loadable/Segmentations/MRMLDM/vtkMRMLSegmentationsDisplayableManager2D.cxx
```diff
@@ -781,11 +793,37 @@ void vtkMRMLSegmentationsDisplayableManager2D::vtkInternal::UpdateDisplayNodePip
...
+        imageData->GetFieldData()->GetAbstractArray(vtkMRMLSegmentationsDisplayableManager2D::GetScalarRangeFieldName()));
+      if (scalarRange)
+        {
+        minimumValue = scalarRange->GetValue(0);
+        maximumValue = scalarRange->GetValue(1);
+        }
+      else
+        {
+        minimumValue = 0.0;
```
lassoan: No need for this, just initialize them alive to these values.

Modules/Loadable/Segmentations/MRMLDM/vtkMRMLSegmentationsDisplayableManager2D.cxx
```diff
@@ -815,8 +853,20 @@ void vtkMRMLSegmentationsDisplayableManager2D::vtkInternal::UpdateDisplayNodePip
...
-      //TODO: Interpolate fractional labelmaps
-      pipeline->Reslice->SetInterpolationModeToNearestNeighbor();
+
+      // Set the interpolation mode from the InterpolationType field if it exists
+      // Default to nearest neighbor interpolation otherwise
+      vtkSmartPointer<vtkIntArray> interpolationType = vtkIntArray::SafeDownCast(
+        imageData->GetFieldData()->GetAbstractArray(vtkMRMLSegmentationsDisplayableManager2D::GetInterpolationTypeFieldName()));
+      if (interpolationType)
+        {
+        pipeline->Reslice->SetInterpolationMode(interpolationType->GetValue(0));
```
lassoan: Same as above, for the other array (simple pointer, initial value, check if there is actually a value in the array)

Modules/Loadable/Segmentations/MRMLDM/vtkMRMLSegmentationsDisplayableManager2D.cxx
```diff
@@ -825,7 +875,36 @@ void vtkMRMLSegmentationsDisplayableManager2D::vtkInternal::UpdateDisplayNodePip
...
+          // Pad the extent of the image by one voxel in each direction
+          int extent[6] = {0,-1,0,-1,0,-1};
+          identityImageData->GetExtent(extent);
+          for (int i=0; i<2; ++i)
+            {
+            --extent[2*i];
+            ++extent[2*i+1];
+            }
+          pipeline->ImageConstantPad->SetInputData(identityImageData);
+          pipeline->ImageConstantPad->SetOutputWholeExtent(extent);
```
lassoan: Why do we need padding?

fedorov:
Kyle, can you provide some screenshots to demonstrate how this will look:
* single layer fractional map
* example of overlapping fractional maps
* user UI elements for controlling the appearance specific to fractional maps
* anything else that could help developers or users see the result without having to compile this PR

Thanks for working on this!

Modules/Loadable/Segmentations/MRMLDM/vtkMRMLSegmentationsDisplayableManager2D.cxx
Sunderlandkyl: In some structures there are non-empty values on the edges of the image, which causes problems with interpolation. Padding the image with empty pixels fixes this issue.

Sunderlandkyl:
one_structure
Single Layer

multiple_structure
Overlapping Fractional Labelmaps

- There are no unique UI elements for controlling fractional labelmap display. They are handled in the same way as binary labelmaps.
- The values for the scalar range, isosurface threshold and interpolation type are attached to the image using field data at the end of the conversion process and do not need to be set manually.

fedorov:
Thanks, this is very helpful! I am a bit confused though. In the first one, there is only one isocontour, but in the bottom there are multiple for the same map (it looks like it at least). What defines how many isocontours are shown?

> * There are no unique UI elements for controlling fractional labelmap display. They are handled in the same way as binary labelmaps.
> * The values for the scalar range, isosurface threshold and interpolation type are attached to the image using field data at the end of the conversion process and do not need to be set manually.

If not very difficult to implement, it might be helpful to be able to control this from the GUI. What defines the threshold and the isocontours? Some people might not want to see isocontours at all, and others may want to see it at 0%, or at some application-defined level. You may also want to integrate scalar bar with fractional labels. They are quite different from scalar labels to be handled without any special UI elements.

lassoan:
In Segmentations module widget you can already change these settings:
* show/hide contour in slice view (Show in 2D as contour), per segment (click-and-hold eye icon)
* show/hide fill in slice view (Show in 2D as fill), per segment (click-and-hold eye icon)
* show/hide as surface in 3D (Show in 3D), per segment (click-and-hold eye icon)
* change opacity of all segments in all views (Opacity slider)

In MRML you can set but there is no GUI for setting:
* 2D contour opacity (per segment) - default is 1
* 2D fill opacity (per segment) - default is 0.5
* 3D opacity (per segment) - default is 1

Probably we should show these opacity settings on the GUI and probably we should make it easy to change all these settings for multiple segments at once. Any suggestions for changing the current GUI to achieve these and fulfill most common use cases would be very welcome.

fedorov:
> In MRML you can set but there is no GUI for setting:
> 
> 2D contour opacity (per segment) - default is 1
> 2D fill opacity (per segment) - default is 0.5
> 3D opacity (per segment) - default is 1

It can be added UI, but it will be quite busy. Maybe for now, as long as they can be controlled in the MRML, it is ok to keep them out of the GUI.

My question was a bit different - there seem to be kinds types of visualizations that are specific to fractional segmentations, and it is not clear whether they are exposed at all (not in UI, but sounds like not in MRML either):

* isocontour threshold level (%) per segment
* number of isocontours per segment
* visualization of the isocontour (or perhaps this one is identical to the control of the binary label contour)

few more question: 
* do you expect fractional segmentations to be scaled between 0 and 1 on input, or scaling will be applied somehow depending on the data?
* what will be the process of loading fractional segmentations? will they be converted from integer/floating point images explicitly by the user using UI/API, or will there be an option in "Add data"? 

Of course, when we load fractional segs from DICOM, it should be straightforward, but with non-DICOM it will be quite cumbersome to initialize things with "Add data".

lassoan:
> number of isocontours per segment

There is always one isocontour per segment, which is used to determine the 2D contour and 3D surface. Threshold is always in the middle of the scalar range of the data (e.g., if scalar range is -128..128, threshold is at 0).

> Do you expect fractional segmentations to be scaled between 0 and 1 on input, or scaling will be applied somehow depending on the data?

Display is the same for fractional and binary labelmap (binary is a special case).
* pixel value<=scalar range min -> opacity=0%
* scalar range min < pixel value <= scalar range max -> opacity linearly interpolated (isocontour/surface is drawn at center of middle of the scalar range)
* pixel value>scalar range max => opacity=100%

By default the scalar range is (min=0, max=0), which corresponds to a binary labelmap.

If a data source specifies a custom scalar range (stored in vtkDataObject as field data), for example -128..127, then there will be a transition in opacity in 2D fill view and contours will be smooth.

> what will be the process of loading fractional segmentations? will they be converted from integer/floating point images explicitly by the user using UI/API, or will there be an option in "Add data"?

Mainly data importers would create fractional labelmaps. But we can have an option in Segmentation module's image import section and in "Add data" dialog (for Segmentation data type) where the user could choose to interpret the data as fractional labelmap (and allow specification of a scalar range).

fedorov:
>>     number of isocontours per segment
> 
> There is always one isocontour per segment, which is used to determine the 2D contour and 3D surface. Threshold is always in the middle of the scalar range of the data (e.g., if scalar range is -128..128, threshold is at 0).

Fractional segmentations I worked with in the past had floating point scalar range from 0 to 1, and the voxel value corresponded to the probability of tissue occurrence in the given voxel. If I understand correctly your explanation, you would assign threshold at 50% probability for such data. I am not sure how this would be useful, especially if 50% is fixed.

> By default the scalar range is (min=0, max=0), which corresponds to a binary labelmap.
> If a data source specifies a custom scalar range

I am confused by this. My understanding of the scalar range is [minimum voxel value in the image, maximum voxel value in the image].

lassoan:
Yes, if you have floating point image then you can set scalar range to (0.0, 1.0).

?> By default the scalar range is (min=0, max=0), which corresponds to a binary labelmap.
If a data source specifies a custom scalar range
> I am confused by this. My understanding of the scalar range is [minimum voxel value in the image, maximum voxel value in the image]. 

Yes, I agree that the nomenclature is not very clear. It's a step/ramp function that maps values in the image to displayed opacity. Can you suggest better names?

We could of course add various display options (color LUT, multiple contours - not just at 50%) but that would not be really appropriate for displaying segmentation. For example, a segment already has a color - how could we colorize a heat map? Also, each segment defines an "object" which has a single boundary - it would be probably quite confusing if we allowed some segments to produce multiple contours/surfaces (one at the most probable boundary and a few optimistic/pessimistic boundaries) - for this we could just create multiple segments.

lassoan:
So far the only use case for fractional labelmaps that we've considered is representing uncertainty in object boundaries of segments (due to partial volume effect, finite sampling, etc). Visualizing arbitrary probabilities that don't represent object boundaries is a different but important use case. It could be displayed as a simple foreground volume but display using segmentations would have the advantage that there is no limitation on how many layers are displayed and probability maps could be edited. So, we could think about generalizing the use of fractional labelmaps in segmentations.

fedorov:
Andras, the use case you didn't consider seems to me more widely applicable. I would suggest you at least try how your implementation will handle EMSegmenter atlases.

I don't know if anyone is editing probability atlases in editor, main use case is visualization.

I appreciate your explanations, but I feel it would be best for me to actually try things out, or see a demo. Don't wait for me to merge this. I think fractional segs are not widely used, so it should not be a lot of damage one way or another. We can discuss later. This conversation helped me understand the capabilities and the use case that motivated this work.

lassoan:
Yes, thanks, it was a useful discussion.

lassoan:
@Sunderlandkyl There are several border/padding options in the reslice image filter. Can you double-check if we really need implement custom padding?

cpinter:
Andriy, a little background for all this.

The fractional labelmap feature was originally an experiment solely within SlicerRT to see how much such representation of an RT structure (same memory footprint as for a binary labelmap but much finer data contained) improves metrics used for treatment planning and evaluation. This experiment has yielded a new segment representation (that as you point out, can be used for other things as well), and during the integration process from a branch to the SlicerRT core, it seemed to be the best decision to add this as a special case for the segmentation displayable manager (instead of subclassing it and unregistering the generic one).

Currently the only way to obtain a fractional labelmap is to convert from a surface representation within a segmentation node, and this conversion and the representation is only available if you install SlicerRT. Also, the conversion is quite slow, and it's not intended for "everyday" use, but this integration is part of the effort to maintain what we developed. So this won't change anything in Slicer core, or according to our plans SlicerRT.

It is good to keep in mind that it is already implemented in SlicerRT, and when the need arises for this representation to be used by a project, it can be used as a starting point.

pieper:
Does the segmentation editor allow brushes with 'soft' edges?  This would
be a very important use case for the fractional segmentations.  For example
painting with a more perfectly round brush or rasterizing a complex drawing
by antialiasing the edges could lead to much nicer segmentation results and
models that are more accurate and smoother.

cpinter:
Agreed, soft brushes would be a good use case for this. As far as I know, there is no soft brush implemented in segment editor, as the representation is only handled by SlicerRT right now. If there is a project that would use (and be able to process) such segmentations then it would be fairly easy to add this feature I think.

pieper:
I don't think there's any particular rush for the soft brush, but typically when people have low-res data their only choice is to upsample in order to get more decently smooth models.  But with soft segmentation they could work with the original data and that would be nice.  It'll be good to keep this in mind for the future.

cpinter:
Definitely!

Sunderlandkyl:
@lassoan From what I've tried, none of the vtkImageReslice border/padding options seem to solve this issue.

jcfr:
Summary of dev hangout:
* good to integrate
* Note: if DICOM flag "is factional" is true, import could be done directly into this representation

cpinter:
Integrated in http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25315

PR #571, opened 2016-08-23
BUG: Add the extension third party dirs to python additional launcher settings

naucoin:
This change will cause an addition to the inner build
AdditionalLauncherSettings.ini file in extensions with a superbuild
set up, adding the top level extension third party lib directory to the python
path settings. This will allow third party python packages to be loaded during
module tests.

Issue https://github.com/SBU-BMI/SlicerOpenCV/issues/12


jcfr:
LGTM :+1: 

Feel free to commit using `git svn dcommit`

naucoin:
Thanks for the discussion on the dev hangout, and done:
http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25305

PR #572, opened 2016-08-23
COMP: use proper variable evaluation in SlicerExtensionCPack

ihnorton:
Should fix UKFTractography extension packaging due to using incorrect fixup script. Has been broken since 195c0d55.

See build error: http://slicer.cdash.org/buildSummary.php?buildid=875941

### DIFF
--- a/CMake/SlicerExtensionCPack.cmake
+++ b/CMake/SlicerExtensionCPack.cmake
@@ -161,7 +161,7 @@ if(APPLE)
 
   set(msg "Checking if extension type is SuperBuild")
   message(STATUS "${msg}")
-  if(DEFINED ${EXTENSION_NAME}_SUPERBUILD)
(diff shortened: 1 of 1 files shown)


ihnorton:
cc @ljod @yrathi

ihnorton:
To clarify, this is necessary because UKFTractography uses `UKFTractography_SUPERBUILD` as an `option`, so it is always defined.

ihnorton:
(CI non-failure due to this changing CMake file)

jcfr:
Thanks for the patch

Integrated in [r25303](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25303)

ihnorton:
Looks good now, thanks: http://slicer.cdash.org/buildSummary.php?buildid=876325

PR #573, opened 2016-08-26
BUG: Fix vtkMRMLNonlinearTransformNodeTest1

jcfr:
This commit fixes the reading of ITKv3 transforms by addressing
a regression introduced in [r24875](https://github.com/Slicer/Slicer/commit/12743c6fcebb67ff9b22319cbe52a322a3738134) (BUG: Fix transform tests by not
using dynamic_cast in vtkITKTransformConverter)

It basically ensures that executable created using compiler (e.g g++4.4.3)
[not supporting the use of dynamic_cast with templated classes instantiated
in a translation unit different than the one where they are defined] still
work as expected.

See #4139

### DIFF
--- a/Libs/MRML/Core/vtkITKTransformConverter.h
+++ b/Libs/MRML/Core/vtkITKTransformConverter.h
@@ -341,8 +341,17 @@ bool vtkITKTransformConverter::SetVTKBSplineParametersFromITKGeneric(
     return false;
     }
 
-  typename BSplineTransformType::Pointer bsplineItk = dynamic_cast< BSplineTransformType* >( warpTransformItk.GetPointer() );
-  if (!bsplineItk.GetPointer())
+  typename BSplineTransformType::Pointer bsplineItk;
+  std::string warpTransformItkName = warpTransformItk->GetNameOfClass();
+  if (warpTransformItkName == "InverseBSplineTransform" ||
+      warpTransformItkName == "BSplineTransform" ||
+      warpTransformItkName == "BSplineDeformableTransform" ||
+      warpTransformItkName == "InverseBSplineDeformableTransform"
+      )
+    {
+    bsplineItk = static_cast< BSplineTransformType* >( warpTransformItk.GetPointer() );
+    }
+  else
     {
     return false;
     }


jcfr:
This commit fixes test `vtkMRMLNonlinearTransformNodeTest1`

More work is still needed to fix `py_BRAINSFitRigidRegistrationCrashIssue4139`

An helpful regex to lookup potential faulty `dynamic_cast` is the following:

```
$ ack "dynamic\_cast\<[ a-zA-Z0-9\_]+Trans[ a-zA-Z0-9\_\*]+>"
[...]

Modules/CLI/ExpertAutomatedRegistration/ITKRegistrationHelper/itkAnisotropicSimilarityLandmarkBasedTransformInitializer.txx
64:  AnisotropicSimilarity3DTransformType *testPtr = dynamic_cast<AnisotropicSimilarity3DTransformType *>(
70:  else if( dynamic_cast<Rigid2DTransformType *>(this->m_Transform.GetPointer() ) )
112:      AnisotropicSimilarity3DTransformType *transform = dynamic_cast<AnisotropicSimilarity3DTransformType *>(
431:      Rigid2DTransformType *transform = dynamic_cast<Rigid2DTransformType *>(

Modules/CLI/ExpertAutomatedRegistration/ITKRegistrationHelper/itkRigidImageToImageRegistrationMethod.txx
35:    this->SetTransform( dynamic_cast<RigidTransformType *>(
43:    this->SetTransform( dynamic_cast<RigidTransformType *>(
105:  return dynamic_cast<const TransformType  *>( Superclass::GetTransform() );
129:    dynamic_cast<RigidTransformType *>( this->GetTransform() );

Modules/CLI/BSplineToDeformationField/BSplineToDeformationField.cxx
45:    dynamic_cast<BSplineDeformableTransformType *>( transformFile->GetTransformList()->front().GetPointer() );

Modules/CLI/ResampleDTIVolume/itkDiffusionTensor3DNonRigidTransform.txx
79:    typename AffineTransform::Pointer newAffine = dynamic_cast<AffineTransform *>( newTransform.GetPointer() );
```

jcfr:
Cc: @thewtex @cpinter @blowekamp @lassoan

blowekamp:
I sent a note to the slicer developer list about the cmake option of disabling itk's hidden visibility. This would make these problematic hidden symbols globally visible where the runtime linker should uniquely resolve them. Was usage of this feature considered?

jcfr:
Thanks for the reminder. Would it make sense to have this option enabled by
default in ITK ? That it would be restored across a broad range of platform
and compiler.

thewtex:
> Thanks for the reminder. Would it make sense to have this option enabled by default in ITK ? That it would be restored across a broad range of platform and compiler.

+1 for enabling the option by default in ITK.

CC: @fbudin69500

thewtex:
PS> The commit LGTM

fbudin69500:
The patch looks good to me. Would disabling ITK's hidden visibility slow down ITK's loading time because more symbols would be exposed?

thewtex:
@fbudin69500 good point -- yes, hidden visibility is very important for load time.

blowekamp:
Regarding ITK's visibility. Yes there will be more global symbols and causing larger symbol tables and some amount of more computation to resolve them. This could be counted and time to measure. But on the other side, it ensures that the same compiled code is used globally, which may have performance benefits.

IMHO, I don't think this will still be significant. As the ITK libraries themselves are rather small and don't (or didn't) instantiate much of ITK. There are likely more symbols in ITK's third party libraries than ITK it's self. Where ITK can generate a large number of symbols is when a filter sequence is instantiated over a large number of type, like is done with some CLI. Or in wrapping ITK.

jcfr:
@thewtex I am looking into the other failure related to `py_BRAINSFitRigidRegistrationCrashIssue4139` and it looks like the following exception is triggered: https://github.com/InsightSoftwareConsortium/ITK/blob/0928b5e3b8487fa4e9242284330279db4ec86233/Modules/Core/Common/include/itkImageBase.hxx#L321-L327

I will be investigating further later this afternoon, in mean time, do you have any idea ? The version of compiler used on that machine is `g++ Ubuntu 4.4.3-4ubuntu5.1`

thewtex:
@jcfr Interesting.  GCC 4.4 is pretty old and is known to have its quirks.

It may need more investigation, but that section of code looks like a candidate for conversion from `dynamic_cast` to `static_cast`...

CC: @hmjohnson

thewtex:
CC: @hjmjohnson

blowekamp:
There are a number of other dynamic_cast in that file. What would be the new guideline on when to use static_cast vs dynamic_cast?

thewtex:
> There are a number of other dynamic_cast in that file. What would be the new guideline on when to use static_cast vs dynamic_cast?

`static_cast` can be used when it is known logically that the cast should always be valid. We can use the `itkDynamicCastInDebugMode` macro, which uses `dynamic_cast` in Debug mode to help with logic errors.

jcfr:
Updates:

* Commit fixing `vtkMRMLNonlinearTransformNodeTest1` integrated as [r25318](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25318)

* Using `static_cast` in `itkImageBase` allow the test `py_BRAINSFitRigidRegistrationCrashIssue4139` to pass. I will come up with an ITK patch later on

blowekamp:
I apologize for diverging into an ITK discussion, but I strongly disagree with the proposed solution. We can't see the forest burning through the smoke.

The ImageBase::CopyInformation is a virtual method of DataObject::CopyInformation and public. There is no logical guarantee that this method can't be passed an arbitrary DataObject. This method is intended to be used as part of the generic pipeline and data object methods. I have written code which does this when generically handing multiple inputs and output in ProcessObject classes.

There are numerous potential dynamic_casts which can fail due to private symbols. With a known failing designed it's not responsive to only fix the ones which fail for our tests. Nor do I believe that it is reasonable or safe to convert many of them to static_casts. There are several cases I can recall where a static_cast was converted to dynamic_casts because of problems when writing or using a filter and messing up the inputs. Yes, even setting inputs in ITK is not type safe, dynamic_casts are needed. Remember that the ProcessObject uses base DataObject pointers and we rely on the dynamic_cast to correctly perform operations as needed in the pipeline.

With the known casting issue for certain compilers, it would be very unexpected behavior in debug vs release mode if itkDynamicCastInDebugMode was used.

I do think that making the ITK symbols global is the easiest and safest approach. We have a CMake flag it would be easy of Slicer to test. I would like to see some analysis supporting that this approach "was" ( it's the way things were ), slower than the current. I am not sure this criticism is fully justified for the generic case.

Another reasonable approach might be to adopt something like VTK's SafeDownCast in some existing itkNew macro. But this would be more work.

This post is long enough. I hope the opinion is welcomed.

lassoan:
I fully agree. I think hiding symbols is only appropriate when building one, monolithic, locked-down application. We cannot use a C++ library without reliable dynamic_cast. The issues that our tests found are just examples, there could be many problems at other places that we don't test thoroughly.

thewtex:
> There is no logical guarantee that this method can't be passed an arbitrary DataObject. This method is intended to be used as part of the generic pipeline and data object methods.

There is an existing logical check that the cast should be passed an ImageBase *. In the current code:

   https://github.com/InsightSoftwareConsortium/ITK/blob/0928b5e3b8487fa4e9242284330279db4ec86233/Modules/Core/Common/include/itkImageBase.hxx#L321-L327

an exception is thrown when the cast fails.

jcfr:
> hiding symbols is only appropriate when building one, monolithic, locked-down application. We cannot use a C++ library without reliable dynamic_cast. The issues that our tests found are just examples, there could be many problems at other places that we don't test thoroughly.

Agreed. I will look into creating a sample project with the ideal behavior, demonstrate it is failing on compilers with known issue, enable visibility of inlined symbols to confirm  it addresses the problem.

I think I will improve the project I created back in December when I was working on having ITK IOs build as shared libraries (See https://github.com/jcfr/ITKIOFactoryRegistrationTest)

> re: failure of EMSegment tests

In the mean time, to continue the infamous saga, this PR will fix remaining failing tests. See #575

jcfr:
Good news is that this MR from @blowekamp already implement a test checking for incorrect behavior of `dynamic_cast`, see http://review.source.kitware.com/#/c/21097/

blowekamp:
@jcfr That patch was merged, and it's been on the dashboard. It is not failing on as many stable builds as I thought it would [1]. It is with Xcode ~7.3, that we see it on ITK dashboard. However, we don't have good and clear coverage of OS X Xcode versions. It may be good to run it on Travis to get a clear picture of the problem [2]. I have enable SHARED libraries on most lhcp-rh6 gcc builds, and we don't see the failing tests there. So this report of the dynamic_cast with gcc 4.4 is news to me ( though I am not following things too closely ).

> Agreed. I will look into creating a sample project with the ideal behavior, demonstrate it is failing on compilers with known issue, enable visibility of inlined symbols to confirm it addresses the problem.

My current theory is with the "ENABLE_SHARED" option enabled "-fvisibility=hidden". Specifically the ITKIOTransformBase is very large and instantiates many core ITK classes, look at the size of it at 15MB. The USE_COMPILER_HIDDEN_VISIBILITY CMake option should disable usage of this flag ( a separate flag could be used for WrapITK, if desired )

Actually, I'll hack at [2] to get it testing this problem with Xcode versions and to demonstrate ITK CMake option. Unless there is another enthusiastic volunteer.

[1] https://open.cdash.org/testSummary.php?project=2&name=itkIOTransformTxtTest&date=2016-08-31
[2] https://github.com/blowekamp/ITK/commit/4ee1ad5271412457772c43b445e4c9d1ea40c80e
[3] https://gcc.gnu.org/wiki/Visibility

Libs/MRML/Core/vtkITKTransformConverter.h
```diff
@@ -341,8 +341,17 @@ bool vtkITKTransformConverter::SetVTKBSplineParametersFromITKGeneric(
...
     }
 
-  typename BSplineTransformType::Pointer bsplineItk = dynamic_cast< BSplineTransformType* >( warpTransformItk.GetPointer() );
-  if (!bsplineItk.GetPointer())
+  typename BSplineTransformType::Pointer bsplineItk;
+  std::string warpTransformItkName = warpTransformItk->GetNameOfClass();
+  if (warpTransformItkName == "InverseBSplineTransform" ||
+      warpTransformItkName == "BSplineTransform" ||
+      warpTransformItkName == "BSplineDeformableTransform" ||
+      warpTransformItkName == "InverseBSplineDeformableTransform"
```
blowekamp: The transforms that 3 template parameters do we have any guarantee that the ParamterType? Dimension or Spline order match?

blowekamp:
I created a Travis CI configuration to build and run the itkIOTransformTxtTest [1]. This test loads a transform and then performs a dynamic_cast. The TransformIO has hidden symbols instantiating the transforms. The builds with Xcode 8, 7.3, 7.2, 7.1, 7.0, and 6.4 all fail with the dynamic_cast. When the USE_COMPILER_HIDDEN_VISIBILITY flag is disabled they 7.3 passes.

I hope this helps. It is news to me that this fails with all those compilers.

[1] https://github.com/blowekamp/ITK/blob/3b2751e0f7f01765577e0a8fd2e84a3cb976ec5a/.travis.yml

ihnorton:
@blowekamp  I think the reason the test is failing with `USE_COMPILER_HIDDEN_VISIBILITY` (i.e. `-fvisibility=hidden` on \*nix systems) is that `itk::AffineTransform` does not have an export specifier in the template class definition. Therefore, the typeinfo is unique and private for each instance in a compilation unit and cannot be coalesced (across shared library and executable in the test case).

Adding `ITK_ABI_EXPORT` to the class definition fixes the test for me (on OSX 10.11): `dynamic_cast succeeded for: 0x102919c90`.

```
diff --git a/Modules/Core/Transform/include/itkAffineTransform.h b/Modules/Core/Transform/include/itkAffineTransform.h
index bf2d138..d4efddf 100644
--- a/Modules/Core/Transform/include/itkAffineTransform.h
+++ b/Modules/Core/Transform/include/itkAffineTransform.h
@@ -99,7 +99,7 @@ template<
   typename TParametersValueType = double,
   unsigned int NDimensions = 3 >
 // Number of dimensions in the input space
-class AffineTransform:
+class ITK_ABI_EXPORT AffineTransform:
   public MatrixOffsetTransformBase< TParametersValueType, NDimensions, NDimensions>
 {
 public:
```

The difference it makes is as follows:

current ITK head:
```
bwh003265:ITK-build-r inorton$ nm -mo lib-bkup/libITKIOTransformBase-4.11.dylib | grep __ZTIN3itk15AffineTransformIdLj
lib-bkup/libITKIOTransformBase-4.11.dylib: 00000000004b9630 (__DATA,__data) non-external (was a private external) __ZTIN3itk15AffineTransformIdLj2EEE
lib-bkup/libITKIOTransformBase-4.11.dylib: 00000000004b9e00 (__DATA,__data) non-external (was a private external) __ZTIN3itk15AffineTransformIdLj3EEE
lib-bkup/libITKIOTransformBase-4.11.dylib: 00000000004ba590 (__DATA,__data) non-external (was a private external) __ZTIN3itk15AffineTransformIdLj4EEE
lib-bkup/libITKIOTransformBase-4.11.dylib: 00000000004bad20 (__DATA,__data) non-external (was a private external) __ZTIN3itk15AffineTransformIdLj5EEE
lib-bkup/libITKIOTransformBase-4.11.dylib: 00000000004bb4b0 (__DATA,__data) non-external (was a private external) __ZTIN3itk15AffineTransformIdLj6EEE
lib-bkup/libITKIOTransformBase-4.11.dylib: 00000000004bbc40 (__DATA,__data) non-external (was a private external) __ZTIN3itk15AffineTransformIdLj7EEE
lib-bkup/libITKIOTransformBase-4.11.dylib: 00000000004bc3d0 (__DATA,__data) non-external (was a private external) __ZTIN3itk15AffineTransformIdLj8EEE
lib-bkup/libITKIOTransformBase-4.11.dylib: 00000000004bcb60 (__DATA,__data) non-external (was a private external) __ZTIN3itk15AffineTransformIdLj9EEE
```

with patch:
```
bwh003265:ITK-build-r inorton$ nm -mo bin/ITKIOTransformInsightLegacyTestDriver | grep __ZTIN3itk15AffineTransformIdLj
bin/ITKIOTransformInsightLegacyTestDriver: 000000010013d630 (__DATA,__data) weak external __ZTIN3itk15AffineTransformIdLj10EEE
bin/ITKIOTransformInsightLegacyTestDriver: 000000010013cf30 (__DATA,__data) weak external __ZTIN3itk15AffineTransformIdLj4EEE
bwh003265:ITK-build-r inorton$ nm -mo lib/libITKIOTransformBase-4.11.dylib | grep __ZTIN3itk15AffineTransformIdLj
lib/libITKIOTransformBase-4.11.dylib: 00000000004b97b0 (__DATA,__data) weak external __ZTIN3itk15AffineTransformIdLj2EEE
lib/libITKIOTransformBase-4.11.dylib: 00000000004b9f80 (__DATA,__data) weak external __ZTIN3itk15AffineTransformIdLj3EEE
lib/libITKIOTransformBase-4.11.dylib: 00000000004ba710 (__DATA,__data) weak external __ZTIN3itk15AffineTransformIdLj4EEE
lib/libITKIOTransformBase-4.11.dylib: 00000000004baea0 (__DATA,__data) weak external __ZTIN3itk15AffineTransformIdLj5EEE
lib/libITKIOTransformBase-4.11.dylib: 00000000004bb630 (__DATA,__data) weak external __ZTIN3itk15AffineTransformIdLj6EEE
lib/libITKIOTransformBase-4.11.dylib: 00000000004bbdc0 (__DATA,__data) weak external __ZTIN3itk15AffineTransformIdLj7EEE
lib/libITKIOTransformBase-4.11.dylib: 00000000004bc550 (__DATA,__data) weak external __ZTIN3itk15AffineTransformIdLj8EEE
lib/libITKIOTransformBase-4.11.dylib: 00000000004bcce0 (__DATA,__data) weak external __ZTIN3itk15AffineTransformIdLj9EEE
```

Having the typeinfo set as **weak external** is required everywhere in order for the linker to coalesce (see also: http://www.russellmcc.com/posts/2013-08-03-rtti.html)

(doing likewise for the other templated transform classes such as `BSplineDeformable` should probably fix the Slicer test issues mentioned here, without requiring the work-around in this PR).

ihnorton:
cc @hjmjohnson who appears to have removed the (a?) precursor to `ITK_ABI_EXPORT` in https://github.com/InsightSoftwareConsortium/ITK/commit/b788601abbe

jcfr:
@ihnorton Very nice. That is most likely one of the last pieces of the puzzle.

After validating this approach on Visual Studio, it will make sense to integrate it.

ihnorton:
A couple other notes:
- GCC 4.5+ (possibly later, I haven't double-checked) reverted to using `strcmp` for the typeinfo comparison so this may be less of an issue on newer platforms.
- at some point Apple used to compile libc++abi with `_LIBCXX_DYNAMIC_FALLBACK` which would allow a similar thing (see https://llvm.org/svn/llvm-project/libcxxabi/trunk/src/private_typeinfo.cpp). I'm not sure when they changed it, but keep in mind for compiler/sdk/OS permutations...

blowekamp:
This sounds like a good approach to get all the symbols to play nicely together!

But careful consideration needs to be made as to which symbols are public(weak), and private. The bug advantage of using the hidden visibility is that many of the instantiated template classes are hidden. This is clearly a major problem for some of the DataObjects like Images and Transform which rely on dynamic_cast, it could be a more subtle problem for other classes.

So this approach will require more manual intervention in determining which template class have hidden or default visibility.

ihnorton:
@blowekamp I'm kind of surprised that this has not been a bigger issue given the number of `dynamic_cast`s within ITK itself, and the number of separate shared libraries...

Regarding @lassoan's point about testing:
- should probably turn on `-Wweak-vtables` on Clang-based platforms which will help catch some of these issues.
- could replace `dynamic_cast` with a customised version that either asserts out in debug mode, or ideally provides a specialization for ITK-derived classes with fallback to `static_cast` based on a `GetNameOfClass` check.

(Alternatively, on Mac, we could rename every binary to `Adobe Illustrator` and everything will Just Work)

blowekamp:
Please consider the follow patch to resolve the problem of the private symbols "emitted" form the TransformIO:

http://review.source.kitware.com/#/c/21496/

This patch pulls out the TransformFactory into a separate module, which is compiled without "hidden" visibility. So all the instantiated itk::Transform class symbols should be publicly visible, along with important classes they use like itk::Image.

The one caveat, is the problem of Windows registering factories multiple time. This was fixed by forcing an "itkclassname" name be registered only once.

The other topic JC referred to:
http://review.source.kitware.com/#/c/21102/

This does a similar thing with the TransformFactory, but places the registration in the module of the transform. While this is a more logical place to hold them there are some other trickiness not yet handled with the visibility of classes the transform use.

The first patch just fixes these apparent issues discussed here from symbols "emitted" from the TransformIO. There likely are other ENABLE_SHARED libraries which could cause problems by "emitting" private symbols. I hope that this is the only factory directly producing a lot of them.

HTH

ihnorton:
@blowekamp your topic branch fixes both the ITK test and the `NonlinearTransformNode1` Slicer test (with Jc's change here reverted).

Maybe the mechanism is obvious to others more familiar with ITK build system, but it took some digging for me to understand the commit comment about `ENABLE_SHARED` (which I now understand as: "most other modules are compiled with `ENABLE_SHARED`, passed to the ITK module CMake macro, which causes `set_target_properties` to be passed for hidden visibility, when `USE_COMPILER_HIDDEN_VISIBILITY=ON`. Therefore compiling without `ENABLE_SHARED` means that we don't pass the hidden-visibility flags, and all template instantiations within this module will have default (exported) visibility").

The double-negative implication was not really clear (to me) from the commit comment only mentioning `ENABLE_SHARED`.

jcfr:
@ihnorton  Thanks for the update. This is great news !

As soon as there is consensus  on the "right" approach:

* http://review.source.kitware.com/#/c/21496/ : `ENH: Separate TransformFactory class into separate module`

or 

* http://review.source.kitware.com/#/c/21499/ : `WIP: Add SafeDownCast to LightObject classes`

We will update Slicer accordingly and revert the `dynamic_cast` hacks in Slicer, EMSegment and BRAINSTools

jcfr:
Closing. Similar fix has been integrated in https://github.com/Slicer/Slicer/commit/c9d952b6e6dc7c8e7e7da99281497782ca1dd4cc / r25365

PR #574, opened 2016-08-29
Improved node saving and loading

lassoan:
* Storable node can decide what is the most appropriate storage node, based on the requested file name and storable node content. This allows simplifications and removal of "special case" hacks throughlout the code base.

* Allow overriding default write file extension in storage nodes (instead of hardcoding DefaultWriteFileExtension string, it is now stored in a member variable). For example, changing default model node saving extension can be achieved by this:
```
msn=slicer.vtkMRMLModelStorageNode()
msn.SetDefaultWriteFileExtension('stl')
slicer.mrmlScene.AddDefaultNode(msn)
```
* Improved composite file extension support (.nii.gz, .seg.nrrd, ...). In "Add data dialog" the most specifically matched extension's reader is selected by default (for example, .nrrd is more specific than .*; .seg.nrrd is more specific than .nrrd). File extension matching has been more robust: instead of using heuristics, such as assuming .gz means a composite file extension
and collecting all possible extension from all storage nodes; now extensions are mostly managed by the storage node, which knows exactly what extensions are possible.

In the future, composite extensions could be used for more file formats where we use the same extension for storing different node types (annotations, markups, labelmap/scalar volumes, etc) and there would be no more needs for heuristics or manual user choices for selecting the correct node type.

### DIFF
--- a/Base/QTCore/qSlicerCoreIOManager.cxx
+++ b/Base/QTCore/qSlicerCoreIOManager.cxx
@@ -88,39 +88,27 @@ qSlicerFileReader* qSlicerCoreIOManagerPrivate::reader(const QString& fileName)c
 //-----------------------------------------------------------------------------
 QList<qSlicerFileReader*> qSlicerCoreIOManagerPrivate::readers(const QString& fileName)const
 {
-  QList<qSlicerFileReader*> matchingReaders;
-  // Some readers ("DICOM (*)" or "Scalar Overlay (*.*))" can support any file,
-  // they are called generic readers. They might not be the best choice to read
-  // the file as it might exist a more specific reader to read it.
-  // So let's add generic readers at the end of the reader list.
-  QList<qSlicerFileReader*> genericReaders;
+  // Use a map so that we can access readers sorted by confidence.
+  // The more specific the filter that was matched, the higher confidence
+  // that the reader is more appropriate (e.g., *.seg.nrrd is more specific than *.nrrd;
+  // *.nrrd is more specific than *.*)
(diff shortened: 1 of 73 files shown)


Base/QTCore/qSlicerCoreIOManager.cxx
```diff
@@ -88,39 +88,27 @@ qSlicerFileReader* qSlicerCoreIOManagerPrivate::reader(const QString& fileName)c
...
     {
-    QStringList matchingNameFilters = reader->supportedNameFilters(fileName);
-    if (matchingNameFilters.count() == 0)
-      {
-      continue;
-      }
-    // Generic readers must be added to the end
-    foreach(const QString& nameFilter, matchingNameFilters)
+    int longestExtensionMatch = 0;
+    QStringList matchedNameFilters = reader->supportedNameFilters(fileName, &longestExtensionMatch);
```
jcfr: Add a comment specifying that `longestExtensionMatch` is updated by `supportedNameFilters`

jcfr:
:+1: Looks good to me

Thanks for working on this. This was long overdue. 

Some possibly related issues:
* http://na-mic.org/Mantis/view.php?id=3373
* http://na-mic.org/Mantis/view.php?id=2479

pieper:
👍

Base/QTCore/qSlicerCoreIOManager.cxx
lassoan: OK, I add this.

lassoan:
Integrated into rev25326

PR #579, opened 2016-09-12
COMP: Fix windows compilation with Slicer_BUILD_PARAMETERSERIALIZER_S…

vovythevov:
…UPPORT

The JsonCpp_LIBRARY needs to be passed to SlicerExecutionModel as a
CMAKE_ARGS because it contains the result of the ${CMAKE_CFG_INTDIR}
variable that needs to be evaluated.

cc: @jcfr, @lassoan

### DIFF
--- a/SuperBuild/External_SlicerExecutionModel.cmake
+++ b/SuperBuild/External_SlicerExecutionModel.cmake
@@ -22,10 +22,11 @@ endif()
 
 if(NOT DEFINED SlicerExecutionModel_DIR AND NOT ${CMAKE_PROJECT_NAME}_USE_SYSTEM_${proj})
 
-  set(EXTERNAL_PROJECT_OPTIONAL_ARGS)
+  set(EXTERNAL_PROJECT_OPTIONAL_CMAKE_CACHE_ARGS)
+  set(EXTERNAL_PROJECT_OPTIONAL_CMAKE_ARGS)
 
   if(APPLE)
-    list(APPEND EXTERNAL_PROJECT_OPTIONAL_ARGS
+    list(APPEND EXTERNAL_PROJECT_OPTIONAL_CMAKE_CACHE_ARGS
       -DSlicerExecutionModel_DEFAULT_CLI_EXECUTABLE_LINK_FLAGS:STRING=-Wl,-rpath,@loader_path/../../../
       )
   endif()
@@ -36,7 +37,7 @@ if(NOT DEFINED SlicerExecutionModel_DIR AND NOT ${CMAKE_PROJECT_NAME}_USE_SYSTEM
 
   macro(_set var type value)
     set(${var} ${value})
-    list(APPEND EXTERNAL_PROJECT_OPTIONAL_ARGS
+    list(APPEND EXTERNAL_PROJECT_OPTIONAL_CMAKE_CACHE_ARGS
       -D${var}:${type}=${value}
       )
   endmacro()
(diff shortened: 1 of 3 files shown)


SuperBuild/External_SlicerExecutionModel.cmake
```diff
@@ -56,9 +56,14 @@ if(NOT DEFINED SlicerExecutionModel_DIR AND NOT ${CMAKE_PROJECT_NAME}_USE_SYSTEM
...
-    _set(JsonCpp_INCLUDE_DIR PATH ${JsonCpp_INCLUDE_DIR})
-    _set(JsonCpp_LIBRARY PATH ${JsonCpp_LIBRARY})
     _set(ParameterSerializer_DIR PATH ${ParameterSerializer_DIR})
+    _set(JsonCpp_INCLUDE_DIR PATH ${JsonCpp_INCLUDE_DIR})
+
+    # JsoncCpp_LIBRARY needs to be added as a CMAKE_ARGS because it contains an
+    # expression that needs to be evaluated
+    list(APPEND EXTERNAL_PROJECT_OPTIONAL_ARGS
+      CMAKE_ARGS -DJsonCpp_LIBRARY:PATH=${JsonCpp_LIBRARY}
+      )
```
jcfr: Nice. This will indeed fix the problem.

But I would prefer that you:
1. Rename `EXTERNAL_PROJECT_OPTIONAL_ARGS` to `EXTERNAL_PROJECT_OPTIONAL_CMAKE_CACHE_ARGS` like it is done in `External_ITKv4.cmake`
2. set `EXTERNAL_PROJECT_OPTIONAL_CMAKE_ARGS` to an empty string outside of the `if`, append optonal args to it. Then, just before the call to `ExternalProject_Add`, you woud do:

```
# Note: Remove this block if non optional CMAKE_ARGS are added to ExternalProject_Add.
if(NOT EXTERNAL_PROJECT_OPTIONAL_CMAKE_ARGS STREQUAL "")
  set(EXTERNAL_PROJECT_OPTIONAL_CMAKE_ARGS
    CMAKE_ARGS
    ${EXTERNAL_PROJECT_OPTIONAL_CMAKE_ARGS})
endif()
```

Doing so will make the code easier to maintain.

Thanks

vovythevov:
@jcfr: Thanks for your suggestion. I was going for the smallest change possible but I like your suggestion better. I updated the commit to reflect that.

jcfr:
:+1: 

Could  you also consider re-setting the Slicer_BUILD_PARAMETERSERIALIZER_SUPPORT to 1 ?

jcfr:
> Could you also consider re-setting the Slicer_BUILD_PARAMETERSERIALIZER_SUPPORT to 1 ? 

After confirming it addresses the problem reported by @lassoan  ...

lassoan:
Thank you! When you test this, delete at least the Slicer-build directory to make sure the build doesn't succeed because of leftover files from previous builds.

vovythevov:
@jcfr, @lassoan: Integrated in [r25351](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25351), [r25352](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25352) and [r25353](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25353)

jcfr:
Very nice :+1:  also thanks for fixing the test 


Nitpick / Note for later:

* Serialization  != Serialazation 

* first 7 chars of SHA are sufficient. Both `5d0b7e07df4ea0b3be00c11d8200166c0f135b2e` and `5d0b7e0` render as 5d0b7e0

* In Slicer commit message, I usually do not reference SHA1 of Slicer ... since when we will transition to git only, we will most-likely rewrite the history, they will become invalid. That said, during the transition, we will also fix such references ...

PR #581, opened 2016-09-15
ENH: Added segment selector combobox to slice view controller

cpinter:
A fourth row is added to slice view controller which is only visible if a segmentation node is present in the scene. It behaves differently than the other rows in the sense that it does not assign the selected node to a role that is specific to the slice view, but allows changing display properties of a segmentation that are global (visibility, opacity, segment visibility, outline/fill). Selecting a segmentation shows it if hidden.

ENH: Added multi-selection option to segment selector widget

ENH: Removed legacy labelmap support

Segmentation node is no longer subclass of labelmap volume node. Temporary color table is removed (color index tag remained, consecutive values are assigned). Utility function ShiftImageDataExtentToZeroStart moved to volume node from segmentation node, may come in handy


cpinter:
Segmentation nodes do not provide a merged labelmap any more to display in the label layer. It caused confusion between labelmaps and segmentations, it was undeterministic in case of overlapping segments, caused labelmap conversion unnecessarily if it was the first node in the volume combobox when switching to the Volumes module (and in other cases when GetImageData was called), etc.

Instead of the label layer support, slice view controller was augmented for segmentations, so that it can be used in the usual way too. This short video demonstrates:
http://screencast.com/t/2BInMbl4OO

cpinter:
I decided that Data probe support will be added in a separate pull request, I think it's cleaner this way.

pieper:
I didn't look into the code, but the video is very convincing!

fedorov:
Csaba, few questions:

* in the demo video you provided, what is the "Tiny Two Labels" ? Is it the legacy label map?
* have you tested the behavior of the node and segment selectors when the node name exceeds the space allocated?
* is there any way the height of the selectors can be reduced? The popup will now consume more than half of the slice viewer height ... I understand this is unrelated to this specific PR, but do you know why do we have the padding above and under the text in the node selector?

cpinter:
Good questions Andrey!
- TinyTwoLabels is a simple labelmap just to show a scenario where we have one of each
- Good point, I'll try
- Indeed, there is a few pixel extra height and I tried to get rid of that and spent an hour or so but had no luck without adding harsh limits. Any idea anyone?

On September 15, 2016 5:07:09 PM Andrey Fedorov <notifications@github.com> wrote:

Csaba, few questions:

  *   in the demo video you provided, what is the "Tiny Two Labels" ? Is it the legacy label map?
  *   have you tested the behavior of the node and segment selectors when the node name exceeds the space allocated?
  *   is there any way the height of the selectors can be reduced? The popup will now consume more than half of the slice viewer height ... I understand this is unrelated to this specific PR, but do you know why do we have the padding above and under the text in the node selector?

-

fedorov:
Thank you Csaba!

To state the obvious - this functionality is essential for helping users navigate the segmentations overlay. Thank you very much for working on this, this is a great improvement over how things are handled right now!

fedorov:
Oh, another suggestion. Trying to load brain atlas is always a good stress test. It would be interesting to see how that volume of segments will be handled.

cpinter:
Andras did a lot of performance improvements over the last week(s) using the brain atlas, so the generic segmentation infrastructure can handle that many segments much better now.

I'll try it too to make sure my changes don't introduce new bottlenecks.

On September 15, 2016 5:27:16 PM Andrey Fedorov <notifications@github.com> wrote:

Oh, another suggestion. Trying to load brain atlas is always a good stress test. It would be interesting to see how that volume of segments will be handled.

-

cpinter:
I found that neither the long node names nor the many segments in the brain atlas cause problems.

I found a bug that I need to fix (combobox not updated correctly after renaming a segment), and I'm still not sure what to do about the height of the rows in the controller popup. Otherwise all seems good to me.

cpinter:
I fixed the bug. If there are no objections then I'll integrate. Thanks for the comments!

cpinter:
I integrated this commit as http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25366
If anybody finds issues please let me know!
Also closing this pull request.

PR #591, opened 2016-09-30
ENH: when using createProgressDialog QProgressDialog needs to be always on top 

che85:
- otherwise it would be hidden behind the main window and gives the impression, that the process finished
- also added another keyword argument windowFlags for specifying other flags if needed

fixes https://github.com/SlicerProstate/mpReview/issues/146

TODO: We could also think about some parameter for disabling the mainWindow for the duration of that process

### DIFF
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -909,7 +909,7 @@ def messageBox(text, parent=None, **kwargs):
       setattr(mbox, key, value)
   return mbox.exec_()
 
-def createProgressDialog(parent=None, value=0, maximum=100, labelText="", windowTitle="Processing...", **kwargs):
+def createProgressDialog(parent=None, value=0, maximum=100, labelText="", windowTitle="Processing...", windowFlags=None, **kwargs):
   """Display a modal QProgressDialog. Go to QProgressDialog documentation
   http://pyqt.sourceforge.net/Docs/PyQt4/qprogressdialog.html for more keyword arguments, that could be used.
   E.g. progressbar = createProgressIndicator(autoClose=False) if you don't want the progress dialog to automatically
@@ -918,7 +918,8 @@ def createProgressDialog(parent=None, value=0, maximum=100, labelText="", window
   Updating label text with progressbar.labelText = "processing XYZ"
   """
   import qt
-  progressIndicator = qt.QProgressDialog(parent if parent else mainWindow())
+  progressIndicator = qt.QProgressDialog(parent if parent else mainWindow(),
+                                         windowFlags if windowFlags else qt.Qt.WindowStaysOnTopHint)
   progressIndicator.minimumDuration = 0
   progressIndicator.maximum = maximum
   progressIndicator.value = value


pieper:
LGTM

lassoan:
The change is OK, and popups may be good choices in very specific workflow-optimized user interfaces (slicelets, ...), but I think we should consistently switch to using waitCursor+status bar for progress reporting, because often we don't know how long an operation lasts and showing popups every time when the user may have to wait would just appear as flicker. CLIs already use the status bar, so we would be more consistent with that. Also, popup can only be used for progress reporting when we don't have anything useful on the screen and don't enable user interaction, while with waitcursor+status bar is applicable in all use cases.

The method name contains "dialog", so its behavior is correct, it shouldn't be changed, but it would be nice to add a similar convenience function for progress reporting on the status bar and use that most of the time.

lassoan:
Actually, I've just realized that if the change is made then it's not possible to specify a top-level progress popup (popup with NULL parent). Solution could be to add a default vale for the parent argument with a special value (for example, "mainWindow") and if the caller does not specify the argument then it would be replaced by mainWindow().

che85:
@lassoan Then that "mainWindow" should be added to all other dialogs as well, right? 

For instance in those lines: 
- https://github.com/Slicer/Slicer/blob/master/Base/Python/slicer/util.py#L713
- https://github.com/Slicer/Slicer/blob/master/Base/Python/slicer/util.py#L724
- https://github.com/Slicer/Slicer/blob/master/Base/Python/slicer/util.py#L735

lassoan:
For those the parent is not overridden (there is no "parent if parent else mainWindow()"), so you can set None as parent without a special default value.

che85:
@lassoan Oh yes there is a "parent if parent else mainWindow()" here https://github.com/Slicer/Slicer/blob/master/Base/Python/slicer/util.py#L772 which will be call from all of the by me mentioned methods

lassoan:
I see. It seems that this behavior is consistent among many similar methods (which is used in many places), so probably it is not worth messing around with it. If somebody needs to set None parent then it can be set manually or creating the window directly without these convenience methods.

So, the implementation in the pull request is good as is. Sorry for taking your time with this digging & discussion.

Using the status bar is a separate matter, it would be nice to explore that at some point.

fedorov:
For the reference, Christian implemented status bar progress reporting in SliceTracker.

che85:
Correct, but I modified the statusbar which is not really clean. It would be interesting for me if there is anything already built into Slicer which would allow that without the need of "hacking".

fedorov:
I agree status progress bar is a separate topic. I just wanted to mention this to Andras.

che85:
I absolutely agree. We have so much space in the Slicer status bar, why not using it more efficient.

The status bar could be used for displaying a lot more information (not only textual):
- download progress,
- module specific processing, where users could easily get the impression that Slicer got stuck
- selftest steps
and many many more.

fedorov:
Some time ago I also suggested we could use status bar for reporting information that currently shows up in the data probe.

I think the way to go about it is prototype things in individual modules/extensions, find out what works well, and then bring it to the attention of the Slicer community. Implementation of the new features like this directly in the application tends to be extremely contentious and time-consuming, due to the fact that it is often very hard to reach consensus on issues that have anything to do with the functionality that affects large community of users (you can search archives about things like where to put logo, how to host Slicer home page, how to implement support for downloads), and few people have knowledge and resources to spend hacking the application core.

che85:
I wouldn't display that information in the status bar unless the DataProbe can be deactivated somehow. Problem here is that the event is fired each time the cursor moves in the slice views and other running processes get paused for the duration of the computation (done in the processEvent method of DataProbe). 

I only noticed that comment https://github.com/Slicer/Slicer/blob/master/Modules/Scripted/DataProbe/DataProbe.py#L194

I don't remember if that's only an issue on El Capitan

jcfr:
@che85 @fedorov What is the status on this PR ? Can we close it ?

che85:
@jcfr funny that you are mentioning that today because TODAY I noticed that there are also similar issues with displaying confirmation message boxes [1] which should also provide an option to the user to set flags like `qt.Qt.WindowStaysOnTopHint` 

[1] https://github.com/Slicer/Slicer/blob/master/Base/Python/slicer/util.py#L887

che85:
I think this should be merged. I will create another PR  for message boxes later.

jcfr:
> I think this should be merged. I will create another PR for message boxes later.

Is it still the case ?  Can we integrate ?

che85:
@jcfr I think so yes. As long as it doesn't collide with QT5.

lassoan:
I've tested this on Windows with Qt5. Adding qt.Qt.WindowStaysOnTopHint window flag makes the progress dialog show up above all windows, even when switching to another application, which is really annoying, since I switch to other applications because Slicer is busy and I don't want Slicer windows to cover my workspace.

image

If qt.Qt.WindowStaysOnTopHint is not specified then everything works correctly. The progress dialog is always shown above the Slicer main window, no matter how I switch between applications and windows.

@che85 Could you check if the current behavior (no qt.Qt.WindowStaysOnTopHint specified) in recent Slicer versions is correct on Mac as well?

jcfr:
> @che85 Could you check if the current behavior (no qt.Qt.WindowStaysOnTopHint specified) is correct in recent Slicer versions on Mac?

Did you have a chance to check ?

jcfr:
> Adding qt.Qt.WindowStaysOnTopHint window flag makes the progress dialog show up above all windows, even when switching to another application, which is really annoying, since I switch to other applications because Slicer is busy and I don't want Slicer windows to cover my workspace.

Considering the last comment of @lassoan 

@che85  Are we ok to abandon this PR ?

che85:
Yeah, let's just do that for now

jcfr:
Thanks for the update.

Following last comment, I am closing.

PR #592, opened 2016-10-02
ENH: adding CloudLasso selection editor effect

Punzo:
Finally I had some time to finish my Slicer Summer project week. @lassoan @cpinter @pieper @jcfr , may you kindly review it? It will be great, if you will also have time for testing it.

A video showing it: https://youtu.be/v6x3uPuLxzM

(perfomance on my laptop with GTX860M are ok, but not always interactive; on my gaming station with GTX970 are always interactive).

There is still a small TO DO in line 722 of qSlicerSegmentEditorCloudLassoEffect.cxx, relative to a a rather small bug (please read the long comment in the code). However, I haven't a lot residual time to dedicate to this project, so for the moment I'll leave the bug (@lassoan if you have any idea how to fix it fast, I'll be more than glad to do it).

Thanks,
Davide.

### DIFF
--- a/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorThresholdEffect.py
+++ b/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorThresholdEffect.py
@@ -129,6 +129,8 @@ class SegmentEditorThresholdEffect(AbstractScriptedSegmentEditorEffect):
     self.thresholdSlider.blockSignals(True)
     self.thresholdSlider.setMinimumValue(self.scriptedEffect.doubleParameter("MinimumThreshold"))
     self.thresholdSlider.setMaximumValue(self.scriptedEffect.doubleParameter("MaximumThreshold"))
+    unitNodeIntensity = slicer.app.applicationLogic().GetSelectionNode().GetUnitNode("intensity");
+    self.thresholdSlider.decimals = unitNodeIntensity.GetPrecision()
     self.thresholdSlider.blockSignals(False)
 
   def updateMRMLFromGUI(self):
--- a/Modules/Loadable/Segmentations/EditorEffects/Resources/qSlicerSegmentationsEditorEffects.qrc
+++ b/Modules/Loadable/Segmentations/EditorEffects/Resources/qSlicerSegmentationsEditorEffects.qrc
@@ -1,9 +1,9 @@
-<!DOCTYPE RCC><RCC version="1.0">
-  <qresource>
-    <file>Icons/CursorBaseArrow.png</file>
-    <file>Icons/NullEffect.png</file>
-    <file>Icons/Erase.png</file>
-    <file>Icons/Paint.png</file>
-    <file>Icons/Rectangle.png</file>
-  </qresource>
+<RCC>
+    <qresource prefix="/">
+        <file>Icons/CursorBaseArrow.png</file>
+        <file>Icons/NullEffect.png</file>
+        <file>Icons/Erase.png</file>
+        <file>Icons/Paint.png</file>
+        <file>Icons/Rectangle.png</file>
+    </qresource>
 </RCC>
--- a/Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorAbstractEffect.h
+++ b/Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorAbstractEffect.h
@@ -223,7 +223,8 @@ public:
   /// \param selectEffectSlot called from the active effect to initiate switching to another effect (or de-select).
   /// \param updateVolumeSlot called to request update of a volume (modifierLabelmap, alignedMasterVolume, maskLabelmap).
   /// \param saveStateForUndoSlot called to request saving of segmentation state for undo operation
-  void setCallbackSlots(QObject* receiver, const char* selectEffectSlot, const char* updateVolumeSlot, const char* saveStateForUndoSlot);
+  void setCallbackSlots(QObject* receiver, const char* selectEffectSlot, const char* updateVolumeSlot,
+                        const char* saveStateForUndoSlot);
 
   /// Called by the editor widget.
   void setVolumes(vtkOrientedImageData* alignedMasterVolume, vtkOrientedImageData* modifierLabelmap,
--- a/Modules/Loadable/Segmentations/MRML/vtkMRMLSegmentEditorNode.cxx
+++ b/Modules/Loadable/Segmentations/MRML/vtkMRMLSegmentEditorNode.cxx
@@ -57,6 +57,10 @@ vtkMRMLSegmentEditorNode::vtkMRMLSegmentEditorNode()
   this->SetHideFromEditors(true);
   this->MasterVolumeIntensityMaskRange[0] = 0.0;
   this->MasterVolumeIntensityMaskRange[1] = 0.0;
+  this->MasterVolumeIntensityMaskRangeLimits[0] = 0.0;
+  this->MasterVolumeIntensityMaskRangeLimits[1] = 0.0;
+  this->MasterVolumeIntensitySingleStep = 0.01;
+  this->MasterVolumeIntensityDecimals = 2.;
 }
 
 //----------------------------------------------------------------------------
@@ -66,6 +70,26 @@ vtkMRMLSegmentEditorNode::~vtkMRMLSegmentEditorNode()
   this->SetActiveEffectName(NULL);
   this->SetMaskSegmentID(NULL);
 }
+
+namespace
+{
+//----------------------------------------------------------------------------
+template <typename T> T StringToNumber(const char* num)
+{
+  std::stringstream ss;
+  ss << num;
+  T result;
+  return ss >> result ? result : 0;
+}
+
+//----------------------------------------------------------------------------
+double StringToDouble(const char* str)
+{
+  return StringToNumber<double>(str);
+}
+
+}// end namespace
+
 
 //----------------------------------------------------------------------------
 void vtkMRMLSegmentEditorNode::WriteXML(ostream& of, int nIndent)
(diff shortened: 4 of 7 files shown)


lassoan:
Thank you, this looks great!

As the effect is highly specialized (mainly in how the thresholding is performed), it should be in the SlicerAstro extension, but I'll make all the necessary API changes in the Slicer core and also port the main drawing feature into the rectangle effect (it'll be renamed to lasso effect or something similar).

Could you try to add the effect in the SlicerAstro module?

Punzo:
Hi Andras, of course (I have just to register the effect in my Astro module, haven't I?); I just thought that the automatic threshold could be useful also for other application. 

How should we procede? shall I remove from this PR the qSlicerSegmentEditorCloudLassoEffect files, and continue the review of the other modifications?

1) setting threshold widgets decimals with the intensity unit precision attribute
2) undoLastState signals/slots

or you will do another PR?

lassoan:
Maybe the best would be to put this effect in a separate extension that SlicerAstro uses and others can install, too, without installing SlicerAstro. If you agree, try to create this separate extension and in the meantime I'll review and integrate the necessary API changes. I'll post my question about these changes here.

Punzo:
Ok, thanks. I'll wait for the API changes.

Punzo:
Hi Andras, after some user feedback, I decide to put the effect into the SlicerAstro extension. In fact, I need to specialize it even more (e.g., initialize the threshold value to an optimal one, etc. ) and I named it AstroCloudLasso.

I tested it and everything works: https://github.com/Punzo/SlicerAstro/tree/cloudLassoSelectionEffect . I'll wait the other API changes before to merge it in my SlicerAstro master (if I can help with them, please, let me know). 

Thanks.

jcfr:
@Punzo  Thanks for the patches :+1:

Punzo:
@lassoan, in https://github.com/Slicer/Slicer/pull/592/commits/ea75f20537bbd144b6b6537fc9834726cbe6eac9, I have just cleaned the PR from the cloudlasso editor effect files for an easier review of the API that I'd like to add (if you still need to look at them: https://github.com/Punzo/SlicerAstro/tree/cloudLassoSelectionEffect/AstroVolume/EditorEffects).

Punzo:
Hi @lassoan, this week I have a deadline (for which I need this PR) and, if you have time, it will be great if we can do the review of the remaining small changes to the API. Thanks in advance.

lassoan:
@Punzo a few questions about the API change:
Is unit linked to the selected image node or global to the scene?
Would it be possible to save some labelmaps (and restore them if needed) in the effect instead of adding access to undo in the effect?

Punzo:
1) in general, the properties of the intensity unit node are global to the scene. However, in my case these will be updated on the properties of the selected volume node (different datacube have different intensity unit, so different number of decimals and so on.). That's why I'd like to insert the https://github.com/Slicer/Slicer/pull/592/files#diff-bb11aa63d106e0d90061bc4049e27094R1425.

2)  in principle yes. I didn't want to use more memory, but I think your suggestion will be better (due the relative bugs in using undoLastState). I have updated the PR removing this API changes.

lassoan:
1) Intensity unit is always specific to the volume, it cannot be a global settings. We often display different volumes at the same time (e.g., CT, dose , MRI, ultrasound), so there is not a single unit that is usable. For dose volumes we have custom field in the volume node that specifies the unit. Do you have something similar? We should probably have a standardized solution for storing intensity unit for all kinds of volumes and the threshold slider should use that.

fedorov:
> We should probably have a standardized solution for storing intensity unit for all kinds of volumes

👍  This is needed for the purposes of quantitative reporting too. It would be great if we could agree upon an application-wide solution for this.

Punzo:
I agree too, but at the moment, all the unit nodes are unique and global (I mean the unit nodes initialized in the settings). If we'd like to change the behavior (allowing to store, etc., multiple unit nodes of the same type), we should discuss how we'd like it.

Punzo:
Maybe we can associate a set of unique units nodes for each volume and we can change how to visualize them in the settings menu'. However, this will be a big change and I don't feel to try to implement it before a discussion.

lassoan:
The changes that we implement now may be very small and simple, we just have to discuss a bit to make sure it can evolve to a full implementation in the future without breaking backward compatibility.

For example, we could just agree in adding a new member to vtkMRMLVolume (or just agree in a custom attribute name) that would store a unit name, which by default could be displayed as a postfix for pixel values. For example HU (for CT images), Gy (for RT dose grids). In the future we can add offset/linear or non-linear scaling, etc., image kind (radiodensity, dose, etc).

Would a single new human-readable "unit" string field for volumes solve immediate needs?

Punzo:
A custom attribute name is fine for me too. However, what about the the relative precision (used in this PR), coefficient, offset, etc.. attributes in the unit nodes API? Shall we store them also as attributes in the volume nodes? or you want to store in the volumes only the relative suffix attribute? I think, all the units attributes are important for quantification (e.g. display in dataprobe) and if this is the case, is not better to use the full API provided by the units nodes? 

Moreover, will these changes affect only the intensity "unit" or also the others?

lassoan:
Scaling (offset and scale for linear scaling; other parameters for non-linear scaling) is volume-specific, too, so that should be stored in the volume node and not in a global unit node. Do you need linear scaling now or it is enough to specify it later?

Display range and precision is often not a global parameter either. For example, in this case we know the range from the scalar range of the pixel data; display precision is deduced from the screen resolution, widget size, and min/max range (approximately ceil(-log((rangeMax-rangeMin)/500))).

I think the only unit node feature we would use is generation of displayable string, but that's kind of a "logic" feature and shouldn't be implemented in MRML nodes.

So, overall, I don't really see any use for unit nodes here.

lassoan:
@fedorov I agree that we should be as close as possible to DICOM's unit specification. Do you have some specific examples for how UCUM is used in DICOM for specifying units?

Punzo:
1) Usually my intensity pixels are in milli or micro (Jensky). Therefore, linear scaling will be great, if we display on the threshold widget the "unit string" as in dataProbe. 

2) I agree.

3) that's true.

4) Ok.

fedorov:
I completely agree that units and scaling etc should be associated with the volume node.

About how to specify scaling, DICOM/UCUM gives flexibility - for example, you can use mm, or you can use m and add scaling 10e-3 in the DICOM Real World Value Mapping (RWVM) module.

@lassoan for specific examples you can look at these json files that encode measurements for conversion into DICOM measurements Structured Report: https://github.com/QIICR/dcmqi/blob/master/doc/sr-tid1500-ct-liver-example.json (see "units" attribute). This specific examples encodes measurements as produced by Slicer LabelStatistics. 

Note how each unit code is a triple of code/coding scheme designator/code meaning, same as done for other concepts in DICOM, such as segmentation category/type and anatomic location. This way use cases where no UCUM code exists (although, I think those should be rare) can be addressed by utilizing alternative (or even custom) coding scheme.

lassoan:
It seems that by adding 4 fields (string CodeValue, string CodingSchemeDesignator, double Slope, double Intercept) would be enough to fully specify unit (although it's redundant, we can add CodeMeaning for completeness). Support for non-linear mapping and multiple real-world value mapping could be added in the future.

Unit would need to be used in other nodes (e.g., scalar values in models), so it would be implemented in a small separate class that can be reused in several nodes.

For now we could go with a very simple hard-coded formatting method of using CodeValue as postfix. In the future we could add a MRML logic class to generate formatted strings with more complex rules.

fedorov:
>although it's redundant, we can add CodeMeaning for completeness

I don't think CodeMeaning is redundant. It may seem redundant in some cases, but it is important to use consistent implementation pattern. In the general case, code value does not match code meaning, and can be completely incomprehensible.

lassoan:
I mean it's redundant to store code meaning in the node, as it can be looked up in the coding scheme dictionary. If we store code meaning in the node then it may lead to inconsistency (the code meaning differs from the dictionary value), which we have to detect and handle somehow. If the meaning is not stored in the node then we may have problems when the code value is not found in Slicer's coding scheme dictionaries. So, neither solution is perfect. Probably it's better to always store the meaning and deal with inconsistencies.

Unfortunately, for UCUM units, code meaning is not usable directly for display, as it is too verbose (for example "Hounsfield unit", "cubic millimeter"). Code value is often usable as is (cm3, mm3), but not always (for example code value of Hounsfield unit is [hnsf'U]). Do you know if there is a method in DICOM to convert a numbers to a displayable string? (for example for Hounsfield Unit it would append the HU postfix).

fedorov:
> I mean it's redundant to store code meaning in the node, as it can be looked up in the coding scheme dictionary. 

This assumes availability and maintenance of dictionary. Note that as of now, DICOM does not define any means to communicate dictionary alongside the objects (we may work to improve this as part of the QIICR project). It is also the case that DICOM mandates inclusion of code meaning in the code tuple. I agree with your point that inconsistency can happen, but it can also happen with the use of dictionary, since I can't imagine we can expect centralized lookup service, and individual tools will need to duplicate dictionaries (and this will not address private coding schemes with the codes outside UCUM or DICOM). In any case, we are in agreement it makes sense to keep code meaning.

> Unfortunately, for UCUM units, code meaning is not usable directly for display, as it is too verbose (for example "Hounsfield unit", "cubic millimeter"). Code value is often usable as is (cm3, mm3), but not always (for example code value of Hounsfield unit is [hnsf'U]). Do you know if there is a method in DICOM to convert a numbers to a displayable string? (for example for Hounsfield Unit it would append the HU postfix).

I don't think DICOM defines anything like that. I will discuss this with David Clunie on Thu at our QIICR meeting, and update this thread then. I think for now code value is a reasonable choice, because it will make sense in most of the cases.

fedorov:
@lassoan if we add units, it makes sense to add a quantity attribute at the same time. The example I referenced earlier in https://github.com/Slicer/Slicer/pull/592#issuecomment-252913866 and also this one https://github.com/QIICR/dcmqi/blob/master/doc/pm-example.json#L9-L12 show how it can be using DICOM concepts.

I also think this discussion diverged a lot from the topic of the request. Should we make a Labs page? Although the advantage of our current approach is that we can have a discussion, unlike in wiki.

pieper:
+1 for including a DICOM-like description of the quantity being measured by
the volume and the units in which it is expressed.

+1 for opening an issue to discuss those specific topics, and then a wiki
Labs page once there is a plan of action.

fedorov:
I added relevant notes to the existing labs page here: http://wiki.slicer.org/slicerWiki/index.php/Documentation/Labs/Units

Punzo:
I close this PR due it is not anymore needed and I'll be really great to have standardised rules for units for each volume!

lassoan:
We'll work units for sure. Thank you for putting up with finally not integrating the API changes, but if an insufficient API is introduced and later has to be changed then it usually leads to a lot of extra work (often several times more than implementing sufficient API for the first place).

Punzo:
I simply agree that it will much better to have the units attributes for the volumes and then we can properly relook where this new infrastructure should be used (such as here for the intensity widget thershold). Thanks as usual for your feedback and help!!!

PR #593, opened 2016-10-03
ENH: prefer executable CLIs by default

fedorov:
This resolves a known issue with the application crashing while using modules
as shared libraries under certain circumstances.

Motivated by this thread:

http://slicer-devel.65872.n3.nabble.com/Slicer-crashes-when-running-registration-without-having-settings-set-to-quot-Prefer-Executable-CLIs--td4037245.html

### DIFF
--- a/Base/QTGUI/qSlicerSettingsModulesPanel.cxx
+++ b/Base/QTGUI/qSlicerSettingsModulesPanel.cxx
@@ -119,7 +119,7 @@ void qSlicerSettingsModulesPanelPrivate::init()
   this->FavoritesMoreButton->setChecked(false);
 
   // Default values
-  this->PreferExecutableCLICheckBox->setChecked(false);
+  this->PreferExecutableCLICheckBox->setChecked(true);
   this->TemporaryDirectoryButton->setDirectory(coreApp->defaultTemporaryPath());
   this->DisableModulesListView->setFactoryManager( factoryManager );
   this->FavoritesModulesListView->setFactoryManager( factoryManager );


jcfr:
Makes sense. Running CLIs in a sandboxed way is a sane default. 

Shared library mode should be reserved for special case where it clearly improves performance.

pieper:
Thanks for taking care of this!

But I think line 123 also needs to had a default value of 'true' here:

https://github.com/Slicer/Slicer/blob/7c97ad2cd412e873ce78568b252e56c6ee51e515/Base/QTApp/qSlicerApplicationHelper.cxx#L121-L123

The value method takes a default so the first time the program is installed it will take on the desired value:

http://doc.qt.io/qt-4.8/qsettings.html#value

fedorov:
@pieper should it also be fixed here: https://github.com/Slicer/Slicer/blob/master/Base/QTGUI/qSlicerApplication.cxx#L924 ?

pieper:
Yes,  would think both places.  Since that default value shows up two places there's some ambiguity about which one will show up first, so maybe it should be #defined someplace to prevent only changing one in the future.

jcfr:
>  #defined someplace to prevent only changing one in the future.
 :+1: 

Please, consider adding `Slicer_PREFER_CLI_EXECUTABLE` to https://github.com/Slicer/Slicer/blob/master/CMake/vtkSlicerConfigure.h.in

jcfr:
or `Slicer_CLI_PREFER_EXECUTABLE`

fedorov:
Guys, sorry, I have just a few days a week relatively free from meetings,and there is too much work to do. The proposed changes of course make sense, but involve changing cmake and multiple things, and doing test builds, so not something I can do quickly. I will get to it, but not right away.

jcfr:
This PR should address the problem: https://github.com/Slicer/Slicer/pull/598

PTAL

jcfr:
Closing. This PR is superseded by #598

PR #594, opened 2016-10-03
Remove tractography parts of vtkTeem, and SlicerDMRI remote module

ihnorton:
Regarding vtkTeem: as far as I can tell, the removed libraries are only used by SlicerDMRI modules. I checked this by doing `git clone` or `svn co` on all registered extension `scmurl`, and grep'ing for `vtkTeem`. The only matches are:

- 'LASegmenter': specifying `vtkTeem_INCLUDES` is not necessary because it does not actually use vtkTeem.
- 'Sequences': only used for vtkNRRD{Reader,Writer}, which are not being removed from base.

The vtkTeem history is being migrated, to the extent possible (goes back as far as svn can provide, to a rename in 2007). See https://github.com/SlicerDMRI/SlicerDMRI/tree/merge_vtkteem

SlicerDMRI will be distributed as an extension.

cc @ljod @pieper


jcfr:
Looks good to me. Thanks for working on this.

ihnorton:
Please commit whenever you have time, then I will add to ExtensionsIndex.

(I would do it, but I don't seem to have svn access anymore. Or I don't remember the password correctly -- any way to reset?)

pieper:
Makes sense to me.

jcfr:
After deleting `Slicer-*` from a local build tree, I reconfigured and it is now rebuilding. I will let you know how it goes.

jcfr:
Integrated in [r25408](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25408) and [r25409](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25409)

ihnorton:
Thanks.

ihnorton:
@jcfr it looks like the subversion-git sync robot isn't running properly.

PR #595, opened 2016-10-04
BUG: Revert default VTK backend to OpenGL

msmolens:
Revert the default VTK backend to OpenGL. Because the OS X factory
machine targets OS X 10.6 (CMAKE_OSX_DEPLOYMENT_TARGET:STRING=10.6), the
Mac nightly builds don't find the required OpenGL version. The error
message seen when running Slicer is:

    vtkCocoaRenderWindow (0x125872400): VTK is designed to work with
    OpenGL version 3.2 but it appears it has been given a context that
    does not support 3.2. VTK will run in a compatibility mode designed
    to work with earlier versions of OpenGL but some features may not
    work.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -517,7 +517,7 @@ endif()
 #-----------------------------------------------------------------------------
 # VTKv7 - Slicer_VTK_COMPONENTS
 #-----------------------------------------------------------------------------
-set(Slicer_VTK_RENDERING_BACKEND "OpenGL2" CACHE STRING "Choose the rendering backend.")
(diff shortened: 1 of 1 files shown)


msmolens:
Integrated in [r25406](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25406).

pieper:
Thanks

msmolens:
Update:
I tested building Slicer/OpenGL2 on [`factory-south`](https://www.slicer.org/slicerWiki/index.php/Documentation/Nightly/Developers/Factory#factory-south.kitware) (OS X 10.11) with `CMAKE_OSX_DEPLOYMENT_TARGET:STRING=10.10` and can confirm that the VTK error about OpenGL 3.2 no longer appears.

PR #597, opened 2016-10-05
ENH: update MultiVolumeImporter hash

fedorov:
Updated MultiVolumeImporter plugin is improved to handle certain types of 4D
Cardiac CT.

Thanks to @lassoan for identifying the use case and contributing the patch.

### DIFF
--- a/SuperBuild.cmake
+++ b/SuperBuild.cmake
@@ -200,7 +200,7 @@ list_conditional_append(Slicer_BUILD_MultiVolumeExplorer Slicer_REMOTE_DEPENDENC
 
 Slicer_Remote_Add(MultiVolumeImporter
   GIT_REPOSITORY ${git_protocol}://github.com/fedorov/MultiVolumeImporter.git
-  GIT_TAG 669594fb993b34a104795cd0b1e1a3340c4311c2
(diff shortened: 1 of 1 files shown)


jcfr:
Building locally. Planning on integrating shortly.

@fedorov Should this be integrated today or wait after the release ?

fedorov:
It is a very small change, and I can't see how it can break anything (nothing in the core depends on it), so I would say integrating now would be good and painless. It's just the last 2 commits from here https://github.com/fedorov/MultiVolumeImporter/commits/master

jcfr:
Thanks for the note

Integrated in [r25422](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25422)

There are a small warning when opening the module, I wonder if this is a symptom of something more important .., 

```
PythonQt: QObject::connect() signal 'mrmlSceneChanged(vtkMRMLScene*)' does not exist on qMRMLNodeComboBox
```

fedorov:
It is probably this line, the purpose of which I don't know at this point.

https://github.com/fedorov/MultiVolumeImporter/blob/master/MultiVolumeImporter.py#L84

I don't think it is important.

PR #598, opened 2016-10-05
BUG: Introduce macro to set "Prefer CLI executable" settings default value

jcfr:
This commit fixes a regression introduced in r25404 (ENH: prefer executable
CLIs by default) and ensures the settings default value is consistently set.


pieper:
👍 Thanks Jc!

fedorov:
Thanks a lot for taking care of this JC!

jcfr:
After confirming it compiled locally and checking that the default was correct after removing the settings, this change has been integrated in [r25416](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25416)

PR #601, opened 2016-10-11
ENH: Added OBJ file save option for model nodes

lassoan:
Reading from .obj files are already supported, users requested ability to also save as .obj file to be able to load the exported models directly into Unity.


jcfr:
Thanks for the patch. After CI is green, :+1: 

Should a small test be added ? 

Could also you also update the wiki page: http://wiki.slicer.org/slicerWiki/index.php/Documentation/Nightly/SlicerApplication/SupportedDataFormat

jcfr:
Very nice. I am building now to confirm test pass.

lassoan:
I've added a test and updated the wiki page.

jcfr:
No problem compiling. Test also pass.

Integrated in [r25421](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25421)

PR #603, opened 2016-10-11
ENH: scripted helper module to suggest SlicerDMRI extension

ihnorton:
We got a request to make more obvious to users where to find the modules that are now in SlicerDMRI extension. Not sure if this can make it in to release (great if so), but we'd like to have it in the nightly for a few months too.

### DIFF
--- a/Modules/Scripted/CMakeLists.txt
+++ b/Modules/Scripted/CMakeLists.txt
@@ -3,6 +3,7 @@ include(SlicerMacroBuildScriptedModule)
 
 set(modules
   DataProbe
+  DMRIInstall
   Editor
   EditorLib
   LabelStatistics
--- a/Modules/Scripted/DMRIInstall/CMakeLists.txt
+++ b/Modules/Scripted/DMRIInstall/CMakeLists.txt
@@ -1,0 +1,14 @@
+#-----------------------------------------------------------------------------
+set(MODULE_NAME DMRIInstall)
+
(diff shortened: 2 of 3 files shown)


ihnorton:
cc @ljod

jcfr:
Great. I will locally test and then integrate.

jcfr:
The text do not wrap and force the module to panel to extend. 

Could you look into word wrap ? See http://doc.qt.io/qt-4.8/qlabel.html#wordWrap-prop

dmriinstall-too-wide

ihnorton:
Updated, now uses wordWrap and does not expand.

jcfr:
Great. Thanks

jcfr:
That I will make it to the release, I will cut it a little later tonight

jcfr:
I tweaked things a little. Instead of using `QLabel`, I used the `ctkFittedTextBrowser`.

As a side I also added a redirect from the DMRIInstall wiki page to the extension page.
See https://www.slicer.org/slicerWiki/index.php?title=Documentation/Nightly/Modules/DMRIInstall&redirect=no

Here is the result:

screenshot from 2016-10-12 00-52-15

jcfr:
Integrated in [r25425](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25425)

ihnorton:
Awesome, thanks.

PR #606, opened 2016-10-12
BUG: ExtensionBuildSystem tests: Do not pass empty CMAKE_OSX_ARCHITECTURES

jcfr:
Since it is valid to have build without an architecture explicitly
set (meaning no -arch x86_64 compiler flag), this commit do not force
an empty value in the CMakeCache.

Reported by: Nicole Aucoin nicole@bwh.harvard.edu


jcfr:
Cc: @naucoin @pieper @ihnorton That should allow you to run the test without architecture flag

jcfr:
Integrated in [r25506](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25506)

PR #608, opened 2016-10-26
BUG: Fix issue with qMRMLNodeComboBox when removing an attribute filter

vovythevov:
Because of the use of the [] operator on the `QHash` map used to check
attributes in the `qMRMLComboBox`, removing an attribute filter could lead
the combobox to never show anything.

The bug was as follow in pseudo-code:
```
 combobox = slicer.qMRMLComboBox()
 combobox.removeAttribute('vtkMRMLVolumeNode', 'MyAttribute')
```
-> Because of the [] operator, the map now has an entry for
`vtkMRMLVolumeNode` with nothing in it, i.e. it expects nodes to have that
attribute (and it can be anything).

The fix is to not use the [] operator when accessing values simply for
reading but to use the values() method instead.

Added a test along to make sure everything works properly.

### DIFF
--- a/Libs/MRML/Widgets/Testing/qMRMLNodeComboBoxTest9.cxx
+++ b/Libs/MRML/Widgets/Testing/qMRMLNodeComboBoxTest9.cxx
@@ -1,0 +1,115 @@
+/*==============================================================================
+
+  Program: 3D Slicer
+
+  See COPYRIGHT.txt
+  or http://www.slicer.org/copyright/copyright.txt for details.
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+  This file was originally developed by Johan Andruejol, Kitware Inc.
+
+==============================================================================*/
+
+// QT includes
+#include <QApplication>
+#include <QTimer>
+
+// CTK includes
+#include <ctkCoreTestingMacros.h>
+
+// qMRML includes
+#include "qMRMLNodeComboBox.h"
+
+// MRML includes
+#include <vtkMRMLNode.h>
+#include <vtkMRMLScene.h>
+#include <vtkMRMLScalarVolumeNode.h>
+
+// VTK includes
+#include <vtkNew.h>
+
+// test the filtering with many cases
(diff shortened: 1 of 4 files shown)


vovythevov:
@jcfr Would you mind taking a look ?

Libs/MRML/Widgets/Testing/qMRMLNodeComboBoxTest9.cxx
```diff
@@ -0,0 +1,158 @@
...
+
+  vtkNew<vtkMRMLScene> scene;
+
+  vtkNew<vtkMRMLScalarVolumeNode> noAttributeNode;
+  scene->AddNode(noAttributeNode.GetPointer());
+
+  const char *testingAttributeName = "testingAttribute";
+  const char *testingAttribute = noAttributeNode->GetAttribute(testingAttributeName);
+  std::cout << "Volume node with no call to SetAttribute, GetAttribute returns " << (testingAttribute ? testingAttribute : "0") << "." << std::endl;
+  if (testingAttribute != NULL)
```
jcfr: Please, consider using the macros provided by `ctkCoreTestingMacros.h`

For an example of use, see https://github.com/commontk/CTK/blob/afe8505a8286a701900fca1c97a4ad8876bbbe1a/Libs/Widgets/Testing/Cpp/ctkAxesWidgetTest1.cpp#L91-L107

jcfr:
Good catch. Makes sense.

After addressing few nitpicks in the test, it will make sense to integrate.

for reference, here is the doc of `[]` operator for `QHash`:


> Returns the value associated with the key as a modifiable reference.
>
> If the hash contains no item with the key, the function inserts a default-constructed value into the 
> hash with the key, and returns a reference to it. If the hash contains multiple items with the key, this 
> function returns a reference to the most recently inserted value.

vovythevov:
Updated to address @jcfr comments.
Will intergrate if CircleCI passes.

jcfr:
very nice.

Nitpicks related to the commit msg 

```
Because of the use of the [] operator on the QHash map to check
attributes in the qMRMLComboBox, removing an attribute filter could lead
the combobox to never show anything.

[...]
```

-> 

```
Since QHash [] operator inserts a default-constructed value into the hash if the key is not found
and because the [] operator was systematically use, removing an attribute filter could lead the combobox to never show anything.

[...]
```

jcfr:
Thanks.

Nitpick: When mentioning the revision, consider using link like this one: http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25481

PR #612, opened 2016-10-28
ENH: restore non-DTI part of Neurosurg.MarkupsTest

ihnorton:
Restoring the non-DTI parts of the
NeuroSurgicalPlanningTutorialMarkupsSelfTest that were removed in:

https://github.com/Slicer/Slicer/commit/2d54809b42b7e0b11402e44e74776be4fdbf9977

Because it is a useful test of a variety of functionality.


jcfr:
Thanks for the patch. Integrated in [r25483](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25483)

PR #614, opened 2016-10-28
BUG: Use Spline order for 3 template parameter

blowekamp:
Fix common mistake of using dimension for spline order.

### DIFF
--- a/Libs/MRML/Core/vtkITKTransformInverse.h
+++ b/Libs/MRML/Core/vtkITKTransformInverse.h
@@ -265,10 +265,10 @@ namespace itk
 
 } // end of ITK namespace
 
-typedef itk::InverseBSplineDeformableTransform< float, VTKDimension, VTKDimension > InverseBSplineTransformFloatITKv3Type;
-typedef itk::InverseBSplineDeformableTransform< double, VTKDimension, VTKDimension > InverseBSplineTransformDoubleITKv3Type;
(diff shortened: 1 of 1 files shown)


lassoan:
Thank you, good catch. Could you please add a #define SPLINE_ORDER 3 or something similar and use that instead of magic number 3?

lassoan:
Thanks! I've just noticed that static const int BSPLINE_TRANSFORM_ORDER = 3; is already defined along with static const unsigned int VTKDimension = 3; in vtkITKTransformConverter.h. So, we should just use BSPLINE_TRANSFORM_ORDER as this way both forward and inverse transforms use the same constant.

blowekamp:
I ran into this after seeing some Transform registration warning. And working on this patch for ITK:
http://review.source.kitware.com/#/c/21677/2

jcfr:
Great. Thanks for the patch.

I will test locally and integrate.

This will address http://na-mic.org/Mantis/view.php?id=4290

jcfr:
Integrated in [r25485](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25485)

blowekamp:
Neither of these patches address the warnings posted Slicer is having. The value of the b-spline order is not changed with this patch. It does address the conceptual conflating for the 3rd template parameter. 

The ITK patch http://review.source.kitware.com/#/c/21677/2 address warnings in SimpleITK related to duplicate registration of BSpline Transforms. And Bugs related to IO of BSplines with different orders.

PR #615, opened 2016-10-28
BUG: don't rely on application state in widget ::init

ihnorton:
QtDesigner widget initialization crashes because Terminologies widget `::init` calls a function that requires application state.

ref https://github.com/Slicer/Slicer/commit/4bac13a687b39a69e2648f281bd737bae3eaf08c#commitcomment-19617853


lassoan:
Thanks, I've just run into this today, too (Qt designer crashed on start).

lassoan:
I've accidentally merged this instead of pulling and svn-dcommitting. Anyway, I had a fix in my working copy, too, I've just committed that. I hope my merge won't break the svn sync.

jcfr:
@ihnorton @lassoan  Thanks for fixing this

jcfr:
> accidentally merged this instead of pulling and svn-dcommitting.

No problem, the bots are now able to deal with this

PR #618, opened 2016-11-03
CircleCI Error: `CircleCI does not build changes to Slicer dependencies`

MayeulChassagnard:
This error appears when the slicer/slicer-build-deps docker image is not updated with the actual version of slicer.

In order to fix the issue, please create a Pull Request like this one:
  https://github.com/thewtex/SlicerDocker/pull/21

This will build and push all images based on the newer version of Slicer

CC @thewtex

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -17,6 +17,6 @@ dependencies:
 
 test:
   override:
-    - "echo 'Notice: CircleCI does not build changes to Slicer dependencies' && if git diff --name-only master | grep -q SuperBuild > /dev/null; then false; fi"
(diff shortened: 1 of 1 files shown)


circle.yml
```diff
@@ -16,6 +16,11 @@ dependencies:
 
 test:
   override:
+    # if the line below produce an error in CircleCI,
+    # you have to update the slicer dependencies image in SlicerDocker
+    # In order to do it, please create a Pull Request like this one https://github.com/thewtex/SlicerDocker/pull/21
+    # This will build and push all images based on the newer version of Slicer
```
thewtex: Folks that receive the build error will not know that this comment is here.

Please try adding

```
 echo "Please create a Pull Request on https://github.com/thewtex/SlicerDocker according to the Update section of the project README.";
```

before the `false` instead.

MayeulChassagnard: Oh, I see! I'm so sorry, I really didn't understand what you were looking for

jcfr:
Thanks for the patch. Integrated in [r25724](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25724)

PR #620, opened 2016-11-07
ENH: Improved crosshair usability, added display in 3D view

lassoan:
Crosshair had several problems/limitations:
1. Crosshair position was often inconsistent in different slice views (for example: when in navigation view, crosshair was always displayed in all slice views because otherwise the user could not drag the crosshair to a new position; but this way when the user changed slices then the crosshair showed up at different RAS positions in each slice view)
2. Navigation mode was very inconvenient: to jump all slices to a specific RAS position the user had to position the mouse to the previous crosshair and drag it to the new position (it was too easy to miss the small navigation area in the middle and when it was missed, brightness/contrast of the image was accidentally modified)
3. By default (if the user just toggled the crosshair button on the toolbar), crosshair was only visible in the current slice view, which was useless (the crosshair just followed the mouse in the current view)
4. Crosshair position was not visible in 3D view

Solution implemented:
1. Only show crosshair in the correct position in all views. If crosshair position is not in the current slice then hide the crosshair from current slice.
2. Crosshair is now moved by holding down shift + moving mouse. No need to grab the previous crosshair position. Navigation mode is replaced by "jump slices" option (enabled by default), which controls if slice views should jump to current crosshair position. If disabled then shift + moving mouse only moves the crosshair.
3. By default slices jump to current crosshair position, so crosshair is visible in all views. User just has to keep Shift key depressed (added hint to tooltip)
4. Implemented display of crosshair in 3D view.

### DIFF
--- a/Base/QTGUI/qSlicerViewersToolBar.cxx
+++ b/Base/QTGUI/qSlicerViewersToolBar.cxx
@@ -43,7 +43,7 @@ qSlicerViewersToolBarPrivate::qSlicerViewersToolBarPrivate(qSlicerViewersToolBar
 {
   this->CrosshairToolButton = 0;
   this->CrosshairMenu = 0;
-  this->CrosshairNavigationAction = 0;
+  this->CrosshairJumpSlicesAction = 0;
 
   this->CrosshairMapper = 0;
   this->CrosshairNoAction = 0;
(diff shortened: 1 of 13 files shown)


pieper:
Really nice addition and fixes.

@millerjv can you have a look?

lassoan:
If there are no objections, I will merge this tomorrow. The improved crosshair is very helpful for seeing correspondence between slice views and 3D views during segmentation, studying anatomical atlases, or for probing (see for example this recent request http://slicer-users.65878.n3.nabble.com/view-scalar-values-of-3d-surface-td4031274.html).

millerjv:
If people are happy with the refined behavior, no issues on my side. 

It looks like a number of the code changes are refactoring the term Navigation for the term Jump Slice. At the 20,000 foot level this seems unnecessary. But if it helps to simplify the number of concepts then that is great.

lassoan:
Yes, exactly the goal is to merge concepts.

Navigation/crosshair had some jump-to-slice features with limitations - only worked in slice views, and UI was not very convenient. Jump-to-slice provided 3D position with some limitations - you could see slice intersection visually, but there was no explicit 3D position provided for probing etc.

Merging the two concepts provides both 3D crosshair position and jump-to-slice consistently, without their respective limitations.

pieper:
Sounds good to me.

jcfr:
Closing. Integrated in [r25521](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25521)

PR #622, opened 2016-11-10
ENH: Refactoring dicom widgets

che85:
- some fixes and simplification of code
- formatting
- removing progressbar as a class member since it could collide with other processes that want to access the same progressbar at the same time

### DIFF
--- a/Modules/Scripted/DICOM/DICOM.py
+++ b/Modules/Scripted/DICOM/DICOM.py
@@ -276,7 +276,7 @@ class DICOMWidget:
     self.layout.addWidget(self.activityFrame)
 
     self.recentActivity = DICOMLib.DICOMRecentActivityWidget(self.activityFrame,detailsPopup=self.detailsPopup)
-    self.activityFrame.layout().addWidget(self.recentActivity.widget)
+    self.activityFrame.layout().addWidget(self.recentActivity)
     self.requestUpdateRecentActivity()
 
 
--- a/Modules/Scripted/DICOMLib/DICOMPlugin.py
+++ b/Modules/Scripted/DICOMLib/DICOMPlugin.py
@@ -1,7 +1,4 @@
-import os
-import qt
 import slicer
-import DICOMLib
 
 
 #########################################################
--- a/Modules/Scripted/DICOMLib/DICOMWidgets.py
+++ b/Modules/Scripted/DICOMLib/DICOMWidgets.py
@@ -20,6 +20,8 @@ This code is slicer-specific and relies on the slicer python module
 for elements like slicer.dicomDatabase and slicer.mrmlScene
 
 """
+
+
 #
 #########################################################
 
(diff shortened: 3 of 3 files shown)


pieper:
Looks good to me - is it ready to merge?  I can do that unless someone else is planning to.

che85:
It should be ready for merge yes

pieper:
Are you working with someone else who plans to merge it?

che85:
No no one else works with me on that.

che85:
@pieper Don't merge it yet. I will go ahead and take a look at the DICOMDetailsPopup since there are methods that might need some refactoring and this could also be useful for the Reporting module

pieper:
Okay - thanks for working on this and let me know when you have things to
commit.

che85:
@pieper I would say, it's ready for merge. I could still do some refactoring, but right now I don't have time for this.

pieper:
There was a problem with a missing end of line on one of the commits (see below), but with some help from @ihnorton I was able to squash and commit to svn.

http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25545

> ERROR from SVN:
> A repository hook failed: Commit blocked by pre-commit hook (exit code 1) with output:
> Log: 'ENH: Refactored DICOMHeaderPopup and inheriting from ctkDICOMObjectListWidget\n \n - removed unused imports and specified ctk imports explicit\n \n From: Christian Herz <christianherz1985@googlemail.com>\n'
> ----------------------------------------------------------------------
> Problems in file "trunk/Modules/Scripted/DICOMLib/DICOMWidgets.py":
> The following line(s) do not contain new-line character:
>  1218:     settings.setValue('DICOM/headerPopup.geometry', self.geometry)
> 
> Found problems with 1 file(s). Cannot commit
> ----------------------------------------------------------------------
>

pieper:
@jcfr do we or could we have a git commit hook that would have caught the missing end of line like the svn one does?

I guess in the future squashing the commits before putting them in svn would have avoided this problem.

jcfr:
We indeed have local git hook.

See https://www.slicer.org/wiki/Documentation/Nightly/Developers/DevelopmentWithGit#Slicer_Setup

They could be improved by updating:

   https://github.com/Slicer/Slicer/blob/hooks/pre-commit

and incrementing the version number here:

  https://github.com/Slicer/Slicer/blob/00a0a3c10bb6db2998bd6a6da09a68b997f9708e/Utilities/SetupForDevelopment.sh#L47-L49

> in the future squashing the commits before putting them

Avoiding having giant commit is still important, it makes tracking of issues, reviewing of changes and overall software maintenance easier.


Hth
Jc

fedorov:
>> in the future squashing the commits before putting them

> Avoiding having giant commit is still important, it makes tracking of issues, reviewing of changes and overall software maintenance easier.

I agree. I think it does make sense to squash commits that are small refinements, but conceptual breakdown of what was done is important to maintain.

pieper:
Yes, this one was problematic because a commit part way through the pull
request violated the svn hooks (no end of line on the last line of the
file).  So the first several commits are one-by-one and the last 4 or 5 had
to be squashed to bypass the missing end-of-line character.

I assume @che85 had the developer setup so that's why I wonder if we can
synchronize what is checked by both git and svn -- or finally switch to git
only! : )

fedorov:
> or finally switch to git only! : )

is it still somewhere in the timeline? There've been talks about this for as long as I remember

pieper:
We've been talking for a long time but it's not been enough of a priority
to invest the time and effort to change.

PR #624, opened 2016-11-16
ENH: Improve Surface Toolbox Normal Filtering. Add MirrorModel functionality.

bpaniagua:
@jcfr Could you please have a look into this PR? Thanks

### DIFF
--- a/Modules/Loadable/Models/Logic/vtkSlicerModelsLogic.cxx
+++ b/Modules/Loadable/Models/Logic/vtkSlicerModelsLogic.cxx
@@ -29,9 +29,12 @@
 #include <vtkNew.h>
 #include <vtkObjectFactory.h>
 #include <vtkPolyDataNormals.h>
-#include <vtkReverseSense.h>
+#include <vtkTransformPolyDataFilter.h>
+#include <vtkTransform.h>
+#include <vtkMatrix4x4.h>
 #include <vtkSmartPointer.h>
 #include <vtkTagTable.h>
+#include <vtkReverseSense.h>
 
 /// ITK includes
 #include <itksys/Directory.hxx>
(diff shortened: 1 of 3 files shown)


lassoan:
Use vtkTransformPolyDataFilter instead of manually going through the points. Use a transformation matrix:
Element[0][0] = flipX ? -1 : 1
Element[1][1] = flipY ? -1 : 1
Element[2][2] = flipZ ? -1 : 1

Manually inverting point coordinates is insufficient, as you need to update associated point/cell data, too (for example surface normal). vtkTransformPolyDataFilter takes care of all that.

Modules/Loadable/Models/Logic/vtkSlicerModelsLogic.cxx
```diff
@@ -438,33 +439,102 @@ void vtkSlicerModelsLogic::TransformModel(vtkMRMLTransformNode *tnode,
...
-    //--- Polygons are not automatically re-stripped.
-    vtkNew<vtkPolyDataNormals> normals;
-    normals->SetInputData(poly.GetPointer());
-    //--- NOTE: This assumes a completely closed surface
-    //---(i.e. no boundary edges) and no non-manifold edges.
-    //--- If these constraints do not hold, the AutoOrientNormals
-    //--- is not guaranteed to work.
+  modelOut->SetPolyDataConnection(normals->GetOutputPort());
+
+  //modelOut->SetAndObservePolyData(poly);
```
bpaniagua: Thanks for the feedback! I have updated the code to use vtkTransformPolyDataFilter and there is still the problem described here http://na-mic.org/Mantis/view.php?id=4297 unless normals are explicitly filtered using the vtkPolyDataNormals. Do you think this is something we should report to the VTK folks?

lassoan:
If you turn the surface inside-out (apply a transformation matrix that has a negative determinant - http://www.vtk.org/doc/nightly/html/classvtkMatrix4x4.html#a5c707371ad2d298e434c01a091359061) then the ordering of triangle points become inverted. In this case you have to apply vtkReverseSense filter to "fix" the surface. This is all expected behavior, not a problem.

lassoan:
Note that there are two somewhat independent issues to manage: cell point ordering and normal direction. See from page 340 of the VTK textbook http://www.kitware.com/products/books/VTKTextbook.pdf

bpaniagua:
@lassoan I have introduced the changes requested. Let me know what you think.

lassoan:
I've merged your changes into SurfaceToolbox.py, so auto-orient normals and mirror options are available in that module.

I haven't merged Models module changes, as it's a much more complex topic.

1. vtkSlicerModelsLogic::TransformModel is not used anywhere and there is already vtkMRMLModelNode::ApplyTransform that transforms a model. vtkMRMLModelNode::ApplyTransform could be made more sophisticated so that it can apply vtkReverseSense aumatically, but then we should also handle it in the displayable manager (to avoid having black models when the transform is applied dynamically and not hardened yet) and it should work for non-linear transforms support presence/absence of normals, etc. Such automatic mechanisms are not trivial to implement and could interfere with modules that are aware that after certain transforms cells or normals have to be reoriented.

2. Mirroring code is so trivial (see in surface toolbox module, just 5-10 lines), just uses 1 or 2 VTK filters, mirroring is rarely needed (in general any kind of "flip" operations should be avoided), SurfaceToolbox module would require mirror operation on vtkPolyData objects and not on model nodes, so using the model logic would just make it more complex. So, it does not seem to be very useful to add mirror operation to models module logic.

bpaniagua:
Hi @lassoan. Thank you!
It makes sense, but by implementing those functionalities in the models logic was @jcfr 's suggestion. By implementing them in the logic we can reach them at any part of Slicer (unlike python code, the logic can be reached from both python and CLI modules). But as you said this case scenario might not be a very likely thing to happen. Thanks!

lassoan:
In general I agree that algorithms should be implemented in module logic classes. However, due to the specific reasons I listed, in this case I think it's better not to add it to the logic.

PR #627, opened 2016-11-18
Fix segmentation core warnings

jcfr:


jcfr:
Cc: @lassoan 

Additionally I am note sure if the warning reported in http://na-mic.org/Mantis/view.php?id=4305 can simply be ignored or if it is a real problem ...

jcfr:
Integrated in [r25594](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25594)

cpinter:
The targetRepresentationName argument is actually used in the function, and the function clearly works, so I can only imagine that the warning is incorrect.

What do you suggest it can mean?

jcfr:
@cpinter Looking at the code, I don't see where `targetRepresentationName` is used. Copied below for convenience:

```
bool vtkSegmentation::CreateRepresentation(const std::string& targetRepresentationName,
                                           vtkSegmentationConverter::ConversionPathType path,
                                           vtkSegmentationConverterRule::ConversionParameterListType parameters)
{
  if (!this->Converter)
    {
    vtkErrorMacro("CreateRepresentation: Invalid converter!");
    return false;
    }
  if (path.empty())
    {
    return false;
    }

  // Set conversion parameters
  this->Converter->SetConversionParameters(parameters);

  // Perform conversion on all segments (do overwrites)
  for (SegmentMap::iterator segmentIt = this->Segments.begin(); segmentIt != this->Segments.end(); ++segmentIt)
    {
    if (!this->ConvertSegmentUsingPath(segmentIt->second, path, true))
      {
      vtkErrorMacro("CreateRepresentation: Conversion failed");
      return false;
      }
    const char* segmentId = segmentIt->first.c_str();
    this->InvokeEvent(vtkSegmentation::RepresentationModified, (void*)segmentId);
    }

  this->InvokeEvent(vtkSegmentation::ContainedRepresentationNamesModified);
  return true;
}
```
Copied from https://github.com/Slicer/Slicer/blob/dee6335f84505200468d5868efb9297aa0358c88/Libs/vtkSegmentationCore/vtkSegmentation.cxx#L956-L988

cpinter:
I see, I didn't notice there were two methods with that name. Yes it's strange that it's not used in the second one. I'm trying to figure out why

cpinter:
It's not necessary because the second argument "path" contains that information as well. Probably it's best to simply remove the first argument. It's only used from qMRMLSegmentationConversionParametersWidget::applyConversion()

jcfr:
Great. Thanks for your help. Fix integrated in [r25597](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25597)

cpinter:
Perfect, thanks!

PR #629, opened 2016-11-25
ENH: update DICOMExtensions

fedorov:
Added new entries for QuantitativeReporting, which supports Parametric Maps and
certain types of Structured Reports, in addition to DICOM.

Also "prettified" JSON formatting.

### DIFF
--- a/Modules/Scripted/DICOM/DICOMExtensions.json
+++ b/Modules/Scripted/DICOM/DICOMExtensions.json
@@ -31,30 +31,30 @@
         ]
       }
     },
-		{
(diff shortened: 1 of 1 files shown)


pieper:
Yes, that should be fine.  I don't recall there being anything 'magic' about Modality.  You should be able to put any tag name and value in that field.  The SlicerHeart one also only has SOPClassUID.

pieper:
Merged as r25578 = ad7cd3967d938950b795310c8cebe2b842ac5df3

PR #630, opened 2016-11-28
Segment color refactor

cpinter:
Addresses the issue of segment color stored in multiple locations (vtkSegment::DefaultColor and vtkMRMLSegmentationDisplayNode::SegmentDisplayProperties::Color). It caused ambiguity and increased complexity when saving and loading. Also it was inconsistent with the recent changes related to terminologies, in that the current terminology information was stored in the segment, but the current color was not (DefaultColor was for keeping track of the original color from DICOM etc.).

### DIFF
--- a/Libs/MRML/Core/vtkMRMLSegmentationDisplayNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLSegmentationDisplayNode.cxx
@@ -55,7 +55,7 @@ vtkMRMLNodeNewMacro(vtkMRMLSegmentationDisplayNode);
 vtkMRMLSegmentationDisplayNode::vtkMRMLSegmentationDisplayNode()
   : PreferredDisplayRepresentationName2D(NULL)
   , PreferredDisplayRepresentationName3D(NULL)
-  , NumberOfAddedSegments(0)
+  , NumberOfGeneratedColors(0)
   , SegmentListUpdateTime(0)
   , SegmentListUpdateSource(0)
   , Visibility3D(true)
@@ -365,19 +365,6 @@ void vtkMRMLSegmentationDisplayNode::SetSegmentDisplayProperties(std::string seg
       }
     }
 
-  // Save cached value of color
-  // TODO: remove this when terminology infrastructure is in place
-  vtkMRMLSegmentationNode* segmentationNode = vtkMRMLSegmentationNode::SafeDownCast(this->GetDisplayableNode());
-  if (segmentationNode)
-    {
-    vtkSegmentation* segmentation = segmentationNode->GetSegmentation();
-    vtkSegment* segment = segmentation->GetSegment(segmentId);
-    if (segment)
-      {
-      segment->SetDefaultColorWithoutModifiedEvent(properties.Color);
-      }
-    }
-
   if (modified)
     {
     this->Modified();
@@ -401,15 +388,26 @@ vtkVector3d vtkMRMLSegmentationDisplayNode::GetSegmentColor(std::string segmentI
 }
 
 //---------------------------------------------------------------------------
-void vtkMRMLSegmentationDisplayNode::SetSegmentColor(std::string segmentID, double r, double g, double b)
+void vtkMRMLSegmentationDisplayNode::GetSegmentColor(std::string segmentID, double* color)
 {
-  // Set color in display properties
-  SegmentDisplayProperties properties;
-  this->GetSegmentDisplayProperties(segmentID, properties);
-  properties.Color[0] = r;
-  properties.Color[1] = g;
-  properties.Color[2] = b;
-  this->SetSegmentDisplayProperties(segmentID, properties);
+  if (!color)
+    {
+    vtkErrorMacro("GetSegmentColor: Invalid output color array");
+    return;
+    }
+  vtkVector3d colorVtk = this->GetSegmentColor(segmentID);
+  color[0] = colorVtk.GetX();
+  color[1] = colorVtk.GetY();
+  color[2] = colorVtk.GetZ();
+}
+
+//---------------------------------------------------------------------------
+void vtkMRMLSegmentationDisplayNode::GetSegmentColor(std::string segmentID, double &r, double &g, double &b)
+{
+  vtkVector3d colorVtk = this->GetSegmentColor(segmentID);
+  r = colorVtk.GetX();
+  g = colorVtk.GetY();
+  b = colorVtk.GetZ();
 }
 
 //---------------------------------------------------------------------------
@@ -420,6 +418,18 @@ void vtkMRMLSegmentationDisplayNode::SetSegmentColor(std::string segmentID, vtkV
   properties.Color[0] = color.GetX();
   properties.Color[1] = color.GetY();
   properties.Color[2] = color.GetZ();
+  this->SetSegmentDisplayProperties(segmentID, properties);
+}
+
+//---------------------------------------------------------------------------
+void vtkMRMLSegmentationDisplayNode::SetSegmentColor(std::string segmentID, double r, double g, double b)
+{
+  // Set color in display properties
+  SegmentDisplayProperties properties;
+  this->GetSegmentDisplayProperties(segmentID, properties);
+  properties.Color[0] = r;
+  properties.Color[1] = g;
+  properties.Color[2] = b;
   this->SetSegmentDisplayProperties(segmentID, properties);
 }
 
@@ -697,25 +707,30 @@ void vtkMRMLSegmentationDisplayNode::SetSegmentDisplayPropertiesToDefault(const 
 
   int wasModifyingDisplayNode = this->StartModify();
 
-  // Set segment color for merged labelmap
-  double defaultColor[3] = { 0.0, 0.0, 0.0 };
-  segment->GetDefaultColor(defaultColor);
-  // Generate color if default color is the default gray
-  this->NumberOfAddedSegments += 1;
-  bool generateNewDefaultColor =
-    (defaultColor[0] == vtkSegment::SEGMENT_COLOR_VALUE_INVALID[0]
-    && defaultColor[1] == vtkSegment::SEGMENT_COLOR_VALUE_INVALID[1]
-    && defaultColor[2] == vtkSegment::SEGMENT_COLOR_VALUE_INVALID[2]);
-  if (generateNewDefaultColor)
+  // Generate color
+  double color[3] = { vtkSegment::SEGMENT_COLOR_VALUE_INVALID[0],
+                      vtkSegment::SEGMENT_COLOR_VALUE_INVALID[1],
+                      vtkSegment::SEGMENT_COLOR_VALUE_INVALID[1] };
+  vtkMRMLSegmentationDisplayNode* segmentationDisplayNode = vtkMRMLSegmentationDisplayNode::SafeDownCast(
+    segmentationNode->GetDisplayNode() );
+  if (segmentationDisplayNode)
     {
-    this->GenerateSegmentColor(defaultColor);
+    segmentationDisplayNode->GetSegmentColor(segmentId, color);
+    }
+  // Generate color if default color is the default gray
+  bool generateNewColor = ( color[0] == vtkSegment::SEGMENT_COLOR_VALUE_INVALID[0]
+                         && color[1] == vtkSegment::SEGMENT_COLOR_VALUE_INVALID[1]
+                         && color[2] == vtkSegment::SEGMENT_COLOR_VALUE_INVALID[2] );
+  if (generateNewColor)
+    {
+    this->GenerateSegmentColor(color);
     }
 
   // Add entry in segment display properties
   vtkMRMLSegmentationDisplayNode::SegmentDisplayProperties properties;
-  properties.Color[0] = defaultColor[0];
-  properties.Color[1] = defaultColor[1];
-  properties.Color[2] = defaultColor[2];
+  properties.Color[0] = color[0];
+  properties.Color[1] = color[1];
+  properties.Color[2] = color[2];
   properties.Visible = true;
   properties.Visible3D = true;
   properties.Visible2DFill = true;
@@ -744,7 +759,7 @@ void vtkMRMLSegmentationDisplayNode::RemoveSegmentDisplayProperties(std::string 
 void vtkMRMLSegmentationDisplayNode::ClearSegmentDisplayProperties()
 {
   this->SegmentationDisplayProperties.clear();
-  this->NumberOfAddedSegments = 0;
+  this->NumberOfGeneratedColors = 0;
   this->Modified();
 }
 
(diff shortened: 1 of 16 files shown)


jcfr:
> IMPORTANT! This solution will not be integrated to Slicer core. Although this works almost perfectly (the only practical problem is that undo/redo cannot restore color any more), it is not a good theoretical solution. Despite the fact that color is a display property, it needs to be stored in the segment so that it has a consistent state with terminology and name. Also mechanisms have been implemented to circumvent the fact that the actual segment color is in the display node, which are undesired. The proposed solution:
> - Store actual color of segment in vtkSegment::Color
> - Allow display node to override segment color (so that different colors can be used in different views etc.) SegmentDisplayProperties::OverrideColor
> - Storage gets/sets the color from/to the segment itself, the override color comes for the scene if necessary

Is the *color* of segment always look up in a table ? 

Instead of associating RGB value with a segment, should it only be associated with an ID (from the terminology module) ?   With that in place, the selection of color would be done at the display level.

cpinter:
@jcfr Segmentations don't have color table nodes any more, so the color property is an RGB color, and cannot be index.
With terminology in place, an RGB color still needs to be stored, as the user can override the default color from terminology (in fact I expect they will do that in most case because the color from terminology is always the same brownish color with slight variations)

Libs/MRML/Core/vtkMRMLSegmentationDisplayNode.cxx
```diff
@@ -365,62 +368,121 @@ void vtkMRMLSegmentationDisplayNode::SetSegmentDisplayProperties(std::string seg
...
-    vtkSegmentation* segmentation = segmentationNode->GetSegmentation();
-    vtkSegment* segment = segmentation->GetSegment(segmentId);
-    if (segment)
-      {
-      segment->SetDefaultColorWithoutModifiedEvent(properties.Color);
-      }
+    vtkErrorMacro("GetSegmentColor: No segmentation node associated to this display node");
+    return;
+    }
+  vtkSegmentation* segmentation = segmentationNode->GetSegmentation();
```
lassoan: For efficient implementation of segmentation sequences (allowing sequence nodes changing in time, replayed with reasonable speed) I had to allow setting Segmentation member from outside and so now it is possible that temporarily Segmentation is NULL. So, it might worth checking if segmentation is NULL.

Libs/MRML/Core/vtkMRMLSegmentationDisplayNode.cxx
```diff
@@ -365,62 +368,121 @@ void vtkMRMLSegmentationDisplayNode::SetSegmentDisplayProperties(std::string seg
...
+//---------------------------------------------------------------------------
+void vtkMRMLSegmentationDisplayNode::GetSegmentColor(std::string segmentID, double* color)
+{
+  if (!color)
+    {
+    vtkErrorMacro("GetSegmentColor: Invalid output color array");
+    return;
+    }
+
+  // Invalidate color
```
lassoan: It would better to not constrain developers to set segmentation node before they can set/get override color. We should probably check first if override color is set for a specific segment ID and if it is set then return the override color. If override color is not found then check the color in segmentation node.

Libs/MRML/Core/vtkMRMLSegmentationDisplayNode.cxx
```diff
@@ -365,62 +368,121 @@ void vtkMRMLSegmentationDisplayNode::SetSegmentDisplayProperties(std::string seg
...
+  return colorVtk;
+}
+
+//---------------------------------------------------------------------------
+void vtkMRMLSegmentationDisplayNode::GetSegmentColor(std::string segmentID, double* color)
+{
+  if (!color)
+    {
+    vtkErrorMacro("GetSegmentColor: Invalid output color array");
+    return;
```
lassoan: should we return with bool?

Libs/MRML/Core/vtkMRMLSegmentationDisplayNode.cxx
```diff
@@ -365,62 +368,121 @@ void vtkMRMLSegmentationDisplayNode::SetSegmentDisplayProperties(std::string seg
...
 
-  if (modified)
+  // Get color from segment
+  segment->GetColor(color);
+
+  // Get display properties
+  this->UpdateSegmentList();
+  SegmentDisplayPropertiesMap::iterator propsIt = this->SegmentationDisplayProperties.find(segmentID);
+  if ( propsIt != this->SegmentationDisplayProperties.end()
+    && propsIt->second.OverrideColor[0] != SEGMENT_COLOR_NO_OVERRIDE
```
lassoan: It might be safer to check (overrideColor[0]>=0 && overrideColor[1]>=0 && overrideColor[2]>=0) to make sure the color is fully valid

Libs/MRML/Core/vtkMRMLSegmentationDisplayNode.cxx
```diff
@@ -365,62 +368,121 @@ void vtkMRMLSegmentationDisplayNode::SetSegmentDisplayProperties(std::string seg
...
 {
-  // Set color in display properties
+  // Set override color in display properties
   SegmentDisplayProperties properties;
-  this->GetSegmentDisplayProperties(segmentID, properties);
-  properties.Color[0] = r;
-  properties.Color[1] = g;
-  properties.Color[2] = b;
-  this->SetSegmentDisplayProperties(segmentID, properties);
+  if (this->GetSegmentDisplayProperties(segmentID, properties))
```
lassoan: we should allow setting segment override color no matter if a segmentation node is set or not (for example, display node may be read from XML before the segment node is read)

Libs/MRML/Core/vtkMRMLSegmentationDisplayNode.cxx
```diff
@@ -365,62 +368,121 @@ void vtkMRMLSegmentationDisplayNode::SetSegmentDisplayProperties(std::string seg
...
-void vtkMRMLSegmentationDisplayNode::SetSegmentColor(std::string segmentID, vtkVector3d color)
+void vtkMRMLSegmentationDisplayNode::SetSegmentOverrideColor(std::string segmentID, vtkVector3d overrideColor)
 {
   SegmentDisplayProperties properties;
-  this->GetSegmentDisplayProperties(segmentID, properties);
-  properties.Color[0] = color.GetX();
-  properties.Color[1] = color.GetY();
-  properties.Color[2] = color.GetZ();
-  this->SetSegmentDisplayProperties(segmentID, properties);
+  if (this->GetSegmentDisplayProperties(segmentID, properties))
```
lassoan: allow always

Libs/MRML/Core/vtkMRMLSegmentationDisplayNode.cxx
```diff
@@ -697,25 +759,26 @@ void vtkMRMLSegmentationDisplayNode::SetSegmentDisplayPropertiesToDefault(const
...
+  bool generateNewColor = ( color[0] == vtkSegment::SEGMENT_COLOR_INVALID[0]
+                         && color[1] == vtkSegment::SEGMENT_COLOR_INVALID[1]
+                         && color[2] == vtkSegment::SEGMENT_COLOR_INVALID[2] );
+  if (generateNewColor)
     {
-    this->GenerateSegmentColor(defaultColor);
+    this->GenerateSegmentColor(color);
+
+    // Set color to segment (no override is specified by default, so segment color is used)
+    segment->SetColor(color);
```
lassoan: Can we avoid this? It's difficult to foresee all the side-effects of modifying the segmentation when a set method on the display node is called. Could GenerateSegmentColor be part of Segmentation node and called from higher level (not from the display node)?
Instead of just generating segment colors for tissue/tissue in the display node, we could actually generate terminology items in segmentation node with default color and segment name: tissue/bone, tissue/bodyfat, tissue/muscle, ... (in the future users could customize the list and in the terminology window there could be a simple quick-select list where the user could select a segment with a single click).

Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
```diff
@@ -388,14 +379,14 @@ int vtkMRMLSegmentationStorageNode::ReadBinaryLabelmapRepresentation4DSpatial(vt
...
-    std::string defaultColorValue;
+    // Color
+    std::string colorString;
     itk::ExposeMetaData<std::string>(
           metadata,
-          GetSegmentMetaDataKey(segmentIndex, KEY_SEGMENT_DEFAULT_COLOR),
-          defaultColorValue);
-    double currentSegmentDefaultColor[3] = {0.0,0.0,0.0};
-    GetSegmentDefaultColorFromString(currentSegmentDefaultColor, defaultColorValue);
+          GetSegmentMetaDataKey(segmentIndex, KEY_SEGMENT_COLOR),
```
lassoan: Is this backward-compatible (work with both Color and DefaultColor)?

Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
```diff
@@ -388,14 +379,14 @@ int vtkMRMLSegmentationStorageNode::ReadBinaryLabelmapRepresentation4DSpatial(vt
           currentSegmentName);
     currentSegment->SetName(currentSegmentName.c_str());
 
-    // DefaultColor
-    std::string defaultColorValue;
+    // Color
+    std::string colorString;
```
lassoan: What happens if the color is not specified? Can we look up a default from terminology?

Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
```diff
@@ -824,25 +813,25 @@ int vtkMRMLSegmentationStorageNode::ReadPolyDataRepresentation(vtkMRMLSegmentati
...
       currentSegmentName = ssCurrentSegmentName.str();
       }
     currentSegment->SetName(currentSegmentName.c_str());
 
-    double defaultColor[3]={1.0, 0.0, 0.0};
-    vtkDoubleArray* defaultColorArray = vtkDoubleArray::SafeDownCast(
-      currentPolyData->GetFieldData()->GetArray(GetSegmentMetaDataKey(SINGLE_SEGMENT_INDEX, KEY_SEGMENT_DEFAULT_COLOR).c_str()) );
-    if (defaultColorArray && defaultColorArray->GetNumberOfTuples() > 0 && defaultColorArray->GetNumberOfComponents() == 3)
+    double color[3]={1.0, 0.0, 0.0};
+    vtkDoubleArray* ColorArray = vtkDoubleArray::SafeDownCast(
```
lassoan: capitalization should be colorArray

Libs/vtkSegmentationCore/vtkSegment.cxx
```diff
@@ -41,7 +41,7 @@
 #include <sstream>
 
 //----------------------------------------------------------------------------
-const double vtkSegment::SEGMENT_COLOR_VALUE_INVALID[4] = {0.5, 0.5, 0.5, 1.0};
+const double vtkSegment::SEGMENT_COLOR_INVALID[4] = {0.5, 0.5, 0.5, 1.0};
```
lassoan: do we need the 4 components? as I see only 3 components are used

Libs/vtkSegmentationCore/vtkSegment.cxx
```diff
@@ -379,11 +379,3 @@ void vtkSegment::ExtendBounds(double partialBounds[6], double globalBounds[6])
     globalBounds[5] = partialBounds[5];
     }
 }
-
-//---------------------------------------------------------------------------
-void vtkSegment::SetDefaultColorWithoutModifiedEvent(double color[3])
```
lassoan: It's so great that it's gone!

Libs/vtkSegmentationCore/vtkSegment.h
```diff
@@ -143,10 +137,9 @@ class vtkSegmentationCore_EXPORT vtkSegment : public vtkObject
   /// This is the default identifier of the segment within segmentation, so needs to be unique within a segmentation
   char* Name;
 
-  /// Default color
-  /// Called default because this is only used initially indicating the original color of the segment,
-  /// but then copies are made for display and this member variable has no effect on it afterwards.
-  double DefaultColor[3];
+  /// Color
+  /// The actual color the segment is shown in. Can be overridden in the display node for advanced purposes
```
lassoan: we could be a bit more specific than "advanced purposes", for example:

The actual color the segment is shown in. Can be overridden in the display node to allow displaying segment in selected views in a different color

Modules/Loadable/Segmentations/MRMLDM/vtkMRMLSegmentationsDisplayableManager2D.cxx
```diff
@@ -809,12 +811,12 @@ void vtkMRMLSegmentationsDisplayableManager2D::vtkInternal::UpdateDisplayNodePip
 
       // Update pipeline actors
       pipeline->PolyDataOutlineActor->SetVisibility(segmentOutlineVisible);
-      pipeline->PolyDataOutlineActor->GetProperty()->SetColor(properties.Color[0], properties.Color[1], properties.Color[2]);
+      pipeline->PolyDataOutlineActor->GetProperty()->SetColor(color[0], color[1], color[2]);
```
lassoan: very small thing - can we just use SetColor(color)?

Libs/MRML/Core/vtkMRMLSegmentationDisplayNode.cxx
cpinter: Let's keep this in mind and I'll implement later what we agreed on.

Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
cpinter: I probably missed this spot... In other places it is.

Libs/MRML/Core/vtkMRMLSegmentationDisplayNode.cxx
lassoan: Yes, I agree that we should commit this as is and improve in the next iteration.

cpinter:
Thanks for the comments, Andras! I addressed them and amended the commit.

Let's talk about two of them and I'll fix them seperately:
* GenerateDefaultColor
* Getting color from terminology if Color header item is missing from segmentation file

lassoan:
Thanks! Both topics can be addressed in a next iteration. If there are no test failures and risk of breaking something for RSNA then it should be merged.

cpinter:
I ran py_SegmentationTest1, as far as I know the only one that could be influenced by segmentations internals at the moment. I can run all of them to be on the safe side.
Also did some manual testing and everything seemed alright.

cpinter:
Integrated in http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25580

PR #631, opened 2016-12-05
ENH: Added ViewGroup property to view nodes

lassoan:
ViewGroup is used for restricting scope of:
* linked slice view property changes (if slices are linked, a property change will only change views in the same group)
* crosshair jump to slice (if crosshair is moved with shift+mousemove and slice jump is enabled, only those slices will be moved that are in the same group as the view where the mouse was)
* slice intersection display (slice intersections will only shown of those slices that are in the same group)

Implementation:
* Added new ViewGroup member to abstract view node (string, empty by default)
* In default layouts "Three over three" and "Four over four" separated top/bottom rows (view IDs changed to Red+, Yellow+, Green+, 1+; unique node IDs were necessary because on layout change view group is not updated, so we cannot share view nodes between layouts the same group ID for all views)
* Updated vtkMRMLSliceNode::JumpAllSlices to only jump those slices that have matching viewgroup
* Updated vtkMRMLSliceLinkLogic::OnMRMLNodeModified(vtkMRMLNode* node) to make aware of groups
* Updated slice displayable manager to only show slice intersections in the same group (and mapped in layout)
* Updated 2D/3D displayable managers to only jump those slices on crosshair update (shift+mousemove) that are in the same group
* Updated GUI:
** Added View group view/edit to Slice information widget (used in View Controllers module).
** Added viewGroup property to simple markups widget (that can restrict slice jumping to a specific view group)

### DIFF
--- a/Libs/MRML/Core/vtkMRMLAbstractViewNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLAbstractViewNode.cxx
@@ -37,6 +37,7 @@ static const char* DEFAULT_AXIS_LABELS[vtkMRMLAbstractViewNode::AxisLabelsCount]
 vtkMRMLAbstractViewNode::vtkMRMLAbstractViewNode()
 {
   this->LayoutLabel = NULL;
+  this->ViewGroup = NULL;
   this->Active = 0;
   this->Visibility = 1;
 
@@ -64,11 +65,8 @@ vtkMRMLAbstractViewNode::vtkMRMLAbstractViewNode()
 //----------------------------------------------------------------------------
 vtkMRMLAbstractViewNode::~vtkMRMLAbstractViewNode()
 {
-  if ( this->LayoutLabel )
-    {
-    delete [] this->LayoutLabel;
-    this->LayoutLabel = NULL;
-    }
+  this->SetLayoutLabel(NULL);
+  this->SetViewGroup(NULL);
 }
 
 //----------------------------------------------------------------------------
@@ -88,6 +86,10 @@ void vtkMRMLAbstractViewNode::WriteXML(ostream& of, int nIndent)
     {
     of << indent << " layoutName=\"" << this->GetLayoutName() << "\"";
     }
+  if (this->GetViewGroup())
+  {
+    of << indent << " viewGroup=\"" << this->GetViewGroup() << "\"";
+  }
 
   of << indent << " active=\"" << (this->Active ? "true" : "false") << "\"";
   of << indent << " visibility=\"" << (this->Visibility ? "true" : "false") << "\"";
@@ -133,13 +135,17 @@ void vtkMRMLAbstractViewNode::ReadXMLAttributes(const char** atts)
     attName = *(atts++);
     attValue = *(atts++);
     if (!strcmp(attName, "layoutLabel"))
-      {
-      this->SetLayoutLabel( attValue );
-      }
+    {
+      this->SetLayoutLabel(attValue);
+    }
     else if (!strcmp(attName, "layoutName"))
       {
       this->SetLayoutName( attValue );
       }
+    else if (!strcmp(attName, "viewGroup"))
+    {
+      this->SetViewGroup(attValue);
+    }
     else if (!strcmp(attName, "backgroundColor"))
       {
       std::stringstream ss;
@@ -295,7 +301,8 @@ void vtkMRMLAbstractViewNode::Copy(vtkMRMLNode *anode)
   vtkMRMLAbstractViewNode *node = (vtkMRMLAbstractViewNode *) anode;
 
   this->SetLayoutLabel(node->GetLayoutLabel());
-  this->SetBackgroundColor ( node->GetBackgroundColor ( ) );
+  this->SetViewGroup(node->GetViewGroup());
+  this->SetBackgroundColor(node->GetBackgroundColor());
   this->SetBackgroundColor2 ( node->GetBackgroundColor2 ( ) );
   // Important: do not use SetActive or RemoveActiveFlagInScene will be called
   this->Active = node->GetActive();
@@ -328,9 +335,11 @@ void vtkMRMLAbstractViewNode::Reset(vtkMRMLNode* defaultNode)
   // automatically.
   // This require a custom behavior implemented here.
   std::string layoutLabel = this->GetLayoutLabel() ? this->GetLayoutLabel() : "";
+  std::string viewGroup = this->GetViewGroup() ? this->GetViewGroup() : "";
   this->Superclass::Reset(defaultNode);
   this->DisableModifiedEventOn();
   this->SetLayoutLabel(layoutLabel.c_str());
+  this->SetViewGroup(viewGroup.empty() ? NULL : viewGroup.c_str());
   this->AxisLabels->Reset();
   for (int i=0; i<vtkMRMLAbstractViewNode::AxisLabelsCount; i++)
     {
@@ -345,6 +354,7 @@ void vtkMRMLAbstractViewNode::PrintSelf(ostream& os, vtkIndent indent)
   Superclass::PrintSelf(os,indent);
 
   os << indent << "LayoutLabel: " << (this->LayoutLabel ? this->LayoutLabel : "(null)") << std::endl;
+  os << indent << "ViewGroup: " << (this->ViewGroup ? this->ViewGroup : "(null)") << std::endl;
   os << indent << "Active:        " << this->Active << "\n";
   os << indent << "Visibility:        " << this->Visibility << "\n";
   os << indent << "BackgroundColor:       " << this->BackgroundColor[0] << " "
--- a/Libs/MRML/Core/vtkMRMLAbstractViewNode.h
+++ b/Libs/MRML/Core/vtkMRMLAbstractViewNode.h
@@ -71,6 +71,16 @@ public:
   inline const char *GetLayoutName();
 
   ///
+  /// An optional identifier to link groups of views. Views that have matching
+  /// ViewGroup value are in the same group.
+  /// By default the value is NULL, which is equivalent with ampty string.
+  /// Empty is a valid value: views that have empty ViewGroup value form a group.
+  /// When changing linked properties in a view then all other views in the group
+  /// are changed accordingly, but other views are not modified.
+  vtkSetStringMacro(ViewGroup);
+  vtkGetStringMacro(ViewGroup);
+
+  ///
   /// Label for the view. Usually a 1 character label, e.g. R, 1, 2, etc.
   /// \sa SetLayoutName()
   vtkSetStringMacro(LayoutLabel);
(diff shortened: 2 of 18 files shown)


fedorov:
Looks very helpful! I will plan to build and test (it looks like the functionality is accessible from the GUI, right?)

lassoan:
Yes, accessible from GUI:
* Choose "Three over three" or "Four over four" layout - the top and bottom rows are two different groups
* You can change view group of individual slice views in the View Controllers module's Slice information section (I didn't add a new widget for changing view group for 3D views, but if you think it's important then I'll implement that)

lassoan:
If there are no comments then I'll merge this tomorrow.

Libs/MRML/Core/vtkMRMLAbstractViewNode.h
```diff
@@ -71,6 +71,16 @@ class VTK_MRML_EXPORT vtkMRMLAbstractViewNode
   inline const char *GetLayoutName();
 
   ///
+  /// An optional identifier to link groups of views. Views that have matching
+  /// ViewGroup value are in the same group.
+  /// By default the value is NULL, which is equivalent with ampty string.
```
vovythevov: Minor typo: "ampty"

Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
```diff
@@ -272,6 +272,34 @@ GetSliceLogicByLayoutName(const char* layoutName) const
...
+  {
+    if (logic->GetSliceModelDisplayNode() == displayNode)
+    {
+      break;
+    }
+
+    logic = 0;
+  }
+
+  return logic;
```
vovythevov: Would it be more clear to do:
```
for (logics->InitTraversal(it);
    (logic = vtkMRMLSliceLogic::SafeDownCast(logics->GetNextItemAsObject(it)));)
  {
  if (logic->GetSliceModelDisplayNode() == displayNode)
    {
    return logic;
    }
  }
return NULL;
```

Libs/MRML/Core/vtkMRMLAbstractViewNode.cxx
```diff
@@ -88,6 +86,10 @@ void vtkMRMLAbstractViewNode::WriteXML(ostream& of, int nIndent)
     {
     of << indent << " layoutName=\"" << this->GetLayoutName() << "\"";
     }
+  if (this->GetViewGroup())
+  {
+    of << indent << " viewGroup=\"" << this->GetViewGroup() << "\"";
+  }
```
vovythevov: Minor style: indentation

Libs/MRML/Core/vtkMRMLAbstractViewNode.cxx
```diff
@@ -133,13 +135,17 @@ void vtkMRMLAbstractViewNode::ReadXMLAttributes(const char** atts)
...
+      this->SetLayoutLabel(attValue);
+    }
     else if (!strcmp(attName, "layoutName"))
       {
       this->SetLayoutName( attValue );
       }
+    else if (!strcmp(attName, "viewGroup"))
+    {
+      this->SetViewGroup(attValue);
+    }
```
vovythevov: Minor style: indentation

Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
vovythevov: Minor style: indentation

Libs/MRML/Logic/vtkMRMLLayoutLogic.cxx
```diff
@@ -1519,6 +1526,17 @@ void vtkMRMLLayoutLogic::ApplyProperty(const ViewProperty& property, vtkMRMLNode
...
+  {
+    vtkMRMLAbstractViewNode* viewNode = vtkMRMLAbstractViewNode::SafeDownCast(view);
+    if (!viewNode)
+    {
+      vtkWarningMacro("Invalid view group property.");
+      return;
+    }
+    viewNode->SetViewGroup(value.c_str());
+  }
   // LayoutLabel
```
vovythevov: Minor style: indentation

Libs/MRML/Widgets/qMRMLSliceInformationWidget.cxx
```diff
@@ -227,6 +232,26 @@ void qMRMLSliceInformationWidget::setSliceVisible(bool visible)
...
+
+  if (!viewGroup.isEmpty())
+    {
+    d->MRMLSliceNode->SetViewGroup(viewGroup.toLatin1());
+    }
+  else
+    {
+    d->MRMLSliceNode->SetViewGroup(NULL);
+    }
+}
```
vovythevov: Minor style: indentation

lassoan:
Thank you, I'll fix these!

fedorov:
I won't be able to test it today. I will run the build overnight or tomorrow.

lassoan:
There is an issue with the DataProbe widget (Python wrapping interestingly cannot distinguish two method names with different input types), I'll fix that tomorrow (it should not interfere with the testing).

I also plan to change the ViewGroup to be an integer instead of a string, as there are just too many string comparisons, null-pointer-checks, special value for matching string names, etc. and also it's not really good that we have to invent a string that can be mistyped, etc.

fedorov:
One issue I noticed is that there is a huge lag updating slice views. I built in debug mode, but I wonder if this is due to this change, or existed before. Even if I remove all groups, I still get a big lag. Is this feature expected to reduce performance? Are there any performance tests to run to evaluate the impact?

Another comment - I wonder where a developer is expected to learn about the behavior implementation decisions made while implementing this feature. It seems logical as I read it, but I will need to remember to search for this PR to read the summary again, when the time comes to use it...

jcfr:
> to learn about the behavior implementation decisions made while implementing this feature

May the information could be added to https://www.slicer.org/wiki/Documentation/Nightly/Developers/Layouts and/or https://www.slicer.org/wiki/Documentation/Nightly/Developers/DisplayableManagers

pieper:
Are there any tests?  I don't see any in the PR.

It would be helpful to do some profiling of switching layouts and other
slice view operations.  I noticed that even on the 4.6.2 release build on a
fast windows machine things are slower than I would expect.

lassoan:
I've added some documentation to headers, but I'll add more.

lassoan:
I didn't want to add any tests before there is an agreed solution. I'll add some.

lassoan:
Performance shouldn't be impacted (we just skip some views), but I'll double-check. I don't think there is a test for shift+mousemove performance. I'll see if I can add it to the performance tests.

There is a significant performance degradation in VTK7 in debug mode because they added a layer of indirection in accessing data arrays (http://www.vtk.org/doc/nightly/html/classvtkGenericDataArray.html). Maybe you see unexpectedly slow performance because of that.

As a sidenote, I think implementing generic array indirection like this was a mistake, because they slowed down and made more complex the entire toolkit just to serve a narrow use case. In release mode the performance degradation is just about 10-15%, so I didn't want to start fighting back on this - there are other changes in VTK I would like to push instead (such as introduction of oriented image data).

lassoan:
I've added a crosshair slice jump speed test to performance tests and tried it on Slicer 4.6.2 and the working copy. There was no measureable speed difference.

Slicer 4.6.2:
number of slice views = 3, fps = 10.8617 (92.0667 ms per frame)
number of slice views = 3, fps = 10.0402 (99.6 ms per frame)
number of slice views = 3, fps = 9.36622 (106.767 ms per frame)
number of slice views = 3, fps = 12.2349 (81.7333 ms per frame)
number of slice views = 3, fps = 12.0192 (83.2 ms per frame)

Slicer trunk with view groups:
number of slice views = 3, fps = 12.7986 (78.1333 ms per frame)
number of slice views = 3, fps = 8.70322 (114.9 ms per frame)
number of slice views = 3, fps = 9.38673 (106.533 ms per frame)
number of slice views = 3, fps = 9.06071 (110.367 ms per frame)
number of slice views = 3, fps = 11.2024 (89.2667 ms per frame)

lassoan:
Integrated in r25591. I think I've addressed all the comments (tested performance, added test, documentation, fixed style issues), but let me know if I should add/change something.

jcfr:
> I think implementing generic array indirection like this was a mistake, because they slowed down and made more complex the entire toolkit just to serve a narrow use case. In release mode the performance degradation is just about 10-15%

@lassoan Thanks for pointing this out. Would you have more details this and/or benchmark to help understand the problem ?

lassoan:
Performance of vtkGenericDataArray is impacted, which means that essentially all VTK filters are slowed down. I've just picked vtkWindowedSincPolyDataFilter and measured computation time on the same inputs with an older VTK7 and latest VTK7 in Slicer and found that computation time is increased by about 10%.

The problem is that generic data arrays added further indirection to accessing data in VTK objects. See these call stacks for very basic operations (that are performed tens of thousands of times in every filter):

vtkPoints::GetPoint:

```
vtkCommon-7.1.dll!vtkAOSDataArrayTemplate<float>::GetTypedComponent(__int64 tupleIdx, int comp)  Line 93 + 0x11 bytes	C++
vtkCommon-7.1.dll!vtkGenericDataArray<vtkAOSDataArrayTemplate<float>,float>::GetTypedComponent(__int64 tupleIdx, int compIdx)  Line 144	C++
vtkCommon-7.1.dll!vtkGenericDataArray<vtkAOSDataArrayTemplate<float>,float>::GetTuple(__int64 tupleIdx, double * tuple)  Line 40 + 0x14 bytes	C++
vtkCommon-7.1.dll!vtkGenericDataArray<vtkAOSDataArrayTemplate<float>,float>::GetTuple(__int64 tupleIdx)  Line 30	C++
vtkCommon-7.1.dll!vtkPoints::GetPoint(__int64 id)  Line 117 + 0x43 bytes	C++
```

vtkFloatArray::SetTuple():

```
vtkCommon-7.1.dll!vtkAOSDataArrayTemplate<float>::SetValue(__int64 valueIdx, float value)  Line 68	C++
vtkCommon-7.1.dll!vtkAOSDataArrayTemplate<float>::SetTypedComponent(__int64 tupleIdx, int comp, float value)  Line 102	C++
vtkCommon-7.1.dll!vtkGenericDataArray<vtkAOSDataArrayTemplate<float>,float>::SetTypedComponent(__int64 tupleIdx, int compIdx, float value)  Line 154	C++
vtkCommon-7.1.dll!vtkGenericDataArray<vtkAOSDataArrayTemplate<float>,float>::SetComponent(__int64 tupleIdx, int compIdx, double value)  Line 160	C++
vtkCommon-7.1.dll!vtkDataArray::SetTuple(__int64 i, const double * source)  Line 511	C++
vtkCommon-7.1.dll!vtkGenericDataArray<vtkAOSDataArrayTemplate<float>,float>::SetTuple(__int64 tupleIdx, const double * tuple)  Line 227 + 0x3c bytes	C++

```

In debug mode the extra layer takes significant amount of time. For example, in a surface smoothing in debug mode 45% of the time is spent in vtkPoints::GetPoint and 21% of the time is spent in vtkPoints::SetPoint().

jcfr:
To follow up on this, after few offline exchanges with @lassoan and the help of the VTK core team. the issue has been addressed and performances have been restored for both Debug and Release build.

As soon the version of VTK in Slicer is updated, we should benefit from the improvements.

For future reference:
* https://gitlab.kitware.com/vtk/vtk/merge_requests/2254
* https://gitlab.kitware.com/vtk/vtk/merge_requests/2267#note_209304

PR #635, opened 2016-12-08
Fix segmentation warnings

jcfr:


jcfr:
@cpinter Here are few more minor fixes. A sanity check would be great.

I also wonder if some code should have really been implemented in the segmentation node (see 5fbaa76 )

lassoan:
OnSegmentAdded/Removed/Modified can be removed. We might have used it earlier but they are not needed anymore (and can be added back easily if needed).

PR #636, opened 2016-12-09
ENH: Implemented DICOMDetailsPopup as window and dialog

che85:
- default: DICOMDetailsDialog which is bound to the mainWindow and slides in from top
  - closing with esc (which could be improved)
- when using DICOMDetailsWindow, it always says on top
  - resize/move event are overridden called and saves the dimensions to the settings for next call
- old ones like DICOMDetailsPopup and DICOMDetailsDock should be removed since they don't improve any usability

image

### DIFF
--- a/Modules/Scripted/DICOM/DICOM.py
+++ b/Modules/Scripted/DICOM/DICOM.py
@@ -158,6 +158,8 @@ class DICOMWidget:
   Slicer module that creates the Qt GUI for interacting with DICOM
   """
 
+  detailWidgetClasses = [DICOMLib.DICOMDetailsWindow, DICOMLib.DICOMDetailsDialog, DICOMLib.DICOMDetailsDock]
+
   def __init__(self, parent=None):
     self.testingServer = None
 
@@ -169,11 +171,6 @@ class DICOMWidget:
     self.updateRecentActivityTimer.singleShot = True
     self.updateRecentActivityTimer.interval = 500
     self.updateRecentActivityTimer.connect('timeout()', self.onUpateRecentActivityRequestTimeout)
-
-    self.widgetTypes = {'window': DICOMLib.DICOMDetailsWindow,
-                        'dialog': DICOMLib.DICOMDetailsDialog,
-                        'modal_dialog': DICOMLib.DICOMDetailsModalDialog,
-                        'dock': DICOMLib.DICOMDetailsDock}
 
     if not parent:
       self.parent = slicer.qMRMLWidget()
@@ -290,19 +287,30 @@ class DICOMWidget:
 
   def getSavedDICOMDetailsWidgetType(self, default="window"):
     widgetType = settingsValue('DICOM/BrowserWidgetType', default)
-    if widgetType in self.widgetTypes.keys():
-      return self.widgetTypes[widgetType]
-    raise ValueError("Widget type %s for DICOMDetails does not exist" %widgetType)
+    widgetClass = self.getDetailsWidgetClassForType(widgetType)
+    if not widgetClass:
+      qt.QSettings().setValue('DICOM/BrowserWidgetType', widgetType)
+      widgetClass = self.getDetailsWidgetClassForType(default)
+    return widgetClass
 
   def setDICOMDetailsWidgetType(self, widgetType):
-    if widgetType in self.widgetTypes.keys():
-      qt.QSettings().setValue('DICOM/BrowserWidgetType', widgetType)
-      self.detailsPopup = self.widgetTypes[widgetType]()
+    if not widgetType in self.getAvailableWidgetTypes():
+      raise ValueError("Widget type '%s' for DICOMDetails does not exist" % widgetType)
     else:
-      raise KeyError("Widget type %s for DICOMDetails does not exist" % widgetType)
+      qt.QSettings().setValue('DICOM/BrowserWidgetType', widgetType)
+      if self.detailsPopup:
+        if self.detailsPopup.isVisible():
+          self.detailsPopup.close()
+        self.detailsPopup = self.getDetailsWidgetClassForType(widgetType)()
 
   def getAvailableWidgetTypes(self):
-    return self.widgetTypes.keys()
+    return [c.widgetType for c in self.detailWidgetClasses]
+
+  def getDetailsWidgetClassForType(self, widgetType):
+    try:
+      return self.detailWidgetClasses[self.getAvailableWidgetTypes().index(widgetType)]
+    except (KeyError, ValueError):
+      return None
 
   def onDatabaseChanged(self):
     """Use this because to update the view in response to things
--- a/Modules/Scripted/DICOM/DICOM.py
+++ b/Modules/Scripted/DICOM/DICOM.py
@@ -170,6 +170,11 @@ class DICOMWidget:
     self.updateRecentActivityTimer.interval = 500
     self.updateRecentActivityTimer.connect('timeout()', self.onUpateRecentActivityRequestTimeout)
 
+    self.widgetTypes = {'window': DICOMLib.DICOMDetailsWindow,
+                        'dialog': DICOMLib.DICOMDetailsDialog,
+                        'modal_dialog': DICOMLib.DICOMDetailsModalDialog,
+                        'dock': DICOMLib.DICOMDetailsDock}
+
     if not parent:
       self.parent = slicer.qMRMLWidget()
       self.parent.setLayout(qt.QVBoxLayout())
@@ -234,7 +239,7 @@ class DICOMWidget:
 
     self.runListenerAtStart = qt.QCheckBox("Start Listener when Slicer Starts")
     self.localFrame.layout().addWidget(self.runListenerAtStart)
-    self.runListenerAtStart.checked =  settingsValue('DICOM/RunListenerAtStart', False, converter=toBool)
+    self.runListenerAtStart.checked = settingsValue('DICOM/RunListenerAtStart', False, converter=toBool)
     self.runListenerAtStart.connect('clicked()', self.onRunListenerAtStart)
 
     # the Database frame (home of the ctkDICOM widget)
@@ -243,7 +248,7 @@ class DICOMWidget:
     self.dicomFrame.setText("DICOM Database and Networking")
     self.layout.addWidget(self.dicomFrame)
 
-    self.detailsPopup = DICOMLib.DICOMDetailsWindow()
+    self.detailsPopup = self.getSavedDICOMDetailsWidgetType()()
 
     # XXX Slicer 4.5 - Remove these. Here only for backward compatibility.
     self.dicomBrowser = self.detailsPopup.dicomBrowser
@@ -282,6 +287,22 @@ class DICOMWidget:
 
     # Add spacer to layout
     self.layout.addStretch(1)
+
+  def getSavedDICOMDetailsWidgetType(self, default="window"):
+    widgetType = settingsValue('DICOM/BrowserWidgetType', default)
+    if widgetType in self.widgetTypes.keys():
+      return self.widgetTypes[widgetType]
+    raise ValueError("Widget type %s for DICOMDetails does not exist" %widgetType)
+
+  def setDICOMDetailsWidgetType(self, widgetType):
+    if widgetType in self.widgetTypes.keys():
+      qt.QSettings().setValue('DICOM/BrowserWidgetType', widgetType)
+      self.detailsPopup = self.widgetTypes[widgetType]()
+    else:
+      raise KeyError("Widget type %s for DICOMDetails does not exist" % widgetType)
+
+  def getAvailableWidgetTypes(self):
+    return self.widgetTypes.keys()
 
   def onDatabaseChanged(self):
     """Use this because to update the view in response to things
--- a/Modules/Scripted/DICOM/DICOM.py
+++ b/Modules/Scripted/DICOM/DICOM.py
@@ -243,7 +243,7 @@ class DICOMWidget:
     self.dicomFrame.setText("DICOM Database and Networking")
     self.layout.addWidget(self.dicomFrame)
 
-    self.detailsPopup = DICOMLib.DICOMDetailsPopup()
+    self.detailsPopup = DICOMLib.DICOMDetailsWindow()
 
     # XXX Slicer 4.5 - Remove these. Here only for backward compatibility.
     self.dicomBrowser = self.detailsPopup.dicomBrowser
--- a/Modules/Scripted/DICOM/DICOM.py
+++ b/Modules/Scripted/DICOM/DICOM.py
@@ -160,6 +160,33 @@ class DICOMWidget:
 
   detailWidgetClasses = [DICOMLib.DICOMDetailsWindow, DICOMLib.DICOMDetailsDialog, DICOMLib.DICOMDetailsDock]
 
+  @staticmethod
+  def getSavedDICOMDetailsWidgetType(default="window"):
+    widgetType = settingsValue('DICOM/BrowserWidgetType', default)
+    widgetClass = DICOMWidget.getDetailsWidgetClassForType(widgetType)
+    if not widgetClass:
+      qt.QSettings().setValue('DICOM/BrowserWidgetType', widgetType)
+      widgetClass = DICOMWidget.getDetailsWidgetClassForType(default)
+    return widgetClass
+
+  @staticmethod
+  def setDICOMDetailsWidgetType(widgetType):
+    if not widgetType in DICOMWidget.getAvailableWidgetTypes():
+      raise ValueError("Widget type '%s' for DICOMDetails does not exist" % widgetType)
+    else:
+      qt.QSettings().setValue('DICOM/BrowserWidgetType', widgetType)
+
+  @staticmethod
+  def getAvailableWidgetTypes():
+    return [c.widgetType for c in DICOMWidget.detailWidgetClasses]
+
+  @staticmethod
+  def getDetailsWidgetClassForType(widgetType):
+    try:
+      return DICOMWidget.detailWidgetClasses[DICOMWidget.getAvailableWidgetTypes().index(widgetType)]
+    except (KeyError, ValueError):
+      return None
+
   def __init__(self, parent=None):
     self.testingServer = None
 
@@ -186,7 +213,7 @@ class DICOMWidget:
     globals()['d'] = self
 
   def enter(self):
-    self.detailsPopup.open()
+    self.onOpenDetailsPopup()
 
   def exit(self):
     if not self.detailsPopup.browserPersistent:
@@ -254,14 +281,14 @@ class DICOMWidget:
     # connect to the 'Show DICOM Browser' button
     self.showBrowserButton = qt.QPushButton('Show DICOM Browser')
     self.dicomFrame.layout().addWidget(self.showBrowserButton)
-    self.showBrowserButton.connect('clicked()', self.detailsPopup.open)
+    self.showBrowserButton.connect('clicked()', self.onOpenDetailsPopup)
 
     # connect to the main window's dicom button
     mw = slicer.util.mainWindow()
     if mw:
       try:
         action = slicer.util.findChildren(mw,name='LoadDICOMAction')[0]
-        action.connect('triggered()',self.detailsPopup.open)
+        action.connect('triggered()',self.onOpenDetailsPopup)
       except IndexError:
         logging.error('Could not connect to the main window DICOM button')
 
@@ -285,32 +312,12 @@ class DICOMWidget:
     # Add spacer to layout
     self.layout.addStretch(1)
 
-  def getSavedDICOMDetailsWidgetType(self, default="window"):
-    widgetType = settingsValue('DICOM/BrowserWidgetType', default)
-    widgetClass = self.getDetailsWidgetClassForType(widgetType)
-    if not widgetClass:
-      qt.QSettings().setValue('DICOM/BrowserWidgetType', widgetType)
-      widgetClass = self.getDetailsWidgetClassForType(default)
-    return widgetClass
-
-  def setDICOMDetailsWidgetType(self, widgetType):
-    if not widgetType in self.getAvailableWidgetTypes():
-      raise ValueError("Widget type '%s' for DICOMDetails does not exist" % widgetType)
-    else:
-      qt.QSettings().setValue('DICOM/BrowserWidgetType', widgetType)
-      if self.detailsPopup:
-        if self.detailsPopup.isVisible():
-          self.detailsPopup.close()
-        self.detailsPopup = self.getDetailsWidgetClassForType(widgetType)()
-
-  def getAvailableWidgetTypes(self):
-    return [c.widgetType for c in self.detailWidgetClasses]
-
-  def getDetailsWidgetClassForType(self, widgetType):
-    try:
-      return self.detailWidgetClasses[self.getAvailableWidgetTypes().index(widgetType)]
-    except (KeyError, ValueError):
-      return None
+  def onOpenDetailsPopup(self):
+    print "onOpenDetailsPopup"
+    if not isinstance(self.detailsPopup, self.getSavedDICOMDetailsWidgetType()):
+      print "is not instance"
+      self.detailsPopup = self.getSavedDICOMDetailsWidgetType()()
+    self.detailsPopup.open()
 
   def onDatabaseChanged(self):
     """Use this because to update the view in response to things
--- a/Modules/Scripted/DICOM/DICOM.py
+++ b/Modules/Scripted/DICOM/DICOM.py
@@ -313,9 +313,7 @@ class DICOMWidget:
     self.layout.addStretch(1)
 
   def onOpenDetailsPopup(self):
-    print "onOpenDetailsPopup"
     if not isinstance(self.detailsPopup, self.getSavedDICOMDetailsWidgetType()):
-      print "is not instance"
       self.detailsPopup = self.getSavedDICOMDetailsWidgetType()()
     self.detailsPopup.open()
 
--- a/Modules/Scripted/DICOMLib/DICOMWidgets.py
+++ b/Modules/Scripted/DICOMLib/DICOMWidgets.py
@@ -918,9 +918,6 @@ class DICOMDetailsPopup(DICOMDetailsBase, ctkPopupWidget):
 
 
 class DICOMDetailsDock(DICOMDetailsBase, qt.QFrame):
-  """
-  TODO: this is probably not usable since the dicom browser takes too much space so it wouldn't make sense to dock it
-  """
 
   def __init__(self, dicomBrowser=None):
     DICOMDetailsBase.__init__(self, dicomBrowser)
@@ -931,7 +928,22 @@ class DICOMDetailsDock(DICOMDetailsBase, qt.QFrame):
                           qt.QDockWidget.DockWidgetClosable)
     slicer.util.mainWindow().addDockWidget(qt.Qt.TopDockWidgetArea, self.dock)
     self.dock.setWidget(self)
+    self.dock.visibilityChanged.connect(self.onVisibilityChanged)
     self.setup()
+
+  def __del__(self):
+    self.dock.visibilityChanged.disconnect(self.onVisibilityChanged)
+
+  def open(self):
+    if not self.dock.visible:
+      self.dock.show()
+
+  def close(self):
+    self.dock.hide()
+
+  def onVisibilityChanged(self, visible):
+    if not visible:
+      self.close()
 
 
 class DICOMPluginSelector(qt.QWidget):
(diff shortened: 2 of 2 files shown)


fedorov:
I personally think this looks very useful!

pieper:
Does persistent mode still work?  Can you drag the browser to a second
monitor?  These were two of the requested features that led to the existing
design.

fedorov:
@pieper I don't know if those things still work, but as a comment, I have not seen radiology reading workstations that would optimize DICOM interface UI for utilization of second monitor (I also admit I probably didn't see too many). I am curious what is the proportion of Slicer users who find separate window for DICOM browser feature useful. But of course we will never know the answer to that question.

@che85 for the sake of efficiency, we should probably just proceed optimizing DICOM workflow in our extensions as we discussed earlier. Features of the main application are optimized to accommodate a broad range of usage scenarios, and we really don't have much data to back choices made for one reason or another. Fortunately, we have sufficient control to optimize individual modules to support our target users, and implement design choices as we see fit, and refine based on the direct feedback from the users we interact with.

che85:
@pieper We should probably use the window instead of the dialog. The mayor reason, why I started changing code again in DICOMWidgets was, that the size/position of the DICOMDetailsPopup was never saved in the settings even when an implementation existed. I will change it back to the window.

Do we at least agree about that the DICOMDetailsPopup should always be in the front? That's another thing I implemented. 

@fedorov I noticed those missing features while implementing our extension. I only wanted that code to go here, because it belongs to the main application

pieper:
As I recall @lassoan had use cases for the second monitor persistent mode and may want to comment.

It seems fine to me that if persistent mode is off then having the popup window act like a conventional dialog box (always on top) would be fine.

I'm not sure why the size and location aren't persisted - as I recall they did originally.

cpinter:
I built and tried on windows, but I didn't seem to notice any change in the DICOM window with my usual Slicer.ini. I tried deleting the ini file to see whether resetting to default will bring out the new feature, but then I got this error when I clicked on the DICOM button:
  File "C:/e/ScD/Slicer-build/lib/Slicer-4.7/qt-scripted-modules/DICOM.py", line 246, in setup
    self.detailsPopup = DICOMLib.DICOMDetailsWindow()
  File "C:/e/ScD/Slicer-build/lib/Slicer-4.7/qt-scripted-modules\DICOMLib\DICOMWidgets.py", line 853, in __init__
    DICOMDetailsBase.__init__(self, dicomBrowser)
  File "C:/e/ScD/Slicer-build/lib/Slicer-4.7/qt-scripted-modules\DICOMLib\DICOMWidgets.py", line 57, in __init__
    self.promptForDatabaseDirectory()
  File "C:/e/ScD/Slicer-build/lib/Slicer-4.7/qt-scripted-modules\DICOMLib\DICOMWidgets.py", line 484, in promptForDatabaseDirectory
    self.onDatabaseDirectoryChanged(databaseDirectory)
  File "C:/e/ScD/Slicer-build/lib/Slicer-4.7/qt-scripted-modules\DICOMLib\DICOMWidgets.py", line 338, in onDatabaseDirectoryChanged
    setDatabasePrecacheTags(self.dicomBrowser)
AttributeError: QWidget has no attribute named 'dicomBrowser'
  File "C:/e/ScD/Slicer-build/lib/Slicer-4.7/qt-scripted-modules/DICOM.py", line 187, in enter
    self.detailsPopup.open()
AttributeError: DICOMWidget instance has no attribute 'detailsPopup'

Also a few comments:
* If "sliding" means that the window animates, then some people possibly won't like it, including myself. Animated windows tend to hijack focus and take time to appear which is inefficient
* I see offsets with fixed pixel values in the code (e.g. DICOMDetailsWindow::open). We should phase these out rather than introducing them, because with today's low- and high-resolution (retina etc.) displays these only cause problems. You implement something that looks good on yours, and it will look bad on the other. I had the same problems with Segmentations module recently.
* Is there a human-readable equivalent to this line: self.alignment = 0x82 ? Such as qt.Qt.FlagNameXY. It reduces readability of the code. If there is none, then at least add comment

cpinter:
I just see that the last comment is not related to new, but only moved code. If you could address that it would still be nice. Thanks!

che85:
@pieper I changed the implementation back to having the window which can still be persistent. The only things that changed are:
- window is always on top (I experienced pretty often, that it disappeared behind the slicer main window)
- resize/move events are used for saving the geometry so that the geometry can be restored when opening the browser again

@cpinter Thanks for the comments. I fixed the bug, that you had on Windows and I also removed the fixed pixel values (which I came from the code, that I moved) 

So no more animation, when opening the browser. But at least, we can use that functionality in slicelets or other cases, which in my opinion is pretty useful.

lassoan:
We used to use the DICOM browser as an interactive way of loading data that we receive intra-operatively from MRI while keep showing the main Slicer screen to clinicians. However, in recent years we haven't had a project with such requirements, and similar project, like Christian's prostate tracker module implemented a good alternative approach of implementing a mini DICOM receiver and browser built into the module GUI.

What I see is that users don't understand the current DICOM browser and they find it very difficult to use. Just a couple of common issues:
1. They would expect to load DICOM files as all other file types. They don't get what is the difference between importing and loading. Often they don't want to add the loaded files permanently to the Slicer DICOM database. Once import is complete it's really difficult to find the data that have just been imported.
2. It's very very slow. It's most annoying for importing, when you don't have a chance to limit your attention to a particular series. Importing of a study may take tens of minutes.
3. Very basic information, such as number of images are missing! This makes it very difficult to find what you are looking for.
4. Error/warning messages are not helpful (Imported 0 patients, 0 studies, 0 series --- why? - sometimes simply because you've already imported that data before).

Potential solutions:
1. Simplify the workflow. Hide the import/loading steps from users and show it as a simple loading operation with maybe one popup window where the user can deselect data that he is not interested in and how to interpret the data - similarly how it is done in the "Add data" dialog. Could we just populate the Add data dialog from data that we collect from DICOM plugins?
2. We have to seriously use DICOMDIR information. Parsing through all the files in a directory structure is not an option. It also means that DICOM plugins should give an approximate confidence value based on DICOM tags available in DICOMDIR.
3. Direct mapping between the Slicer database and DICOM tags is a flawed concept. It's OK to have a tag cache that stores verbatim DICOM information (we have this already). But patient/study/series table should contain derived data, such as nicely formatted patient name, series description that is constructed from various fields, aggregate data, such as number of series in a study or number of frames in a series. This would also allow us to store non-DICOM data in the Slicer database (that we can convert to DICOM format on export). We could have a single-click save button that saves the scene in the database, much easier browsing of data (you wouldn't need to resort to finding scene and data files in a directory structure but you could just easily find your data in the Slicer database).
4. Simplify architecture. Having a generic widget in CTK and trying to customize it in Slicer hasn't really worked well. It greatly limits us that we have to consider many other unknown potential uses of the DICOM database and widget. I think we should prototype a completely new DICOM widget and workflow, independent from the CTK DICOM implementation (it's a good starting point, so we could fork that and may merge back once we have a solid working concept).

pieper:
Obviously this all very related to the discussion in
https://github.com/Slicer/Slicer/pull/504

It's good to see the momentum to take a fresh look at all this, keeping
some good parts but changing out the painful parts.

Adding to Andras's wish list:

* do the 'interpretation' step in the background / in parallel whenever
possible so that we already have answers by the time the user tries to
load.  Cache these so they only need to happen once.

* figure out the most common use cases (e.g. load scalar volume) and factor
them out into an optimized paths.  The logic for mapping dicom instances
into volumes is spread out across far too many code bases (in the end much
of the work is still being done by gdcm).

* unify the concepts of query and interpretations of metadata, but separate
from storage.  DICOMweb has metadata queries that can provide DICOMDIR-like
information so I would like to consider adopting that model to separate the
data storage from the application level processing.

* add thumbnails and other ways of previewing the data before loading!

* keep an eye on other developments in the field and try to leverage other
people's work.  As you guys know I'm working on similar issues in the
javascript domain and I'd prefer to see a unified solution that works for
both slicer and browser based applications.  I honestly find the javascript
code cleaner and more powerful, so I'd like it if we could focus our
efforts there and make it compatible with the needs of the Slicer core.

lassoan:
@pieper Would the javascript component be just a frontend that allows user to browse database content? Does DICOMweb have a backend implementation for metadata parsing and storage - accessible through web browsers and multiple Slicer instances (preferably free, open-source, easy-to-install - so that we could bundle with Slicer packages)?

cpinter:
@che85 Thanks for the fix! I updated, and I don't get the error message any more. However, I cannot seem to invoke the new kind of docked behavior. Is there some setting I need to enable?

Also unfortunately there is a problem with always on top. When I clicked Import, the dialog appeared behind the browser, and I couldn't move or close the browser, because the path selector dialog that was modal, prevented me from doing that.

lassoan:
It's extremely difficult to enforce top-level window ordering reliably on all operating systems, windowing systems, and use cases (application switch with mouse, keyboard shortcuts; window maximize/minimize; drag-and-drop).

What do you think about putting the DICOM browser into the view layout? By default it could be a 1-up layout (only the DICOM browser is shown) but you could easily define layouts that have the browser and other viewers. This way we could utilize the entire window area without worrying about window z ordering.

pieper:
@lassoan yes, there are already (at least) two DICOMweb implementations that are close to the criteria you mentioned [1,2].  But also the rest services are pretty straightforward [3] and if we follow their structure when implementing our API we can be pretty flexible about the backend, either connecting to a real server or to a local cache.  

You should try the study browser/search in LesionTracker (and the annotation tools) and see if it doesn't already have a lot of the features we would want (or at least be something we should build on and contribute back to).  OHIF uses orthanc (via DICOMweb) as a back-end.  We know the OHIF developers well and they are very eager to work with Slicer's 3D capabilities.  I have been in discussions with them already about how they would like to be able to use the functionality of the Slicer DICOMPlugins for the same basic purposes we use them for so to me it's natural that we evolve Slicer's DICOM interface in this direction.

By the way, I just checked an LesionTracker's study browser and the cornerstone viewer/annotation tools actually already work in Slicer 4.6!  I had thought we would need to wait until we have Qt5 and the newer webkit (I'm sure that will still help make us even more compatible).  I think it would be very easy to add hooks to move data back-and-forth between the LesionTracker code and Slicer.

Easy to try with the code below - make yourself an account (free and easy) and give it a try!

>>> v = qt.QWebView()
>>> v.show()
>>> v.setUrl(qt.QUrl('http://lesiontracker.ohif.org/'))

image

image

[1] http://orthanc-server.com/static.php?page=dicomweb is AGPL, which means we wouldn't distribute it with Slicer proper, but since it's an independent program that we would connect to via http the license wouldn't be viral.

[2] https://github.com/dcm4che/dcm4che/blob/master/README.md 

[3] https://hcintegrations.ca/2015/09/11/dicomweb-cheatsheet/

[4] http://lesiontracker.ohif.org/

cpinter:
@che85 @fedorov Can you please answer my question about how to invoke the new dockwidget functionality? It doesn't seem to work at all on Windows. Please don't integrate until this is cleared up.

lassoan:
It would be also great if you could upload a few screenshots of what we are expected to see (the screenshot at the top doesn't show the docking mode).

che85:
@cpinter I added a new commit for making the docking mode usable. 

If you want to try it, just replace line https://github.com/Slicer/Slicer/blob/master/Modules/Scripted/DICOM/DICOM.py#L246

with 

```
self.detailsPopup = DICOMLib.DICOMDetailsDock()
```

P.S. I removed the flag for setting the window always on top so that we don't need to take care of the window z ordering for now

cpinter:
Thanks, Christian, for the prompt fixes! I'll try it tonight.

I'm btw OK with always on top, except for this obscuring issue I mentioned. I wish there was a property (that works on all platforms) for prioritizing modal windows. But in general I think it will cause less trouble if the window is not always-on-top for now.

che85:
Yeah I agree. I didn't know that this would cause so many problems.

fedorov:
@che85 as we discussed last week, why not just make the DICOM Browser part of the main window layout? Why does it have to be a popup at all?

che85:
@fedorov I don't know what's the opinion of the others about that

- 1st click on dcm toolbar icon
  - hide all slice views
  - set the module widget to full width
  - display DICOM browser in that module widget
- 2nd click on dcm toolbar icon restore the old view

But when clicking on the dcm toolbar icon, the DICOM module is opened which also has its own UI regarding the server/recent activity, so where would that go?

@pieper, @cpinter, @lassoan any thoughts?

cpinter:
My first thoughts:
* Adding meaningful things to the DICOM module panel makes sense; so far it didn't contain anything useful to me, so when I closed the DICOM browser there was a module panel that had no meaning
* I don't like the idea of hiding the views, because 1. it would be an orthogonal mechanic to the view layouts, and I'm pretty sure it will cause issues, 2. updating the views when modifying the layout could take seconds if you have complex models etc. loaded, 3. Setting full width to the module panel is unprecedented (as far as I know), and if the user does something unexpected (they always do) instead of clicking the button the second time, then it might be hard to restore
* I have a hunch that the user won't think to click on the DCM button the second time, it doesn't seem intuitive

So to me having the DICOM browser in the actual view layout seems the best choice, although it is alluring to put it in the module panel.

che85:
@cpinter The 2nd click was my own idea. 

You mean having the DICOM browser in the actual view layout (like a dock)? My impression is, that the DICOM browser itself is so complex (space consuming) that user could get annoyed by that. One good thing about it is that the user can undock it.

cpinter:
What I mean is what was suggested before: having one or more new layouts in which the DICOM browser appears. Same as we have with tables and the quantitative layouts with the graphs.

fedorov:
>  full width to the module panel is unprecedented

That's an interesting argument. I think for users who start using Slicer coming from more clinically-oriented tools, most of the things in Slicer are unprecedented. Giving them full width for DICOM browser would actually be more consistent with how other tools provide this functionality.

But I am not pushing for changing the behavior of Slicer DICOM Browser module. We should experiment with this kind of things in our customized module.

cpinter:
The "unprecedented" argument in my interpretation means that it will be hard to assess the implications of it. Slicer is a very complex sandbox-like application, which, as opposed to most clinical software, has a completely open workflow meaning the user can do any steps in almost order, so keeping the behaviors consistent and simple is vital so that the application state is not messed up at any moment. If we introduce one more UI mechanic, then this becomes increasingly harder.

cpinter:
@lassoan and @fedorov suggested the same what I did at the end of my comment, which involves substantially changing the behavior of the DICOM browser. The advantage of that is that we use an existing paradigm instead of introducing a new one. The only downside as I can see is that the layout list grows a bit longer.

fedorov:
Csaba, I agree with your argument for the "generic use case" of Slicer.

However, I find that Slicer users who are most used to clinical tools do not really appreciate the "sandbox" and "open workflow" concepts. Those users need a locked up single-module solution or a slicelet that removes all bells and whistles that can cause problems or create confusion, or not utilize screen space effectively. 

As long as generic Slicer components allow to hide unnecessary items and reconfigure things to optimize implementation of that single-module solution (as you do, for example, in Segment Editor), we can reuse Slicer components. If such reuse is not possible, we will need to "hack around", re-implement things, etc, as we also do already for some other Slicer features. As you know, integrating things into Slicer core brings a lot of delays, so there is always a choice between doing things, and trying to do things the right way.

cpinter:
I don't see how this argument favors the 'make a module panel whole window' versus 'add the browser in the layout'. Making Slicer itself harder to use does not make the slicelets better.

fedorov:
I agree it is confusing. That is why I advocate pilots in separate modules that provide a good test and demonstration of improved features before suggesting changes into the core application.

che85:
I am implementing that in Quantitative Reporting, but it depends on this PR.

fedorov:
> I don't see how this argument favors the 'make a module panel whole window' versus 'add the browser in the layout'

@cpinter to explain that point, the issue with the DICOM Browser module panel is that it provides no value, introduces confusion, and steals screen space.

fedorov:
@cpinter getting away from controversies and back to this specific PR - please let us know if you have any other concerns merging it after testing the updated code.

lassoan:
There are several options that may work better than the current method:
A. Show DICOM module widget on the left + patient browser in the layout
B. Show DICOM module widget on the left + hide central widget (slice+3D viewers)
C. Show DICOM browser in a floating dockable widget (like Python interactor)
D. Show DICOM browser in a modal popup (like Add data)

A. Layout switch takes time. We would need to remember what was the previous layout and switch back when done with DICOM loading (not very clean; what happens if the user changes the layout while in the DICOM module).

B. Layout redraw takes time. So far none of the modules manipulated visibility of the central widget, but if modules start doing that then it may become messy (e.g., when you switch between two modules that hide the central widget, switch would take a long time because there would be several view redraws).
Note that you can easily try this:
slicer.util.mainWindow().centralWidget().hide()
slicer.util.mainWindow().centralWidget().hide()

C. Automatic placement (overlaid on central widget + if the module GUI is docked then probably over that area, too) might be non-trivial. It's always above docket widgets but the user may raise other floating docking widgets above, so emulating modal popup behavior might not be very easy.

D. The only limitation I see is that it's not possible to keep the DICOM browser open while working with other modules, but I would not mind this too much.

Based on all these, I think option D is the simplest and most robust. It could be opened similarly to the Add data dialog (from application main menu, toolbar, etc.) and not through the module selector.

fedorov:
@lassoan if I understand correctly, the proposed PR will enable both C and D options from your list.

cpinter:
Note that modal (option D) is different from always-on-top, because in that case the user would not be able to click on anything that pops up behind the browser (like closing Slicer).

Also how about not switching to the DICOM module when clicking the DCM button? Then the module panel with limited meaning would not show up at all. It could be shown by switching to the DICOM module similarly to all other modules.

fedorov:
> Note that modal (option D) is different from always-on-top, because in that case the user would not be able to click on anything that pops up behind the browser (like closing Slicer).

Christian, would it be possible to add the modal mode? 

Again, for us most important is what we will be able to do to configure DICOM Browser widget for the purposes of our module/slicelet. I am ambivalent about the choices for the default behavior of the DICOM Browser module, as it is unlikely we will use it without customizations. So please choose any defaults as you see suitable (e.g., modal vs always-on-top).

> Also how about not switching to the DICOM module when clicking the DCM button? Then the module panel with limited meaning would not show up at all. It could be shown by switching to the DICOM module similarly to all other modules.

That is fine with me.

Aren't we all in agreement that the proposed PR is an improvement over what is done in Slicer right now? We can always improve further.

cpinter:
I can try it most easily on my home computer (because the branch is cloned and built there already - hence I said evening), but feel free to integrate if it's urgent.

fedorov:
Csaba, it is not urgent. I just want to make sure there is a clear path forward with this PR, and we don't lose the intent of this PR in the discussions of related ideas.

cpinter:
As it did not work on Windows (what most users use), and it is not yet confirmed that is does now, I think it's safest to integrate after it's tested. If someone else is willing to test on Windows before I can do that myself, I have no objections.

fedorov:
No rush Csaba, we definitely appreciate your testing on Windows, and we want to make sure you and @lassoan do not object integration. This can wait until you test and approve.

che85:
@fedorov 
> Christian, would it be possible to add the modal mode?

Yes I can add this functionality until tonight so that @cpinter can test everything on Windows

che85:
Ok that's what we have now:

1. DICOMDetailsDialog (which is bound to the mainWindow and slides in from top)
  i. can be persistent and if not can be closed via esc key
  ii. metadata, import and SendDialog are displayed on top of it which also block the mainWindow and browser once opened
2. DICOMDetailsModalDialog
  i. stays always on top blocking the mainWindow
  ii. see 1. ii.
  iii. on resize/move extend and position will be saved for restore
3. DICOMDetailsWindow (non modal dialog which doesn't stay on top
  i. doesn't stay on top and doesn't block mainWindow
  ii. see 1.ii.
4. DICOMDetailsDock
  i. is un/dockable and can be positioned as preferred
  ii. see 1.ii

DICOMHeaderPopup alias metadata information dialog is now a qDialog for ensuring that it never disappears in the background

What we use as the default is your decision. I tend to using the DICOMDetailsDock or DICOMDetailsModalDialog
  
@cpinter let me know how the testing goes.

Other than that I could also imagine to have some setting in the Application Settings of Slicer so that users can customize their preferred DICOMDetails(Window/ModalDialog/Dialog/Dock)

che85:
P.S. the window z ordering should be fixed with the last commit

cpinter:
Sounds good! I like the idea of an application setting, potentially storing the last used mode so that it appears the same way the next time it's opened.

How does one switch between the modes?

I'll test it asap and let you know!

che85:
For now you would have to replace DICOMDetailsPopup[1] with your preferred DICOMDetails... class

[1] https://github.com/Slicer/Slicer/blob/master/Modules/Scripted/DICOM/DICOM.py#L246

cpinter:
Probably this is why I couldn't try the new features. Sorry I missed this earlier!

fedorov:
@che85 would you mind adding an API call to configure the preferred choice of DICOMDetails... class? Definitely exposing that in a setting option is a good one.

che85:
Yes I can do that

lassoan:
I've tested the new DICOM widget types.

* DICOMDetailsDialog and DICOMDetailsModalDialog: With every open/close, the window keeps appearing a bit higher than previously; after a few iterations the 'X' button is out of screen; it's tricky to bring the window back to normal.
* I cannot find DICOMDetailsModalDialog class.
* DICOMDetailsPopup is quite annoying: appears at wrong position somewhere near the middle of the screen and it's quite small; takes a long time to disappear (about 1-2sec) and flashes before it starts the animation. I don't think we should keep this mode.

The DICOMDetailsDock window is interesting. With the current DICOM widget layout it's not very convenient to use in any docked position, but if we improve the widget's layout (and automatically optimize it to docked positions, for example switch between horizontal/vertical layout, remove some columns; make it persistent when docked) then docking could be a better option than a simple popup.

If you fix the positioning bug then it should be OK for now to merge this so that we can play with many options, but eventually we should reduce the number of classes to 1 or 2 (for example, a docking widget that the Slicer main window would use; a popup window that modules/slicelets could show). And of course as soon as possible we should add dedicated menu/toolbar button to show the DICOM browser and not waste space by showing the DICOM module GUI.

che85:
@lassoan Did you checkout the most recent commit?

lassoan:
I think it's too early to expose these options to users. We should improve things before that - for example not using the DICOM module GUI for showing the DICOM browser; improving the layout of the DICOM window to make it more usable when docked; etc.

che85:
In the most recent commit DICOMDetailsPopup is not even available anymore, because there is no point of having it anymore. It doesn't bring any new/improved functionality to the others

fedorov:
> I think it's too early to expose these options to users.

Sure, that is fine. I only offered this in case it is helpful. I am fine merging without exposing these options in the settings.

lassoan:
I was testing the version available at 2pm. I now see the extra commit since then.

che85:
@lassoan I cannot reproduce the issue that you are describing. My DICOMDetailsModalDialog and DICOMDetailsDialog always appear at the same position unless I move the modal dialog

lassoan:
Using self.detailsPopup = DICOMLib.DICOMDetailsDialog() the problem is still reproducible with the latest version:
http://screencast.com/t/zBy5agos

che85:
That doesn't make sense. DICOMDetailsDialog is no modal window. It should be displayed and bound like in the screenshot on top without and buttons for closing.

fedorov:
@che85 you should be able to test this on windows without having to build slicer, right?

che85:
Yes you only need to replace the two python files

lassoan:
Fortunately, DICOMDetailsWindow behaves correctly and that is used currently.

cpinter:
I updated to latest and tried again. Each mode seems to work well, and I don't experience the position issue Andras reported. I agree with Andras that probably vertical view would be better in case of the docking mode, but we can worry about this later.
I think you can go ahead and integrate. Thanks for working on this, and for the quick fixes!

fedorov:
Thank you @cpinter and @lassoan for testing this, and also for the productive discussion of this topic. 

I will do a test myself on Windows, and if I don't see any issues I will merge tonight.

fedorov:
I tested it on Windows 7 SP1 (the only windows I have), and I was able to reproduce the problem @lassoan describedL DICOMDetailsDialog does not show up at toolbar, and is changing its position towards the top of the screen on every open/close.

I suggested we merge, since it is only DICOMDetailsWindow what would be exposed to the user, and its functionality is already an improvement over the current one in Slicer master, but @che85 prefers to investigate this further, so we will wait to hear back from Christian.

che85:
The same bugs appeared on my Windows with a) every new call of the dialog ended up a bit higher and b) the dialog was never displayed the way it is displayed on Mac OS on top of the mainWindow which also is bound to the MainWindow

That's the reason why I removed the DICOMDetailsDialog and renamed DICOMDetailsModalDialog to DICOMDetailsDialog.

One feature that I added is: if window mode is used, and the DICOM Browser disappears behind Slicer, one can just click on the DCM icon in the toolbar and the window will be activated and raised to the front.

I also added static methods to the DICOMWidget so that one can easily decide which widget type he/she wants to see:

```
# dock
from DICOM import DICOMWidget
DICOMWidget.setDICOMDetailsWidgetType("dock")
```

```
# dialog
from DICOM import DICOMWidget
DICOMWidget.setDICOMDetailsWidgetType("dialog")
```

```
# window
from DICOM import DICOMWidget
DICOMWidget.setDICOMDetailsWidgetType("window")
```

If you want to add a dropdown for that in the Slicer settings, just add the list that you get from 
```
DICOMWidget.getAvailableWidgetTypes()
```

I am happy with the solution and I tested it on Windows and Mac.

P.S. Those are a lot of commit messages. Maybe you consider squashing them.

fedorov:
Thank you Christian!

@lassoan @cpinter I am traveling today. If you are happy with the proposed solution, would be great if you could merge, otherwise I should be able to get to it tonight, after we get ok from you.

lassoan:
I'll test this now on Windows and integrate if I don't find any problem.

lassoan:
I've tested and didn't find any issue. I'll integrate this.

The behavior is still not fully clean, as the window is only restored to the exact same state if you moved the window after it was opened. However, this is not a regression: the inconsistency is there in the current trunk version, too. You may try window geometry&state&layout save/restore methods implemented in ctkSettings. Maybe they work more predictably.

che85:
@lassoan Let me check this out. We can still merge tomorrow. When you merge now, it won't be in the nightly anyway, I assume.

lassoan:
OK, I'll wait then. Please also squash the commits or just include here a commit comment that I can use for the squashed commit.

che85:
@lassoan I tried with ctkSettings and it seems that the methods _saveState_ and  _restoreState_ are not accessible from python

```
settings = ctk.ctkSettings()

settings.value
# output: <qt slot value of ctkSettings instance at 0x14143f3b0>

settings.saveState
AttributeError: ctkSettings has no attribute named 'saveState'
```

che85:
@lassoan I implemented a mixin class, that implements the behavior for our needs and I was also able to reuse it for the DICOMHeaderPopup so that I removed some code. Now it should work fine.

For testing, you can remove the settings as follows.
```
settings = qt.QSettings()
settings.beginGroup("DICOM/HeaderPopup")
settings.remove("")
settings.endGroup()
settings.beginGroup("DICOM/SlicerDICOMBrowser")
settings.remove("")
settings.endGroup()
```

I think it should be fine now (hopefully).

fedorov:
@lassoan can this be merged?

lassoan:
It works correctly now!
If you squash the commits or send me a commit message that I should apply to the squashed commits then I merge it.

che85:
@lassoan I am just figuring out how to squash commits ... will hopefully commit soon

che85:
@lassoan Maybe you can just add the following commit message:

```
ENH: DICOM Browser implementation as window, dialog, dock

- window mode: 
  - DICOM Browser lost behind mainWindow user can just 
    click on DCM icon for bringing it to the front
- DICOMWidget: 
  - added setter/getter for widget type
- new mixin class: SizePositionSettingsMixin
  - saving/restoring dialog size and position
  - used for DICOM Browser and DICOMHeaderPopup
- taking care of z window ordering
```

lassoan:
Thanks, I'll merge this now.
I've also sent a pull request to add python wrapping of ctkSettings methods:
https://github.com/commontk/CTK/pull/683

che85:
Thanks a lot to all for the helpful discussion

lassoan:
Committed r25611

lassoan:
Thanks for your contribution. We should keep improving the DICOM GUI (remove the redundant DICOM module GUI and show the popup directly, etc.).

jcfr:
> Thanks for your contribution. We should keep improving the DICOM GUI 

Ditto. @che85 Thanks for your patience and contributions :+1: 

>  sent a pull request to add python wrapping of ctkSettings methods:

Integrated in [r25612](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25612)

pieper:
+1

fedorov:
Thanks to all for working on this!

PR #637, opened 2016-12-13
ENH: Update pydicom from 0.9.8 to 0.9.9

lassoan:
The main reason is that 0.9.8 fails to load files that have special characters in their name. The problem is fixed in 0.9.9.

0.9.9 doesn't need patching of setup.py anymore (in 0.9.8, first 3 lines of setup.py had to be removed to avoid loading some old dependencies).


pieper:
LGTM - thanks!

jcfr:
Thanks. 

Integrated in [r25607](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25607) and [r25608](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25608)

PR #638, opened 2016-12-13
ENH: Added DICOMPatcher module for fixing up DICOM files before importing

lassoan:


pieper:
👍 especially the icon!

lassoan:
Committed r25609

PR #640, opened 2016-12-19
BUG: fix runtime error related to creation of DICOM database directory

fedorov:
In some situations, user settings may point to a temporary directory with multiple
nesting levels, which cannot be created by os.mkdir().

To solve the issue, when exception occurs, fall back to creating the DICOM DB
directory in the default location.

Also fix a typo for the name of the DICOM DB location in developer mode.

cc: @che85

### DIFF
--- a/Modules/Scripted/DICOMLib/DICOMWidgets.py
+++ b/Modules/Scripted/DICOMLib/DICOMWidgets.py
@@ -491,18 +491,22 @@ class DICOMDetailsBase(VTKObservationMixin, SizePositionSettingsMixin):
         But, if the application is in testing mode, just pick a temp directory.
     """
     if slicer.app.commandOptions().testingEnabled:
-      databaseDirectory = os.path.join(slicer.app.temporaryPath, 'tempDICOMDatbase')
+      databaseDirectory = os.path.join(slicer.app.temporaryPath, 'tempDICOMDatabase')
     else:
       databaseDirectory = self.settings.value('DatabaseDirectory')
-      if not databaseDirectory:
-        documentsLocation = qt.QDesktopServices.DocumentsLocation
-        documents = qt.QDesktopServices.storageLocation(documentsLocation)
-        databaseDirectory = os.path.join(documents, "SlicerDICOMDatabase")
-        message = "DICOM Database will be stored in\n\n{}\n\nUse the Local Database button in " \
-                  "the DICOM Browser to pick a different location.".format(databaseDirectory)
-        slicer.util.infoDisplay(message, windowTitle='DICOM')
+      documentsLocation = qt.QDesktopServices.DocumentsLocation
+      documents = qt.QDesktopServices.storageLocation(documentsLocation)
+      defaultDatabaseDirectory = os.path.join(documents, "SlicerDICOMDatabase")
+      message = "DICOM Database will be stored in\n\n{}\n\nUse the Local Database button in " \
+                "the DICOM Browser to pick a different location.".format(defaultDatabaseDirectory)
     if not os.path.exists(databaseDirectory):
-      os.mkdir(databaseDirectory)
+      try:
+        os.mkdir(databaseDirectory)
+      except OSError:
+        databaseDirectory = defaultDatabaseDirectory
+        if not os.path.exists(databaseDirectory):
+          os.mkdir(databaseDirectory)
+        slicer.util.infoDisplay(message, windowTitle='DICOM')
     self.onDatabaseDirectoryChanged(databaseDirectory)
 
   def onTableDensityComboBox(self, state):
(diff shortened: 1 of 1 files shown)


Modules/Scripted/DICOMLib/DICOMWidgets.py
```diff
@@ -491,18 +491,22 @@ def promptForDatabaseDirectory(self):
...
+      documents = qt.QDesktopServices.storageLocation(documentsLocation)
+      defaultDatabaseDirectory = os.path.join(documents, "SlicerDICOMDatabase")
+      message = "DICOM Database will be stored in\n\n{}\n\nUse the Local Database button in " \
+                "the DICOM Browser to pick a different location.".format(defaultDatabaseDirectory)
     if not os.path.exists(databaseDirectory):
-      os.mkdir(databaseDirectory)
+      try:
+        os.mkdir(databaseDirectory)
+      except OSError:
+        databaseDirectory = defaultDatabaseDirectory
```
che85: @fedorov I would assign defaultDatabaseDirectory here because we could run into problems causing the variable to be referenced before assigned

fedorov: You mean it could happen with testingEnabled, but failing to create the directory? I guess it is not impossible, but in this case wouldn't it make sense to move definition of defaultDatabaseDirectory to the top of the function?

che85: Oh yeah makes sense. I didn't see the message variable...

pieper:
LGTM

Modules/Scripted/DICOMLib/DICOMWidgets.py
```diff
@@ -491,18 +491,22 @@ def promptForDatabaseDirectory(self):
...
-        slicer.util.infoDisplay(message, windowTitle='DICOM')
+      documentsLocation = qt.QDesktopServices.DocumentsLocation
+      documents = qt.QDesktopServices.storageLocation(documentsLocation)
+      defaultDatabaseDirectory = os.path.join(documents, "SlicerDICOMDatabase")
+      message = "DICOM Database will be stored in\n\n{}\n\nUse the Local Database button in " \
+                "the DICOM Browser to pick a different location.".format(defaultDatabaseDirectory)
     if not os.path.exists(databaseDirectory):
-      os.mkdir(databaseDirectory)
+      try:
+        os.mkdir(databaseDirectory)
```
lassoan: os.makedirs works recursively, use that to take care of nested directories
(permission or other issues may still prevent directory creation, so it's fine to keep exception handling)

Modules/Scripted/DICOMLib/DICOMWidgets.py
```diff
@@ -491,18 +491,22 @@ def promptForDatabaseDirectory(self):
...
     else:
       databaseDirectory = self.settings.value('DatabaseDirectory')
-      if not databaseDirectory:
-        documentsLocation = qt.QDesktopServices.DocumentsLocation
-        documents = qt.QDesktopServices.storageLocation(documentsLocation)
-        databaseDirectory = os.path.join(documents, "SlicerDICOMDatabase")
-        message = "DICOM Database will be stored in\n\n{}\n\nUse the Local Database button in " \
-                  "the DICOM Browser to pick a different location.".format(databaseDirectory)
-        slicer.util.infoDisplay(message, windowTitle='DICOM')
+      documentsLocation = qt.QDesktopServices.DocumentsLocation
```
lassoan: I had to spend several minutes reviewing this couple of lines of code and I'm still not sure about intent and correctness. Please put variable initializations where to the closest possible place where variables are actually used (similarly as it was in the original version).

Modules/Scripted/DICOMLib/DICOMWidgets.py
fedorov: @pieper is there any particular reason why `os.makedirs` was not used in the first place? If there is no some hidden logic, I will push the change.

Modules/Scripted/DICOMLib/DICOMWidgets.py
fedorov: I did not move initialization of the variables, they remain in the same places where they were in the original version. It looks like I moved them, but I just changed the spacing. I think the location is alright, not sure what can be improved here.

fedorov: It might be easier to look at the actual function, not the change set: https://github.com/fedorov/Slicer/blob/3da900772d24cbf809725576de4bfceaf9bdffe5/Modules/Scripted/DICOMLib/DICOMWidgets.py#L489-L510

Modules/Scripted/DICOMLib/DICOMWidgets.py
pieper: +1 to os.mkdirs

fedorov:
I updated the code to rearrange few variables and use os.makedirs().

fedorov:
@lassoan I believe I addressed your concerns. Please let me know if any further changes are needed. Thank you.

lassoan:
Thank you.

fedorov:
Thanks!

Integrated in r25633 = 2a6e43ecd5bff6e235db298dc38354279381b9e7

PR #642, opened 2016-12-22
BUG: itkhdf5_cpp library component name is cpplibraries

fbudin69500:
If cpplibraries is not given as a COMPONENT name, hdf5_cpp does
not get packaged in Slicer.


fbudin69500:
This patch will not solve completely itkhdf5* packaging issue.
The corresponding ITK patch is under review: http://review.source.kitware.com/#/c/21890/

jcfr:
Should we wait to update ITK before integrating ?

fbudin69500:
Both fixes are technically independent, but maybe this patch should also update the ITK commit hash to fully fix the issue. In that case, waiting until the corresponding ITK patch is integrated is a good idea.

fbudin69500:
The issue has been corrected in ITK: https://github.com/InsightSoftwareConsortium/ITK/commit/f535cf3cd87178b06eb682358a6d19fd7d5ed7da

pieper:
LGTM - @ihnorton can you commit it?

fbudin69500:
To have the itkhdf5_cp library packaging issue fix, ITK also needs to be updated in Slicer, either by cherry-picking the commit I mentioned above, or by creating a new branch from the current master with the necessary additional commits specific to Slicer.

pieper:
Ah, thanks @fbudin69500 

@jcfr are you able to look at this?

ihnorton:
The patch is tiny. Seems like we should just pull it into the Slicer ITK fork if we can't do a full update cycle right now. (don't think I have access to the ITK github branch, though)

pieper:
Hi Isaiah -

I added you to the Slicer/ITK team if you want to try doing the update.
Thanks!

-Steve

jcfr:
@fbudin69500 Thanks for the fix

I will update the Slicer/ITK fork later today.

For future reference, the current branch is:

 https://github.com/Slicer/ITK/tree/slicer-v4.11.0-2016-11-07-883ecb4
 https://github.com/Slicer/Slicer/blob/0fa5202aadf70664dbdbf6aa0f1f5265a573333b/SuperBuild/External_ITKv4.cmake#L32-L35

Instructions to update fork are here:

  https://www.slicer.org/wiki/Documentation/Nightly/Developers/Versioning#Project_fork

jcfr:
Integrated in [r25631](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25631)

PR #643, opened 2017-01-10
Update DICOMUtils.py

kmader:
adding a function to easily load lists of series by UID without having to manually create a ```dicomWidget``` object. Also changing the ```loadPatientByUID``` object to use the new function.


cpinter:
Looks good to me. Thanks for the contribution! If there are no other comments today I'll integrate.

pieper:
👍

cpinter:
Integrated in http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25634

Thanks for the contribution!

PR #647, opened 2017-01-16
Add Volumetric mesh support

agirault:
See: https://www.na-mic.org/Wiki/index.php/2017_Winter_Project_Week/Support_for_volumetric_meshes

With the help of @lassoan @jcfr @cpinter @curtislisle @vovythevov @pieper (Thank you!)

# Features & Results
### ENH : Add volumetric mesh support in vtkMRMLModelNod

Update vtkMRMLModelNode to store a vtkPointSet instead of a
vtkPolyData. vtkPointSet can be a base class for vtkPolyData
(surface mesh) as well as vtkUnstructuredGrid (volumetric mesh).

The safest way to know if the mesh is unstructuredGrid is to
check if `vtkMRMLModelNode::GetUnstructuredGrid()` is not NULL,
but since it requires to update the pipeline, vtkMRMLModelNode
also has a member variable called `MeshTypeHint`.

For backward compatibility, keep the public polydata-based API
and wrap around the mesh-based API.

Update vtkMRMLModelDisplayNode, vtkMRMLModelDisplayableManager,
vtkMRMLModelSliceDisplayableManager, and qMRMLModelDisplayNodeWidget
to use the mesh-based API.

volumetricmesh1

###	ENH: Add support for vtkUnstructuredGrid IO
Update vtkMRMLModelStorageNode and qSlicerModelsReader to
be able to read and write unstructured grid based models using
the vtkUnstructuredGrid readers and writers (.vtk and .vtu).

Implement `GetAssociatedDataNode()` similar to the one in
vtkMRMLSegmentationStorageNode to offer the correct supported
write file types.

volumetricmeshio

###	ENH: Implement vtkMRMLProceduralColorNode::GetLookupTable
Migrate conversion of ColorTransferFunction to LookupTable
from ModelDisplayableManager to ProceduralColorNode, to
simplify code in the displayable manager, and to be able to
reuse that conversion code for ModelSliceDisplayableManager.

### ENH: Consolidate AutoScalarRange & ScalarRangeFlag

AutoScalarRange - in vtkMRMLDisplayNode -  would define if the
ModelDisplayNode would reset its ScalarRange to map the range
of the mesh active scalar or not (manually set by the user).

ScalarRangeFlag - in vtkMRMLDisplayNode - would define if the
ModelDisplayableManager would map scalar range based on the
active scalar if UseDataScalarRange (same that AutoScalarOn),
would use the DataType range if UseDataTypeScalarRange, would
use the color node lookup table range if UseColorNodeScalarRange,
or would use the display node scalar range (aka its OWN range
defined by AutoScalarRange value) if UseDisplayNodeScalarRange.

To consolidate those two class members which show redundancies,
they are replaced by a unique ScalarRangeFlag which can take the
following values:
* UseManualScalarRange (new, equivalent to AutoScalarRangeOff)
* UseDataScalarRange (as previously, equivalent to AutoScalarRangeOn)
* UseColorNodeScalarRange (as previously)
* UseDataTypeScalarRange (as previously)
Note the removal of `UseDisplayNodeScalarRange`.

For backward compatibility, the AutoScalarRange API is conserved
and wraps Set/GetScalarRangeFlag.

The range update based on this ScalarRangeFlag is then made in
vtkMRMLModelDisplayNode::UpdateMeshPipeline instead of
vtkMRMLModelDisplayableManager::SetModelDisplayProperty, which
simplifies greatly the code in the displayable manager by only
calling `mapper->SetScalarRange(modelDisplayNode->GetScalarRange())`,
and allows that call in vtkMRMLModelSliceDisplayableManager to
now return the correct range based on that option.

vtkMRMLModelSliceDisplayableManager is also updated to instantiate
the vtkMapper in a similar way than it is made in the 3D displayable
manager for visualization consistency (deep copy of the lut, etc...)

volumetricmesh2

### ENH: Add Thresholding for vtkMRMLModelDisplayNode

volumetricmeshthreshold

### ENH: Add model mesh type in qMRMLModelInfoWidget

volumetricmeshinfo

# Needed tests
* Work made on vtkMRMLDisplayNode and vtkMRMLModelDisplayNode could impact a lot of classes. Should probably test with fiber bundles (@ihnorton ?).

# Future work
* Clean up and consolidate qSlicerModelsModuleWidget (display properties, representations, wireframe, clipping... just move options around to make it more usable)
* Add clipping option to smoothen the cut (using vtkClipDataSet vs vtkExtractGeometry). Not urgent since the Slicer view can display a smooth cut.
* Update extensions like [CleaverExtension](https://www.slicer.org/wiki/Documentation/Nightly/Extensions/CleaverExtension
) or [BodyCentricCubicMesh](https://www.slicer.org/wiki/Documentation/Nightly/Modules/BodyCentricCubicMesh) module, import edition/creation work from [IA_FEMesh](https://www.slicer.org/wiki/Modules:IA_FEMesh-Documentation-3.6).


Libs/MRML/Core/vtkMRMLModelDisplayNode.cxx
```diff
@@ -31,20 +35,23 @@ vtkMRMLModelDisplayNode::vtkMRMLModelDisplayNode()
 {
   this->PassThrough = vtkPassThrough::New();
   this->AssignAttribute = vtkAssignAttribute::New();
+  this->ThresholdFilter = vtkThreshold::New();
+  this->Threshold = false;
```
lassoan: A name that includes "enabled" (ThresholdEnabled or EnableThreshold) would be more clear.

Libs/MRML/Core/vtkMRMLModelDisplayNode.h
```diff
@@ -74,7 +75,20 @@ class VTK_MRML_EXPORT vtkMRMLModelDisplayNode : public vtkMRMLDisplayNode
...
-  virtual void UpdatePolyDataPipeline();
+  virtual void UpdateMeshPipeline();
+
+  vtkGetMacro(Threshold,bool);
+  vtkSetMacro(Threshold,bool);
+  vtkBooleanMacro(Threshold,bool);
+
+  /// Set the threshold range of the model display node.
+  /// \sa GetThresholdMin(), GetTresholdMax()
+  void SetThresholdRange(double min, double max);
```
lassoan: Could you add a Get/SetThresholdRange with double[2] input/output?

Libs/MRML/Core/vtkMRMLModelNode.cxx
```diff
@@ -141,64 +173,130 @@ void vtkMRMLModelNode::SetAndObservePolyData(vtkPolyData *polyData)
...
     return;
     }
 
-  vtkAlgorithm* oldPolyDataAlgorithm = this->PolyDataConnection ?
-    this->PolyDataConnection->GetProducer() : 0;
+  vtkAlgorithm* oldMeshAlgorithm = this->MeshConnection ?
+    this->MeshConnection->GetProducer() : 0;
 
-  this->PolyDataConnection = newPolyDataConnection;
+  this->MeshConnection = newMeshConnection;
```
lassoan: If would be simpler and safer to use a vtkSmartPointer. Or, probably even better, you could probably use object observer macros - for example vtkSetAndObserveMRMLObjectEventsMacro takes care all register/unregister event observe/unobserve.

Libs/MRML/Core/vtkMRMLModelNode.cxx
```diff
@@ -141,64 +173,130 @@ void vtkMRMLModelNode::SetAndObservePolyData(vtkPolyData *polyData)
...
+  this->MeshType = vtkMRMLModelNode::PolyDataMeshType;
+  this->SetMeshConnection(newPolyDataConnection);
+  this->GeometryFilter->SetInputConnection(0);
+}
+
+//---------------------------------------------------------------------------
+void vtkMRMLModelNode
+::SetUnstructuredGridConnection(vtkAlgorithmOutput *newUnstructuredGridConnection)
+{
+  this->MeshType = vtkMRMLModelNode::UnstructuredGridMeshType;
```
lassoan: Do we need to read/write this->MeshType from/to XML?

Libs/MRML/Core/vtkMRMLModelDisplayNode.h
agirault: for the `GetThresholdRange` with double[2] output, the values min and max are retrieved from `this->ThresholdFilter` directly so I can't return a double* to the double[2] range since it would be a local variable. Would the signature `GetThresholdRange(double range[2])` be alright? The user would need to take care of instantiating the double[2] and pass it as an argument.

Libs/MRML/Core/vtkMRMLModelNode.cxx
agirault: I'm not familiar with the `ReadXML` and `WriteXML` processes. If the read calls `SetUnstructuredGridConnection` or `SetPolydataConnection` then no because MeshType will be set. If not it looks easy to add.

Libs/MRML/Core/vtkMRMLModelDisplayNode.cxx
agirault: 👍

Libs/MRML/Core/vtkMRMLDisplayNode.cxx
```diff
@@ -165,8 +164,6 @@ void vtkMRMLDisplayNode::WriteXML(ostream& of, int nIndent)
 
   of << indent << " scalarRangeFlag=\"" << this->ScalarRangeFlag << "\"";
```
agirault: 👍

Libs/MRML/Core/vtkMRMLModelNode.cxx
agirault: Would you have an example somewhere I could use?

Libs/MRML/Core/vtkMRMLProceduralColorNode.cxx
```diff
@@ -129,6 +134,38 @@ void vtkMRMLProceduralColorNode::ProcessMRMLEvents ( vtkObject *caller,
 }
 
 //-----------------------------------------------------------
+vtkLookupTable* vtkMRMLProceduralColorNode::GetLookupTable()
+{
+  this->ConvertedCTFtoLUT->SetNumberOfTableValues(0);
+
+  // since setting the range is a no-op on color transfer functions,
+  // copy into a color look up table with, for now 256 entries
+  const int numEntries = 256;
```
lassoan: Make it number of color table entries a member variable, initialized to 256 by default. No need to add GUI to change it.

Libs/MRML/Core/vtkMRMLProceduralColorNode.h
```diff
@@ -88,6 +88,12 @@ class VTK_MRML_EXPORT vtkMRMLProceduralColorNode : public vtkMRMLColorNode
   /// Only compares the color map (x->RGB mapping)
   static bool IsColorMapEqual(vtkColorTransferFunction* tf1, vtkColorTransferFunction* tf2);
 
+  ///
+  /// Reimplemented vtkMRMLColorNode::GetLookupTable() to convert
+  /// the transfer function to a look up table with 256 entries
```
lassoan: do not hardcode 256 (see above) - it's just as simple to define a member variable

Libs/MRML/DisplayableManager/vtkMRMLModelSliceDisplayableManager.cxx
```diff
@@ -406,60 +408,87 @@ ::UpdateDisplayNodePipeline(vtkMRMLDisplayNode* displayNode, const Pipeline* pip
...
+    {
+    mapper->SetInputConnection( pipeline->Transformer->GetOutputPort() );
+    mapper->SetScalarVisibility(modelDisplayNode->GetScalarVisibility());
+    if (mapper->GetScalarVisibility())
+      {
+      // The renderer uses the lookup table scalar range to
+      // render colors. By default, UseLookupTableScalarRange
+      // is set to false and SetScalarRange can be used on the
+      // mapper to map scalars into the lookup table. When set
+      // to true, SetScalarRange has no effet and it is necessary
```
lassoan: typo: effet -> effect

Libs/MRML/Core/vtkMRMLModelDisplayNode.cxx
agirault: Fixed in recent commits.

Libs/MRML/Core/vtkMRMLModelNode.cxx
lassoan: Search for vtkSetAndObserveMRMLObjectEventsMacro in the Slicer source code. I think it's used in vtkMRMLNode. Simpler versions (without event observations) are used in many more places.

agirault:
@lassoan @jcfr after the email I sent to Slicer devs mailing list, and after asking @ihnorton, it looks like [vtkMRMLUnstructuredGridNode](https://github.com/Slicer/Slicer/blob/eefeac9286f48552e8418a11f412b2823a09b407/Libs/MRML/Core/vtkMRMLUnstructuredGridNode.h) is not used (as well as its display node). Should we get rid of it in this PR?

jcfr:
> Should we get rid of it in this PR?

Sounds good to me. And considering you are adding support for unstructured grid, now is the ideal time for removing this dead/unused code.

Libs/MRML/Core/vtkMRMLModelNode.cxx
lassoan: For example, it might be needed when somebody tries to retrieve the mesh type before the mesh is read from file. It's a short time, so most likely nothing tries to access mesh type during that short transient. Just try to think about it and test saving/loading thoroughly and see if you need to do something.

lassoan:
Please add a few basic tests. At least try if you can read an unstructured grid from file and write it to file and that you can get/set various mesh types in random order in the MRML node and you get consistent results.

Libs/MRML/Core/vtkMRMLProceduralColorNode.cxx
agirault: No GUI, but a Get/Set API?

lassoan: Yes, exactly.

Libs/MRML/Core/vtkMRMLModelNode.cxx
agirault: So should I just switch to a smartpointer?

lassoan: vtkSmartPointer only manages register/unregister. vtkSetAndObserveMRMLObjectEventsMacro manages register/unregister and observe/unobserved events as well. So, the best would be to use vtkSetAndObserveMRMLObjectEventsMacro.

ihnorton:
I will test with our code today, please hold off merging for a while. Thanks. (cc @ljod)

Libs/MRML/Core/vtkMRMLProceduralColorNode.cxx
agirault: fixed in more recent commits.

Libs/MRML/Core/vtkMRMLProceduralColorNode.h
agirault: fixed in more recent commits.

Libs/MRML/DisplayableManager/vtkMRMLModelSliceDisplayableManager.cxx
agirault: fixed in more recent commits. (as well as in vtkMRMLDisplayableManager)

Libs/MRML/Core/vtkMRMLModelNode.cxx
agirault: I need a little more help to get familiar with what's expected of these. The fact that those classes in the Model module handle VTK algorithm and algorithm outputs more than VTK datasets raises the following questions on my end:
* Do I need to add the member `vtkIntArray ObservedEvents` to vtkMRMLModelNode and to vtkMRMLModelDisplayNode?
* What events do I need to observe? All [these](https://github.com/Slicer/Slicer/blob/eefeac9286f48552e8418a11f412b2823a09b407/Modules/Loadable/VolumeRendering/MRML/vtkMRMLVolumePropertyNode.cxx#L28-L33)?
* In vtkMRMLModelNode, should I call the `vtkSetAndObserveMRMLObjectEventsMacro` on the `vtkAlgorithmOutput` (GetMeshConnection()), or the `vtkPointSet` (GetMesh(), which runs the pipeline) ? What about the 
* In vtkMRMLModelDisplayNode, should I call the `vtkSetAndObserveMRMLObjectEventsMacro` on the model node, the input mesh, or the output mesh? Knowing that there can be 3 different output meshes (from Passthrough, AssignAttributes, or Threshold)
* Where I am supposed to call `vtkObserveMRMLObjectEventsMacro` and `vtkUnObserveMRMLObjectEventsMacro` ?

Libs/MRML/Core/vtkMRMLModelNode.cxx
agirault: Indeed, if a vtkMRMLModelNode is instantiated and GetMeshType is called before a mesh is set, then the value will not make sense. Should I had an `Empty` or `None` entry to initialize MeshType when the model does not have a mesh?

lassoan: Difficult to tell what's the best solution. Probably keeping just the two actual types is simpler (less chance for inconsistency) and by default assuming that we have NULL polydata int he node is more backward-compatible. So, unless you find a use case that's actually broken with the current implementation, I would keep things as they are now. My comment was intended just for giving some thoughts and testing use cases around reading/writing of the node. For example, make sure that there is no crash if the mesh file is not available (it was deleted, etc); or it's the wrong type.

agirault:
@lassoan I kept the API for Get/Set PolyData/PolyDataConnection in vtkMRMLModelNode for backward compatibility, but not in vtkMRMLModelDisplayNode, as it was not called anywhere in Slicer. Do I also need to add wrapping functions in that class? It would include:
* GetInputPolydata (wrapping GetInputMesh())
* GetInputPolydataConnection (wrapping GetInputMeshConnection())
* GetOutputPolydata (wrapping GetOutputMesh())
* GetOutputPolydataConnection (wrapping GetOutputMeshConnection())
* UpdatePolydataPipeline (wrapping UpdateMeshPipeline())

lassoan:
It's not nice to get polydata from the display node. Unfortunately, in the past a few classes hijacked the model display node and implemented display pipeline in the display node (instead of implementing display pipeline properly in a custom displayable manager).

For example:
vtkAlgorithmOutput* vtkMRMLDiffusionTensorVolumeSliceDisplayNode
::GetOutputPolyDataConnection()
{
  return this->DiffusionTensorGlyphFilter->GetOutputPort();
}

While this situation is not nice, I would not touch this in this pull request. I would recommend to keep remain fully compatible with the old model display node API (keep the same method name, parameters, and behavior).

ihnorton:
I needed some [fixes](https://github.com/agirault/Slicer/compare/volumetric-mesh-support...ihnorton:pr647_fixes?expand=1) to get this to build to the point where I can test.

After that, the [changes for SlicerDMRI](https://github.com/SlicerDMRI/SlicerDMRI/compare/master...ihnorton:update_for_pr647?expand=1) are small. I'm still working through getting full functionality back because down-casting from pointset seems to silently drop associated tensors somewhere, so our color pipeline doesn't work (solid colors only).

This is a bit more disruptive than I expected in terms of API changes, but so far the adjustment doesn't seem too bad (provided I can fix the coloring). I haven't checked whether anyone else outside of core is using these APIs, but if only us and SlicertRT then it should be fine.

There are some other modules that might have used these APIs (e.g. NIRAL DTI stuff), but I think most of them don't anyway since the VTK update.

Libs/MRML/Core/vtkMRMLModelNode.cxx
lassoan: > Do I need to add the member vtkIntArray ObservedEvents to vtkMRMLModelNode and to vtkMRMLModelDisplayNode?

Whatever events you set, will be observed, others will be unobserved. You would probably always set the same events, so you could store in a member variable, but creating a vtkIntArray is a light operation and mesh would not be set very frequently, so I don't think you need to have a vtkIntArray class member variable.

> What events do I need to observe? All these?
Observe the same events as observed before (only vtkCommand::ModifiedEvent).

> In vtkMRMLModelNode, should I call the vtkSetAndObserveMRMLObjectEventsMacro on the vtkAlgorithmOutput (GetMeshConnection()), or the vtkPointSet (GetMesh(), which runs the pipeline) ?
The macro is developed for simple VTK object referencing and observation. It may or may not be fully applicable here. You can have a look at what the macro does and choose the simplest solution that works.

lassoan:
@ihnorton In the long term it could be better to move towards custom displayable managers for DTI stuff, because model display node and displayable managers becoming too complex to maintain and enhance. You can use Segmentations or Transforms display nodes and displayable managers as examples.

agirault:
> It's not nice to get polydata from the display node. Unfortunately, in the past a few classes hijacked the model display node and implemented display pipeline in the display node (instead of implementing display pipeline properly in a custom displayable manager). - @lassoan 

In the case of Models, I thought it was actually quite nice to be able to have "Assign Attributes" and "Threshold" in the Display Node, so it wouldn't have to be duplicated in each Displayable Manager. What else would the display node for models handle if not that pipeline? Just have "ThresholdEnabled" as a class member?

> I needed some fixes to get this to build to the point where I can test. After that, the changes for SlicerDMRI are small. [...] This is a bit more disruptive than I expected in terms of API changes, but so far the adjustment doesn't seem too bad.

Sorry you had to do this, I actually just updated the tests too on my end. Neither of those changes (in tests or SlicerDMRI) should be necessary if I re-implement the old ModelDisplayNode API for PolyData like discussed above with Andras to maintain full compatibility.

> I'm still working through getting full functionality back because down-casting from pointset seems to silently drop associated tensors somewhere, so our color pipeline doesn't work (solid colors only). 

That sounds like a more serious issue. Thanks for looking into it.

agirault:
@lassoan the more I see it the more I wonder if we shouldn't get rid of the GeometryFilter which extracts the surface mesh of the unstructured grid when calling GetPolyData if the point set is an unstructured grid, instead of returning null. I think it might be an unexpected result for users to have GetPolyData return a non-null pointer if there is not polydata point set.

lassoan:
I agree, it's cleaner to not perform the outer surface extraction automatically for unstructured grids.

lassoan:
> In the case of Models, I thought it was actually quite nice to be able to have "Assign Attributes" and "Threshold" in the Display Node, so it wouldn't have to be duplicated in each Displayable Manager. What else would the display node for models handle if not that pipeline? Just have "ThresholdEnabled" as a class member?

MRML nodes are supposed to be responsible for data storage and interaction only. All processing is the responsibility of Logic classes, all rendering related tasks is the responsibility of displayable managers. These rules ensure clear roles&responsibilities, which results in easy maintenance and development.

I agree that for performance reasons it is better to share part of the processing and rendering pipeline between multiple displayable managers and display nodes are convenient place for that. But it wouldn't be much more difficult to share part of the pipeline between multiple displayable managers.

agirault:
> While this situation is not nice, I would not touch this in this pull request. I would recommend to remain fully compatible with the old model display node API (keep the same method name, parameters, and behavior). - @lassoan 

This forces me to add a `MeshType` member in the display node too, since there is no handle on the vtkMRMLModelNode to know what the mesh type is when retrieving mesh connection.

lassoan:
Could the display node just call a method of the data node to get the mesh type?

agirault:
> I agree, it's cleaner to not perform the outer surface extraction automatically for unstructured grids.

Done. I also fixed the issue of the vtkTriangleFilter in the qMRMLModelInfoWidget.

> Could the display node just call a method of the data node to get the mesh type? - @lassoan

The ModelDisplayNodes have no handle on the ModelNode, it is the other way around, but I believe I found a satisfying implementation (see patched 1st commit).

> I needed some fixes to get this to build to the point where I can test. After that, the changes for SlicerDMRI are small. [...] This is a bit more disruptive than I expected in terms of API changes, but so far the adjustment doesn't seem too bad. - @ihnorton

I restored the API for Get/Set polydata on vtkMRMLModelDisplayNode so the changes you made for Slicer tests and SlicerDMRI should not be necessary anymore.

I'm now working on improving the tests. My funding and time is now very limited for that project so I'll do my best but will be restricted pretty quickly. I think the main issue to tackle will be

jcfr:
@agirault Thanks again for your effort on this :+1:  Your improvement will be very welcome and appreciated by many. 

* After @ihnorton  confirms the restored API behave as expected 
and, 
* as suggested by @lassoan , after adding tests that reasonably exercise the API

... we should be in good shape for integration.

All of that said, in addition of mentioning possible future improvements in the commit message, it would also be great to also summarize on the wiki some of the great suggestion that couldn't be implemented in this PR. (e.g removing rending pipeline from nodes, sharing logic by DM, dedicated DM for fiber, ...)

Libs/MRML/Core/vtkMRMLModelDisplayNode.h
```diff
@@ -73,23 +90,48 @@ class VTK_MRML_EXPORT vtkMRMLModelDisplayNode : public vtkMRMLDisplayNode
...
+  double GetThresholdMin();
+  double GetThresholdMax();
 
 protected:
   vtkMRMLModelDisplayNode();
   ~vtkMRMLModelDisplayNode();
   vtkMRMLModelDisplayNode(const vtkMRMLModelDisplayNode&);
   void operator=(const vtkMRMLModelDisplayNode&);
 
-  virtual void ProcessMRMLEvents(vtkObject *caller,
```
ihnorton: Why is this removed?

ihnorton:
- I don't think backward compatibility is practical for anyone who overrides the removed API in a subclasses, because those virtual methods in subclasses will not called. Our `ModelDisplayNode` subclasses override `Get*` in order to run data through filters before display (which should probably change).
- Taking out `ProcessMRMLEvents` significantly changes the event propagation and should probably be restored.
- The rest is ok for us.

Libs/MRML/Core/vtkMRMLModelDisplayNode.h
agirault: When running model tests and debugging, it appeared that `UpdatePolyDataPipeline()` (now `UpdateAssignedAttribute()`) was called literally a hundred times before a mesh would even be set and observed by the model node, since that `ProcessMRMLEvents()` would just call it everytime an object of the scene calls `ModifiedEvent` (from my understanding?)
Instead, I only called it when `SetActiveScalarName` and `SetActiveScalarLocation` were changed, and would call `AssignAttribute::Update()` to ensure the changes would propagate through the pipeline. Also, if the model node is modified, that means the input mesh is modified, and changes should automatically be propagated along the pipeline.

Even though I first removed it for convenienve during debugging, the tests for the models kept working, as well as the Models module GUI and its event propagation, so I figured it would be a good performance increase not to run that pipeline which appeared to often run for no good reason... but I didn't thoroughly test all the remaining tests so if there are failing ones I must be missing someone in my reasoning :)

agirault:
> The rest is ok for us: this changeset plus locally restoring ProcessMRMLEvents passes all tests. - @ihnorton 

I see the tensors part is commented out: did that resolved the initial issue you discussed?

Thanks for your reviews. I left a message on your preceding comment about `ProcessMRMLEvents` : indeed it appears I am not correctly understanding how, why, and when the overall pipeline should be updated, or why this needs to be called so often.

ihnorton:
> I see the tensors part is commented out: did that resolved the initial issue you discussed?

No, the main issue was that I didn't overload and/or calling all of the renamed methods. So, for example, when the model displayable manager called `DisplayNode->GetPolyData*`, it would get the original polydata, not the output connection from the filtering pipeline (adds tubes/glyphs and tensor->scalar coloring computation).

The commented section is vestigial code that was crashing due to an ordering issue, I think. With some other small changes, that section is no longer necessary (as far as I can tell).

Libs/MRML/Core/vtkMRMLModelDisplayNode.h
lassoan: ProcessMRMLEvents() is only called when an observed object is modified. Probably it is called whenever the associated model node modified, but it definitely should not observe all the nodes of the scene. I'll check if there is any obvious reason why there many updates, but if we change anything in this then probably the best to do it later, once this pull request has been merged.

lassoan: I've added a breakpoint into vtkMRMLModelDisplayNode::ProcessMRMLEvents and I don't see any unreasonable ModifiedEvent calls. Could you check in your stack trace where those hundreds of ModifiedEvent calls originated from?

ihnorton: > was called literally a hundred times before a mesh would even be set and observed by the model node

LOL. The Slicer event model is best understood as *eventually consistent*.

agirault: > I've added a breakpoint into vtkMRMLModelDisplayNode::ProcessMRMLEvents and I don't see any unreasonable ModifiedEvent calls. Could you check in your stack trace where those hundreds of ModifiedEvent calls originated from?

Ok I figured out what happened: I was running the model tests and the `EXERCISE_ALL_BASIC_MRML_METHODS` as well as other subsets preceding the test I was debugging would just keep triggering `ProcessMRMLEvents`, for a good reason I guess.

However, three things happen in `UpdatePolyDataPipeline()` (now `UpdateAssignedAttribute()`)
1. reassign the scalar name and location in the `AssignAttribute` filter
2. Update the `AssignAttribute` filter
3. if scalar range mode is not manual, update scalar range based on data, colornode, or data type (within `UpdateScalarRange()`)

Updating the Assignattribute filter should only happen when ActiveScalarName or ActiveScalarLocation are updated. Would it make sense to call `UpdateScalarRange()` only in `ProcessMRMLEvents()` ?

agirault: @lassoan Once I put it back in `ProcessMRMLEvents, this warning is called 10 times when I load a volume: https://github.com/Slicer/Slicer/pull/647/files#diff-3e0d8da9d31fcf1472c2e8e9293b19a8R289

Should I just get rid of those warnings?

lassoan: I'm not sure what warnings you are referring to.

lassoan: > Updating the Assignattribute filter should only happen when ActiveScalarName or ActiveScalarLocation are updated.

Usually if you set the same value for a parameter in a VTK filter then the filter detects that and does not do any actual update. So, probably you don't need to check if ActiveScalarName or ActiveScalarLocation is changed.

lassoan:
I don't see any blocking issues. If there are no objections, I'll merge this on Monday. If we find problems then we can address them in further commits.

@agirault, please squash your commits into a single one (or send me a consolidated commit log message and I squash the commits on merge).

jcfr:
> I don't see any blocking issues.

Great

> merge this on Monday

:+1: 

A a side note, @agirault should have commit access and should be able to merge on Monday :fireworks:

lassoan:
Great, then @agirault will merge it.

agirault:
Before I merge this, I'll need to add back `ProcessMRMLEvents`, correct?

lassoan:
Yes, I would keep the changes to what is necessary to reduce chance of breaking things and give chance to users to start testing earlier. We can implement further cleanups and fixes in separate commits later.

Libs/MRML/Core/vtkMRMLModelDisplayNode.h
agirault: > I'm not sure what warnings you are referring to.

```c++
vtkWarningMacro("Can not use data scalar range since model display node"
                      << "does not have an active scalar name.");
```
 @ https://github.com/Slicer/Slicer/pull/647/files#diff-3e0d8da9d31fcf1472c2e8e9293b19a8R289

(If you scroll a little bit towards the top it should be shown as highlighted in Github. For some reason a website header covers the anchor.)

agirault:
> please squash your commits into a single one - @lassoan 

Just curious about why we do not keep them as separate? I rebased them to have 8 distinct commits, which I feel helps for separating each independent improvements/changes and document them per commit, where one unique commit would have a huge description and a lot of changes.

jcfr:
If each one of them independently improve the code while ensuring it still
compile, it makes sense to keep them separate.

lassoan:
Yes, exactly.

Most often you need some consolidation of commits. For example, if you make changes based on code review comments then you don't want to commit the initial version and subsequent improvements separately, but commit the final version only.

lassoan:
If those warnings detect actual problems, then fix those problems. If the warning is logged when everything is in order, then fix the checks that lead to the warning (to only log warning when something is not right) or remove the warning completely.

agirault:
Last suggested adjustments made, integrated in:
* [r25660](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25660) - ENH: Add volumetric mesh support in vtkMRMLModelNode
* [r25661](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25661) - STYLE: Use actorProperties in Model displayable manager
* [r25662](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25662) - ENH: Implement vtkMRMLProceduralColorNode::GetLookupTable
* [r25663](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25663) - ENH: Consolidate AutoScalarRange & ScalarRangeFlag
* [r25664](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25664) - ENH: Add Thresholding for vtkMRMLModelDisplayNode
* [r25665](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25665) - ENH: Improve MRML Model tests
* [r25666](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25666) - ENH: Get rid of vtkMRMLUnstructuredGridNode
* [r25667](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25667) - STYLE: Update ModelsModule contributors & acknowledgement

Thank you, everyone, for your help.

jcfr:
:tada: Thanks for your contribution :+1: and thanks for everyone for reviewing and moving things forward.

lassoan:
@agirault Could you write about this new feature on the mailing list and maybe also to post it on ResearchGate (https://www.researchgate.net/project/3D-Slicer)?

jcfr:
A blog is on the way :)

lassoan:
@agirault Unstructured grid related tests are failing:
http://slicer.cdash.org/viewTest.php?onlydelta&buildid=960672

agirault:
Working on it now.

agirault:
Quick question: why did GitHub show that the tests passed on CircleCI for this branch? Does it not run CTest? I thought it would and merged based on that assumption. 

1. Removing tests using unstructured grid not properly removed in [r25666](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25666): https://github.com/Slicer/Slicer/pull/660
2. [qSlicerModelsModuleWidgetTest](http://slicer.cdash.org/testDetails.php?test=7814256&build=960672) and [qMRMLModelDisplayNodeWidgetTest1](http://slicer.cdash.org/testDetails.php?test=7814255&build=960672) fail because of the tresholder being called while there is not yet an input (see below). I'll work on this now:
```
ERROR: In C:\D\N\Slicer-0-build\VTKv7\Common\ExecutionModel\vtkDemandDrivenPipeline.cxx, line 710
vtkCompositeDataPipeline (000002472107B300): Input port 0 of algorithm vtkThreshold(0000024720E7BBE0) has 0 connections but is not optional.
```
3. Last test to investigate will be [vtkSlicerSubjectHierarchyModuleLogicTest](http://slicer.cdash.org/testDetails.php?test=7814223&build=960672). Looks like an issue with transform hardening on model in SubjectHierarchy, I'll have to give it a look afterward.

jcfr:
> Why did GitHub show that the tests passed on CircleCI for this branch? Does it not run CTest? 

Thanks to the work for @MayeulChassagnard and @thewtex , as soon #618 is integrated, the test will be indeed  be executed.

It will  be finalized and integrated next week.

Thanks for your patience

agirault:
> Thanks to the work for @MayeulChassagnard and @thewtex , as soon #618 is integrated, the test will be indeed be executed. It will be finalized and integrated next week.

👍

agirault:
Blog post out : https://blog.kitware.com/3d-slicer-adds-volumetric-mesh-support/

Should we disseminate on the mailing list and research gate?

jcfr:
> Should we disseminate on the mailing list and research gate?

Yes

pieper:
Nice post!

lassoan:
Yes, very nice. Please send a short summary and blogpost link to the mailing list and ResearchGate.

agirault:
I'll do this tomorrow morning. Thank you.

agirault:
I'll do this tomorrow morning. Thank you.
Just checking: are we talking about the users or developers mailing list?

jcfr:
Both, and mention the feature is available in the nightly build.
Thanks

On Feb 1, 2017 8:12 PM, "Alexis Girault" <notifications@github.com> wrote:

I'll do this tomorrow morning. Thank you.
Just checking: are we talking about the users or developers mailing list?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/Slicer/Slicer/pull/647#issuecomment-276835958>, or mute
the thread

agirault:
Sounds good. I could use the summary from the blog in the emails:

> With the collaboration of multiple 3D Slicer community members reunited during the NA-MIC 2017 Winter Project Week, volumetric mesh visualization support was brought back to the 3D Slicer 4.x series.
> 
> The basic capabilities gravitate around an enhanced MRML model node class, which now wraps a vtkPointSet data structure, a base class of vtkPolyData (surface mesh) and vtkUnstructuredGrid (volumetric mesh), as well as improved display functionalities for clipping, thresholding, 3D and 2D visualization and scalar range selection. These capabilities offer better consistency and usability.
> 
> These features, which are already available in the Slicer Nightly build, will be a part of the 4.7 release of 3D Slicer and will facilitate upcoming work to bring more volumetric mesh capabilities (such as mesh creation and conversion to other data structures) to the platform.
>
> Find more about those features on the dedicated blog post describing volumetric mesh support: https://blog.kitware.com/3d-slicer-adds-volumetric-mesh-support/

The second paragraph could be for the developers mailing list only.

agirault:
I was not in the slicer-users mailing list, but can't reach the website to register: http://massmail.bwh.harvard.edu/mailman/listinfo/slicer-users

jcfr:
Where did you get this link ? We need to update it to be:

http://massmail.spl.harvard.edu/mailman/listinfo/slicer-users

Where `bwh` is changed by `spl`.

As far as I know, it is not possible to have a redirection and we need to update all references.

agirault:
> Where did you get this link ?

https://www.slicer.org/wiki/Documentation/Nightly/FAQ/General#Where_can_I_find_someone_to_help_me_use_Slicer.3F

jcfr:
> https://www.slicer.org/wiki/Documentation/Nightly/FAQ/General#Where_can_I_find_someone_to_help_me_use_Slicer.3F

Thanks. Reference updated.

PR #649, opened 2017-01-20
ENH: Upgrade the OpenIGTLink library to version 3.0.

tokjun:
See:

http://www.na-mic.org/Bug/view.php?id=4325


jcfr:
Thanks for the update. I will review and integrate this afternoon.

jcfr:
Integrated in [r25652](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25652)

PR #650, opened 2017-01-21
ENH: Added CLI output display to module widget

lassoan:
Problem:
It is often difficult to find out why a CLI failed or get detailed output.
(user would need to open the log window and find the messages in the log,
but most users don't even know that there is a log window).

Solution:
Expand button is added next to the status message.
User can click there to show/hide all details.
Error output is displayed in red, at the top. Standard output is displayed black.
Window is automatically sized to show all content with minimal need for scrolling.

### DIFF
--- a/Libs/MRML/CLI/vtkMRMLCommandLineModuleNode.h
+++ b/Libs/MRML/CLI/vtkMRMLCommandLineModuleNode.h
@@ -119,6 +119,12 @@ public:
   /// \sa GetStatus(), IsBusy()
   const char* GetStatusString() const;
 
+  void SetOutputText(const std::string& text, bool modify = true);
+  const std::string GetOutputText() const;
+
+  void SetErrorText(const std::string& text, bool modify = true);
+  const std::string GetErrorText() const;
+
   /// Return true if the module is in a busy state: Scheduled, Running,
   /// Cancelling, Completing.
   /// \sa SetStatus(), GetStatus(), BusyMask, Cancel()
--- a/Base/QTCLI/qSlicerCLIProgressBar.cxx
+++ b/Base/QTCLI/qSlicerCLIProgressBar.cxx
@@ -26,6 +26,10 @@
 #include <QLabel>
 #include <QProgressBar>
 
+// CTK includes
+#include <ctkExpandButton.h>
+#include <ctkFittedTextBrowser.h>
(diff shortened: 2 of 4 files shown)


fedorov:
Good idea!

Can you include a screenshot? Will this show only the output relevant to the specific CLI, or just a shortcut to the unified log output?

lassoan:
Expand button is next to status field, by default details are hidden:
defaulthidden

Short error message example:
shorterror

Longer output example:
longmessage

fedorov:
Thanks, looks helpful!

Libs/MRML/CLI/vtkMRMLCommandLineModuleNode.h
```diff
@@ -119,6 +119,12 @@ class VTK_MRML_CLI_EXPORT vtkMRMLCommandLineModuleNode : public vtkMRMLNode
   /// \sa GetStatus(), IsBusy()
   const char* GetStatusString() const;
 
+  void SetOutputText(const std::string& text, bool modify = true);
```
jcfr: Could you document the news methods like what is done for `SetStatus` and `GetStatus` ?

pieper:
+1!

jcfr:
Very nice :+1:

jcfr:
Make also use to include the CTK update to include the change for `ctkFittedBrowser`

lassoan:
The CTK fix is not required but it makes the output nicer (there is no unnecessary vertical scrollbar for wide output).

lassoan:
I've been thinking about auto-show, too. Finally I haven't added that because I was not sure how this automatic showing could be harmonized with manual show/hide. Any suggestions?
For example, we could always show the error output if the extension returns with error (regardless of user choice).

cpinter:
Nice! Thanks!
Auto-show error sounds good.

jcfr:
> always show the error output if the extension returns with error (regardless of user choice).

That would be a reasonable default. Adding an option `ExpandOutputLogOnError` (or flag `NotifyUserOnError` on the CLI node) could also be sensitive.

I guess we could move forward with integration as it is and revisit "auto-showing" later ...

lassoan:
I've added automatic showing of error output if CLI completes with errors and integrated in r25646.

PR #652, opened 2017-01-22
STYLE: Expose utility methods in qMRMLColor

agirault:
Move GetColorIndexByName from vtkMRMLColorTableNode
to vtkMRMLColorNode to expose it in qMRMLColorModel
and qMRMLColorTableView.

by @finetjul

### DIFF
--- a/Libs/MRML/Core/vtkMRMLColorNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLColorNode.cxx
@@ -375,6 +375,31 @@ const char *vtkMRMLColorNode::GetColorName(int ind)
 }
 
 //---------------------------------------------------------------------------
+int vtkMRMLColorNode::GetColorIndexByName(const char *name)
+{
+  if (name == NULL)
+    {
+    vtkErrorMacro("vtkMRMLColorNode::GetColorIndexByName: need a non-null name as argument")
+    return -1;
+    }
+
+  if (!this->GetNamesInitialised())
+    {
+    this->SetNamesFromColors();
+    }
+
+  std::string strName = name;
+  for (unsigned int i = 0; i < this->GetNumberOfColors(); ++i)
+    {
+    if (strName == this->GetColorName(i))
+      {
+      return i;
+      }
+    }
+  return -1;
+}
+
+//---------------------------------------------------------------------------
 std::string vtkMRMLColorNode::GetColorNameWithoutSpaces(int ind, const char *subst)
 {
   std::string name = std::string(this->GetColorName(ind));
--- a/Libs/MRML/Core/vtkMRMLColorNode.h
+++ b/Libs/MRML/Core/vtkMRMLColorNode.h
@@ -110,9 +110,14 @@ public:
       TypeModifiedEvent = 20002
     };
 
-  ///
-  /// Get the 0th based nth name of this colour
+  /// Get name of a color from its index (index is 0-based)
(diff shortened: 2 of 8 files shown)


Libs/MRML/Core/vtkMRMLColorNode.cxx
```diff
@@ -375,6 +375,23 @@ const char *vtkMRMLColorNode::GetColorName(int ind)
 }
 
 //---------------------------------------------------------------------------
+int vtkMRMLColorNode::GetColorIndexByName(const char *name)
+{
+  if (this->GetNamesInitialised() && name != NULL)
```
lassoan: if name is NULL then log error and return with error immediately

lassoan: Shouldn't we just initialize color names if they have not been initialized yet? 

For example, as it is done in vtkMRMLColorNode::GetColorName(int ind):

  if (!this->GetNamesInitialised())
    {
    this->SetNamesFromColors();
    }

Libs/MRML/Core/vtkMRMLColorNode.cxx
```diff
@@ -375,6 +375,23 @@ const char *vtkMRMLColorNode::GetColorName(int ind)
...
+    {
+    std::string strName = name;
+    for (unsigned int i = 0; i < this->Names.size(); i++)
```
lassoan: To be consistent, either use public APIs (this->GetNumberOfColors() and this->GetColorName()) or use member variables directly (this->Names.size() and this->Names[i]).

Also, a minor thing: get number of colors only once and store it in a variable.

Libs/MRML/Core/vtkMRMLColorNode.h
```diff
@@ -110,10 +110,15 @@ class VTK_MRML_EXPORT vtkMRMLColorNode : public vtkMRMLStorableNode
       TypeModifiedEvent = 20002
     };
 
-  ///
   /// Get the 0th based nth name of this colour
```
lassoan: I know it was not your change, but this description does not make sense. Something like this would be more clear:
/// Get name of a color from its index (index is 0-based)

lassoan:
Thank you for your contribution. I see that the code I commented on was there before (only at a different place), so if you don't want to modify anything in this commit (to not break something accidentally), it's fine. But if you have the time, it would be nice to make some improvements.

agirault:
> Thank you for your contribution. I see that the code I commented on was there before (only at a different place), so if you don't want to modify anything in this commit (to not break something accidentally), it's fine. But if you have the time, it would be nice to make some improvements.

No problem at all. Those are just changes made by other developers in an old fork of Slicer I'm trying to get up to date, so I haven't really dug deep in those changes, I just figured I'd make those merge requests after rebasing & resolving conflicts if things run. If it's a couple changes like this it's easy to do!

agirault:
@lassoan please let me know if this is okay with you now.

lassoan:
Thank you, looks good to me.

agirault:
Integrated in [r25649](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25649).

agirault:
Integrated in [r25649](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25649).

PR #653, opened 2017-01-22
BUG: Add support for selecting No module in module selector toolbar

agirault:
Because the NoModuleAction is not observed by ctkMenuComboBox, the
toolbar shall clear the text of the current action manually.

In case the previous action was the NoModuleAction, it would not
have been added into the history, however it shall still trigger
the moduleSelected signal to show in the module panel.

by @finetjul

### DIFF
--- a/Base/QTGUI/qSlicerModuleSelectorToolBar.cxx
+++ b/Base/QTGUI/qSlicerModuleSelectorToolBar.cxx
@@ -265,9 +265,19 @@ void qSlicerModuleSelectorToolBar::onModuleSelected(const QString& name)
 void qSlicerModuleSelectorToolBar::actionSelected(QAction* action)
 {
   Q_D(qSlicerModuleSelectorToolBar);
+  if (action == 0)
+    {
+    // Because the NoModuleAction is not observed by ctkMenuComboBox, the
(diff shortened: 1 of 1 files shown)


jcfr:
Thanks

Base/QTGUI/qSlicerModuleSelectorToolBar.cxx
```diff
@@ -265,9 +265,19 @@ void qSlicerModuleSelectorToolBar::onModuleSelected(const QString& name)
...
+  if (action == 0)
+    {
+    // Because the NoModuleAction is not observed by ctkMenuComboBox, the
+    // toolbar shall clear the text of the current action manually.
+    d->ModulesComboBox->clearActiveAction();
+    }
   QAction* lastAction = d->lastSelectedAction();
   if (action == lastAction)
     {
+    // In case the previous action was the NoModuleAction, it would not
```
lassoan: It's not very nice that we keep emitting moduleSelected signals even when the module is not changed. Have you considered making ctkMenuComboBox smarter instead of adding workarounds in this class?

lassoan:
Removed repeated signals when no module selected repeatedly and committed in r27197.

PR #654, opened 2017-01-22
ENH: Expose functions for reproducing 3D view shortcuts

agirault:
Before this, there was no simple way to ask programaticaly for the view
to reset the camera like it does when pressing the key pad shortcut keys.
Now the user can do this directly from python. This is potentially useful
for the scripted modules as well.
by @vovythevov

### DIFF
--- a/Libs/MRML/Widgets/qMRMLThreeDView.cxx
+++ b/Libs/MRML/Widgets/qMRMLThreeDView.cxx
@@ -41,6 +41,7 @@
 // MRML includes
 #include <vtkMRMLViewNode.h>
 #include <vtkMRMLScene.h>
+#include <vtkMRMLCameraNode.h>
 
 // VTK includes
 #include <vtkCollection.h>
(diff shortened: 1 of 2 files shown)


pieper:
👍

Libs/MRML/DisplayableManager/vtkThreeDViewInteractorStyle.cxx
```diff
@@ -792,6 +783,56 @@ void vtkThreeDViewInteractorStyle::SetModelDisplayableManager(
 }
 
 //----------------------------------------------------------------------------
+void vtkThreeDViewInteractorStyle::RotateToPosterior(bool inverse)
```
lassoan: Axis labels have been made configurable. In several places in the code, labels or orientations are still hardcoded, but we should really move away from that.

Therefore, we shouldn't use vtkThreeDViewInteractorStyle::RotateToPosterior(bool inverse), but something like this:

vtkThreeDViewInteractorStyle::RotateToViewAxis(const std::string& axisLabel)

axisLabel is "L", "R", ... is one of the labels in vtkMRMLAbstractViewNode::AxisLabels

Libs/MRML/DisplayableManager/vtkThreeDViewInteractorStyle.cxx
```diff
@@ -792,6 +783,56 @@ void vtkThreeDViewInteractorStyle::SetModelDisplayableManager(
...
+    inverse ? vtkMRMLCameraNode::Inferior : vtkMRMLCameraNode::Superior);
+}
+
+//----------------------------------------------------------------------------
+void vtkThreeDViewInteractorStyle::
+ResetCamera(bool resetRotation, bool resetTranslation, bool resetDistance)
+{
+  if (!this->CameraNode)
+    {
+    return;
```
lassoan: log warning instead of just silently fail

Libs/MRML/DisplayableManager/vtkThreeDViewInteractorStyle.cxx
agirault: Sounds good. I'll look into that. However, it looks like `AxisLabel` can be modified to customize the labels with `SetAxisLabel`. Should we pass an integer which defined the index of the axis instead? Or just go through all the values of `AxisLabel` and check if it exists, then infer how to rotate based on the axis id?

Also, @vovythevov, do you remember why you implemented this in the interactor style as well as the qMRMLThreeDView? Thank you.

lassoan: You can get the Nth axis label by calling const char* vtkMRMLAbstractViewNode::GetAxisLabel(int labelIndex). If the view node is not accessible or you would always need to rotate to Nth axis then you can have a method like:
vtkThreeDViewInteractorStyle::RotateToViewAxis(int direction) // direction is in 0..5, same way as in abstract view node axis index (-X, +X, ..., -Z, +Z)
or
vtkThreeDViewInteractorStyle::RotateToViewAxis(int axis, bool positiveDirection) --- where axis is in 0..2

agirault: Sure. Do we want this in `vtkThreeDViewInteractorStyle` as well as `qMRMLThreeDView`?

lassoan: I guess it could be enough to just have it in qMRMLThreeDView.

agirault:
Changes made. In the future, it could be interesting to consolidate [`ctkAxesWidget::Axis`](http://www.commontk.org/docs/html/classctkAxesWidget.html#afdcef6c48213bc075ac6224e68176c5d), [`vtkMRMLCameraNode::Direction`](https://www.slicer.org/doc/html/classvtkMRMLCameraNode.html#a5e55d1409f0ebbeb8bfb3b9f5f2ee52a), and [`vtkMRMLAbstractViewNode`](https://www.slicer.org/doc/html/classvtkMRMLAbstractViewNode.html#a98eebd4ca00ff50feffb4c062a29a985) order of labels.

Libs/MRML/Widgets/qMRMLThreeDView.cxx
```diff
@@ -239,6 +240,93 @@ void qMRMLThreeDView::addDisplayableManager(const QString& displayableManagerNam
 }
 
 //------------------------------------------------------------------------------
+void qMRMLThreeDView::rotateToViewAxis(unsigned int axisId)
+{
+  vtkThreeDViewInteractorStyle* style =
+    vtkThreeDViewInteractorStyle::SafeDownCast(this->interactorStyle());
+  if (!style)
+    {
+    QDebug(QtCriticalMsg) << "qMRMLThreeDView::RotateToViewAxis: "
```
lassoan: Why don't you use simply qCritical or qWarning?

Libs/MRML/Widgets/qMRMLThreeDView.cxx
```diff
@@ -239,6 +240,93 @@ void qMRMLThreeDView::addDisplayableManager(const QString& displayableManagerNam
...
+    }
+}
+
+//------------------------------------------------------------------------------
+void qMRMLThreeDView::rotateToViewAxis(const std::string& axisLabel)
+{
+  Q_D(qMRMLThreeDView);
+  if (!d->MRMLViewNode)
+    {
+    return;
```
lassoan: log warning

Libs/MRML/Widgets/qMRMLThreeDView.cxx
```diff
@@ -239,6 +240,93 @@ void qMRMLThreeDView::addDisplayableManager(const QString& displayableManagerNam
...
+
+  for (unsigned int i = 0; i < vtkMRMLAbstractViewNode::AxisLabelsCount; ++i)
+    {
+    if (axisLabel == std::string(d->MRMLViewNode->GetAxisLabel(i)))
+      {
+      this->rotateToViewAxis(i);
+      return;
+      }
+    }
+  QDebug(QtWarningMsg) << "qMRMLThreeDView::RotateToViewAxis: "
```
lassoan: qWarning

Libs/MRML/Widgets/qMRMLThreeDView.cxx
```diff
@@ -239,6 +240,93 @@ void qMRMLThreeDView::addDisplayableManager(const QString& displayableManagerNam
...
+
+//------------------------------------------------------------------------------
+void qMRMLThreeDView
+::resetCamera(bool resetRotation, bool resetTranslation, bool resetDistance)
+{
+  vtkThreeDViewInteractorStyle* style =
+    vtkThreeDViewInteractorStyle::SafeDownCast(this->interactorStyle());
+  if (!style)
+    {
+    QDebug(QtCriticalMsg) << "qMRMLThreeDView::RotateToViewAxis: "
```
lassoan: qWarning

lassoan:
Thank you, it looks good! If you make those few trivial changes then it should be ready to merge.

agirault:
Integrated in [r25651](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25651)

PR #655, opened 2017-01-22
ENH: Improve UpdateDisplayNodeFromVolumeNode

agirault:
When using this method, one would think that if the displayNode already
has a volume property node and a region of interest node, no other node
will be created. This change makes the method behave that way and gets
rid of the method overload.

### DIFF
--- a/Modules/Loadable/VolumeRendering/Logic/vtkSlicerVolumeRenderingLogic.cxx
+++ b/Modules/Loadable/VolumeRendering/Logic/vtkSlicerVolumeRenderingLogic.cxx
@@ -906,40 +906,51 @@ vtkMRMLVolumeRenderingDisplayNode* vtkSlicerVolumeRenderingLogic
 // if needed create vtkMRMLVolumePropertyNode and vtkMRMLAnnotationROINode
 // and initioalize them from VolumeNode
 //----------------------------------------------------------------------------
-void vtkSlicerVolumeRenderingLogic::UpdateDisplayNodeFromVolumeNode(
-                                          vtkMRMLVolumeRenderingDisplayNode *displayNode,
-                                          vtkMRMLVolumeNode *volumeNode,
-                                          vtkMRMLVolumePropertyNode **propNode,
-                                          vtkMRMLAnnotationROINode **roiNode)
+void vtkSlicerVolumeRenderingLogic
+::UpdateDisplayNodeFromVolumeNode(vtkMRMLVolumeRenderingDisplayNode *displayNode,
+                                  vtkMRMLVolumeNode *volumeNode,
+                                  vtkMRMLVolumePropertyNode *propNode /* = NULL */,
+                                  vtkMRMLAnnotationROINode *roiNode /* = NULL */)
 {
+  if (displayNode == NULL)
+    {
+    vtkErrorMacro("vtkSlicerVolumeRenderingLogic::UpdateDisplayNodeFromVolumeNode: "
+                  << "display node pointer is null.")
+    return;
+    }
 
   if (volumeNode == NULL)
     {
     displayNode->SetAndObserveVolumeNodeID(NULL);
     return;
     }
-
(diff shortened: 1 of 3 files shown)


jcfr:
"Improve UpdateDisplayNodeFromVolumeNode from python" -> "Improve UpdateDisplayNodeFromVolumeNode for python"

jcfr:
commit message:

```
When using this method, one would think that if the paramNode already has
a volume property node and a region of interest node, no other node will
be created. This change makes the method behave correctly.
```

-> 

```
This commit fixes two issues:

* When using this method, one would think that if the paramNode already has
a volume property node and a region of interest node, no other node will
be created. This change makes the method behave correctly.

* It also ensures the method is available from python removing the "inline"
keyword.
```

Modules/Loadable/VolumeRendering/Logic/vtkSlicerVolumeRenderingLogic.h
```diff
@@ -217,9 +217,10 @@ class VTK_SLICER_VOLUMERENDERING_MODULE_LOGIC_EXPORT vtkSlicerVolumeRenderingLog
                                        vtkMRMLVolumeNode *volumeNode,
                                        vtkMRMLVolumePropertyNode **propNode,
                                        vtkMRMLAnnotationROINode **roiNode);
+
   /// Utility function that calls UpdateDisplayNodeFromVolumeNode()
```
lassoan: Could you add some more details what this method does?

Modules/Loadable/VolumeRendering/Logic/vtkSlicerVolumeRenderingLogic.cxx
```diff
@@ -948,6 +948,18 @@ void vtkSlicerVolumeRenderingLogic::UpdateDisplayNodeFromVolumeNode(
 }
 
 //----------------------------------------------------------------------------
+void vtkSlicerVolumeRenderingLogic
+::UpdateDisplayNodeFromVolumeNode(vtkMRMLVolumeRenderingDisplayNode *paramNode,
+                                  vtkMRMLVolumeNode *volumeNode)
+{
+  vtkMRMLVolumePropertyNode *propNode =
+    paramNode ? paramNode->GetVolumePropertyNode() : NULL;
```
lassoan: Shouldn't we return with an error if paramNode is NULL? (log an error and/or return with a bool value)

agirault: It looks that it actually was the default behavior: when calling this function, it would just pass a NULL vtkMRMLVolumePropertyNode and a NULL vtkMRMLAnnotationROINode as 3rd and 4th args to the same function (UpdateDisplayNodeFromVolumeNode) with 4 args.

I guess we could trigger a warning if they're missing from the input vtkMRMLVolumeRenderingDisplayNode. I'm not 100% sure of how this code is expected to behave since I did not write that commit, but I'll adapt to whichever once we figure it out.

lassoan: Just from looking at the code it seems to me that Slicer will crash if paramNode is NULL (in the 4-argument UpdateDisplayNodeFromVolumeNode a method of paramNode is called, without checking first if paramNode is NULL or not). Add checking of vtkMRMLVolumeRenderingDisplayNode *paramNode/displayNode and immediately log error and return if it's NULL.

agirault:
@lassoan I refactored that commit to:
- check that displayNode is not NULL or log error
- get rid of overloaded function by defining default NULL rvalue for propNode and roiNode
- only create new propNode and roiNode if none already exist
- update method description

lassoan:
Thank you it looks good! :+1:

agirault:
Integrated in [r25650](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25650).

PR #656, opened 2017-01-22
ENH: Add vtkMRMLApplicationLogic::FitSlicesToVolume

agirault:
Volume reader has an option ("fitSlicesToVolumes") to control the
fitting of the slices at loading.

### DIFF
--- a/Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
+++ b/Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
@@ -355,6 +355,12 @@ void vtkMRMLApplicationLogic::SetInteractionNode(vtkMRMLInteractionNode *interac
 }
 
 //----------------------------------------------------------------------------
+void vtkMRMLApplicationLogic::PropagateVolumeSelection(int layer, int fit)
+{
+  this->Internal->PropagateVolumeSelection(layer, fit);
+}
+
+//----------------------------------------------------------------------------
 void vtkMRMLApplicationLogic::PropagateVolumeSelection(int fit)
 {
   this->PropagateVolumeSelection(
@@ -380,12 +386,6 @@ void vtkMRMLApplicationLogic::PropagateLabelVolumeSelection(int fit)
 }
 
 //----------------------------------------------------------------------------
-void vtkMRMLApplicationLogic::PropagateVolumeSelection(int layer, int fit)
-{
-  this->Internal->PropagateVolumeSelection(layer, fit);
-}
-
-//----------------------------------------------------------------------------
 void vtkMRMLApplicationLogic::PropagateTableSelection()
 {
   if ( !this->Internal->SelectionNode || !this->GetMRMLScene() )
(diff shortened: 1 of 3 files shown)


Modules/Loadable/Volumes/qSlicerVolumesReader.cxx
```diff
@@ -145,6 +145,11 @@ bool qSlicerVolumesReader::load(const IOProperties& properties)
     {
     options |= properties["discardOrientation"].toBool() ? 0x10 : 0x0;
     }
+  // 0: do not fit slices
```
pieper: Wouldn't an enum make things clearer?

Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
```diff
@@ -439,6 +438,43 @@ void vtkMRMLApplicationLogic::FitSliceToAll(bool onlyIfPropagateVolumeSelectionA
 }
 
 //----------------------------------------------------------------------------
+void vtkMRMLApplicationLogic::FitSlicesToVolume(const char* volumeNodeID)
+{
+  if (this->Internal->SliceLogics.GetPointer() == 0)
+    {
+    return;
+    }
+  std::string volume(volumeNodeID ? volumeNodeID : "");
```
lassoan: if volumeNodeID is NULL then just log a warning and return here

Modules/Loadable/Volumes/qSlicerVolumesReader.cxx
```diff
@@ -186,7 +191,14 @@ bool qSlicerVolumesReader::load(const IOProperties& properties)
...
       if (appLogic)
         {
-        appLogic->PropagateVolumeSelection(); // includes FitSliceToAll by default
+        if (fit == 1)
+          {
+          appLogic->PropagateVolumeSelection(); // includes FitSliceToAll by default
+          }
+        else if (fit == 2)
+          {
+          appLogic->FitSlicesToVolume(node->GetID());
```
lassoan: If you choose fit==2 (by the way, I agree with Steve, these should be enums instead of magic numbers) then what would select the new volume as foreground/background/label in any of the slice views?

agirault: > then what would select the new volume as foreground/background/label in any of the slice views?

I don't think I quite understand your question.

lassoan: If fit==1 then this call both selects the volume in the background layer and fits slices.
However, if fit==2 then only ->FitSlicesToVolume(node->GetID()) is called, which method does not selects the volume to be shown in any slice viewer.

jcfr:
Having an other look at this, I do not understand the use of the second option ``2: fit slice to loaded volume``.

Since a loaded volume is always set as active (see [here](https://github.com/Slicer/Slicer/blob/721be1ba7c622443d876025bd2b30d6f60348247/Modules/Loadable/Volumes/qSlicerVolumesReader.cxx#L179-L186) and https://github.com/Slicer/Slicer/commit/5138b20577767e5b55990d9c810532e2d4f5b965), it will always be propagated and fitted to all views.

jcfr:
Cc: @agirault

PR #657, opened 2017-01-22
ENH: Add optional checkable column in qMRMLColorModel

agirault:
* Remove LabelInColorColumn to be more consistent with qMRMLSceneModel
columns.
* Fix updateColorFromItem() that was always forcing the color opacity to be
1.0 because QColor rgba was always opaque.

by @finetjul & @vovythevov

### DIFF
--- a/Libs/MRML/Widgets/qMRMLCheckableNodeComboBox.h
+++ b/Libs/MRML/Widgets/qMRMLCheckableNodeComboBox.h
@@ -26,7 +26,11 @@
 
 class qMRMLCheckableNodeComboBoxPrivate;
 
+/// This class provides the ability to select(check) multiple nodes at a time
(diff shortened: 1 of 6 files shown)


jcfr:
Integrated in [25839](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25839)

cpinter:
Is it somehow possible that these recent changes made the header labels disappear? In 4.6.2 they looked like this
image
and now they look like this
image

cpinter:
(I created an issue for this so that it is kept in sight http://www.na-mic.org/Bug/view.php?id=4353)

cpinter:
@agirault @jcfr Fixed in https://github.com/Slicer/Slicer/commit/4605982fd30958813df3518cf63136cc666b6f85

jcfr:
Thanks for the fix.

PR #659, opened 2017-01-24
ColumnIndex bug in LabelStatistics.py

kmader:
if there is no ```self.colorNode``` defined (it is None) in ```LabelStatisticsLogic```, the columnIndex variable is never initialized


jcfr:
Thanks for the patch.

Integrated in [r25658](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25658)

lassoan:
@kmader Thank you for your contribution. Note that the Editor and Label Statistics modules are in maintenance mode - bugs are fixed but no new features will be added. I would recommend to transition to the vastly improved Segment Editor and the Segment Statistics modules instead.

PR #660, opened 2017-01-25
TEST: remove obsolete unstructured grid tests

agirault:
vtkMRMLUnstructuredGridNode and its components were removed in r25666 [1], within the pull request #647, but the tests were not undefined in `Libs/MRML/Core/Testing/CMakeLists.txt`.

[1] http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25666

Removes the following failing tests from CDash [2]:
* vtkMRMLSceneImportTest_unstruct_grid.mrml
* vtkMRMLSCeneTest2_unstruct_grid.mrml
* vtkMRMLUnstructuredGridDisplayNodeTest1
* vtkMRMLUnstructuredGridNodeTest1
* vtkMRMLUnstructuredGridStorageNodeTest1

[2] http://slicer.cdash.org/viewTest.php?onlydelta&buildid=960672

Replace scene tests using unstructured grid by scene tests using
volumetric mesh.
* vtkMRMLSceneImportTest_volumetric_mesh.mrml
* vtkMRMLSCeneTest2_volumetric_mesh.mrml

Issue reported by @lassoan

### DIFF
--- a/Libs/MRML/Core/Testing/CMakeLists.txt
+++ b/Libs/MRML/Core/Testing/CMakeLists.txt
@@ -201,9 +201,6 @@ simple_test( vtkMRMLTransformDisplayNodeTest1 )
 simple_test( vtkMRMLTransformNodeTest1 )
 simple_test( vtkMRMLTransformStorageNodeTest1 )
 simple_test( vtkMRMLUnitNodeTest1 )
-simple_test( vtkMRMLUnstructuredGridDisplayNodeTest1 )
-simple_test( vtkMRMLUnstructuredGridNodeTest1 )
-simple_test( vtkMRMLUnstructuredGridStorageNodeTest1 )
 simple_test( vtkMRMLVectorVolumeDisplayNodeTest1 )
 simple_test( vtkMRMLVectorVolumeNodeTest1 )
 simple_test( vtkMRMLViewNodeTest1 )
@@ -233,7 +230,7 @@ set( ScenesToTest
   multi_display.mrml
   name.mrml
   SlicePipeline.mrml
-  unstruct_grid.mrml
+  volumetric_mesh.mrml
   vol_and_cube_camera.mrml
   vol_and_cube.mrml
   vol.mrml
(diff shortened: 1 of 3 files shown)


agirault:
@lassoan

lassoan:
Do you have a test scene that contains an unstructured grid?

I've tried to load Libs\MRML\Core\Testing\TestData\PentaHexa.vtk (unstructured grid) and found 2 issues:
* Slicer did not recognize that it can load the .vtk file as an unstructured grid. Maybe you could add .vtk extension to the supported extensions of model node, or at least convert PentaHexa.vtk to .vtu file format and replace the .vtk file with that.
* The object appeared partially hollow because backface culling was enabled by default. Probably backface culling should be disabled by default when we load an unstructured grid file.

lassoan:
One more small thing: I think we agreed that it would be better to move the "Clip" checkbox in the Models module into the Clipping section. Sorry for dumping these findings in this unrelated ticket, these are just very small things that I though could be fixed very quickly, without creating new tickets. But if you prefer I can create a new ticket for each finding.

agirault:
> Do you have a test scene that contains an unstructured grid?

I do not, sadly I have been lacking time to work on this project.

> Slicer did not recognize that it can load the .vtk file as an unstructured grid. Maybe you could add .vtk extension to the supported extensions of model node.

I just tried and it works for me. The reader for unstructured grid does support .vtk here: https://github.com/Slicer/Slicer/blob/master/Libs/MRML/Core/vtkMRMLModelStorageNode.cxx#L169-L180

> The object appeared partially hollow because backface culling was enabled by default. Probably backface culling should be disabled by default when we load an unstructured grid file.

👍 

> I think we agreed that it would be better to move the "Clip" checkbox in the Models module into the Clipping section. 

Agreed, just ran out of time, and it wasn't a priority to integrate the volumetric mesh support.

agirault:
Updated vtkMRMLSceneImportTest and vtkMRMLSCeneTest2 to use model node with unstructured grid. @lassoan I used the same .vtk file you mentioned, and the tests pass on my machine.

agirault:
Updated vtkMRMLSceneImportTest and vtkMRMLSCeneTest2 to use model node with unstructured grid, see new commit. @lassoan I used the same .vtk file you mentioned, and the tests pass on my machine.

agirault:
Updated vtkMRMLSceneImportTest and vtkMRMLSCeneTest2 to use model node with unstructured grid, see new commit. @lassoan I used the same .vtk file you mentioned, and the tests pass on my machine. Could we merge these?

lassoan:
Looks good to me, thank you!

agirault:
Integrated in [r25678](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25678).

PR #661, opened 2017-01-25
BUG: Fix model display threshold pipeline call

agirault:
The ThresholdFilter should not be updated if there is no input mesh
since the model display pipeline won't have an input algorithm. Doing
the same check than the one made for the AssignAttribute filter when
changing active scalar information fixes the issue.

Error:
```
ERROR: In C:\D\N\Slicer-0-build\VTKv7\Common\ExecutionModel\vtkDemandDrivenPipeline.cxx, line 710
vtkCompositeDataPipeline (000002472107B300): Input port 0 of algorithm vtkThreshold(0000024720E7BBE0) has 0 connections but is not optional.
```

Fixed tests:
* qSlicerModelsModuleWidgetTest
* qMRMLModelDisplayNodeWidgetTest1

Issue reported by @lassoan

### DIFF
--- a/Libs/MRML/Core/vtkMRMLModelDisplayNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLModelDisplayNode.cxx
@@ -138,6 +138,7 @@ vtkPointSet* vtkMRMLModelDisplayNode::GetOutputMesh()
 
   vtkAlgorithmOutput* outputConnection = this->GetOutputMeshConnection();
   vtkAlgorithm* producer =  outputConnection? outputConnection->GetProducer() : 0;
+  producer->Update();
   return vtkPointSet::SafeDownCast(
     producer ? producer->GetOutputDataObject(outputConnection->GetIndex()) : 0);
 }
@@ -186,14 +187,10 @@ void vtkMRMLModelDisplayNode::SetThresholdRange(double min, double max)
 {
   vtkMTimeType mtime = this->ThresholdFilter->GetMTime();
   this->ThresholdFilter->ThresholdBetween(min, max);
-  if (mtime >= this->ThresholdFilter->GetMTime())
+  if (this->ThresholdFilter->GetMTime() > mtime)
     {
-    return;
+    this->Modified();
     }
-
-  this->ThresholdFilter->Update();
-
-  this->Modified();
 }
 
 //---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


Libs/MRML/Core/vtkMRMLModelDisplayNode.cxx
```diff
@@ -186,7 +186,7 @@ void vtkMRMLModelDisplayNode::SetThresholdRange(double min, double max)
 {
   vtkMTimeType mtime = this->ThresholdFilter->GetMTime();
   this->ThresholdFilter->ThresholdBetween(min, max);
-  if (mtime >= this->ThresholdFilter->GetMTime())
+  if (mtime >= this->ThresholdFilter->GetMTime() || !this->GetInputMesh())
```
lassoan: The node content (threshold value) is still modified, so we should not return - regardless of having a valid input mesh or not.
We should just avoid calling this->ThresholdFilter->Update(). By the way this->ThresholdFilter->Update() call should not be necessary, because VTK takes care of running the filters when data is requested (and it can actually slow things down that we force an update instead of letting the renderer to compute data when needed).
Could you try removing this->ThresholdFilter->Update() and see if everything works well? The code could be simplified to this then:
<pre>
vtkMTimeType mtime = this->ThresholdFilter->GetMTime();
this->ThresholdFilter->ThresholdBetween(min, max);
if (this->ThresholdFilter->GetMTime() > mtime) 
  {
  this->Modified();
  }
</pre>

agirault: @lassoan indeed: done. I did the same for the AssignAttribute filter but it needed a couple more tweaks to work since the output data object of the filter was needed to find the data scalar range.

agirault:
@lassoan indeed. It's now done. It needs a couple more tweaks to work since the output data object of the filters is needed to find the data scalar range in UpdateScalarRange, as well as in GetOutputMesh. See new commit.

agirault:
@lassoan indeed. It's now done. It needs a couple more tweaks to work since the output data object of the filters is needed to find the data scalar range in `UpdateScalarRange`, as well as in `GetOutputMesh`. See new commit. Let me know if we can merge.

lassoan:
This looks good to me, thank you!

agirault:
Integrated in [r25679](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25679).

PR #662, opened 2017-01-29
BUG: Fix vtkMRMLModelNode ApplyTransform & GetMesh

agirault:
Reverts regression introduced in r25660 [1] by applying the transformation
to the mesh data object directly if the mesh was set with `SetAndObserveMesh`
(which uses vtkTrivialProducer to generate its mesh connection) and was not
set as a mesh connection being part of a pipeline.

[1] http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25660

Also call `vtkAlgorithm::Update()` in `vtkMRMLModelNode::GetMesh()` to
ensure the mesh connection is up to date when getting its output data
object.

Also remove changes to `MeshType` and calls to `Modified()` in `GetPolyData()`
and `GetUnstructuredGrid()`. Those were triggered when data inconsistency
was detected, but a Get() method should never modify an object. Keep
warnings only.

Fixes `vtkSlicerSubjectHierarchyModuleLogicTest`:
http://slicer.cdash.org/testDetails.php?test=7812827&build=962480

Thank you @lassoan for the recommendations.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLModelNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLModelNode.cxx
@@ -207,9 +207,17 @@ vtkPointSet *vtkMRMLModelNode::GetMesh()
 {
   vtkAlgorithm* producer = this->MeshConnection ?
     this->MeshConnection->GetProducer() : 0;
+
+  if (!producer)
+    {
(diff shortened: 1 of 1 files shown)


lassoan:
Perfect, thank you. Committed r25681.

PR #663, opened 2017-02-02
Transform interaction

vovythevov:
This PR add a widget that allow user to interact with transform directly in the 3D view.

To support this feature, the following infrastructure was added/modified in Slicer:
- Improved ```GetRASBounds()```:
The ```vtkMRMLDisplayableTransform``` method ```GetRASBounds()``` was improved to allow the user choose whether they want the bounds of the node transformed (when the node is observing a transform) or not. This is useful when you want to be able to have a cube that emcopasses the bounds of the node.
Also, a boolean return value was added to be able to check easily if the bounding box returned is valid.
**Note**: Because it wasn't very straight forward to do for the Segmentation node, it is **NOT** supported for this node. A warning is emitted if you try to use that option on such node. Testing was added with ```SlicerRASBoundsTest.py```

 - Added Transform logic methods:

1.  ```GetNodesRASBounds()```:
    This method is fairly straight forward and returns the RAS bounds of a list of nodes. It is tested in ```vtkSlicerTransformLogicTest3.cxx```.

 2. ```GetTransformedNodes()```
    This method returns the nodes that are observing a particular transform. It also offers the option to do so recursively so any nodes under a child transform may be returned as well. It is tested in ```vtkSlicerTransformLogicTest2```.

  - Add TransformReferenceAddedEvent and TransformReferenceRemovedEvent
    The transform node lacked a good way to be notified when a node would start observing it. These new event remediate to this by being sent on the transform node whenever another node set it as its transform. This is tested in ```vtkMRMLTransformNodeOnNodeReferenceAddTest```.

  - Minor fix of sample data
    This is simply to be able to use the ```CTA abdomen (Panoramix)``` dataset in testing because it is not centered and thus offers better testing for transforms.

The transform interaction itself has the following features:
  - The widget is shown/hidden based on the ```vtkMRMLTransformDisplayNode::VisualizationMode``` and the ```Visible``` option.  The visualization mode option was chosen instead of a separate "Interactive" option because the interactive mode is so far incompatible with the other modes of visualization for the transform. In my opinion, it integrates better in the interface as well.

transformmodule

 - If no transform is present, the widget will have a default size so the user can still interact with it.

defaultsize

 - Whenever a node that has valid bounds is placed under the transform, the widget will automatically
resize itself to the bounds of the node. 
onecube
If multiple nodes (with valid bounds) are set under the transform, the widget will resize itself to encompass all the nodes.
twonodes

 - GUI/Widget synchronization
Of course, whenever the user interacts with the widget, the transform node will be updated. Likewise, the
whenever the transform node is modified the widget will be updated.
 
  - Support for parent transforms
Parent transforms are supported and allow one to be able to move node individually or by group like so:
withparents

  - User interaction:
The interaction is the standard provided by the widget. Options are exposed to the user through the transform module interface (enable/disable Scaling/Translation/Rotation).

Interaction infrastructure:
 - The underlying widget used is vtkBoxWidget2.
This widget provides a very nice interface as you can directly set to it the desired transform. 

 - No non-linear transforms:
As of now, due to the limitations of the widget, the interaction does not support non-linear transforms.

 - Do not allow interactive and transform visualization at the same time
    The interaction takes precedence over the transform visualization.

  - Default scale size of widget:
The default size is hard-coded for now in the displayable manager. It could made into a parameter later easily however.

 - Testing:
The widget is tested through the ```SlicerTransformInteractionTest1```. To do so, an internal method to access the widget was added to the ```vtkMRMLTransformsDisplayableManager3D```.

Limitations and future work:

  - 2D interaction
Unfortunately, VTK does not have an appropriate 2D widget that would work. The only remotely close widget vtkBorderWidget does not handle rotations correctly. To have the 2D interaction working, one would first have to write a VTK 2D widget for it.


vovythevov:
@lassoan, @jcfr: 
I finally got the PR in for the 3D interaction. Would you guys mind taking a look ?
Thanks !

jcfr:
Thanks for moving this forward. Will be reviewing shortly. :+1:

lassoan:
Could you please resolve the conflicts with the master?

Libs/MRML/Core/vtkMRMLDisplayableNode.cxx
```diff
@@ -319,8 +319,9 @@ void vtkMRMLDisplayableNode::SetDisplayClassVisibility(const char* nodeClass, in
 }
 
 //---------------------------------------------------------------------------
-void vtkMRMLDisplayableNode::GetRASBounds(double bounds[6])
+bool vtkMRMLDisplayableNode
+::GetRASBounds(double bounds[6], bool useTransform)
```
lassoan: You cannot change this method, as it is overridden in many MRML nodes in many extensions. Also, the method name clearly states that bounds are returned in RAS coordinate system. You shouldn't add a flag that completely changes the method.

Instead, you could add a new GetBounds(double bounds[6]) method.

bool return value is not necessary, as it is a common VTK convention that a bounds is invalid if any of the bounds are inverted (bool valid = bounds[0]<=bounds[1] || bounds[2]<=bounds[3] || bounds[4]<=bounds[5]). Returning a bool value would be redundant, as you would need to check validity of the bounds aunway.

Libs/MRML/Core/vtkMRMLTransformDisplayNode.h
```diff
@@ -49,6 +49,7 @@ class VTK_MRML_EXPORT vtkMRMLTransformDisplayNode : public vtkMRMLDisplayNode
     VIS_MODE_GLYPH,
     VIS_MODE_GRID,
     VIS_MODE_CONTOUR,
+    VIS_MODE_INTERACTIVE,
```
lassoan: This is not an alternative visualization mode. We should be able to modify the transform regardless of visualizing it.

Libs/MRML/Core/vtkMRMLTransformableNode.cxx
```diff
@@ -117,7 +117,18 @@ void vtkMRMLTransformableNode::SetAndObserveTransformNodeID(const char *transfor
     transformNodeID = 0;
     }
 
+  vtkMRMLTransformNode* previousNode = this->GetParentTransformNode();
   this->SetAndObserveNodeReferenceID(this->GetTransformNodeReferenceRole(), transformNodeID);
+  if (tnode)
```
lassoan: SetAndObserveTransformNodeID is just a convenience method for setting a parent transform. If parent transform s set using general node reference API (calling SetAndObserveNodeReferenceID directly) then you miss the notification. If you want to make sure you don't miss an added reference then override vtkMRMLNode::OnNodeReferenceAdded.

There is already GetReferencingNodes() method in vtkMRMLScene that can give you a list of all nodes that refer to a particular node. So, it is not very nice that you add a less generic mechanism to get the same information.

I would recommend to not add these two new events, but only add a single NodeReferenceModified event to the MRML scene, which would be invoked when any node references in the scene are modified. Then in your code you could observe this event and retrieve the current list of referencing nodes for the transform.

Libs/MRML/Core/vtkMRMLDisplayableNode.cxx
vovythevov: I agree for the return value.
Yes, if extensions rely on this method it might be too much burden for them to change it. I'll add a method with a different name then.

Libs/MRML/Core/vtkMRMLTransformDisplayNode.h
vovythevov: The reasons I ended up adding it as a visualization mode (contrary to what we decided at NA-MIC week) are:
 - I wasn't sure we should support both a visualization and the interaction at the same time. 
 - The interaction needs to rely on the Visibility and SliceIntersectionVisibility flags too. As of now, in the interface you cannot unselect a visualization mode (no visualization mode =VIS_MODE_NONE). So you wouldn't be able to have interaction without a visualization mode. 

So to make it a separate variable I think we would need to:
 - Have interaction and a visualization mode at the same time
 - Add a visualization mode VIS_MODE_NONE (or another better/appropriate name) to be able to only have the interaction on.
What do you think ?

lassoan: You can add EditorWidgetVisibility and EditorWidgetSliceVisibility flags and handle those completely independently from the Visibility and SliceIntersectionVisibility flags.

Also, please add a separate displayable manager for the widget (similar to vtkMRMLThreeDReformatDisplayableManager) that only checks EditorWidgetVisibility and EditorWidgetSliceVisibility flags and the transform type. A separate displayable manager is appropriate, because the widget only supports linear transforms (while the transform displayable manager supports all transform types) and we'll have to add separate widgets for TPS, Bspline, etc. It also results in having two simpler, independent classes instead of forcing two completely unrelated features into one class.

vovythevov: That makes sense, thanks for the suggestion !

vovythevov:
@lassoan, @jcfr: I updated the PR with Andras suggestions:
 - Rebased onto master
 - Created GetNodeBounds method instead of modifying GetRASBounds
 - Fix the transform event TransformReferenceModifiedEvent
 - Moved the transform functionnality to its own displayable manager

Applications/SlicerApp/Testing/Python/CMakeLists.txt
```diff
@@ -298,6 +298,8 @@ if(Slicer_USE_QtTesting AND Slicer_USE_PYTHONQT)
   slicer_add_python_unittest(SCRIPT RSNAQuantTutorial.py)
   slicer_add_python_unittest(SCRIPT slicerCloseCrashBug2590.py)
   slicer_add_python_unittest(SCRIPT SlicerOrientationSelectorTest.py)
+  slicer_add_python_unittest(SCRIPT SlicerTransformInteractionTest1.py)
+  slicer_add_python_unittest(SCRIPT SlicerRASBoundsTest.py)
```
jcfr: Should this test be named `SlicerNodeBoundsTest` ?

Libs/MRML/Core/vtkMRMLDisplayableNode.h
```diff
@@ -194,8 +194,16 @@ class VTK_MRML_EXPORT vtkMRMLDisplayableNode : public vtkMRMLTransformableNode
   virtual void SetDisplayClassVisibility(const char* nodeClass, int visible);
 
   /// Get bounding box in global RAS the form (xmin,xmax, ymin,ymax, zmin,zmax).
+  /// This method retuns the bounds of the object with any transforms that may
```
jcfr: retuns -> returns

Libs/MRML/Core/vtkMRMLDisplayableNode.h
```diff
@@ -194,8 +194,16 @@ class VTK_MRML_EXPORT vtkMRMLDisplayableNode : public vtkMRMLTransformableNode
...
+  /// be applied to it.
+  /// \sa GetNodeBounds()
   virtual void GetRASBounds(double bounds[6]);
 
+  /// Get bounding box in global RAS the form (xmin,xmax, ymin,ymax, zmin,zmax).
```
jcfr: RAS the form  ->  RAS form

Libs/MRML/Core/vtkMRMLDisplayableNode.h
```diff
@@ -194,8 +194,16 @@ class VTK_MRML_EXPORT vtkMRMLDisplayableNode : public vtkMRMLTransformableNode
...
+  /// This method always returns the bounds of the untransformed object.
```
jcfr: untransformed object. ->  untransformed displayable node.

Libs/MRML/Core/vtkMRMLModelNode.cxx
```diff
@@ -726,19 +726,23 @@ void vtkMRMLModelNode::ApplyTransform(vtkAbstractTransform* transform)
 //---------------------------------------------------------------------------
 void vtkMRMLModelNode::GetRASBounds(double bounds[6])
 {
-  this->Superclass::GetRASBounds( bounds);
+  double localBounds[6];
```
jcfr: localBounds is unused ?

Libs/MRML/Core/vtkMRMLSegmentationNode.cxx
```diff
@@ -460,8 +460,16 @@ void vtkMRMLSegmentationNode::GetRASBounds(double bounds[6])
     {
     this->Segmentation->GetBounds(bounds);
     }
+
 }
 
+//---------------------------------------------------------------------------
+void vtkMRMLSegmentationNode::GetNodeBounds(double bounds[6])
+{
+  Superclass::GetNodeBounds(bounds);
```
jcfr: this->Superclass::GetNodeBounds(bounds);

Libs/MRML/Core/vtkMRMLSegmentationNode.h
```diff
@@ -75,9 +75,17 @@ class VTK_MRML_EXPORT vtkMRMLSegmentationNode : public vtkMRMLDisplayableNode
   /// Get unique node XML tag name (like Volume, Model)
   virtual const char* GetNodeTagName() {return "Segmentation";};
 
-  /// Get bounding box in global RAS in the form (xmin,xmax, ymin,ymax, zmin,zmax).
+  /// Get bounding box in global RAS the form (xmin,xmax, ymin,ymax, zmin,zmax).
```
jcfr: global RAS the form -> global RAS form

Libs/MRML/Core/vtkMRMLSegmentationNode.h
```diff
@@ -75,9 +75,17 @@ class VTK_MRML_EXPORT vtkMRMLSegmentationNode : public vtkMRMLDisplayableNode
...
+  /// This method retuns the bounds of the object with any transforms that may
+  /// be applied to it.
+  /// \sa GetNodeBounds()
   virtual void GetRASBounds(double bounds[6]);
 
+  /// Get bounding box in global RAS the form (xmin,xmax, ymin,ymax, zmin,zmax).
```
jcfr: global RAS the form -> global RAS form

Libs/MRML/Core/vtkMRMLVolumeNode.cxx
```diff
@@ -902,8 +902,20 @@ void vtkMRMLVolumeNode::GetRASBounds(double bounds[6])
...
+}
+
+//---------------------------------------------------------------------------
+void vtkMRMLVolumeNode::GetNodeBounds(double bounds[6])
+{
+  vtkMRMLVolumeNode::GetBoundsInternal(bounds, NULL, false);
+}
 
+//---------------------------------------------------------------------------
+void vtkMRMLVolumeNode::GetBoundsInternal(double bounds[6],
```
jcfr: GetBoundsInternal -> GetBounds ?

Libs/MRML/Core/vtkMRMLVolumeNode.h
```diff
@@ -139,9 +139,17 @@ class VTK_MRML_EXPORT vtkMRMLVolumeNode : public vtkMRMLDisplayableNode
 
   ///
   /// Get bounding box in global RAS the form (xmin,xmax, ymin,ymax, zmin,zmax).
+  /// This method retuns the bounds of the object with any transforms that may
```
jcfr: in global RAS the form  ->  in global RAS form

Libs/MRML/Core/vtkMRMLVolumeNode.h
```diff
@@ -139,9 +139,17 @@ class VTK_MRML_EXPORT vtkMRMLVolumeNode : public vtkMRMLDisplayableNode
...
+  /// be applied to it.
   /// \sa GetSliceBounds(), GetIJKToRASMatrix(), vtkMRMLSliceLogic::GetVolumeSliceBounds()
+  /// \sa GetNodeBounds()
   virtual void GetRASBounds(double bounds[6]);
 
+  /// Get bounding box in global RAS the form (xmin,xmax, ymin,ymax, zmin,zmax).
```
jcfr: global RAS the form -> global RAS form

Modules/Loadable/Annotations/MRML/vtkMRMLAnnotationROINode.h
```diff
@@ -153,8 +153,14 @@ class  VTK_SLICER_ANNOTATIONS_MODULE_MRML_EXPORT vtkMRMLAnnotationROINode
   /// Get bounding box in global RAS the form (xmin,xmax, ymin,ymax, zmin,zmax).
   /// The deafult implementation in the model node would not work correctly, as the polydata
   /// in this class is used for storing the centerpoint position and radius.
+  /// \sa GetNodeBounds()
   virtual void GetRASBounds(double bounds[6]);
 
+  /// Get bounding box in global RAS the form (xmin,xmax, ymin,ymax, zmin,zmax).
```
jcfr: global RAS the form  -> global RAS form

Modules/Loadable/Transforms/Logic/vtkSlicerTransformLogic.h
```diff
@@ -126,6 +130,31 @@ class VTK_SLICER_TRANSFORMS_MODULE_LOGIC_EXPORT vtkSlicerTransformLogic : public
   static bool GetTransformedPointSamplesAsVectorImage(vtkImageData* outputVectorImage, vtkMRMLTransformNode* inputTransformNode,
     vtkMatrix4x4* ijkToRAS, bool transformToWorld = true);
 
+  /// Return the list of node that are transformed by the given node.
```
jcfr: the list of node -> the list of nodes

Libs/MRML/Core/vtkMRMLSegmentationNode.cxx
lassoan: Could you rename this method to GetBounds()?
node->GetBounds() is more compact and readable than node->GetNodeBounds() and it's also consistent with the corresponding VTK method name.

Libs/MRML/Core/vtkMRMLTransformDisplayNode.cxx
```diff
@@ -88,6 +88,12 @@ vtkMRMLTransformDisplayNode::vtkMRMLTransformDisplayNode()
     this->ContourLevelsMm.push_back(level);
     }
 
+  this->EditorVisibility = 0;
+  this->EditorSliceIntersectionVisibility = 0;
+  this->TranslationEnabled = 1;
```
lassoan: It would be better to prefix these with "Editor":
this->EditorTranslationEnabled
this->EditorRotationEnabled
this->ScalingEnabled

Also, use bool type for Boolean variables and true/false to set their value (earlier it was a limitation of the TCL wrapper that it could not handle bool type, but this limitation does not exist anymore)

Libs/MRML/Core/vtkMRMLSegmentationNode.cxx
```diff
@@ -460,8 +460,16 @@ void vtkMRMLSegmentationNode::GetRASBounds(double bounds[6])
     {
     this->Segmentation->GetBounds(bounds);
     }
+
 }
 
+//---------------------------------------------------------------------------
+void vtkMRMLSegmentationNode::GetNodeBounds(double bounds[6])
+{
```
lassoan: Actually, GetRASBounds was incorrectly implemented for vtkMRMLSegmentationNode.

this->Segmentation->GetBounds(bounds) => this should be vtkMRMLSegmentationNode::GetBounds() result.

vtkMRMLSegmentationNode::GetRASBounds() can be computed as usual, by transforming GetBounds() result to RAS.

Libs/MRML/Core/vtkMRMLTransformableNode.cxx
```diff
@@ -141,6 +141,45 @@ void vtkMRMLTransformableNode::ProcessMRMLEvents ( vtkObject *caller,
...
+    this->InvokeCustomModifiedEvent(
+      vtkMRMLTransformableNode::TransformModifiedEvent, reference->GetReferencedNode());
+    }
+
+  vtkMRMLTransformNode* transformNode =
+    vtkMRMLTransformNode::SafeDownCast(reference->GetReferencedNode());
+  if (transformNode)
+    {
+    transformNode->InvokeEvent(
+      vtkMRMLTransformNode::TransformReferenceModifiedEvent, this);
```
lassoan: Have you looked into implementing this in a more generic way, at the vtkMRMLScene level? What I don't like about how it is implemented now that if we want to know if referencing nodes are changed for other node references then we would need to reimplement that from scratch. In contrast, if we implement notification about node reference changes at the scene level, then that mechanism can be used in the future for all kinds of node references.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
```diff
@@ -1286,22 +1286,50 @@ void vtkMRMLMarkupsNode::GetRASBounds(double bounds[6])
...
+void vtkMRMLMarkupsNode::GetNodeBounds(double bounds[6])
+{
+  vtkMath::UninitializeBounds(bounds);
+  int numberOfMarkups = this->GetNumberOfMarkups();
+  if (numberOfMarkups == 0)
+    {
+    return;
+    }
+  double markup_RAS[4] = { 0, 0, 0, 1 };
+  this->GetMarkupPoint(0, 0, markup_RAS);
```
lassoan: Markup fiducials only have one point per markup, but in the future one markup can consist of multiple points, so it's not enough to just check the first one.
If you implement GetBounds in MarkupFiducial node then there you may check only the first point (it'll be faster than the generic vktMRMLMarkupsNode GetBounds, because it doesn' have to check how many points are there in each markup and iterate through them).
So,
1. move this GetBounds() implementation into vtkMRMLMarkupFiducialNode, and
2. Either don't implement GetBounds for vtkMRMLMarkupsNode (leave it to child classes) or implement it properly (iterate through all points in all markups)

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
```diff
@@ -1286,22 +1286,50 @@ void vtkMRMLMarkupsNode::GetRASBounds(double bounds[6])
...
+
+  bounds[0] = bounds[1] = markup_RAS[0];
+  bounds[2] = bounds[3] = markup_RAS[1];
+  bounds[4] = bounds[5] = markup_RAS[2];
+  for (int i = 1; i < numberOfMarkups; i++)
+    {
+    this->GetMarkupPoint(i, 0, markup_RAS);
+    if (markup_RAS[0] < bounds[0]) { bounds[0] = markup_RAS[0]; }
```
lassoan: Use http://www.vtk.org/doc/nightly/html/classvtkBoundingBox.html . It would make the code much more compact (instead of having 6 lines for adding a point to a bound you would call bound->AddPoint(markup_RAS))

Modules/Loadable/Transforms/MRMLDM/vtkMRMLTransformsDisplayableManager2D.cxx
```diff
@@ -398,10 +398,6 @@ void vtkMRMLTransformsDisplayableManager2D::vtkInternal::UpdateDisplayNodePipeli
 void vtkMRMLTransformsDisplayableManager2D::vtkInternal::AddObservations(vtkMRMLTransformNode* node)
 {
   vtkEventBroker* broker = vtkEventBroker::GetInstance();
-  if (!broker->GetObservationExist(node, vtkMRMLDisplayableNode::TransformModifiedEvent, this->External, this->External->GetMRMLNodesCallbackCommand() ))
```
lassoan: Why was this removed?

Modules/Loadable/Transforms/MRMLDM/vtkMRMLTransformsDisplayableManager3D.cxx
```diff
@@ -47,6 +53,9 @@
 #include <vtkRenderWindowInteractor.h>
 #include <vtkRenderer.h>
 #include <vtkSmartPointer.h>
+#include <vtkTransform.h>
+
+const double DEFAULT_SCALE[3] = {100.0, 100.0, 100.0};
```
lassoan: Usually scale is in the 0..1 range. Either rename to DEFAULT_SCALE_PERCENT (if the range is 0..100) or change the range to 0..1.

Modules/Loadable/Transforms/MRMLDM/vtkMRMLTransformsDisplayableManager3D.cxx
```diff
@@ -340,10 +350,6 @@ void vtkMRMLTransformsDisplayableManager3D::vtkInternal::UpdateDisplayNodePipeli
 void vtkMRMLTransformsDisplayableManager3D::vtkInternal::AddObservations(vtkMRMLTransformNode* node)
 {
   vtkEventBroker* broker = vtkEventBroker::GetInstance();
-  if (!broker->GetObservationExist(node, vtkMRMLDisplayableNode::TransformModifiedEvent, this->External, this->External->GetMRMLNodesCallbackCommand() ))
```
lassoan: why is the change?

Modules/Loadable/Transforms/MRMLDM/vtkMRMLTransformsDisplayableManager3D.cxx
lassoan: Is this DEFAULT_SCALE used?

Modules/Loadable/Transforms/Widgets/qMRMLTransformDisplayNodeWidget.cxx
```diff
@@ -552,6 +559,50 @@ void qMRMLTransformDisplayNodeWidget::setContourVisualizationMode(bool activate)
 }
 
 //-----------------------------------------------------------------------------
+void qMRMLTransformDisplayNodeWidget::setInteractive(bool enabled)
```
lassoan: Why not call set EditorVisibility? It makes it hard to follow data and process flow if you suddenly change name of a concept for no good reason.
The best would be to choose a name that is the same everywhere (displayable manager, MRML nodes, preferable even match or be similar to what is displayed to the user).

Libs/MRML/Core/vtkMRMLTransformDisplayNode.cxx
vovythevov: Make sense.

Libs/MRML/Core/vtkMRMLTransformableNode.cxx
vovythevov: That's a good point, yes, that would be great !
Isn't this something that should be tackled in another issue/PR though ? It seems like it would be better to do this as a separate branch.

Libs/MRML/Core/vtkMRMLSegmentationNode.cxx
vovythevov: Sure.

Note that in my opinion it's confusing to have another method named GetRASBounds. It implies that this method does NOT return the bounds in the RAS coordinate which it does (albeit not transformed by any vtkMRMLTransformNode applied to the node). That was why I overloaded the GetRASBounds originally.

Libs/MRML/Core/vtkMRMLVolumeNode.cxx
vovythevov: If I rename GetNodeBounds to GetBounds, I'll keep this to GetBoundsInternal.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
vovythevov: I agree with you, it seems that the method wasn't implemented to cover all the cases originally.
I'll improve it.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
vovythevov: Make sense. I just wanted to limit my footprint :) (See also comment https://github.com/Slicer/Slicer/pull/663/files#r99369997)

Modules/Loadable/Transforms/MRMLDM/vtkMRMLTransformsDisplayableManager2D.cxx
vovythevov: This is a leftover from when I was using the same displayable manager than the visualization modes.
As to why I removed this event, It looks to me as it's just a duplicate of the event vtkMRMLTransformableNode::TransformModifiedEvent defined a few lines lower, so I removed it.

vovythevov: This is a leftover from when I was using the same displayable manager than the visualization modes.
As to why I removed this event, It looks to me as it's just a duplicate of the event vtkMRMLTransformableNode::TransformModifiedEvent defined a few lines lower, so I removed it.

Modules/Loadable/Transforms/MRMLDM/vtkMRMLTransformsDisplayableManager3D.cxx
vovythevov: See https://github.com/Slicer/Slicer/pull/663/files#r99370938

Libs/MRML/Core/vtkMRMLTransformableNode.cxx
lassoan: Yes, doing it in a separate pull request could make sense. Till then, remove the automatic observation of transform chain updates and determine the boundary when the editing widget is turned on (and maybe when you enter the Transforms module and when a transforms is applied in the transforms module). 

We could even keep this behavior in the long term: update bounds when you show the widget (and maybe add a button that would update the bounds so that the user does not have to hide/show the widget to update bounds).

Modules/Loadable/Transforms/Widgets/qMRMLTransformDisplayNodeWidget.cxx
vovythevov: Good catch !

Modules/Loadable/Transforms/MRMLDM/vtkMRMLTransformsDisplayableManager3D.cxx
vovythevov: Good catch ! It's a leftover. I need a better clean up of this displayable manager.

Libs/MRML/Core/vtkMRMLModelNode.cxx
vovythevov: Good catch !

Libs/MRML/Core/vtkMRMLSegmentationNode.cxx
vovythevov: I'll fix the method then.

Libs/MRML/Core/vtkMRMLSegmentationNode.cxx
lassoan: In general, nodes return information without applying the parent transform, see for example Volume node's GetOrigin() or Markup fiducial node's GetNthFiducialPosition - they all provide position in the node's coordinate system, not in RAS. Having _RAS_ in the name of GetRASBounds indicates that it's an exception. Probably not the best name, but it's not too bad and it would be difficult to change, so for now I would not touch it.

Modules/Loadable/Transforms/MRMLDM/vtkMRMLTransformsDisplayableManager2D.cxx
lassoan: I would recommend not to complicate this pull request with any changes that are not necessary and have a slight chance of breaking something. It would be better to submit a separate pull request for this later.

Libs/MRML/Core/vtkMRMLTransformableNode.cxx
vovythevov: +1 Good plan !
Added an issue to track this here: http://na-mic.org/Mantis/view.php?id=4338

Modules/Loadable/Transforms/MRMLDM/vtkMRMLTransformsDisplayableManager2D.cxx
vovythevov: Make sense, I will add separately then.

vovythevov:
@jcfr, @lassoan: New update:
 - No automatic bounds update of the widget, instead use an event fired on the transform display node.
 - Update Editor* variables (name update and type change to boolean).
 - Cleaned up displayable manager.

jcfr:
Nice. Before we move forward with integration, would it be possible to consolidate/squash in fewer commits ?

vovythevov:
@jcfr: Consolidated most commits into two commits; all the logic into one and the UI change in the other.

jcfr:
The following warning is reported at startup time:

```
Bad plane coordinate system


Algorithm vtkPlaneSource(0x4004c40) returned failure for request: vtkInformation (0x4007700)
  Debug: Off
  Modified Time: 89819
  Reference Count: 1
  Registered Events: (none)
  Request: REQUEST_DATA
  FORWARD_DIRECTION: 0
  ALGORITHM_AFTER_FORWARD: 1
  FROM_OUTPUT_PORT: 0

```

lassoan:
This error is unrelated. It's due to model node changes by Alexis (introduced in some of the late patches). I'll ask Alexis to have a look.

jcfr:
Integrated in [r25695](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25695) to [r25699](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25699)

jcfr:
> This error is unrelated. It's due to model node changes by Alexis (introduced in some of the late patches). I'll ask Alexis to have a look.

@agirault When you have a chance, could you look at the warning reported when Slicer starts ? 

```
Bad plane coordinate system


Algorithm vtkPlaneSource(0x4004c40) returned failure for request: vtkInformation (0x4007700)
  Debug: Off
  Modified Time: 89819
  Reference Count: 1
  Registered Events: (none)
  Request: REQUEST_DATA
  FORWARD_DIRECTION: 0
  ALGORITHM_AFTER_FORWARD: 1
  FROM_OUTPUT_PORT: 0

```

agirault:
@lassoan this looks like it was introduced by the call to the mesh connection `Update()` on the in `getMesh()` that you recommended adding. I don't have time to look at it this week, I can let you know if I do next week.

lassoan:
Yes, it's caused by that Update(). I'll investigate and let you know what I find.

jcfr:
Thanks @lassoan

PR #667, opened 2017-02-07
EHN: OpenIGTLinkIF is capable of processing RGB image

leochan2009:
Refer to Issue 4339 from bug tracker


SuperBuild/External_OpenIGTLink.cmake
```diff
@@ -40,6 +40,7 @@ if(NOT DEFINED OpenIGTLink_DIR AND NOT ${CMAKE_PROJECT_NAME}_USE_SYSTEM_${proj})
       -DBUILD_TESTING:BOOL=OFF
       -DBUILD_SHARED_LIBS:BOOL=ON
       -DOpenIGTLink_PROTOCOL_VERSION_2:BOOL=ON
+      -DOpenIGTLink_PROTOCOL_VERSION_3:BOOL=ON
```
lassoan: Is it intentional that both OpenIGTLink_PROTOCOL_VERSION_2 and OpenIGTLink_PROTOCOL_VERSION_3 are on?

leochan2009: Hi Andras,

i intentionally turn the OpenIGTLink_PROTOCOL_VERSION_3 on, actually the the OpenIGTLink_PROTOCOL_VERSION_2 could be turned off. The Header version is still by default version1. So even with the protocol version3 turned on, all the feature of openigtlink will be the same as version2, only if the user set the protocol header to version2, there will be some difference.  As the same time, the command message from version3 will be available.

lassoan:
Thank you. Committed r25700.

PR #669, opened 2017-02-10
BUG: fix some DICOMDiffusionPlugin issues

ihnorton:
- 0.75 confidence is too high: non-DWI GE scans can have these tags set, no way to discriminate.
- set `loadable.selected` so that matches will actually be loadable (per @lassoan)
- check the first three volumes because some (Siemens) scans won't have some tags for b0 volume.

### DIFF
--- a/Modules/Scripted/DICOMPlugins/DICOMDiffusionVolumePlugin.py
+++ b/Modules/Scripted/DICOMPlugins/DICOMDiffusionVolumePlugin.py
@@ -88,9 +88,12 @@ class DICOMDiffusionVolumePluginClass(DICOMPlugin):
     for vendor in self.diffusionTags:
       matchesVendor = True
       for tag in self.diffusionTags[vendor]:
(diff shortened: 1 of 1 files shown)


fedorov:
are we duplication work between this plugin and https://github.com/fedorov/MultiVolumeImporter/blob/master/MultiVolumeImporterPlugin.py#L34-L50 ?

I didn't know about this plugin until today!

fedorov:
I need to run now, but will look a bit closer later.

pieper:
LGTM - thanks!

lassoan:
LGTM, too. Thank you.

fedorov:
LGTM too, sorry, I mis-interpreted the purpose of the code I was looking at.

jcfr:
Thanks for the fix :+1: 

Integrated in [r25704](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25704)

PR #671, opened 2017-02-16
Dataprobe.py - Check for attribute _text_ on parent

joshy:
We are using the DataProbeWidget on a view on our own. Without the proposed change the processEvents throws errors because there exists not _text_ attribute on the parent. The change checks for the attribute _text_ and executes the code only when the attribute exists. 
It will not break any existing code.

Thanks for considering this change.


jcfr:
Thanks for PR :+1: 

(And apologize for the failing CircleCI, it is on our side and we are working on addressing it)

jcfr:
Et voila. Integrated as http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25709

joshy:
@jcfr Cool! Thanks for being super fast!

PR #675, opened 2017-03-03
ENH: add "build-only" key for extension dependency resolution.

ihnorton:
Extension with the configuration key "build_only" will be ignored by dependency resolution.
This allows us to hide the warning for a compile time-only extension which does not create
any installable build products, such as Eigen.

Fixes:
http://www.na-mic.org/Bug/view.php?id=3704

### DIFF
--- a/Base/QTCore/qSlicerExtensionsManagerModel.cxx
+++ b/Base/QTCore/qSlicerExtensionsManagerModel.cxx
@@ -116,7 +116,8 @@ public:
     ScreenshotsColumn,
     EnabledColumn,
     ArchiveNameColumn,
-    MD5Column
+    MD5Column,
+    BuildOnlyColumn
     };
 
   enum ItemDataRole{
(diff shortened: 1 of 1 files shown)


pieper:
LGTM

ihnorton:
I updated to add the key to the item model from which `::extensionMetadata(...)` is constructed, otherwise I guess it would be ignored. I'm not sure how to test this locally without setting up a full extension server.

jcfr:
:+1: Thanks for the patch

The following test could be updated to test this feature: https://github.com/Slicer/Slicer/blob/afeba986ebb8b93367d0fb4c0ed9eb04a0e78c66/Extensions/CMake/Testing/CMakeLists.txt#L65-L79

ihnorton:
@jcfr I'm not really following how those tests would apply here.

jcfr:
I will try to clarify:

Since the tests I pointed in my first comment both CDash and extension servers are mocked.  You are right, there are basically no way to test the installation from Slicer.

That said, since you extended the extension description file with a new field `build_only`, the [following tests](https://github.com/Slicer/Slicer/blob/afeba986ebb8b93367d0fb4c0ed9eb04a0e78c66/Extensions/CMake/Testing/CMakeLists.txt#L18-L32) should probably be updated:
* `slicer_extension_description_spec_defaults_test`
* `slicer_extension_metadata_set_defaults_test`
* `slicer_extract_extension_description_test`
* `slicer_generate_extension_description_test`


All of that said, what do you think of instead if we instead introduce `build_depends` ?

ihnorton:
> All of that said, what do you think of instead if we instead introduce build_depends ?

I have no preference as long as the Eigen install failure message goes away. The message is annoying and can be confusing to users -- everything should still work fine, but that's not obvious. We now install UKFTractography by default, which depends on Eigen, so the message pops up for every install.

The bug on mantis has been open for a while, and this seemed like the simplest solution. However if someone is going to work on adding `build_depends`, that's great.

lassoan:
I strongly prefer build_depends name. Negative (DISABLE_...) and exclusive (..._ONLY) names lead to contradictions.

PR #677, opened 2017-03-10
COMP: fix cmake extension build issue

ihnorton:
work around https://gitlab.kitware.com/cmake/cmake/issues/15448
to fix https://www.na-mic.org/Bug/view.php?id=4311

### DIFF
--- a/Extensions/CMake/CMakeLists.txt
+++ b/Extensions/CMake/CMakeLists.txt
@@ -76,6 +76,10 @@ include(ExternalProject)
 find_package(Slicer REQUIRED)
 find_package(Git REQUIRED)
 find_package(Subversion REQUIRED)
+# force in to CACHE to work around https://gitlab.kitware.com/cmake/cmake/issues/15448
+if (DEFINED Subversion_SVN_EXECUTABLE)
+  set(Subversion_SVN_EXECUTABLE ${Subversion_SVN_EXECUTABLE} CACHE FILEPATH "SVN command line client")
+endif()
 
 #-----------------------------------------------------------------------------
 # Options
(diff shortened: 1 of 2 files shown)


Extensions/CMake/CMakeLists.txt
```diff
@@ -76,6 +76,10 @@ include(ExternalProject)
 find_package(Slicer REQUIRED)
 find_package(Git REQUIRED)
 find_package(Subversion REQUIRED)
+# force in to CACHE to work around https://gitlab.kitware.com/cmake/cmake/issues/15448
+if (DEFINED Subversion_SVN_EXECUTABLE)
+  set(Subversion_SVN_EXECUTABLE ${Subversion_SVN_EXECUTABLE} CACHE FILEPATH "SVN command line client")
+endif()
```
pieper: Does it matter that GIT_EXECUTABLE is marked as advanced but the svn executable is not?

ihnorton: We use upstream FindSubversion, [which does `mark_as_advanced`](https://github.com/Kitware/CMake/blob/29ec8d65f2815bbe4b47c59c5483a9934b92898b/Modules/FindSubversion.cmake#L68).

The main goal is to put these variables in the cache so that `mark_as_advanced` doesn't overwrite them upon repeated calls. I could put the `set(...)` calls in the same place as SVN if you prefer, but we have our own FindGit (which is called first by the extension build system), so it seemed better to change there. `ExternalProject_Add` ends up using the built-in version from CMake, but as long as the variable is already set in the CACHE it is not overwritten.

pieper: Yes, makes sense just doublechecking.

LGTM - thanks for figuring this all out.

jcfr:
Thanks. The change makes sense.  I will revert https://github.com/Slicer/Slicer/commit/b25d2f40 and apply a similar change for GIT_EXECUTABLE.

jcfr:
Nevermind, you already took care of it in your commit.

jcfr:
Integrated in [r25781](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25781)

PR #679, opened 2017-03-13
Fix sh warnings

jcfr:
This PR fixes the following warnings:

```
/home/jcfr/Projects/Slicer/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx:849:42: warning: unused variable 'mostSuitablePluginByConfidenceNumbers' [-Wunused-variable]
  qSlicerSubjectHierarchyAbstractPlugin* mostSuitablePluginByConfidenceNumbers =
                                         ^
/home/jcfr/Projects/Slicer/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginLogic.cxx:296:32: warning: unused variable 'shNode' [-Wunused-variable]
  vtkMRMLSubjectHierarchyNode* shNode = vtkMRMLSubjectHierarchyNode::GetSubjectHierarchyNode(scene);
                               ^
/home/jcfr/Projects/Slicer/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginLogic.cxx:355:102: warning: implicit conversion of NULL constant to 'vtkIdType' (aka 'long long')
      [-Wnull-conversion]
      qSlicerSubjectHierarchyPluginHandler::instance()->pluginsForAddingNodeToSubjectHierarchy(node, NULL);
                                                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~       ^~~~
                                                                                                     0

```

@cpinter Are these benign warnings or do they reveal some other issues ? Thanks for checking

### DIFF
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
@@ -845,10 +845,6 @@ void qMRMLSubjectHierarchyTreeView::selectPluginForCurrentItem()
     return;
     }
 
-  // Check if the user is setting the plugin that would otherwise be chosen automatically
-  qSlicerSubjectHierarchyAbstractPlugin* mostSuitablePluginByConfidenceNumbers =
-    qSlicerSubjectHierarchyPluginHandler::instance()->findOwnerPluginForSubjectHierarchyItem(currentItemID);
-
   // Set new owner plugin
   d->SubjectHierarchyNode->SetItemOwnerPluginName(currentItemID, selectedPluginName.toLatin1().constData());
 }
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginLogic.cxx
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginLogic.cxx
@@ -345,7 +345,8 @@ void qSlicerSubjectHierarchyPluginLogic::addSupportedDataNodesToSubjectHierarchy
 
     // If there is a plugin that can add the data node to subject hierarchy, then add
     QList<qSlicerSubjectHierarchyAbstractPlugin*> foundPlugins =
-      qSlicerSubjectHierarchyPluginHandler::instance()->pluginsForAddingNodeToSubjectHierarchy(node, NULL);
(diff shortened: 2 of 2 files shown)


cpinter:
It's great you asked! qSlicerSubjectHierarchyPluginLogic.cxx:296 is actually needed, see the 5-line comment above the line, where I explain why it is needed.

Please keep the call (and the comment!), but you can scrap declaring the variable, like this:
vtkMRMLSubjectHierarchyNode::GetSubjectHierarchyNode(scene);

I can commit it if you'd lke.

jcfr:
Thanks for the review.

> Please keep the call (and the comment!),

Makes sense. Just wanted to check.

> I can commit it if you'd lke.

I will integrate.

cpinter:
Thank you!

cpinter:
I think I owe a more detailed explanation.
The new subject hierarchy node is special in the sense that it is technically not a singleton, but in practice it is considered a singleton. The reason for considering it a singleton, is that because there is one main scene in Slicer, there should be one subject hierarchy. At the same time it cannot be a singleton node, because when scene import happens, then two subject hierarchy nodes need to co-exist in the scene for a moment, because both contain the hierarchies of their corresponding scene (one for the main Slicer scene, the other is for the imported one, and both are legit). These nodes are merged together at the end of scene import (hence the call in the function that yielded the warning), but also it is checked and merged if needed at the next occasion of accessing the hierarchy by the static GetSubjectHierarchyNode(scene) call. This is why everywhere in the code this static function is called. Not just because it's simpler than getting the first SH node from the scene, but because of this consolidation step.
I hope it is clearer now. I also hope it is not problematic for you, because it would be extremely hard to change :) If you have comments (maybe I misunderstand the "singletonness" of a MRML node), please let me know.

lassoan:
@cpinter If you feel the need for explanation then probably it's a good idea to put this explanation (or a shortened version) into the code as well.

cpinter:
It's there, see SH node header. Also that's why I added 5 lines of comments above the call. I wanted to explain it to Jc personally, because apparently he didn't see those comments, and if somebody, then he should be aware of this.

jcfr:
> It's there, see SH node header. Also that's why I added 5 lines of comments above the call. I wanted to explain it to Jc personally, because apparently he didn't see those comments, and if somebody, then he should be aware of this.

I saw the comment ... but wasn't sure if it was legacy code or required by the SH 2.0. 

@cpinter Thanks for your explanation, it makes sense.

Warning fixed in [r25830](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25830)

PR #682, opened 2017-03-16
Fix/clean-up tensor slice glyph issues from model pipeline rework

ihnorton:
Second commit allows wildcard search in the Data module, which I find useful when looking up nodes because it allows typing a small part of the name.

(cc @ljod)

### DIFF
--- a/Libs/MRML/Core/vtkMRMLDiffusionTensorDisplayPropertiesNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLDiffusionTensorDisplayPropertiesNode.cxx
@@ -36,7 +36,7 @@ vtkMRMLDiffusionTensorDisplayPropertiesNode::vtkMRMLDiffusionTensorDisplayProper
 
   // Default display is FA (often used) and line glyphs (quickest to render)
   this->ScalarInvariant = this->FractionalAnisotropy;
-  this->GlyphGeometry = this->Ellipsoids;
+  this->GlyphGeometry = this->Lines;
   this->ColorGlyphBy = this->FractionalAnisotropy;
 
   // Glyph general parameters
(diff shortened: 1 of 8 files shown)


ihnorton:
(I am planning to move these DT/DWI volume nodes and displayablemanager to SlicerDMRI soon. There is some entanglement with the slice logic, i/o, and volumes module, but I think I can work around that now, except for the Volumes integration. Probably we will create a "Diffusion Volumes" module that reuses most of the existing widgets)

pieper:
LGTM

jcfr:
I am updating the CircleCI image to that we can confirm build works.

In the mean time, could you split the commit a6ba53df2faced6fa66b7 into cleanup and fix .. so that we can more easily understand what the fix is. 

Otherwise, could you explain in the commit message of b82908095abf which problem is being fixed ?

Thanks

jcfr:
@ihnorton  Feel free to restart the CircleCI build .. up-to-date images have been pushed on dockerhub

ihnorton:
Commits updated. CircleCI failed once to due timeout, then again just now due to either OOM killer or superbuild update warning (or both).

ihnorton:
committed in:
http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25851
http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25850
http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25849
http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25848

jcfr:
Thanks for integrating.

> CircleCI failure is unrelated as far as I can tell. Committed in:

Indeed, this was due to the recent changes to "External_*" files.

PR #683, opened 2017-03-17
BUG: Fixed scene views with subject hierarchy 2.0

cpinter:


cpinter:
I'd like to confirm with the core developers ( @jcfr @pieper @lassoan and others if you have time ) that my changes in vtkMRMLScene.cxx and vtkMRMLSegmentationNode.cxx are fine. Let me explain why I did those:
* vtkMRMLScene.cxx:1140 and 1175: the ReferencedIDChanges map contained an invalid entry in every case, namely it associated the last created node's ID with empty string. I don't think it makes sense, as it is not a node ID change
* vtkMRMLScene.cxx:1353: In the case of loading a volume, the test scene created by vtkSlicerVolumesLogic::AddArchetypeVolume assigned ID with number 1 to the volume. Then, when adding it to the main MRML scene, the old ID with number 1 was found, and the ID change was stored always from 1 to the number of the last loaded volume. This I think is also an invalid ID change, as no ID was changed in the scene that stored this change. So if the ID is cleared when removing the node from the test scene, the initial ID is not found, thus this invalid change is not stored
* vtkMRMLSegmentationNode.cxx: This is basically reverting commit rev25625 that introduced display node copying and creation in the data node's copy. In the case of restoring the scene view, it caused a problem like this: if the segmentation node was before its display node in the snapshot scene, a new display node was created when copying the segmentation node, and so after restoring, two display nodes were present instead of one. So the segmentation was not visible. I checked other data nodes and none of the ones I looked at copy display properties in the copy function of the data node. I suggest taking care of this at the higher level, where the data node's copy is called (for example as it is done in the clone node SH plugin)

cpinter:
After these changes that prevent invalid reference changes to be stored (that are applied when resolving the unresolved items that are saved like that in the snapshot SH - need to do that in order to keep uniqueness of the SH item IDs, and to be prepared for real data node ID changes), and preventing the second segmentation display node from being created, scene view restoring seems to work well.

lassoan:
Thank you for the fixes.

Segmentation is particularly sensitive to sharing display nodes, because it stores display information not in simple variables but in a dynamically allocated structure. For example, if a segment is removed from the segmentation node then it is removed from the display node, too. Therefore even transient changes in the data node (remove segment then add back the segment) may cause display information loss.

I couldn't find any place where the usual Copy (without copying the display node) could cause any problems, so let's do the right thing (integrate your changes that remove display node copy) and if any issue comes up then fix it at a higher level (clone display node immediately after cloning segmentation node).

Libs/MRML/Core/vtkMRMLScene.cxx
```diff
@@ -1137,7 +1137,7 @@ vtkMRMLNode*  vtkMRMLScene::AddNodeNoNotify(vtkMRMLNode *n)
       // their references.
       std::string newId(sn->GetID());
       std::string oldId(n->GetID() ? n->GetID() : sn->GetID());
-      if (oldId != newId)
+      if (oldId != newId && !oldId.empty())
```
lassoan: Can oldId be empty? (by default node->ID is initialized to NULL and not "")
Shouldn't we check for GetID()==NULL instead?

Libs/MRML/Core/vtkMRMLScene.cxx
```diff
@@ -1350,6 +1350,7 @@ void vtkMRMLScene::RemoveNode(vtkMRMLNode *n)
     this->RemoveReferencesToNode(n);
     }
 
+  n->SetID(NULL);
```
lassoan: It is difficult to ensure that there are no side effects of this.
Would it be possible to fix this issue in volume loading by erasing the node ID after the volume node is removed from the test scene?

Libs/MRML/Core/vtkMRMLSubjectHierarchyNode.cxx
```diff
@@ -3008,14 +3069,28 @@ vtkMRMLSubjectHierarchyNode* vtkMRMLSubjectHierarchyNode::GetSubjectHierarchyNod
...
-        //      (e.g. by removing all the subject hierarchy nodes from the scene and when adding a
-        //      new one, calling addSupportedDataNodesToSubjectHierarchy in the plugin logic)
+        // (e.g. by removing all the subject hierarchy nodes from the scene and when adding a
+        // new one, calling addSupportedDataNodesToSubjectHierarchy in the plugin logic)
         vtkErrorWithObjectMacro(scene, "vtkMRMLSubjectHierarchyNode::GetSubjectHierarchyNode: Failed to merge subject hierarchy nodes");
         return firstShNode;
         }
+
+      // Trigger re-building the trees in the views
+      firstShNode->Internal->EventsDisabled = false;
```
lassoan: It is always safer to store the old value of EventsDisabled and restore to the old value instead of hardcoding to false (the behavior is incorrect if a method that temporarily disables events calls another one that temporarily disables events, because this method re-enables events).
Name of the flag could be more specific, too, such as ItemUpdateEventsDisabled.

Libs/MRML/Core/vtkMRMLScene.cxx
cpinter: I don't see the difference. It is empty in both cases, thus invalid

Libs/MRML/Core/vtkMRMLScene.cxx
cpinter: That's what I wanted to do, but SetID is a private method. Any suggestions?

My line of thinking was that if a node is removed from the scene, then its ID - which is scene-specific - will be unused, and then re-assigned anyway when it is added to another scene. And without a scene, the ID is unused. So I think there should be no side-effects, but I cannot be 100% sure...

Libs/MRML/Core/vtkMRMLScene.cxx
lassoan: oldId will never be empty. If sn->GetID() is NULL then these lines will crash:
std::string newId(sn->GetID()); 
std::string oldId(n->GetID() ? n->GetID() : sn->GetID());

cpinter: Oh this is the other instance, not the line that actually caused the problem, I get it now! Yes, let's check if it's NULL

Libs/MRML/Core/vtkMRMLSubjectHierarchyNode.cxx
cpinter: (Note to myself: there is another of the same at line 1800)

Libs/MRML/Core/vtkMRMLScene.cxx
lassoan: You are right, it should be the behavior, we just have to avoid potential regressions.
I've checked all RemoveNode calls in the Slicer source code and only found vtkMRMLScene::Undo() and vtkMRMLScene::Redo() relying on the node IDs to be preserved after RemoveNode. Fortunately, vtkMRMLScene is a friend of vtkMRMLNode, so we can save the node ID before RemoveNode and restore node ID after the call.

cpinter: Thank you, this is very useful! I'll amend the commit accordingly.

cpinter:
@lassoan I amended the commit. Can you please check the changes in vtkMRMLScene::Undo and Redo? Thanks!

pieper:
The changes look good to me.

For what it's worth the MRMLScene level undo/redo isn't exposed in the GUI and we should probably mark it deprecated.  But trying to keep the behavior consistent, as you have done, is still a good idea.

http://na-mic.org/Bug/view.php?id=1430

jcfr:
@cpinter As a side note, we just released that commit c30b08c658721db11f039e7ba09b7df47a30534a introduced a regression preventing regular label map from imported as segmentation.

Cc: @jbvimort

cpinter:
@jcfr I don't believe there is anything in this pull request that would fix labelmap import to segmentations. I will test it myself though.
Currently I'm looking into the automated tests. Some tests failed, but when I tried to track it down I found that they failed in release but not in debug. Then I decided to rebuild everything from scratch, had some difficulties, but now they seem to be finishing fine. When it's done, and resolved the question of the failing tests, I'll look into the buggy importing.

jcfr:
Makes sense. Thanks @cpinter

jcfr:
Will also try to find out the culprit on our side, a git bisect led us to the commit I referenced earlier.

cpinter:
It's completely possible I broke it in the process. Importing labelmap to segmentation is fairly straightforward so should be easy to fix

cpinter:
@jcfr @pieper I've identified a few problems around the node ID clearing at the end of RemoveNode.
1. First, vtkMRMLNodeTest1 fails, because when clearing the scene, it expects 0 events fo be fired, but actually two events are fired (IDChangedEvent + Modified), because SetID is called. It's not surprising, and if there are no further problems with this, then I'd just change this expected value to 2
2. There are crashes in the tests, usually because there is an strcmp using the node->GetID() char*, and if it's NULL, then there is a crash. I think adding those checks for the ID would be an acceptable fix for these

There may be other things I bump into, but what do you think about these two so far? Thanks!

pieper:
@cpinter I think both your points are fair.

Regarding the second, the test (and other code) should be robust for a null ID string since that's the default for a newly constructed node [1].

[1] https://github.com/pieper/Slicer/blob/master/Libs/MRML/Core/vtkMRMLNode.cxx#L35

cpinter:
The Slicer core tests are all fine now.

Unfortunately it seems that many EMSegment tests rely on node IDs staying the same after removing from the scene, because they use sub-scenes in a specific way.
@jcfr I see you have made many comments to EMSegment, so if you have a suggestion I'm all ears :)

cpinter:
My best idea so far is to add an argument to RemoveNode to be able to keep the ID after removal, the default value of which would be to remove the ID. It could be called like that from EMSegment.

I think it would really make sense to clear the ID (which is specific to a certain scene) from a node when removing from the scene. The AddArchetypeVolume bug clearly illustrates what kind of problems it can cause (i.e. using the ID of a node assigned to it in a mini test-scene when adding it to the main scene, and thus making the main scene believe that the ID needs to be saved as if another scene was imported).

The other option would be to somehow indicate to the scene that a node ID is to be saved in such a way. As far as I know it is only useful when importing a scene or restoring a scene view, but if this changed ID map is invalid, then it can cause problems.

jcfr:
This discussion makes me think about the topic I started some time ago while working on [4081](http://na-mic.org/Mantis/view.php?id=4081) and [3462](http://na-mic.org/Mantis/view.php?id=3462)

See https://github.com/jcfr/Slicer/compare/add-vtkEventSpy...jcfr:4081-ensure-reference-are-up-to-date-on-import#diff-2680a67c76c231e09095db0f63b8bb69

cpinter:
I actually implemented the "other solution" from my previous comment, as it seemed much more stable and less prone to regressions and backward compatibility issues than clearing the node ID. See the latest commit in this branch. I ran every test, and no tests fail that don't fail on the dashboard (except for py_SlicerSslTest).

If you agree that the ReferencedIDChanges is only supposed to keep track of changed IDs for scene import and scene view restore, then I'll integrate. Also everything seems to be fine after some testing.

jcfr:
Before moving forward with the integration, I would like to see the changes to the MRMLScene done in a dedicated topic with test(s) added to ``vtkMRMLSceneTestX.cxx`` (or similar core tests).

cpinter:
I added the test cases in vtkMRMLSceneTest1, and integrated in rev25853.

jcfr:
Thanks for adding the test :+1: 

((Nitpick: I would have preferred to see a dedicated commit adding the test and fixing the issue with the scene first, and then the integration of the subject hierarchy changes.
For example, it would have been nice to have the commit `ENH: Added test for changes in scene related to ReferencedIDChanges`  integrated on its own. Doing make this key changes easier to grasp.))

cpinter:
I think we misunderstood each other. Your request to see the test before integration made me believe you wanted the test as part of the commit. Otherwise I could have integrated when I was done with the changes, and added the test after that separately (and not wait with the integration).
I hope it's not a big problem though!

jcfr:
No problem at all.

To clarify for later, let's make sure to fix MRML infrastructure core issue independently of larger changes.

jcfr:
Since associated changes have been integrated. Closing this PR.

See [r25853](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25853)

cpinter:
I see your point now! Not just the test, but also the actual MRML fix should have been separated. Makes sense, I'll do that from now on.
In this case the bug only caused issues with the SH changes, so I guess that's why I committed together.

PR #684, opened 2017-03-20
Allow customization external projects

jcfr:


jcfr:
Note that existing build tree have to be reconfigured using: 

```
cmake -USlicer_VTKv7_GIT_REPOSITORY -USlicer_VTKv7_GIT_TAG .
```

jcfr:
Integrated in [r25846](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25846)

PR #689, opened 2017-03-22
STYLE: "become"

tdiprima:
Please consider this small fix.  Changed "be come" to "become".
Thank you :)

### DIFF
--- a/Extensions/Testing/EditorExtensionTemplate/EditorEffectTemplate/EditorEffectTemplate.py
+++ b/Extensions/Testing/EditorExtensionTemplate/EditorEffectTemplate/EditorEffectTemplate.py
@@ -9,7 +9,7 @@ from EditorLib import LabelEffect
 #
 # The Editor Extension itself.
 #
-# This needs to define the hooks to be come an editor effect.
+# This needs to define the hooks to become an editor effect.
 #
 
 #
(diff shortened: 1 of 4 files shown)


jcfr:
Thanks for the fix :+1:  

Integrated in [r25860](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25860)

lassoan:
Note that this is template is for the old editor module. Template for the current segment editor is available here:
https://github.com/Slicer/Slicer/tree/master/Utilities/Templates/Modules/ScriptedSegmentEditorEffect
The easiest way to use it is to run the ExtensionWizard module in Slicer to generate an extension and within that add a 'scriptedsegmenteditoreffect' module.

lassoan:
@jcfr Duplication between [1] and [2] is confusing and difficult to maintain. Should we change extension testing to use [1] and delete [2]; or change extension wizard to use [2] and delete [1]?

[1] https://github.com/Slicer/Slicer/tree/master/Utilities/Templates/Modules
[2] https://github.com/Slicer/Slicer/tree/master/Extensions/Testing

jcfr:
Templates in https://github.com/Slicer/Slicer/tree/master/Extensions/Testing are full example integrating multiple modules that are automatically generated using the Slicer templates generator targets.

See here for more details: https://github.com/Slicer/Slicer/pull/493#issuecomment-219512388

PR #691, opened 2017-03-27
ENH: Display thumbnails in SampleData module

lassoan:
It makes the module nicer and the button easier to hit.
sampledata-thumbnails


lassoan:
@ihnorton - Probably you could make much nicer thumbnails for the DTI/DWI volumes. You can either attach images here or replace them after the PR is merged.

lassoan:
Thumbnails were generate using ScreenCapture module (single step + capture all views option). Note that view controllers are hidden when capture all views option is enabled, which makes the capture much simpler and nicer (no colored bars).

pieper:
LGTM - very nice.

jcfr:
Niceee :+1:

jcfr:
As a side note, the CircleCI bot fails way to often ... we will look into transitioning to CircleCI 2.0 which natively integrate with docker.

lassoan:
Merged.
https://github.com/Slicer/Slicer/commit/471c00b3b643d7ebc1cd58226d3e232ad637b94a

PR #694, opened 2017-03-28
ENH: Switch continuous integration to CircleCI 2.0

jcfr:
**DO NOT INTEGRATE**

This is still work in progress. We need to add the "store_artifacts" step


jcfr:
Closing. Fix integrated in [r27128](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27128)

PR #696, opened 2017-04-01
Support per view visibility of slicer intersection

jcfr:
This commit updates the SliceModel displayable manager to allow hiding part of slice intersection on selected views.

For example, let's consider this scene with:
* 1 model
* slice intersection enabled
* slice views visible in two 3D views

setting_viewids_start

Now, with this change, we can (for example) show the red slice intersection only on the green slice viewer.

The following tweaks have been done:
* model is removed from View1
* "red slice node" is hidden in ThreeD view 2
* only intersection of red slice with  green slice is shown

setting_viewids_done

### DIFF
--- a/Libs/MRML/Core/vtkMRMLDisplayNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLDisplayNode.cxx
@@ -833,7 +833,7 @@ void vtkMRMLDisplayNode::AddViewNodeID(const char* viewNodeID)
 }
 
 //-------------------------------------------------------
-void vtkMRMLDisplayNode::RemoveViewNodeID(char* viewNodeID)
+void vtkMRMLDisplayNode::RemoveViewNodeID(const char* viewNodeID)
 {
   if (viewNodeID == NULL)
     {
--- a/Libs/MRML/Core/vtkMRMLDisplayNode.h
+++ b/Libs/MRML/Core/vtkMRMLDisplayNode.h
@@ -434,7 +434,7 @@ public:
   void AddViewNodeID(const char* viewNodeID);
   /// Remove View Node ID for the view to display this node in.
   /// \sa ViewNodeIDs, AddViewNodeID(), RemoveAllViewNodeIDs()
-  void RemoveViewNodeID(char* viewNodeID);
+  void RemoveViewNodeID(const char* viewNodeID);
   /// Remove All View Node IDs for the views to display this node in.
   /// \sa ViewNodeIDs, AddViewNodeID(), RemoveViewNodeID()
   void RemoveAllViewNodeIDs();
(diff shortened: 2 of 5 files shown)


lassoan:
With this is change there would be several view combinations that we could not set, because there are no visible flags for all view node classes. For example, you cannot specify for a node to show in a slice view and none of the table views.

One solution could be to have a separate view list for all node classes, but this would be very complex to implement and use, as the list of lists would need to be dynamic, as view types can be registered anytime, even by extensions. Also it's not clear what to do with sub-types, e.g., does a modified 3D view subclass counts as a 3D view or a new class?

Proposed alternative solution: You can already fully customize views with group IDs. If you don't want a slice intersection to appear in another slice view then just put them into separate groups.

jcfr:
Thanks for the review. I agree with your comment, we need to keep things simple.

It turns out that grouping doesn't work in that case. Indeed, all slice viewer need to be visible but only a subset of the slice intersection need to be represented in a given slice viewers. 

I revisited the approach minimizing the changes.

### Example 1

All slice visible in 3D but slice intersections visible only in Red Slice viewers:

```
for name in ["Yellow", "Red", "Green"]:
  model_display_node = getNode("%s Display" % name)
  model_display_node.AddViewNodeID("vtkMRMLSliceNodeRed")

  slice_node = getNode("vtkMRMLSliceNode%s" % name)
  slice_node.AddThreeDViewID("vtkMRMLViewNode1")
```

all_slice_visible_in_3d_but_all_slice_intersections_only_in_red_view



### Example 2

All slice visible in 3D but only yellow slice intersection visible in Red Slice views:

```python
for name in ["Yellow", "Red"]:
  model_display_node = getNode("%s Display" % name)
  model_display_node.AddViewNodeID("vtkMRMLSliceNodeRed")

model_display_node = getNode("Green Display")
model_display_node.AddViewNodeID("invalid")

for name in ["Yellow", "Red", "Green"]:
  slice_node = getNode("vtkMRMLSliceNode%s" % name)
  slice_node.AddThreeDViewID("vtkMRMLViewNode1")
```

all_slice_visible_in_3d_but_only_yellow_intersection_in_red

jcfr:
Check that  this can be achieve with additional display node

lassoan:
With an additional vtkMRMLModelDisplayNode node you can specify anything you need. Except there is a bug (that should be easy to fix): slice intersection view visibility seem to ignore the specified ViewIDs. Probably just a standard GetVisibility(viewNodeID) check has to be added.

Example (variables are created using NodeInfo module in Debugging tools extension):

```
>>> Created variable: Red Display (vtkMRMLModelDisplayNode1) -> reddisplay
d=slicer.vtkMRMLModelDisplayNode()
d.Copy(reddisplay)
slicer.mrmlScene.AddNode(d)
>>> Created variable: Red Volume Slice (vtkMRMLModelNode1) -> redvolumeslice
redvolumeslice.AddAndObserveDisplayNodeID(d.GetID())
d.SetSliceIntersectionVisibility(1) # works!
d.SetSliceIntersectionThickness(3) # works!
d.SetVisibility(1) # works!
d.AddViewNodeID("vtkMRMLSliceNodeYellow") # partly works: slice disappears from 3D view but slice intersection is still visible on green slice view

```

lassoan:
Closing this, as slice intersection visibility is already fully customizable using multiple display nodes.

PR #697, opened 2017-04-03
BUG: print out RGB color values instead of pointers

nicoleaucoin:
In PrintSelf of the MRML display node, the pointers were being
printed rather than the RGB color values. EdgeColor was also not being
printed.


pieper:
LGTM - 👍

jcfr:
Thanks @nicoleaucoin , we will integrate as soon as the CI returns green.

jcfr:
Thanks again. 

Integrated in [r25912](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25912)

PR #701, opened 2017-04-11
ENH: extracted class DICOMReferencesDialog and simplified/refactored code 

che85:
I thought that might help for improving readability

image

### DIFF
--- a/Modules/Scripted/DICOMLib/DICOMWidgets.py
+++ b/Modules/Scripted/DICOMLib/DICOMWidgets.py
@@ -685,13 +685,12 @@ class DICOMDetailsBase(VTKObservationMixin, SizePositionSettingsMixin):
     if missingFileCount == allFileCount:
       return
 
-    nDicomPlugins = len(slicer.modules.dicomPlugins)
-    progress = slicer.util.createProgressDialog(parent=self, value=0, maximum=nDicomPlugins)
-    step = 0
+    plugins = self.pluginSelector.selectedPlugins()
 
     loadEnabled = False
-    plugins = self.pluginSelector.selectedPlugins()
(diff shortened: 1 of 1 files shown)


pieper:
Looks reasonable.  Thanks!

jcfr:
Thanks for the PR.

The docker images have been rebuilt. It means CircleCI is expected to be green.

Modules/Scripted/DICOMLib/DICOMWidgets.py
```diff
@@ -906,11 +847,81 @@ def onNodeAdded(caller, event, calldata):
...
+                "confirm if you would like to load the following referenced datasets."
+
+  def __init__(self, parent, loadables):
+    super(DICOMReferencesDialog, self).__init__(parent)
+    self.loadables = loadables
+    self.checkboxes = dict()
+    self.setup()
+
+  def setup(self):
+    self._setBasicProperties()
```
jcfr: If that helps, you create the UI using designer and include it as resources.
For example, see https://github.com/Slicer/Slicer/blob/eefeac9286f48552e8418a11f412b2823a09b407/Modules/Scripted/ExtensionWizard/CMakeLists.txt#L18-L20 that is referencing a ui file loaded dynamically.

It is also possible to install it in location relative the python script and load it from there.

che85:
I think that's ready for merge ;)

jcfr:
Thanks. Integrated in [r25916](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25916) to [r25919](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25919)

PR #702, opened 2017-04-13
Use header metadata to initialize volume scalar range

agirault:
Co-Authored with @jcfr 

# Motivation
We're working on a project where we need to speed up the loading of multiple large volumes (Gigabytes) in a custom version of Slicer.
* Our first bottleneck was - _FYI_ - tackled by simply using uncompressed data vs compressed data, where we encountered a 50% to 70% improvement (working with an NVMe SSD where reading from disk is much faster than uncompressing).
* Our second bottleneck was caused by the computation of the volume scalar range, needed to automatically calculate the window/level. The following stack trace obtained during one of our profilings shows a summary of the calls and their cost in CPU time, where `GetDisplayScalarRange` takes around **40%** (in our use case, using uncompressed data, and could go up to **47%**. With uncompressed data, the cost was around **17%**).

> * 97.0% - [qSlicerIOManager::loadNodes](https://github.com/Slicer/Slicer/blob/eefeac9286f48552e8418a11f412b2823a09b407/Base/QTGUI/qSlicerIOManager.cxx#L406-L459)
>   * 86.2% - [qSlicerCoreIOManager::loadNodes](https://github.com/Slicer/Slicer/blob/eefeac9286f48552e8418a11f412b2823a09b407/Base/QTGUI/qSlicerIOManager.cxx#L421)
>     * 82.5% - [vtkSlicerVolumesLogic::AddArchetypeVolume](https://github.com/Slicer/Slicer/blob/683f5016320b4b5769aa6887f2232c294750d9c9/Modules/Loadable/Volumes/Logic/vtkSlicerVolumesLogic.cxx#L629-L779)
>       * 74.9% - [vtkMRMLStorageNode::ReadData](https://github.com/Slicer/Slicer/blob/683f5016320b4b5769aa6887f2232c294750d9c9/Modules/Loadable/Volumes/Logic/vtkSlicerVolumesLogic.cxx#L700)
>         * 44.7% - vtkMRMLVolumeNode::SetAndObserveImageData
>           * 43.1% - [vtkMRMLScalarVolumeDisplayNode::CalculateAutoLevels](https://github.com/Slicer/Slicer/blob/868c60c58d7c4841462fd15cdb178262a645ce44/Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx#L416) 
>             * 39.2% - [vtkMRMLScalarVolumeDisplayNode::GetDisplayScalarRange](https://github.com/Slicer/Slicer/blob/868c60c58d7c4841462fd15cdb178262a645ce44/Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx#L792)
>               * 39.2% - [vtkDataSet::GetScalarRange](https://github.com/Slicer/Slicer/blob/868c60c58d7c4841462fd15cdb178262a645ce44/Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx#L693)
>             * 3.9% - [vtkStreamingDemandDrivenPipeline::Update](https://github.com/Slicer/Slicer/blob/868c60c58d7c4841462fd15cdb178262a645ce44/Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx#L724)
>               * 3.9%- vtkImageAccumulate::RequestData
>         * 29.6% - vtkStreamingDemandDrivenPipeline::Update
>           * 26.0% - itk::ImageFileReader::GenerateData()
>       * 5.8% - vtkMRMLScene::EndState
>     * 2.6% - vtkMRMLApplicationLogic::vtkInternal::PropagateVolumeSelection
>   * 10.7% - [qProgressDialog::setValue](https://github.com/Slicer/Slicer/blob/eefeac9286f48552e8418a11f412b2823a09b407/Base/QTGUI/qSlicerIOManager.cxx#L418)

# Objective
Offer a way to use the metadata min/max information from the volume file headers to speed up the data loading if it exists.

# Current support
Multiple file formats define properties in the header file to store the scalar range, and ITK adds these info to the metadata dictionary that can be retrieved using `itkObject::GetMetaDataDictionary()` :
- nifti: [cal_min, cal_max]
  - [format standard](https://nifti.nimh.nih.gov/nifti-1/documentation/nifti1fields/nifti1fields_pages/cal_maxmin.html)
  - [itk support](https://github.com/InsightSoftwareConsortium/ITK/blob/f9110a0046a9d07a921767ba000443ad006a2c9b/Modules/IO/NIFTI/src/itkNiftiImageIO.cxx#L887-L893)
- nrrd/nhrd: [min, max]
  - [format standard](http://teem.sourceforge.net/nrrd/format.html#min)
  - [itk support](https://github.com/InsightSoftwareConsortium/ITK/blob/f9110a0046a9d07a921767ba000443ad006a2c9b/Modules/ThirdParty/NrrdIO/src/NrrdIO/NrrdIO.h.in#L1238-L1239)
- mha/mhd : [ElementMin, ElementMax]
  - [format standard](https://itk.org/Wiki/MetaIO/Documentation#Tags_Added_by_MetaImage)
  - [itk support](https://github.com/InsightSoftwareConsortium/ITK/blob/f9110a0046a9d07a921767ba000443ad006a2c9b/Modules/IO/Meta/src/itkMetaImageIO.cxx#L393-L401)

Also, the class `vtkITKArchetypeImageSeries[Scalar]Reader` already retrieves the metadata (1) and sets it to the `vtkMRMLVolumeNode` (2):
1. https://github.com/Slicer/Slicer/blob/868c60c58d7c4841462fd15cdb178262a645ce44/Libs/vtkITK/vtkITKArchetypeImageSeriesReader.cxx#L816-L826
2. https://github.com/Slicer/Slicer/blob/eefeac9286f48552e8418a11f412b2823a09b407/Libs/MRML/Core/vtkMRMLVolumeArchetypeStorageNode.cxx#L407

### DIFF
--- a/Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx
@@ -796,7 +796,7 @@ void vtkMRMLScalarVolumeDisplayNode::CalculateAutoLevels()
 
     window = max-min;
     level = 0.5*(max+min);
-    lower = this->GetLevel();
+    lower = level;
     upper = range[1];
     }
   else
--- a/Libs/vtkITK/vtkITKArchetypeImageSeriesReader.cxx
+++ b/Libs/vtkITK/vtkITKArchetypeImageSeriesReader.cxx
@@ -16,18 +16,24 @@
 #include "vtkITKArchetypeImageSeriesReader.h"
 
 // VTK includes
+#include <vtkDataArray.h>
 #include <vtkImageData.h>
 #include <vtkMath.h>
 #include <vtkMatrix4x4.h>
 #include <vtkInformation.h>
 #include <vtkInformationVector.h>
+#include <vtkNew.h>
 #include <vtkObjectFactory.h>
+#include <vtkPointData.h>
 #include <vtkStreamingDemandDrivenPipeline.h>
 
 // ITK includes
+#include <itkNiftiImageIO.h>
+#include <itkNrrdImageIO.h>
 #include <itkMetaDataDictionary.h>
 #include <itkMetaDataObjectBase.h>
 #include <itkMetaDataObject.h>
+#include <itkMetaImageIO.h>
 #include <itkTimeProbe.h>
 
 // STD includes
@@ -811,7 +817,8 @@ int vtkITKArchetypeImageSeriesReader::RequestInformation(
 
 
   vtkDataObject::SetPointDataActiveScalarInfo(outInfo,
-                                              scalarType, numberOfComponents);
+                                              scalarType,
+                                              numberOfComponents);
 
   // Copy the MetaDataDictionary from the ITK layer to the VTK layer
   if (imageIO.GetPointer() != NULL)
@@ -823,7 +830,111 @@ int vtkITKArchetypeImageSeriesReader::RequestInformation(
     this->Dictionary = itk::MetaDataDictionary();
     }
   ParseDictionary();
+
+  // Store the meta data scalar range if it exists
+  std::string range_keys[2];
+  this->GetScalarRangeMetaDataKeys(imageIO, range_keys);
+  std::string min_str;
+  std::string max_str;
+  bool hasMin = itk::ExposeMetaData<std::string>(this->Dictionary, range_keys[0], min_str);
+  bool hasMax = itk::ExposeMetaData<std::string>(this->Dictionary, range_keys[1], max_str);
+  if (hasMin && hasMax)
+    {
+    this->MetaDataScalarRangeMinima.clear();
+    this->MetaDataScalarRangeMaxima.clear();
+    double value = 0.0;
+    std::istringstream min_sstr(min_str);
+    while (min_sstr >> value)
+      {
+      this->MetaDataScalarRangeMinima.push_back(value);
+      }
+    std::istringstream max_sstr(max_str);
+    while (max_sstr >> value)
+      {
+      this->MetaDataScalarRangeMaxima.push_back(value);
+      }
+    }
+
   return 1;
+}
+
+//----------------------------------------------------------------------------
+void vtkITKArchetypeImageSeriesReader::GetScalarRangeMetaDataKeys(itk::ImageIOBase::Pointer imageIO,
+                                                                  std::string range_keys[2])
+{
+    if (dynamic_cast<itk::NrrdImageIO*>(imageIO.GetPointer()))
+      {
+      // http://teem.sourceforge.net/nrrd/format.html#min
+      range_keys[0] = "min";
+      range_keys[1] = "max";
+      }
+    else if (dynamic_cast<itk::NiftiImageIO*>(imageIO.GetPointer()))
+      {
+      // https://nifti.nimh.nih.gov/nifti-1/documentation/nifti1fields/nifti1fields_pages/cal_maxmin.html
+      range_keys[0] = "cal_min";
+      range_keys[1] = "cal_max";
+      }
+    else if (dynamic_cast<itk::MetaImageIO*>(imageIO.GetPointer()))
+      {
+      // https://itk.org/Wiki/MetaIO/Documentation#Tags_Added_by_MetaImage
+      range_keys[0] = "ElementMin";
+      range_keys[1] = "ElementMax";
+      }
+}
+
+//----------------------------------------------------------------------------
+void vtkITKArchetypeImageSeriesReader::SetMetaDataScalarRangeToPointDataInfo( vtkImageData* data )
+{
+    if (data == NULL)
+      {
+      vtkWarningMacro("No image data specified, can't set scalar range information.")
+      return;
+      }
+
+    // If no metadata scalar range found
+    if (this->MetaDataScalarRangeMinima.empty() ||
+       this->MetaDataScalarRangeMaxima.empty())
+      {
+      return;
+      }
+
+    // If metadata scalar range not consistent
+    int nbrOfComponents = data->GetNumberOfScalarComponents();
+    if (nbrOfComponents != this->MetaDataScalarRangeMinima.size() ||
+        nbrOfComponents != this->MetaDataScalarRangeMaxima.size())
+      {
+      vtkWarningMacro("Mismatch between image data and meta data number of scalar components, ignoring metada scalar range.");
+      return;
+      }
+
+    // Check that scalar info exists
+    vtkDataArray* scalars = data->GetPointData()->GetScalars();
+    vtkInformation* scalarInfo = scalars ? scalars->GetInformation() : NULL;
+    if (scalarInfo == NULL)
+      {
+      return;
+      }
+
+    // Write meta data scalar range in scalars information
+    vtkNew<vtkInformationVector> rangeInfoVector;
+    rangeInfoVector->SetNumberOfInformationObjects(nbrOfComponents);
+    double scalarRange[2];
+    bool scalarRangeIsValid = false; // all values should not be equal to 0
+    for (unsigned int i = 0; i < nbrOfComponents; ++i)
+      {
+      scalarRange[0] = this->MetaDataScalarRangeMinima[i];
+      scalarRange[1] = this->MetaDataScalarRangeMaxima[i];
+      rangeInfoVector->GetInformationObject(i)->Set(vtkDataArray::COMPONENT_RANGE(), scalarRange, 2);
+      if (scalarRange[0] != 0.0 || scalarRange[1] != 0.0)
+        {
+        scalarRangeIsValid = true;
+        }
+      }
+
+    if (scalarRangeIsValid)
+      {
+      scalarInfo->Set(vtkAbstractArray::PER_COMPONENT(), rangeInfoVector.Get());
+      }
 }
 
 //----------------------------------------------------------------------------
--- a/Libs/vtkITK/vtkITKArchetypeDiffusionTensorImageReaderFile.cxx
+++ b/Libs/vtkITK/vtkITKArchetypeDiffusionTensorImageReaderFile.cxx
@@ -147,6 +147,7 @@ int vtkITKArchetypeDiffusionTensorImageReaderFile::RequestData(
   data->SetOrigin(0, 0, 0);
   data->SetSpacing(1, 1, 1);
   data->SetExtent(extent);
+  this->SetMetaDataScalarRangeToPointDataInfo(data);
   vtkNew<vtkFloatArray> tensors;
   tensors->SetName("ArchetypeReader");
 
--- a/Libs/vtkITK/vtkITKArchetypeImageSeriesReader.h
+++ b/Libs/vtkITK/vtkITKArchetypeImageSeriesReader.h
@@ -23,6 +23,7 @@ Version:   $Revision$
 class vtkMatrix4x4;
 
 // ITK includes
+#include "itkImageIOBase.h"
 #include "itkMetaDataDictionary.h"
 #include "itkSpatialOrientation.h"
 
@@ -792,6 +793,12 @@ protected:
 
   int          OutputScalarType;
   unsigned int NumberOfComponents;
+
+  std::vector<double> MetaDataScalarRangeMinima;
+  std::vector<double> MetaDataScalarRangeMaxima;
+  void GetScalarRangeMetaDataKeys(itk::ImageIOBase::Pointer imageIO,
+                                  std::string range_keys[2]);
+  void SetMetaDataScalarRangeToPointDataInfo(vtkImageData* data);
 
   double DefaultDataSpacing[3];
   double DefaultDataOrigin[3];
--- a/Libs/vtkITK/vtkITKArchetypeImageSeriesScalarReader.cxx
+++ b/Libs/vtkITK/vtkITKArchetypeImageSeriesScalarReader.cxx
@@ -84,6 +84,7 @@ int vtkITKArchetypeImageSeriesScalarReader::RequestData(
   data->AllocateScalars(outInfo);
   data->SetExtent(outInfo->Get(
     vtkStreamingDemandDrivenPipeline::WHOLE_EXTENT()));
+  this->SetMetaDataScalarRangeToPointDataInfo(data);
 
 /// SCALAR MACRO
 #define vtkITKExecuteDataFromSeries(typeN, type) \
--- a/Libs/vtkITK/vtkITKArchetypeImageSeriesVectorReaderFile.cxx
+++ b/Libs/vtkITK/vtkITKArchetypeImageSeriesVectorReaderFile.cxx
@@ -126,6 +126,8 @@ void vtkITKArchetypeImageSeriesVectorReaderFile::ExecuteDataWithInformation(vtkD
     default:
         vtkErrorMacro(<< "UpdateFromFile: Unknown data type " << this->OutputScalarType);
       }
+
+    this->SetMetaDataScalarRangeToPointDataInfo(data);
     }
   else
     {
--- a/Libs/vtkITK/vtkITKArchetypeImageSeriesVectorReaderSeries.cxx
+++ b/Libs/vtkITK/vtkITKArchetypeImageSeriesVectorReaderSeries.cxx
@@ -147,6 +147,8 @@ void vtkITKArchetypeImageSeriesVectorReaderSeries::ExecuteDataWithInformation(vt
       {
       vtkErrorMacro(<< "Exception from vtkITK MegaMacro: " << e << "\n");
       }
+
+    this->SetMetaDataScalarRangeToPointDataInfo(data);
     }
 }
 


lassoan:
> Would we only want to store the metadata range (read from the header) or store the cached scalar range in general (either from the header, or from vtkImageData::GetScalarRange) ?

You must not store a "metadata range" separately anywhere, as VTK filters and mappers only care about what is stored in vtkImageData. If scalar range is not available in vtkImageData then filters and mappers recompute it anyway.

Fortunately, there is a way to set the scalar range in the vtkImageData after it is read from file:
<pre>
r=vtk.vtkNrrdReader()
r.SetFileName('something.nrrd')
r.Update()
img=r.GetOutput()

info = img.GetPointData().GetScalars().GetInformation()
infoVec = vtk.vtkInformationVector()
info.Set(vtk.vtkDataArray.PER_COMPONENT(), infoVec)
numberOfComponents = img.img.GetNumberOfScalarComponents()
infoVec.SetNumberOfInformationObjects(numberOfComponents)
for i in range(numberOfComponents):
  scalarRange = [10,100] # Just an example
  infoVec.GetInformationObject(i).Set(vtk.vtkDataArray.COMPONENT_RANGE(), scalarRange, 2 )

print info.Has(vtk.vtkDataArray.PER_COMPONENT())
print img.GetScalarRange() # returns [10,100], does not recompute scalar range
</pre>
 (see vtkDataArray::ComputeRange(double range[2], int comp) for a C++ implementation)

> Where to read the metadata range?
Everywhere a reader completed reading info into a vtkImageData object, update the vtkInformationObject as shown above with the cached scalar range values.

> Where to store the metadata range?
In vtkImageData and only in vtkImageData. It does not matter where the scalar range info comes from (file header, computed from data, ...).

> metadata range name? HeaderMinMax, HeaderRange, MetaDataMinMax, MetaDataRange ...
There will be no additional variable for storing scalar range.

> Right now the concept of scalar range is only defined in the scalar volume display node with GetDisplayScalarRange. Is that something that should be moved to the scalar volume node since it is more a property of the data than a display property (like window/level) ?
Fortunately, you'll not need to change anything. vtkImageData stores the scalar range, same as before.

----

So, this will be a really simple change. Just add the scalar range in every image storage node as the example code shows above.

pieper:
I agree with the analysis of @lassoan.

agirault:
Great, thanks @lassoan. Do we still want an option in the reader to allow the use of the Metadata range or not? As well as a function that can returns the key name based on the file extension?

lassoan:
> Do we still want an option in the reader to allow the use of the Metadata range or not?

I don't think we need to expose this option on the GUI (too low level detail), but may be useful to make it configurable in vtkMRMLVolumeArchetypeStorageNode.

> As well as a function that can returns the key name based on the file extension?

It would be more robust to determine metadata key name based on the reader class name instead of trying to guess from the file extension.

agirault:
> I don't think we need to expose this option on the GUI (too low level detail), but may be useful to make it configurable in vtkMRMLVolumeArchetypeStorageNode.

Sounds good

> It would be more robust to determine metadata key name based on the reader class name instead of trying to guess from the file extension.

So far all the volumes I read in our use case (nrrd, nhdr, mha, nii) used `vtkITKArchetypeImageSeriesReader`. `vtkNrrdReader` was never used, and I did not find a reader specific to other volume file formats.

Also, to store that information in the vtkImageData, do you think I should make use of the output information vector in [`vtkITKArchetypeImageSeriesReader::RequestInformation`](https://github.com/Slicer/Slicer/blob/eefeac9286f48552e8418a11f412b2823a09b407/Libs/vtkITK/vtkITKArchetypeImageSeriesReader.cxx#L190)? It is already used for [`SPACING`](https://github.com/Slicer/Slicer/blob/eefeac9286f48552e8418a11f412b2823a09b407/Libs/vtkITK/vtkITKArchetypeImageSeriesReader.cxx#L604), [`ORIGIN`](https://github.com/Slicer/Slicer/blob/eefeac9286f48552e8418a11f412b2823a09b407/Libs/vtkITK/vtkITKArchetypeImageSeriesReader.cxx#L605), [`EXTENT`](https://github.com/Slicer/Slicer/blob/eefeac9286f48552e8418a11f412b2823a09b407/Libs/vtkITK/vtkITKArchetypeImageSeriesReader.cxx#L618).

pieper:
Two other thoughts  about your original bottleneck issues:

1)  if computing the scalar range is a problem because it's used to calculate window/level you could look at using an approximation of the scalar range instead (i.e. randomly sample 5% of the voxels and you'd probably get essentially the same window/level values).  This could be implemented in the display node pretty easily.

2) The scalar range calculation in vtk is pretty complicated due to templating [1] but as far as I can tell it's not parallelized (is it?).  Since scalar range computation often comes up as a bottleneck it might have the most impact to implement a custom path (partial specialization) for parallel implementation.  This would be a win for many use cases, but even in the window/level calculation case it would be a general improvement since very few data files will include the scalar range info.  Also window/level is estimated for dynamically generated volumes so speeding it up in general would have many benefits.

[1] https://github.com/Kitware/VTK/blob/08b0ed1006d7c5183be18087fb0de478f0e9afed/Common/Core/vtkDataArrayPrivate.txx#L289

fedorov:
If you can disclose this - what kind of images are those? Do you know anything about the range you expect?

Another idea to what Steve said - you can take a small sub-region of the image, and calculate the range based on that region. 

You can also add a tool for the user that would allow to select small rectangle and calculate w/l based on the intensity distribution there. We used this approach for parametric maps where absolute range is fixed, but there is a lot of spatial variability and it needs to be adjusted for looking at variation within small areas.

agirault:
@pieper: agreed with both of those. However, in our use case, we control the data being loaded and can ensure that metadata is properly set. Like Andras said, that should be an easy feature to implement (just like we read the origin or spacing from the file metadata), and would offer the best performance.

All those points are valid, but we won't have the time to look into those other options for our use case because of funding limitations. We could create feature requests on Mantis for your suggestions?

pieper:
@agirault yes, it's important to solve your real use case.  I see on your todo list you plan to add writing the scalar range to the storage node output.  That way it will help other users when they re-read their data back to slicer.

lassoan:
> So far all the volumes I read in our use case (nrrd, nhdr, mha, nii) used vtkITKArchetypeImageSeriesReader. vtkNrrdReader was never used, and I did not find a reader specific to other volume file formats.

vtkITKArchetypeImageSeriesReader has an itk::ImageFileReader inside that should be able to tell you which reader was actually used. Maybe the GetImageIO() method or something similar.

> Also, to store that information in the vtkImageData, do you think I should make use of the output information vector in vtkITKArchetypeImageSeriesReader::RequestInformation? It is already used for SPACING, ORIGIN, EXTENT.

I don't think there is a method to change image scalar range information (vtkChangeImageInformation supports only origin, spacing, extent, etc), so probably you need to read in the image into a vtkImageData object, delete the reading pipeline, and change the scalar range in the vtkImageData object.

agirault:
@lassoan so should I put this in `vtkMRMLVolumeArchetypeStorageNode::ReadDataInternal` or in `vtkITKArchetypeImageSeriesReader` ?

lassoan:
> should I put this in vtkMRMLVolumeArchetypeStorageNode::ReadDataInternal or in vtkITKArchetypeImageSeriesReader ?

As low level as possible. In this case, vtkITKArchetypeImageSeriesReader is preferred.

Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx
```diff
@@ -796,7 +796,7 @@ void vtkMRMLScalarVolumeDisplayNode::CalculateAutoLevels()
 
     window = max-min;
     level = 0.5*(max+min);
-    lower = this->GetLevel();
+    lower = range[0];
```
lassoan: Why is this change? Is the behavior the same as before?

Libs/MRML/Core/vtkMRMLVolumeArchetypeStorageNode.cxx
```diff
@@ -443,6 +446,49 @@ int vtkMRMLVolumeArchetypeStorageNode::ReadDataInternal(vtkMRMLNode *refNode)
 
   vtkNew<vtkImageData> iciOutputCopy;
   iciOutputCopy->ShallowCopy(ici->GetOutput());
+  vtkInformation* imgInfo = iciOutputCopy->GetPointData() ?
```
lassoan: this is unreadable, split it into multiple lines

Libs/MRML/Core/vtkMRMLVolumeArchetypeStorageNode.cxx
```diff
@@ -443,6 +446,49 @@ int vtkMRMLVolumeArchetypeStorageNode::ReadDataInternal(vtkMRMLNode *refNode)
...
+    std::string min_str;
+    std::string max_str;
+    if (itk::ExposeMetaData<std::string>(volNode->GetMetaDataDictionary(), range_keys[0], min_str) &&
+        itk::ExposeMetaData<std::string>(volNode->GetMetaDataDictionary(), range_keys[1], max_str))
+      {
+      double value;
+
+      std::vector<double> min_vec;
+      std::istringstream min_sstr(min_str);
+      while (min_sstr >> value) min_vec.push_back(value);
```
lassoan: add braces and split into multiple lines to make it more readable

Libs/MRML/Core/vtkMRMLVolumeArchetypeStorageNode.cxx
```diff
@@ -443,6 +446,49 @@ int vtkMRMLVolumeArchetypeStorageNode::ReadDataInternal(vtkMRMLNode *refNode)
...
+
+      std::vector<double> max_vec;
+      std::istringstream max_sstr(max_str);
+      while (max_sstr >> value) max_vec.push_back(value);
+
+      int nbrOfComponents = iciOutputCopy->GetNumberOfScalarComponents();
+      if (nbrOfComponents == min_vec.size() && nbrOfComponents == max_vec.size())
```
lassoan: Log warning if there is a mismatch

Libs/MRML/Core/vtkMRMLVolumeArchetypeStorageNode.cxx
agirault: Yeah sorry, this is still experimental/testing for now, I'm just pushing things to keep them online (that's why I have 3 ugly commits): I wanted to know where we want to put it first: in the storagenode or in the reader?

agirault:
> As low level as possible. In this case, vtkITKArchetypeImageSeriesReader is preferred. - @lassoan 

I looked a little more, but I couldn't find where to access the vtkImageData from that class: it looks it only instantiates the reader through `RequestInformation`, and the vtkImageData is instantiated later on. Some metadata is read in there the same way I read the min/max, but only if the archetype is a dicom file: https://github.com/Slicer/Slicer/blob/eefeac9286f48552e8418a11f412b2823a09b407/Libs/vtkITK/vtkITKArchetypeImageSeriesReader.cxx#L1044-L1154

lassoan:
In vtkITKArchetypeImageSeriesReader::RequestInformation you can simply set the scalar range in the output info, similarly how spacing and origin is set:
```
outInfo->Set(vtkDataObject::SPACING(), spacing, 3);
outInfo->Set(vtkDataObject::ORIGIN(), origin, 3);
```

agirault:
Ah! That's the one I was referring to a couple comments ago:
> Also, to store that information in the vtkImageData, do you think I should make use of the output information vector in [`vtkITKArchetypeImageSeriesReader::RequestInformation`](https://github.com/Slicer/Slicer/blob/eefeac9286f48552e8418a11f412b2823a09b407/Libs/vtkITK/vtkITKArchetypeImageSeriesReader.cxx#L190)? It is already used for [`SPACING`](https://github.com/Slicer/Slicer/blob/eefeac9286f48552e8418a11f412b2823a09b407/Libs/vtkITK/vtkITKArchetypeImageSeriesReader.cxx#L604), [`ORIGIN`](https://github.com/Slicer/Slicer/blob/eefeac9286f48552e8418a11f412b2823a09b407/Libs/vtkITK/vtkITKArchetypeImageSeriesReader.cxx#L605), [`EXTENT`](https://github.com/Slicer/Slicer/blob/eefeac9286f48552e8418a11f412b2823a09b407/Libs/vtkITK/vtkITKArchetypeImageSeriesReader.cxx#L618).

Will it matter if that is the outputinfo, and not the point data scalars information?

lassoan:
Good point, you need to put range in the point data information. I think that should be done in derived classes, such as vtkITKArchetypeImageSeriesScalarReader::RequestData.

agirault:
Yes, I just testing, it did not work on the output information vector. I'll give it a go on the subclasses, but that would mean I'd have to duplicate my code to retrieve them?
* vtkITKArchetypeDiffusionTensorImageReaderFile::RequestData
* vtkITKArchetypeImageSeriesScalarReader::RequestData
* vtkITKArchetypeImageSeriesVectorReaderFile::ExecuteDataWithInformation
* vtkITKArchetypeImageSeriesVectorReaderSeries::ExecuteDataWithInformation

Wouldn't it be easier to leave it in the storage node now?

lassoan:
You can retrieve range from metadata in the base class in RequestInformation and store in some member variables. You may implement a helper method in the base class that sets the range in the point data information and call that in RequestData method in each child class, but setting may be different for vector, scalar, tensor images, so maybe there will not be much to reuse.

Implementing in the storage node would not be optimal, because it would reduce cohesion of the storage node and reader classes (the storage node would need to know a lot of internal details of vtkITKArchetypeImageSeriesReader, which could have been all nicely encapsulated in the vtkITKArchetypeImageSeriesReader).

Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx
agirault: The behavior is different, but I assume is correct: we are now setting the initial value for the lower threshold to the minimum scalar value.

The previous behavior was setting the lower threshold to the PREVIOUS level value:

* initial:
  - window=256
  - level=128
  - threshold_min = VTK_SHORT_MIN
  - treshold_max = VTK_SHORT_MAX 
* after computeAutoWindowLevel (x1):
  - window = max-min
  - level = 0.5x(max+min)
  - threshold_min = **getlevel = 128 (previous value from initialisation)**
  - threshold_max = max
* after computeAutoWindowLevel (x2):
  - window = max-min
  - level = 0.5*(max+min)
  - threshold_min = **getlevel = 0.5x(old_max+old_min)  (previous value from 1st computeWL)**
  - threshold_max = max

lassoan: Thanks for the explanation. The current behavior doesn't make much sense to me. Thresholding is disabled by default, so probably that's why it has not been causing problems.

@pieper @millerjv What do you think? Could have been a reason for using the previous level value as minimum threshold in vtkMRMLScalarVolumeDisplayNode::CalculateAutoLevels?

agirault:
There is no RequestData in VectorReaderSeries and VectorReaderFile though

lassoan:
There is ExecuteDataWithInformation or similar methods instead.

Libs/vtkITK/vtkITKArchetypeImageSeriesReader.cxx
```diff
@@ -823,9 +826,76 @@ int vtkITKArchetypeImageSeriesReader::RequestInformation(
...
+      {
+      this->MetaDataScalarRangeMinima.push_back(value);
+      }
+    std::istringstream max_sstr(max_str);
+    while (max_sstr >> value)
+      {
+      this->MetaDataScalarRangeMaxima.push_back(value);
+      }
+    }
+
```
agirault: This is due to the different factories being tested in the following order:
* vtkITKArchetypeDiffusionTensorImageReaderFile
* vtkITKArchetypeImageSeriesVectorReaderSeries
* vtkITKArchetypeImageSeriesVectorReaderFile
* vtkITKArchetypeImageSeriesScalarReader

I'll check how much time this adds up, but my intuition is not that much.

Libs/vtkITK/vtkITKArchetypeImageSeriesReader.cxx
```diff
@@ -823,9 +826,76 @@ int vtkITKArchetypeImageSeriesReader::RequestInformation(
...
   return 1;
 }
 
+
+//----------------------------------------------------------------------------
+void vtkITKArchetypeImageSeriesReader::SetMetaDataScalarRangeToPointDataInfo( vtkImageData* data )
```
agirault: This is due to the different factories being tested in the following order:
* vtkITKArchetypeDiffusionTensorImageReaderFile -> calls requestdata
* vtkITKArchetypeImageSeriesVectorReaderSeries -> doesn't call requestdata
* vtkITKArchetypeImageSeriesVectorReaderFile -> doesn't call requestdata
* vtkITKArchetypeImageSeriesScalarReader -> calls requestdata

While `RequestInformation` shouldn't take much time for those 4 readers, I wonder if having request data in `vtkITKArchetypeDiffusionTensorImageReaderFile` isn't taking some precious time, and if it is necessary or if we can invalidate that reader without requesting the data?

lassoan:
Parsing the header and extracting basic info should not take more than a few ten milliseconds. But instead of guessing you always have to run a profiler to identify bottlenecks.

I would worry about automatic width/level computation time instead, which consists of building an intensity histogram and running bimodal analysis on it.

agirault:
> Parsing the header and extracting basic info should not take more than a few ten milliseconds. But instead of guessing you always have to run a profiler to identify bottlenecks.

I did, it looks fine, no bottlenecks at all.

Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx
pieper: @lassoan I think the level was chosen as the default lower threshold because it is used when making a slice model visible in the threeDView and in the paradigm use case you'd want to see bright tissue while making the air transparent.  Since thresholding is off at initialization you are right the default value has little impact.

jcfr:
Thanks everyone for reviewing 👍 

Considering this first set of changes is reasonable for everyone, I suggest to move forward with its integration.

Note also that the suggested improvements (writing of min/max in metadata for format that support it, random sampling, ...) are valid but outside of the scope of this contribution.

Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx
agirault: @pieper I see, thanks for the precision. Shouldn't it be the current level value however instead of the previous one?

Libs/vtkITK/vtkITKArchetypeImageSeriesReader.cxx
```diff
@@ -823,10 +830,114 @@ int vtkITKArchetypeImageSeriesReader::RequestInformation(
     this->Dictionary = itk::MetaDataDictionary();
     }
   ParseDictionary();
+
+  // Store the meta data scalar range if it exists
+  std::string range_keys[2] = {"",""};
```
lassoan: std::string is initialized to empty string by default, no need for = {"",""}

Libs/vtkITK/vtkITKArchetypeImageSeriesReader.cxx
```diff
@@ -823,10 +830,114 @@ int vtkITKArchetypeImageSeriesReader::RequestInformation(
...
+    }
+
   return 1;
 }
 
 //----------------------------------------------------------------------------
+void vtkITKArchetypeImageSeriesReader::GetScalarRangeMetaDataKeys(itk::ImageIOBase::Pointer imageIO,
+                                                                  std::string range_keys[2])
+{
+    if (dynamic_cast<itk::NrrdImageIO*>(imageIO.GetPointer()))
```
lassoan: Is dynamic_cast working in ITK on all platforms?

Libs/vtkITK/vtkITKArchetypeImageSeriesReader.cxx
agirault: will remove

Libs/vtkITK/vtkITKArchetypeImageSeriesReader.cxx
agirault: Answer provided by @thewtex : there used to be an issue on Mac but that was addressed with ITK 4.11 ([here](http://review.source.kitware.com/#/c/21805/)) and Slicer is there.

agirault:
@lassoan thanks. The MR should already be properly divided in 2 commits.

Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx
pieper: @agirault yes, the default was not too important so nobody noticed - the option was only ever turned on in the GUI and the slider is right there to adjust (which is almost always needed in practice no matter what default we pick).

agirault: I can just revert this commit if you'd like then.

pieper: I don't think it's a huge issue either way, but maybe the best is to use the current value of Level.  That seems to fix the bug and be mostly consistent with the old behavior.

Libs/vtkITK/vtkITKArchetypeImageSeriesReader.cxx
agirault: done

Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx
finetjul: +1

agirault: done

agirault:
Corrected the level value as asked by @pieper. If there are no other comments I'll merge after the checks have passed.

thewtex:
LGTM. Well done! :+1:

agirault:
Thank you everybody for your reviews. Integrated in [r25937](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25937) and [r25938](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25938).

PR #703, opened 2017-04-19
ENH: Integration of Subject hierarchy into the Data module

cpinter:
The name of the subject hierarchy is not indicative to a user at all, and they may have difficulties finding out what the name
actually covers. In the same time there is the Data module with an indicative name, but which has limited uses, and is mostly used
to build transform hierarchies and for development purposes. It seems to make sense to "merge" the two modules, so that the now
stable and fast subject hierarchy can be used where it is easiest to find.
- Made the Data module widget tabbed, with Subject and Transform tabs. Subject contains the subject hierarchy, Transform contains
  the MRML tree from the current Data module (ModelHierarchy and AnnotationHierarchy were not migrated, as Model hierarchies are
  supported by subject hierarchy, and if display properties are to be set, then the Models module is needed anyway, and Annotations
  is considered deprecated - it is also supported by SH by default too)
- Subject hierarchy made a hidden module, and all content removed from its module widget (so that it does not need to be maintained
  in two places
- Data module icon replaced with the subject hierarchy icon, and Subject hierarchy replaced by Data in the favorite module toolbar
- Added a button to "Explore Loaded Data" in Welcome module, and also updated text


cpinter:
So I broke the lines in the commit, and now it's unreadable here :)

cpinter:
image

image

image

fedorov:
In my view, there is significant semantic difference between Data ("bag of everything" and "disorganized by intent") and Subject hierarchy. I wonder if this integration of the two modules will cause confusion for the users.

Where would you fit something like "needle", or "color lookup table", or "star"? There is no formal "data model" of what you can have in Slicer, and "Subject hierarchy" implies subject-centered data model.

Maybe "Data" should have been called "MRML hierarchy" from the beginning, and that would reflect the relationship to Subject hierarchy better.

fedorov:
But to be explicitly clear - I do not formally oppose this PR. I am just not sure if it will help or make things even more confusing...

cpinter:
The nodes that are not included in subject hierarchy (color, view, display, etc) would be in the "Transform" tab. That tab includes the tree that is in the current Data module.

The problem with MRML hierarchy is that its name is meaningless to the user (which is another level of confusing). Same for the current "Subject hierarchy" name. The user reads it and registers "no information" in his/her mind.

The main intent of this PR is to
1. Make it clear what subject hierarchy does
2. Make the Data module usable
3. Spur discussion about this whole issue

lassoan:
Good point about non-transformable nodes showing up in the "Transform" hierarchy. We could have 3 views: Subjects, Transforms, All (or something similar).
@cpinter what do you think
* Would it be easy to hide all non-transformable nodes from the Transform view?
* Would it be easy to show a flat list of all nodes in a new tab?

cpinter:
> Would it be easy to hide all non-transformable nodes from the Transform view?
Yes, I think it would

> Would it be easy to show a flat list of all nodes in a new tab
Yes, I think it would :) The question is then what operations are available in this flat list of all nodes? I think the easiest would be to show the transform hierarchy here as well, but with all nodes visible.

@jcfr @pieper @ihnorton @vovythevov If you have any opinion please don't hold it back! Thanks

pieper:
This seems like a step in the right direction to me.  The Data module has never lived up to its potential and I agree that merging with the Subject Hierarchy makes a lot of sense.

cpinter:
@fedorov I really don't want to integrate this until there is no consensus, and I think it would help if you'd elaborate on what you meant by this being more confusing. In my mind, what was confusing was that there were two modules where the content of the scene could be explored, and one of those modules had a meaningless name. I'd like to understand your point of view too. Thanks!

fedorov:
@cpinter thanks for following up! Upon reading all the comments, and re-reading your explanation, I agree it is a good way forward. I agree it is a great idea to merge these two concepts. 

I had a red eye flight on the night before making my earlier comment, and, in the retrospect, my brain was clearly not up to the task after a sleepless night!

cpinter:
Thanks a lot, Andrey! I'm happy that we agree on that this is a good step forward. Having heard two "good step forward"s, do you guys have any ideas for further steps forward that I should potentially take before this goes into the trunk?

In any case, I will let this settle for a few more days (and I have two bugs to fix before integrating this), giving others a chance to chime in or raise concerns.

Thanks!

fedorov:
Maybe we could come up with a better name for the "Transform" tab? I know it may make sense from the MRML architecture perspective, but for a user entering that module - I think this name can be quite puzzling. Maybe tabs should be named "Subject hierarchy" and "Transformable nodes"?

lassoan:
"Transformable nodes" sounds very technical to me, too. A few more candidates:

Subject hierarchy | Transformable nodes | All nodes
Subject hierarchy | Transforms | All
Subject hierarchy | Transform hierarchy | All

cpinter:
It makes sense to call the Subject tab Subject hierarchy instead.

Transformable nodes however I think only covers half the story. Because the main reason a user ever enters the Transform... tab, is to see the hierarchy of the transforms. So maybe I'd call it Transform hierarchy instead. The problem with that one (as you also felt it when you suggested transformable nodes) that it suggests that it only contains transforms, while it actually contains all nodes (not just transformable nodes), and indicating the transform hierarchy in a tree. If we decide to rename then Subject hierarchy and Transform hierarchy, then I guess the hierarchy is redundant and we're back where we started.

I agree that the name of the Transform is problematic, but I feel like both Transform hierarchy and Transformable nodes are both suboptimal too. The name that expresses exactly what it is would be "All nodes in a transform hierarchy", which is obviously also bad. How about "All / Transform"? I'll keep thinking though. Any suggestions?

cpinter:
@lassoan By making the Transform tab only show transformable and transform nodes, it would lose the "All-ness", and by adding an All, the user could still see all nodes (I guess there would be no hierarchy whatsoever in All right?).
I like this scenario.

lassoan:
Subjects | Transforms | All   => nice and short
Subject hierarchy | Transform hierarchy | All nodes => most verbose, maybe a bit more clear?
Subject hierarchy | Transforms | All => something in between

pieper:
I like "Subjects | Transforms | All".

jcfr:
> I like "Subjects | Transforms | All".

👍 

with

Name of the module: Data
Title of module: Data (MRML Scene)


> was confusing was that there were two modules where the content of the scene could be explored

👍 

> So I broke the lines in the commit, and now it's unreadable here :)

If you use `* this a bullet point`, it should show as expected in the PR

fedorov:
I personally prefer "Subject hierarchy | Transform hierarchy | All nodes". In my view, "prefer" should correspond to the choice that would lower the chances of users being confused. I am not sure being nice and short are important considerations for selecting the names for those tabs.

jcfr:
> I am not sure being nice and short are important considerations for selecting the names for those tabs.

Agreed. Being explicit is important.

The changes added by @cpinter are great 👍  and I would like to see them integrated sooner than later. If the name of these tabs is confusing, we will be able to change them quite easily.

fedorov:
I completely agree. My comment was not meant to delay integration - please proceed with any name choice, I was just making a suggestion. Whatever we do, it will not be the most confusing thing in Slicer!

cpinter:
Thanks for all the feedback! I will
* Make the tabs called "Subject hierarchy | Transform hierarchy | All nodes"
** I'll need to add a new "All nodes" tab then, which will contain every node (with the defautl transform view, because the only other options are model- or annotation hierarchy), while the Transform hierarchy tab will contain only the transformable nodes
* I need to fix two bugs
** The context menu hint popup doesn't seem to work, so I need to fix that (it's a new feature I added last week to SH)
** The header in the Transforms tab's MRML tree view is "1" on my build, I will loko into fixing that
* Integrate

cpinter:
I made all the changes I wrote in the previous comment. I will integrate shortly unless there are issues.

cpinter:
Integrated in https://github.com/Slicer/Slicer/commit/e52ad0afa2c24fed3858d37c1a5cee3ab543f4c0

cpinter:
Updated Data node and SH documentation
https://www.slicer.org/wiki/Documentation/Nightly/Modules/Data (please note that any duplication will be removed when the automatic generation is done the next time)
https://www.slicer.org/wiki/Documentation/Nightly/Modules/SubjectHierarchy (basically empty now)

PR #705, opened 2017-04-21
ENH: add module to PYTHONPATH during registration

ihnorton:
Allows python modules to depend on resources provided by other modules.

There are multiple stages in the module initialization process: registration, instantiation, and loading. Instantiation is when the scripted module is parsed, so any import dependencies must be resolveable at that time. However, instantiation is done alphabetically, with no dependency resolution -- because the module factory does not know anything about dependencies (the module itself supplies the dependencies function). Adding to PYTHONPATH during registration allows to import dependency-provided resources during instantiation.

ref: https://discourse.slicer.org/t/scripted-module-dependency-loaded-too-late-for-one-module-but-for-another-one-not/142

### DIFF
--- a/Base/QTGUI/qSlicerScriptedLoadableModuleFactory.cxx
+++ b/Base/QTGUI/qSlicerScriptedLoadableModuleFactory.cxx
@@ -43,6 +43,24 @@
 //----------------------------------------------------------------------------
 bool ctkFactoryScriptedItem::load()
 {
+#ifdef Slicer_USE_PYTHONQT
+  if (!qSlicerCoreApplication::testAttribute(qSlicerCoreApplication::AA_DisablePython))
+    {
+    // By convention, if the module is not embedded, "<MODULEPATH>" will be appended to PYTHONPATH
+    if (!qSlicerCoreApplication::application()->isEmbeddedModule(this->path()))
+      {
+      QDir modulePathWithoutIntDir = QFileInfo(this->path()).dir();
+      QString intDir = qSlicerCoreApplication::application()->intDir();
+      if (intDir ==  modulePathWithoutIntDir.dirName())
+        {
+        modulePathWithoutIntDir.cdUp();
+        }
+      qSlicerCorePythonManager * pythonManager = qSlicerCoreApplication::application()->corePythonManager();
+      pythonManager->appendPythonPaths(QStringList() << modulePathWithoutIntDir.absolutePath());
+      }
+    }
+#endif
+
   return true;
 }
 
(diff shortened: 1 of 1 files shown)


lassoan:
I haven't tested this but it looks good. Thank you!

fedorov:
Thank you Isaiah!

jcfr:
To be accurate, loading is done alphapatically but instanciation is done
based on the dependency resolution.

jcfr:
To follow up, the change looks good. 

Generally speaking, loading is done alphabetically (or as the module a discover on the file system) and instantiation is done based on the dependency resolution.

This is true for c++ based loadable module. Since the dependent shared library are automatically loaded by the system "DLL" loaded, getting the value of `dependencies` property is always possible.

Now for python scripted loadable module, if a Slicer scripted module `B` imports a Slicer scripted module `A` in the global namespace (at the top of the script), it needs to be available as soon as the script for `B` is imported to check its content.

For that reason, The patch of @ihnorton makes sense.






### Idea for later

Moving forward it could make sense to formalize the dependencies of each module (e.g in a Json file) and indicate:
* if Slicer explicitly require this module to function decently
* if the module logic need to be instantiated and observing the scene for the module to be useful

With that, we will be able to now which module to instantiate and in which order. 

The good news is that Qt5 offers a way associate metadata with Qt based c++ plugin and read these without having to load the associated library and its dependencies.

For python, we could have a companion json file.

ihnorton:
@jcfr I guess you are probably referring to the CTK internal `::load()` function (loads the DLL/so for shared library, and was previously a no-op for scripted), but I think the commit comment is accurate as written relative to what Slicer calls each of these steps. See https://github.com/Slicer/Slicer/blob/656c1670e8800b7787b7b21b9425b250903ea62a/Applications/SlicerApp/Main.cxx#L176-L217

(`::instantiateModules` loops over the keys alphabetically, but each call to `loadModule` does dependency resolution first)

jcfr:
> I think the commit comment is accurate

Indeed, it is. 

May be you could just specify that while the instantiation is done in arbitrary order the "loading" of each module is done based on the dependency graph captured by the `dependencies` property.

> instantiateModules loops over the keys alphabetically,

Each module is instantiated in [the order of the key](https://github.com/Slicer/Slicer/blob/eefeac9286f48552e8418a11f412b2823a09b407/Base/QTCore/qSlicerAbstractModuleFactoryManager.cxx#L374),

```
foreach (const QString& moduleName, d->RegisteredModules.keys())
{
```

then `instantiateModule` for each module will collect the dependent modules

See https://github.com/Slicer/Slicer/blob/eefeac9286f48552e8418a11f412b2823a09b407/Base/QTCore/qSlicerAbstractModuleFactoryManager.cxx#L394-L424

which are later [loaded in the right order](https://github.com/Slicer/Slicer/blob/eefeac9286f48552e8418a11f412b2823a09b407/Base/QTCore/qSlicerModuleFactoryManager.cxx#L115-L169):

```
  // Load the modules the module depends on.
  // There is no cycle check, so be careful
  foreach(const QString& dependency, instance->dependencies())
    {
    // no-op if the module is already loaded
    bool dependencyLoaded = this->loadModule(dependency, name);
    if (!dependencyLoaded)
      {
      qWarning() << "When loading module " << name << ", the dependency"
                 << dependency << "failed to be loaded.";
      return false;
      }
}
```

Base/QTGUI/qSlicerScriptedLoadableModuleFactory.cxx
```diff
@@ -77,6 +61,22 @@ qSlicerAbstractCoreModule* ctkFactoryScriptedItem::instanciator()
     }
 #endif
 
+  return true;
+}
+
+//----------------------------------------------------------------------------
+qSlicerAbstractCoreModule* ctkFactoryScriptedItem::instanciator()
```
cpinter: Can you please rename this function? I think there it is misspelled (I know it was the same before, but now that you are here...). Thanks!
https://en.wiktionary.org/wiki/instantiator

jcfr: It would make sense, that said we would have to rename the function in CTK .. https://github.com/commontk/CTK/search?utf8=%E2%9C%93&q=instanciator&type=

This spelling originate from [french](https://fr.wiktionary.org/wiki/instanciation) ... 

To avoid breaking other project, I would prefer not to change it.

cpinter: If it's too much hassle, then I'm OK with not fixing it.

ihnorton:
committed in:
https://github.com/Slicer/Slicer/commit/abe2f0bd6911b5a0a15e69701802ab7ffdeefcf2

PR #706, opened 2017-04-21
ENH: Install SSL dlls into bin folder

lassoan:
SSL library loading failure came up on my computer again: SSL libraries failed to load on all installed Slicer releases (even those that worked before). I've confirmed that copying libeay32.dll, ssleay32.dll into SlicerApp-real.exe's folder solves the problem. Interestingly, a system restart solves the problem, too (until the problem comes up again).

This commit makes the SSL libraries installed in SlicerApp-real.exe's folder (c:\Program Files\Slicer 4.7.0-2017-04-20\bin) instead of lib folder (c:\Program Files\Slicer 4.7.0-2017-04-20\lib\Slicer-4.7\ssleay32.dll). This should solve this problem permanently.


lassoan:
Committed r25942

PR #707, opened 2017-04-23
ENH: Generate (Extension)Config.cmake

lassoan:
Generate (Extension)Config.cmake to allow an extension import targets from another extension.

This solves the issue of PerkTutor extension's loadable modules depending on Sequences extension loadable modules.


pieper:
I didn't test this, but it sounds good!

CMake/SlicerExtensionGenerateConfig.cmake
```diff
@@ -0,0 +1,77 @@
+# Generate the <Extension>Config.cmake file in the build tree.
+# This file makes it possible an extension to depend on targets defined
```
jcfr: ... This file makes it possible **for** an extension ...

jcfr:
👍 

I suggest to:
* look at https://github.com/SBU-BMI/SlicerOpenCV/blob/master/CMake/SlicerOpenCVConfig.cmake.in 
* provide a way to disable generation of the file for extension willing to provide their own

CMake/SlicerExtensionGenerateConfig.cmake
```diff
@@ -1,9 +1,20 @@
...
 # This file is based on SlicerGenerateSlicerConfig.cmake.
 #
-# Example: add these lines to the CMakeLists.txt to use vtkMRMLSequenceNode
+# Example:
+#
+# Add these lines near the end of top-level CMakeLists.txt in Sequences extension
+# to allow other extensions to use it.
+#
+#   # Pass some custom content to the SequencesConfig.cmake file (optional)
+#   set(Sequences_CUSTOM_CONFIG "MESSAGE(STATUS \"Test\")")
```
jcfr: Nitpick:

->  `set(Sequences_CUSTOM_CONFIG "message(STATUS \"Test\")")`

lassoan:
> look at https://github.com/SBU-BMI/SlicerOpenCV/blob/master/CMake/SlicerOpenCVConfig.cmake.in

It is already very similar. What we don't have in the common script:
* Setting of SlicerOpenCV_SHARE_DIR. I was not sure what packaging has to do with it and how to implement it generally for all extensions, but I think it's not necessary if extensions depend on simple modules. For cases when the extension exposes shared non-Slicer libraries, such as OpenCV, the config file can be hand-crafted (or the config file generation scripts can be improved in the Slicer core, without impacting extensions).
* Prevent importing multiple times (using SlicerOpenCV_TARGETS_IMPORTED). Is this necessary? I've tried to call FIND_PACKAGE(Sequences) multiple times and there did not seem to be any problem.

> provide a way to disable generation of the file for extension willing to provide their own

Default <Extension>Config.cmake is generated by adding this line to the top-level CMakeLists.txt file:
include(${Slicer_EXTENSION_GENERATE_CONFIG})
If the line is not added then the extension can generate its own config file.

I've also added ${EXTENSION_NAME}_CUSTOM_CONFIG variable so that extensions can inject some custom content to the config file to allow customization it generating a config from scratch.

lassoan:
Everything seems to work well (I'll integrate in a few hours if there is no objection). The only limitation is that to test extensions that depend on other extensions, you would need to provide --launcher-additional-settings of the current extension and all extensions that it depends on.

Potential solution:
* Make slicer launcher accept multiple files in --launcher-additional-settings argument (really easy; see implementation here: https://github.com/lassoan/AppLauncher/commit/18034ddf251b3a9856487a13eff1bf96e2f0bbd2)
* Modify SlicerBlockAdditionalLauncherSettings.cmake to pass additional settings of all extensions that the extension depends on.

@jcfr What do you think, should I implement this or do you have better suggestion?

lassoan:
I've committed the changes in this pull request into r25944. Support for multiple additional launcher settings will be submitted in a separate pull request.

jcfr:
> Support for multiple additional launcher settings will be submitted in a separate pull request.

Makes sense.

PR #710, opened 2017-04-25
BUG: Fix ConvertDeviceToXYZ() 1 pixel offset

vovythevov:
The issue appeared when using an annotation. When the annotation point is
dragged at the bottom of a slice view, the annotation Z position would
jump 1 pixel higher.
For example in the axial view, the Z coordinate of the annotation would
jump from 0.0 to 1.0 (when no module is loaded).

Here is a small python code exerpt that can be used to showcase the bug:
```Python
def PrintP1(caller, event):
    p1 = [0, 0, 0]
    caller.GetPosition1(p1)
    print(p1)

ruler = slicer.mrmlScene.AddNode(slicer.vtkMRMLAnnotationRulerNode())
ruler.SetPosition1([0, 0, 0])
ruler.SetPosition2([10, 0, 0])
ruler.AddObserver(vtk.vtkCommand.ModifiedEvent, PrintP1)
```

Fixing the computation of Z in ConvertDeviceToXYZ() fixes this.

### DIFF
--- a/Libs/MRML/DisplayableManager/vtkMRMLAbstractSliceViewDisplayableManager.cxx
+++ b/Libs/MRML/DisplayableManager/vtkMRMLAbstractSliceViewDisplayableManager.cxx
@@ -96,7 +96,7 @@ void vtkMRMLAbstractSliceViewDisplayableManager::ConvertDeviceToXYZ(
   int numberOfRows = sliceNode->GetLayoutGridRows();
 
   float tempX = x / windowWidth;
-  float tempY = (windowHeight - y) / windowHeight;
+  float tempY = (windowHeight - 1 - y) / windowHeight;
 
   float z = floor(tempY*numberOfRows)*numberOfColumns + floor(tempX*numberOfColumns);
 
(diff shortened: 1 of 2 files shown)


pieper:
Sounds reasonable.  Did you try it with various light box modes?

vovythevov:
@pieper: With the annotation, no as they are not supported with lightbox.
Otherwise, I manually looked at the data probe value in light box and normal mode and it seems good.

pieper:
@vovythevov good point about annotations being disabled in lightbox (I remember now there are other issues there).  Using the dataprobe to confirm makes sense.

lassoan:
Looks good, thank you!

The computation is correct in most other displayable managers except vtkMRMLMarkupsDisplayableManager2D.cxx. Could you please apply the same fix to vtkMRMLMarkupsDisplayableManager2D.cxx? Thanks!

jcfr:
> vtkMRMLMarkupsDisplayableManager2D.cxx

@lassoan  Are you talking about OnClickInRenderWindowGetCoordinates ?

lassoan:
No, not in OnClickInRenderWindowGetCoordinates but here:

https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/Markups/MRMLDM/vtkMRMLMarkupsDisplayableManager2D.cxx#L1270

and (just a comment now, but somebody may uncomment it in the future)

https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/Markups/MRMLDM/vtkMRMLMarkupsDisplayableManager2D.cxx#L1182

vovythevov:
@lassoan: Good catch ! Thanks !
I updated my commit to include these.

vovythevov:
Fixed in [r25981](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25981)

PR #715, opened 2017-04-27
BUG: updating macro names in the build system

dzenanz:
Replacing slicerMacroBuildQtModule by slicerMacroBuildLoadableModule, bug #3332
Rename SlicerMacroBuildQtModule.cmake into SlicerMacroBuildLoadableModule.cmake, bug #4180
For backwards compatibility, there is a file with old name which just includes the new file


CMake/SlicerMacroBuildLoadableModule.cmake
```diff
@@ -0,0 +1,326 @@
...
+    endforeach()
+  endif()
+
+endmacro()
+
+
+#
+# Deprecated macro - Will be removed for 4.4
+#
+macro(slicerMacroBuildQtModule)
```
vovythevov: I would consider keeping this in the SlicerMacroBuildQtModule instead of moving it here. This way, when you do remove the macro, you have only one file to remove.

CMake/SlicerMacroBuildQtModule.cmake
```diff
@@ -1,325 +1 @@
...
-  set(QTMODULE_OPTIONS)
-  foreach(option ${options})
-    if(${QTMODULE_${option}})
-      list(APPEND QTMODULE_OPTIONS ${option})
-    endif()
-  endforeach()
-
-  slicerMacroBuildLoadableModule(${QTMODULE_ARGS} ${QTMODULE_OPTIONS})
-endmacro()
+include(${CMAKE_CURRENT_LIST_DIR}/SlicerMacroBuildLoadableModule.cmake)
```
vovythevov: I would also think about adding a deprecation message to be displayed whenever this file is included. This would "force" people to include the correct (SlicerMacroBuildLoadableModule) cmake file.
@jcfr: What do you think ?

lassoan:
There are many synonyms for module types. I'm not sure that Loadable module is more accurate than Qt module.

For example, one categorization is this:
* that there are CLI and loadable modules
* within loadable modules, there are two types: Qt loadable modules and Python scripted loadable modules

I don't like the inconsistency in module node type naming, so there is work to do around here, but it's not as simple as renaming a few macros. Overall, I think it would be better to improve the developer documentation first, make sure the naming is consistent there, and then make sure the naming is consistent in the source code as well.

CMake/SlicerMacroBuildLoadableModule.cmake
dzenanz: This [bug report](http://www.na-mic.org/Bug/view.php?id=3332) calls for renaming.

dzenanz:
I thought that he division of module types was:
* CLI
* Loadable
* Scripted

and that this division has been active for quite a while.

lassoan:
There are lots of file and class names with "ScriptedLoadable". Are we going to change all those, too?

dzenanz:
@jcfr Decide what to do (potentially at the next developer hangout) and either merge or close the pull request. Unless somebody wants to take over, in which case they are welcome.

lassoan:
@dzenanz I appreciate your effort and would like to see more consistent naming, I would like to just make sure we are moving to the right direction.

In the source code almost everywhere this module naming convention is used:
* CLI
* Loadable
* ScriptedLoadable

This is consistent with naming convention of most classes that have a scripted variant, for example
EditorEffect, ScriptedEditorEffect
qSlicerFileWriter,qSlicerScriptedFileWriter
vtkMRMLDisplayableManager, vtkMRMLScriptedDisplayableManager

Therefore, your change of SlicerMacroBuildQtModule.cmake => SlicerMacroBuildLoadableModule.cmake makes naming more consistent.

If everybody agrees on the CLI, Loadable, ScriptedLoadable naming convention then this PR is a step to the right direction and so it should be merged.

pieper:
If we're going back to clean up the naming I'd say that "ScriptedLoadable"
is redundant and should always be "Scripted" (as in the source code
Modules/{CLI,Loadable,Scripted}).

lassoan:
There are ScriptedLoadable and ScriptedCLI modules as well (or at least there were  - see https://www.slicer.org/wiki/Documentation/Nightly/Developers/SlicerExecutionModel/Python - and probably we will make them work again), so actually ScriptedLoadable name is not redundant.

Scripted subdirectory name is correct if it contains all scripted modules (both Loadable and CLI).

jcfr:
> better to improve the developer documentation first, make sure the naming is consistent there, and then make sure the naming is consistent in the source code as well.

Agreed

I will add this as a topic for next week developer meeting. See https://www.slicer.org/wiki/Developer_Meetings/20170502

jcfr:
@dzenanz Thanks again for your help, we will discuss this next week and move forward with your contribution afterward.

pieper:
Yes, it would be very nice to revive the ability to use the
SlicerExecutionModel from python, that was a very productive way to develop
modules.  But as they existed in Slicer3 they aren't really proper CLI
modules because they have access to the full Slicer python environment and
could manipulate MRML and GUI code directly.  So they are more of a
shortcut for creating the GUI of a Scripted Module than they are a proper
CLI module.  A distinguishing feature of a CLI module is the ability to use
it from a batch script.

I look forward to discussing the concepts and coming up with some concrete
definitions, even if the borders between categories are somewhat arbitrary.

jcfr:
File has been renamed. Change integrated in [r26475](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26475)

PR #719, opened 2017-04-28
FIX: Minor - Check for null node in markup widget setMRMLFiducialNode

vovythevov:


lassoan:
Thank you, LGTM.

vovythevov:
Fixed in [r25990](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25990)

PR #721, opened 2017-05-04
ENH: Allow disabling slice view interactions

lassoan:
For interventional applications it is often necessary to **prevent the user from manipulating a slice view** (change selected slice, adjust window width/level, etc).

SetActionEnabled/GetActionEnabled methods have been added to specify which actions are allowed in a slice view (default behavior is the same as before: all actions are allowed).

Example:
```
interactorStyle = slicer.app.layoutManager().sliceWidget('Red').sliceView().sliceViewInteractorStyle()
interactorStyle.SetActionEnabled(interactorStyle.BrowseSlice, False)
```


pieper:
LGTM,  +1

lassoan:
Thanks for the review. Merge in r25997.

PR #722, opened 2017-05-05
BUG: Update Volume loading logic to consider default nodes properties

jcfr:
This commit enables customization of the LabelMapVolumeDisplayNode
attributes by ensuring the custom properties set on the corresponding
default node are effectively re-used.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLScene.cxx
+++ b/Libs/MRML/Core/vtkMRMLScene.cxx
@@ -1250,7 +1250,8 @@ vtkMRMLNode* vtkMRMLScene::AddNode(vtkMRMLNode *n)
 }
 
 //------------------------------------------------------------------------------
-vtkMRMLNode* vtkMRMLScene::AddNewNodeByClass(std::string className, std::string nodeName)
+vtkMRMLNode* vtkMRMLScene::AddNewNodeByClass(
+    std::string className, std::string nodeBaseName /* = "" */)
 {
   if (className.empty())
     {
@@ -1259,9 +1260,9 @@ vtkMRMLNode* vtkMRMLScene::AddNewNodeByClass(std::string className, std::string 
     }
   vtkSmartPointer<vtkMRMLNode> nodeToAdd =
       vtkSmartPointer<vtkMRMLNode>::Take(this->CreateNodeByClass(className.c_str()));
-  if (!nodeName.empty())
+  if (!nodeBaseName.empty())
     {
-    nodeToAdd->SetName(nodeName.c_str());
+    nodeToAdd->SetName(nodeBaseName.c_str());
     }
   return this->AddNode(nodeToAdd);
 }
--- a/Libs/MRML/Core/vtkMRMLScene.cxx
+++ b/Libs/MRML/Core/vtkMRMLScene.cxx
@@ -1250,6 +1250,23 @@ vtkMRMLNode* vtkMRMLScene::AddNode(vtkMRMLNode *n)
 }
 
 //------------------------------------------------------------------------------
+vtkMRMLNode* vtkMRMLScene::AddNewNodeByClass(std::string className, std::string nodeName)
+{
+  if (className.empty())
+    {
+    vtkErrorMacro("AddNewNodeByClass: className is an emptry string");
+    return NULL;
+    }
+  vtkSmartPointer<vtkMRMLNode> nodeToAdd =
+      vtkSmartPointer<vtkMRMLNode>::Take(this->CreateNodeByClass(className.c_str()));
+  if (!nodeName.empty())
+    {
+    nodeToAdd->SetName(nodeName.c_str());
+    }
+  return this->AddNode(nodeToAdd);
+}
+
+//------------------------------------------------------------------------------
 void vtkMRMLScene::NodeAdded(vtkMRMLNode *n)
 {
   this->InvokeEvent(this->NodeAddedEvent, n);
--- a/Libs/MRML/Core/vtkMRMLScene.h
+++ b/Libs/MRML/Core/vtkMRMLScene.h
@@ -178,12 +178,16 @@ public:
   /// the scene. It ensures that the new node properties are initialized
   /// considering its default nodes.
   ///
-  /// The method calls CreateNodeByClass(), AddNode() and vtkMRMLNode::SetName().
+  /// The method calls CreateNodeByClass(), vtkMRMLNode::SetName() and AddNode().
   ///
-  /// \sa CreateNodeByClass()
-  /// \sa AddDefaultNode(vtkMRMLNode*)
-  /// \sa AddNode()
-  vtkMRMLNode* AddNewNodeByClass(std::string className, std::string nodeName = "");
+  /// \note Instead of calling SetName() after creating the node, prefer
+  /// passing \a nodeBaseName, indeed the method AddNode() ensures that the
+  /// final node name is unique by appending a suffix.
+  ///
+  /// \sa CreateNodeByClass(), vtkMRMLNode::SetName(), AddNode()
+  /// \sa AddDefaultNode()
+  /// \sa GenerateUniqueName()
+  vtkMRMLNode* AddNewNodeByClass(std::string className, std::string nodeBaseName = "");
 
   /// Add a copy of a node to the scene.
   vtkMRMLNode* CopyNode(vtkMRMLNode *n);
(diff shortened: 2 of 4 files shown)


jcfr:
I am also planning on updating the rest of the volume factories to consistently use `CreateNodeByClass` instead of `vtkNew` (which doesn't initialize the default properties). 

Before doing so, I wanted to get some feedback on the approach. Thanks

jcfr:
A use case is to for example overwrite the `SliceIntersectionThickness` in an extension using code like this one:

```
vtkNew<vtkMRMLLabelMapVolumeDisplayNode> defaultLabelMapVolumeDisplayNode;
defaultLabelMapVolumeDisplayNode->SetSliceIntersectionThickness(1);
d->MRMLScene->AddDefaultNode(defaultLabelMapVolumeDisplayNode.GetPointer());
```

jcfr:
An other approach would be to update the `vtkMRMLNodeNewMacro` macro and provide a custom implementation for the `New` function.

lassoan:
Looks good to me.

Another approach (not necessarily a replacement; it can be used in addition to default node customization) is to use composite file extensions to customize file loading behavior.

For example Sequences module stores 4D (3D+t) volume sequence in a nrrd file. However, when you load this sequence, you don't want to the user to manually specify to interpret this nrrd file as a volume sequence. Instead, we use a composite file extension: .seq.nrrd. It indicates that the file is a nrrd file and any nrrd-compatible software can read it. However, .seq specifies a sub-type that the Sequence module's file reader plugin recognizes. Slicer automatically chooses that file reader as default that has the best fitting file extensions (taking into account composite extension).

Another example, where composite file extensions should be used (but unfortunately not used yet) is Markups fiducial storage in fcsv files. The files could be nicely loaded by Excel, but due to the unrecognized file extensions, it will not be loaded. Using a file extension like .fid.csv would solve the issue.

Last example: Slicer tries to guess if a volume is a labelmap by searching for arbitrary strings in the file name (such as -label), which is not very robust and have unintended side effects. Instead, we could save labelmap volume nodes as .label.nrrd or something similar.

Similarly for any other custom volume nodes we could specify composite file extensions. To load and display them in specific ways.

lassoan:
> An other approach would be to update the vtkMRMLNodeNewMacro macro and provide a custom implementation for the New function.

That would just add one more new macro (there is already ::New(), vtkNew, and vtkSmartPointer) with very slightly different behavior. I don't think it would make things simpler, more clear, or more robust. Also, it would not be available in Python, so it would be more difficult to port code between C++/Python.

lassoan:
I see that syntax of using CreateNodeByClass is too complex, though. Instead of a macro, we could add a new method:

    vtkMRMLNode* vtkMRMLScene::AddNewNodeByClass(const char* className)

The method would already add the node to the scene, so the scene would already take ownership and this there would be no need for using smart pointers. It could be used like this:

    vtkMRMLLabelMapVolumeDisplayNode* lmdisplayNode = vtkMRMLLabelMapVolumeDisplayNode::SafeDownCast(
      scene->AddNewNodeByClass("vtkMRMLLabelMapVolumeDisplayNode"))

This would also work just as well from Python (without the need of calling node.UnRegister()).
The client code would be also simpler as there would be no need for calling create and add node, just one.

The name could be shortened from AddNewNodeByClass to AddNodeByClass, but I'm not sure if it makes it clear enough that a new node is created.

jcfr:
I started to add `AddNewNodeByClass` and just realized it would change the behavior for cases like these ones.

That said, setting the properties (e.g Name, CenterImage, ...) after the node is added should not be an issue. Indeed, these node are all added to the mini-scene and the NodeModified event wouldn't have any overhead ... 

```
vtkNew<vtkMRMLDiffusionWeightedVolumeNode> dwiNode;
dwiNode->SetName(volumeName.c_str());
nodeSet.Scene->AddNode(dwiNode.GetPointer());
dwiNode->SetAndObserveDisplayNodeID(dwdisplayNode->GetID());
```

or 

```
vtkNew<vtkMRMLNRRDStorageNode> storageNode;
  storageNode->SetCenterImage(options & vtkSlicerVolumesLogic::CenterImage);
  nodeSet.Scene->AddNode(storageNode.GetPointer());
  vectorNode->SetAndObserveStorageNodeID(storageNode->GetID());
```

lassoan:
You are right, if you set node name before then Slicer may change the name to make it unique (it is debatable if this is a good default behavior, but we won't change it now). So, it could be useful to add an optional "name" argument:

    vtkMRMLNode* vtkMRMLScene::AddNewNodeByClass(const char* className, const char* nodeName = NULL)

If there is a large amount of property changes then Start/EndModify can be used to reduce number of Modified() events to 1.

jcfr:
@lassoan  Et voila. That should do it. When you have a chance, let me know what you think.

Libs/MRML/Core/vtkMRMLScene.h
```diff
@@ -162,6 +172,19 @@ class VTK_MRML_EXPORT vtkMRMLScene : public vtkObject
   /// into the already existing singleton node. That node is then returned.
   vtkMRMLNode* AddNode(vtkMRMLNode *nodeToAdd);
 
+  /// \brief Instantiate and add a node to the scene.
+  ///
+  /// This is the preferred way to create and add a new node to
+  /// the scene. It ensures that the new node properties are initialized
+  /// considering its default nodes.
+  ///
+  /// The method calls CreateNodeByClass(), AddNode() and vtkMRMLNode::SetName().
```
lassoan: The actual order is CreateNode, SetName, AddNode

Libs/MRML/Core/vtkMRMLScene.h
```diff
@@ -162,6 +172,19 @@ class VTK_MRML_EXPORT vtkMRMLScene : public vtkObject
   /// into the already existing singleton node. That node is then returned.
   vtkMRMLNode* AddNode(vtkMRMLNode *nodeToAdd);
 
+  /// \brief Instantiate and add a node to the scene.
+  ///
+  /// This is the preferred way to create and add a new node to
+  /// the scene. It ensures that the new node properties are initialized
+  /// considering its default nodes.
```
lassoan: Add a note about unique name generation, using the name that is provided. Maybe the name argument could be called "nodeBaseName" to emphasize that this string will be used to create a unique node name in the scene.

Libs/MRML/Core/vtkMRMLScene.cxx
```diff
@@ -1250,6 +1250,23 @@ vtkMRMLNode* vtkMRMLScene::AddNode(vtkMRMLNode *n)
 }
 
 //------------------------------------------------------------------------------
+vtkMRMLNode* vtkMRMLScene::AddNewNodeByClass(std::string className, std::string nodeName)
```
lassoan: It helps if you indicate default argument values in the cxx file - for example
 ...AddNewNodeByClass(std::string className, std::string nodeName /* ="" */)

jcfr:
@lassoan If the last commit looks good, I will squash the last two and integrate.

Libs/MRML/Core/vtkMRMLScene.h
```diff
@@ -178,12 +178,16 @@ class VTK_MRML_EXPORT vtkMRMLScene : public vtkObject
...
-  /// The method calls CreateNodeByClass(), AddNode() and vtkMRMLNode::SetName().
+  /// The method calls CreateNodeByClass(), vtkMRMLNode::SetName() and AddNode().
   ///
-  /// \sa CreateNodeByClass()
-  /// \sa AddDefaultNode(vtkMRMLNode*)
-  /// \sa AddNode()
-  vtkMRMLNode* AddNewNodeByClass(std::string className, std::string nodeName = "");
+  /// \note Instead of calling SetName() after creating the node, prefer
+  /// passing \a nodeBaseName, indeed the method AddNode() ensures that the
+  /// final node name is unique by appending a suffix.
```
lassoan: Final node name is unique in the scene by appending a suffix if needed.

jcfr: Thanks. All set.

jcfr:
Integrated in [r26000](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26000) 🎆 

and [r26001](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26001)

PR #726, opened 2017-05-21
Fix typo in License.txt

mhalle:
Likely a Word character set issue.


pieper:
Generally it's not a good idea to change license text, even to correct typos but this seems harmless enough so let's go ahead.

mhalle:
This wasn't even a typo.  It was a formatting error that resulted from a conversion of a single close quote in the Word original into an ASCII single quote.

The license didn't change, we just made the ASCII version of it more faithful to its original format.

--Mike

jcfr:
Integrated in [r26043](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26043)

PR #730, opened 2017-05-26
ENH: Added schema file support for defining table column properties

lassoan:
Data type, default value, long name, description, unit, and any number of custom properties can now be defined for each column in a vtkMRMLTableNode. The information is currently used for:
* Long name, description, and unit are shown in table view.

* Data type is used for creating numeric arrays when reading the table.csv
* Default value is used for initializing values in new rows.

Properties are stored in a schema file, which is a CSV or TSV file with the composite file extension .schema.csv (or .schema.csv or .schema.txt) with the same base filename as the data table file.


pieper:
Very nice!

lassoan:
Committed r26052

PR #732, opened 2017-06-08
ENH: add vtkNRRDWriter helper function, and toolTip criteria in findChildren

ihnorton:
These are some small utilities/refactoring I needed recently.

### DIFF
--- a/Libs/vtkTeem/vtkNRRDWriter.cxx
+++ b/Libs/vtkTeem/vtkNRRDWriter.cxx
@@ -172,22 +172,9 @@ int vtkNRRDWriter::VTKToNrrdPixelType( const int vtkPixelType )
     }
   }
 
-
-
-//----------------------------------------------------------------------------
-// Writes all the data from the input.
-void vtkNRRDWriter::WriteData()
-{
-  this->WriteErrorOff();
-  if (this->GetFileName() == NULL)
-    {
-    vtkErrorMacro("FileName has not been set. Cannot save file");
-    this->WriteErrorOn();
-    return;
-    }
-
+void* vtkNRRDWriter::MakeNRRD()
+  {
   Nrrd *nrrd = nrrdNew();
-  NrrdIoState *nio = nrrdIoStateNew();
   int kind[NRRD_DIM_MAX];
   size_t size[NRRD_DIM_MAX];
   unsigned int nrrdDim, baseDim, spaceDim;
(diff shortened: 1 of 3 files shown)


Libs/vtkTeem/vtkNRRDWriter.cxx
```diff
@@ -243,9 +230,9 @@ void vtkNRRDWriter::WriteData()
                       << this->GetFileName() << ":\n" << err);
     // Free the nrrd struct but don't touch nrrd->data
     nrrd = nrrdNix(nrrd);
-    nio = nrrdIoStateNix(nio);
+    //nio = nrrdIoStateNix(nio); // TODO?
```
pieper: Shouldn't you just delete this line since the nio is in the WriteData method?

ihnorton: Done. Thanks for the catch.

pieper:
LGTM - 👍

Libs/vtkTeem/vtkNRRDWriter.h
```diff
@@ -74,6 +74,8 @@ class VTK_Teem_EXPORT vtkNRRDWriter : public vtkWriter
   /// Method to set unit for each axis
   void SetAxisUnit(unsigned int axis, const char* unit);
 
+  void* MakeNRRD();
```
jcfr: docstring

Base/Python/slicer/util.py
```diff
@@ -156,7 +156,7 @@ def showStatusMessage(message, duration = 0):
   mw.statusBar().showMessage(message, duration)
   return True
 
-def findChildren(widget=None, name="", text="", title="", className=""):
```
jcfr: Out of curiosity, what is the use case ? Is it just for convenience when using the python interactor ? 

I would advise against using the tooltip for production code ... it would be too brittle (e.g could break unexpectedly after fixing a typo in a tooltip).

ihnorton: It's the only reliable way that I found to get a widget from a CLI GUI. Yes, only useful for testing/prototyping.

lassoan: We could set the qt object name based on the parameter name. It would be a very simple and reliable method for finding widgets.
@ihnorton Would you be interested in implementing that? It's probably just adding a few lines in qSlicerCLIModuleUIHelper::createTagWidget.

ihnorton: Nice, thanks for the suggestion, that works well. See fe3669f
I removed the toolTip commit.

pieper: 👍

ihnorton:
If there are no objections I'll merge this and #733 later today.

jcfr:
Sounds good. Thanks.
Jc

ihnorton:
integrated:
https://github.com/Slicer/Slicer/commit/40a264477687d5c2eb492c9249221025dd8ffd52
https://github.com/Slicer/Slicer/commit/b2dc90231c09b19f42a98c2ca6c046e37bd0daec

PR #733, opened 2017-06-08
RFC: "ENH: add 'AllowMemoryTransfer' CLI setting"

ihnorton:
Request for comment...

This allows a CLI to control whether in-memory (MRMLID) transfer is attempted.
Implemented as a parameter for simplicity. Could be implemented at the execution model level, but
not sure it's worth the churn (etc.) that would cause.

### DIFF
--- a/Base/QTCLI/qSlicerCLIModule.cxx
+++ b/Base/QTCLI/qSlicerCLIModule.cxx
@@ -102,6 +102,11 @@ vtkMRMLAbstractLogic* qSlicerCLIModule::createLogic()
     logic->DeleteTemporaryFilesOff();
     }
 
+  if (d->Desc.GetParameterDefaultValue("AllowInMemoryTransfer") == "false")
+    {
+    logic->SetAllowInMemoryTransfer(0);
+    }
+
   return logic;
 }
 
--- a/Base/QTCLI/vtkSlicerCLIModuleLogic.h
+++ b/Base/QTCLI/vtkSlicerCLIModuleLogic.h
@@ -76,6 +76,10 @@ public:
   void SetDeleteTemporaryFiles(int value);
   int GetDeleteTemporaryFiles() const;
 
+  // Control use of in-memory data transfer by this specific CLI.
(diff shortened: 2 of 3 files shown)


Base/QTCLI/qSlicerCLIModule.cxx
```diff
@@ -102,6 +102,11 @@ vtkMRMLAbstractLogic* qSlicerCLIModule::createLogic()
     logic->DeleteTemporaryFilesOff();
     }
 
+  if (d->Desc.GetParameterDefaultValue("AllowMemoryTransfer") == "false")
```
jcfr: @jbvimort  That will be an other instance of the code to update from `GetParameterDefaultValue` to `GetParameterValue` when you are done with you changes

Base/QTCLI/vtkSlicerCLIModuleLogic.h
```diff
@@ -76,6 +76,9 @@ class Q_SLICER_BASE_QTCLI_EXPORT vtkSlicerCLIModuleLogic :
   void SetDeleteTemporaryFiles(int value);
   int GetDeleteTemporaryFiles() const;
 
+  void SetAllowMemoryTransfer(int value);
```
jcfr: Add docstring

jcfr:
Nice.

In a next iteration, I wonder if we should improve the SEM spec with an extra per input/output flag.

Base/QTCLI/vtkSlicerCLIModuleLogic.h
lassoan: I think `AllowInMemoryTransfer` would be a bit more clear name. Otherwise I like having this option (and would even consider removing in-memory transfer at some point).

ihnorton: Ok, changed.

ihnorton:
integrated: https://github.com/Slicer/Slicer/commit/4f7b44b067767a001abb30cd85b39d11fd4fcbe8

PR #734, opened 2017-06-14
Dcmtk for scalar volume

pieper:
For discussion with @fedorov

### DIFF
--- a/Applications/SlicerApp/Testing/Python/DICOMReaders.py
+++ b/Applications/SlicerApp/Testing/Python/DICOMReaders.py
@@ -65,25 +65,6 @@ class DICOMReadersLogic(ScriptedLoadableModuleLogic):
   def __init__(self):
     ScriptedLoadableModuleLogic.__init__(self)
 
-  def compareNodes(self,volumeNode1,volumeNode2):
-    """
-    Given two mrml volume nodes, return true of the numpy arrays have identical data
-    and other metadata matches.  Returns empty string on match, otherwise
-    a string with a list of differences separated by newlines.
-    """
-    volumesLogic = slicer.modules.volumes.logic()
-    comparison = ""
-    comparison += volumesLogic.CompareVolumeGeometry(volumeNode1, volumeNode2)
-    image1 = volumeNode1.GetImageData()
-    image2 = volumeNode2.GetImageData()
-    if image1.GetScalarType() != image2.GetScalarType():
-      comparison += "First volume is %s, but second is %s" % (image1.GetScalarTypeAsString(), image2.GetScalarTypeAsString())
-    array1 = slicer.util.array(volumeNode1.GetID())
-    array2 = slicer.util.array(volumeNode2.GetID())
-    if not numpy.all(array1 == array2):
-      comparison += "Pixel data mismatch\n"
-    return comparison
-
 
 class DICOMReadersTest(ScriptedLoadableModuleTest):
   """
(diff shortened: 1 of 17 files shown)


Applications/SlicerApp/Testing/Python/DICOMReaders.py
```diff
@@ -0,0 +1,189 @@
...
+  """This class should implement all the actual
+  computation done by your module.  The interface
+  should be such that other python code can import
+  this class and make use of the functionality without
+  requiring an instance of the Widget
+  """
+  def __init__(self):
+    ScriptedLoadableModuleLogic.__init__(self)
+
+  def compareNodes(self,volumeNode1,volumeNode2):
```
fedorov: Should this be a bit more extensive? I think it is reasonable to expect a situation where reading of the attributes related to the image geometry can be interpreted differently by different readers (specifically, this can become ambiguous in interpreting MF, or handling slice thickness in single frame objects).

Libs/MRML/Core/vtkMRMLVolumeArchetypeStorageNode.h
```diff
@@ -82,6 +83,13 @@ class VTK_MRML_EXPORT vtkMRMLVolumeArchetypeStorageNode : public vtkMRMLStorageN
   /// instance to turn off compression.
   virtual void ConfigureForDataExchange();
 
+  ///
+  /// Provide a uniform way to populate the volume nodes's itk
+  /// metadatadictionary from the reader.  Since itk::MetaDataDictionary
+  /// is not exposed in python, this method allows it to be set indirectly
+  /// using only wrapped types.
+  static void SetMetaDataDictionaryFromReader(vtkMRMLVolumeNode*, vtkITKArchetypeImageSeriesReader*);
```
fedorov: Just a side note - I didn't know about this feature, or where it is used, but I searched, and it seems that it is used in MRMLIDIO, and specifically for communicating diffusion-related attributes: https://github.com/Slicer/Slicer/blob/master/Libs/MRML/IDImageIO/itkMRMLIDImageIO.cxx#L313-L328. I wonder if this is working, or if anyone is using it. @ihnorton FYI!

Modules/Scripted/DICOM/DICOM.py
```diff
@@ -67,14 +69,26 @@ def __init__(self, parent):
...
-    # set the dicom pre-cache tags once all plugin classes have been initialized
-    qt.QTimer.singleShot(0, DICOMLib.setDatabasePrecacheTags)
+      qt.QTimer.singleShot(0, self.performPostModuleDiscoveryTasks)
 
   def setup(self):
     pluginHandlerSingleton = slicer.qSlicerSubjectHierarchyPluginHandler.instance()
     pluginHandlerSingleton.registerPlugin(slicer.qSlicerSubjectHierarchyDICOMPlugin())
 
+  def performPostModuleDiscoveryTasks(self):
+    """Since dicom plugins are discovered while they application
```
fedorov: typo: they -> the ?

Applications/SlicerApp/Testing/Python/DICOMReaders.py
pieper: Thanks for catching this - yes, I meant to flesh that out and and realize now that it would be good to add another option to the "approach" options where we load both volume and then compares them.

pieper: Realized of course I should be using the logic method for this comparison.

https://github.com/Slicer/Slicer/blob/82c5b5a8052cf802dbfd2cb6ea1c2dc1591798e1/Modules/Loadable/Volumes/Logic/vtkSlicerVolumesLogic.h#L211-L223

Libs/MRML/Core/vtkMRMLVolumeArchetypeStorageNode.h
ihnorton: Yes, it's used any time a CLI runs on a DWI NRRD.

pieper:
Merged as of r26131

PR #736, opened 2017-06-16
ENH: Removed legacy BSplineToDeformationField module

lassoan:
BSplineToDeformationField could convert certain bspline transform type to displacement field. The module was limited what transform types it can handle.

Conversion for all transform types was already available for users in the transforms module (and for other modules in the transforms module logic).

Remove BSplineToDeformationField module to reduce user confusion (see https://discourse.slicer.org/t/bspline-to-deformation-field-error/503) and reduce overall Slicer core size and maintenance workload.


fedorov:
The reason this module was added was to support EMSegmenter needs. I don't know what is the current status of EMSegmenter and whether it is still used there.

lassoan:
Thanks for the hint. I've checked the source code of all current extensions in the ExtensionIndex and none of them use this module.

EMSegmenter uses BRAINSTransformConvert now. BRAINSTransformConvert essentially does the same, converts transforms to displacement fields, but works for current bspline transform format.

fedorov:
Great - sorry for the false alarm! I am glad this tool can finally be phased out.

lassoan:
Committed as r26100

PR #737, opened 2017-06-16
ENH: Removed terminology support from Colors module

lassoan:
Terminologies are now managed by a dedicated module (Terminologies) and used in segmentations modules.

Fixes #4320 (https://issues.slicer.org/view.php?id=4320)


lassoan:
Committed r26105

PR #738, opened 2017-06-21
BUG: Fixed warping of markup fiducial labels

lassoan:
Warping of markup fiducial labels was caused by using perspective projection in overlay renderer's camera. See details in https://github.com/commontk/CTK/pull/720.

Before the fix:
image

After the fix:
image

By switching to parallel projection, default scaling of widgets have to be adjusted, which is addressed in this pull request.

Merge this exactly at the same time when switching to the fixed CTK version!


lassoan:
Integrated in r26107

PR #740, opened 2017-06-22
BUG: Catch improper class names in ExtensionWizard

AndrewBeers:
Hello!

This is another small bug. When using the ExtensionWizard, one can click the "Add Module to Extension" to generate a module template under a provided module "name." For example, if a module name is provided as "TestModule" in the menu below..

image

Then a file, TestModule.py will be created. That file will contain some pre-built classes, like 

```
class TestModule(ScriptedLoadableModule):

class TestModuleWidget(ScriptedLoadableModuleWidget):
```

The problem is that one can provide ANY string to that menu. Based on the prompt, I would initially assume that I was providing the final module name -- for example, "Extension Wizard". However, this generates bad class names that won't compile, e.g.:

`class Extension WizardWidget(ScriptedLoadableModuleWidget):`

Pressing "OK" with such a name will immediately lead to an error box for hard-to-decipher reasons at a first pass, e.g.

image

While most coders might be able to spot the problem in the console, some probably won't be able to. I know I missed it the first time I tried Extension Wizard, leading me to believe that it was just broken that time around and then I should download a new version later (read: 6 months). There's a bunch of different solutions to this. For example, one could make the prompt box more explicit about what 'Name' means, or automatically convert invalid class names into valid class names. In the interest of simplicity, however, I've just added an error-checker that detects invalid class names and gives users a relevant error message.

image

It might be something to think about more in-depth for the templating interface in general, but in the meantime this fix will likely work and perhaps not scare off new potential Slicer developers.

All the best,
Andrew

### DIFF
--- a/Modules/Scripted/ExtensionWizard/ExtensionWizardLib/CreateComponentDialog.py
+++ b/Modules/Scripted/ExtensionWizard/ExtensionWizardLib/CreateComponentDialog.py
@@ -1,6 +1,7 @@
 import os
 import slicer
 import qt, ctk
+import re
 
 #=============================================================================
 #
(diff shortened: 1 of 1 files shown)


pieper:
LGTM - 👍

jcfr:
Thanks for the patch. Testing locally  and integrating.

jcfr:
Thanks @AndrewBeers  . I simplified the approach and integrated in [r26136](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26136)

PR #743, opened 2017-07-20
ENH: require Shift key release after translate before slice-follow

ihnorton:
This reduces the risk of inadvertently changing all the slice positions when doing something unrelated (with a 1- or 2- button mouse).

### DIFF
--- a/Libs/MRML/DisplayableManager/vtkSliceViewInteractorStyle.cxx
+++ b/Libs/MRML/DisplayableManager/vtkSliceViewInteractorStyle.cxx
@@ -52,6 +52,7 @@ vtkSliceViewInteractorStyle::vtkSliceViewInteractorStyle()
 {
   this->ActionState = vtkSliceViewInteractorStyle::None;
   this->ActionsEnabled = vtkSliceViewInteractorStyle::AllActionsMask;
+  this->ShiftKeyUsedForPreviousAction = false;
 
   this->StartActionEventPosition[0] = 0;
   this->StartActionEventPosition[1] = 0;
@@ -114,6 +115,13 @@ void vtkSliceViewInteractorStyle::OnKeyPress()
 //----------------------------------------------------------------------------
 void vtkSliceViewInteractorStyle::OnKeyRelease()
 {
+
+  std::string key = this->Interactor->GetKeySym();
+
+  if (((key.find("Shift") != std::string::npos)) && this->ShiftKeyUsedForPreviousAction)
+    {
+    this->ShiftKeyUsedForPreviousAction = false;
+    }
   this->Superclass::OnKeyRelease();
 }
 
@@ -375,6 +383,7 @@ void vtkSliceViewInteractorStyle::OnLeftButtonUp()
 {
   if (this->ActionState == this->Translate)
     {
+    this->ShiftKeyUsedForPreviousAction =  true;
     this->EndTranslate();
     }
   else if (this->ActionState == this->Blend)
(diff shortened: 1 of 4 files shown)


lassoan:
From the code it's not very obvious - could you tell what exactly is changed?
What are the operations that are now better separated?

ihnorton:
Shift-mouse-move scrolls the slice views to correspond with the mouse position (it also does some projection scrolling in 3D -- doesn't work very well).

The problem is the following sequence:
-shift-click to pan
- then release the shift key before the mouse button
- the Shift-mouse-move scroll behavior is invoked and the slice positions jump somewhere random (very annoying)

With this PR, when doing some *non-scroll* shift operation, the shift key must be released once before the default shift-mouse-move scrolling behavior is resumed.

lassoan:
I see, thanks for the explanation.

Maybe renaming "WasShiftkeyed" to "ShiftKeyUsedForPreviousAction" would make things a bit more clear.

The other half of the issue - accidentally moving the mouse cursor before clicking - is still not solved. Would there be a way to address both issues, while keeping synchronized slice browsing easily available?

> scrolling in 3D -- doesn't work very well

The only limitation I know about is that when you display multiple slice views then the planes may drift if you click on them. Have you experienced any other issue?

cpinter:
Based on the description it makes sense to me.

I'd also rename the WasShiftkeyed variable, and add doxygen comment in the header.

I cannot imagine a solution for the "other half of the problem" without introducing another mouse button click or something, but brainstorming about it here is a good idea.

ihnorton:
> The other half of the issue - accidentally moving the mouse cursor before clicking - is still not solved. Would there be a way to address both issues, while keeping synchronized slice browsing easily available?

The best (and simplest) idea I can think of is to change the shift-mouse-move behavior to require essentially a "double-shift" to invoke: second shift and hold within some timeout like 500ms. But I don't know if this would annoy Ron or some other users with a long-developed muscle memory.

cpinter:
> "double-shift"

I think it's too complex to be the default. However it could be an option in application settings for those who encounter this problem regularly (everybody has different habits doing the clicks and button holds)

lassoan:
Something like double-shift would be good, but I agree with Csaba that it's too complicated and not used commonly. Would there something simpler?

Another approach would be to store default crosshair action in application settings. You can set it to None in the crosshair menu on the toolbar but by default it is "move slices".

ihnorton:
> The only limitation I know about is that when you display multiple slice views then the planes may drift if you click on them. Have you experienced any other issue?

I just find it unintuitive and hard to use, but I know that's very subjective 😬 . One issue is that the "in-plane" slice essentially always goes quickly out of the view. For example, if I start with an approximately anterior view, the coronal slice disappears quickly, and I have to move the mouse way over to the side of the screen to bring it back (but possibly PEBCAK or misunderstanding on my part).

I *think* it would be simpler if the scrolling was locked to whichever slice plane the pointer-projection intersects when the action starts. For example, if the pointer is "on" the coronal slice, then the coronal would be locked and scrolling would only happen for the ax and sag planes. But this might be confusing in other ways, and I haven't tried to implement it.

> Something like double-shift would be good, but I agree with Csaba that it's too complicated and not used commonly. Would there something simpler?

Shift-click-release before moving? Also complex and breaks muscle memory.

Anyway, it doesn't happen enough to motivate me to add a bunch of code and configuration complexity to support multiple modes. This PR fixes an issue I do hit semi-often, enough to corss some annoyance threshold (I think Adam was also complaining about it a few months ago on the mailing list).

lassoan:
I agree that the patch is an improvement, I don't see any drawbacks, and it seems that there is no easy way of making it better, so it makes sense to merge as is.

ihnorton:
Comments addressed and integrated.
https://github.com/Slicer/Slicer/commit/dc5ec3f81a047bfa317bfcddf2402d1d6af62fc9

jcfr:
Thanks @ihnorton

PR #744, opened 2017-07-21
BUG: Avoid crash during UI buildup if a ButtonGroup has no checked button

michaelschwier:
I experienced Slicer crashing when loading a module with a `<string-enumeration>` which had no `<default>` element. Reason was that during UI buildup the pointer to the checked button in the ButtonGroup was fetched to get the text of the checked button. However, in this case the pointer to the button was NULL. Returning an empty string in this case seems to solve the issue.

### DIFF
--- a/Base/QTCLI/qSlicerCLIModuleUIHelper.cxx
+++ b/Base/QTCLI/qSlicerCLIModuleUIHelper.cxx
@@ -141,6 +141,7 @@ QString ButtonGroupWidgetWrapper::checkedValue()
     return button->text();
   }
   else {
+    // Handle case where <string-enumeration> has no <default> element
     return QString("");
   }
 }
--- a/Base/QTCLI/qSlicerCLIModuleUIHelper.cxx
+++ b/Base/QTCLI/qSlicerCLIModuleUIHelper.cxx
@@ -137,8 +137,12 @@ QButtonGroup* ButtonGroupWidgetWrapper::buttonGroup()const
 QString ButtonGroupWidgetWrapper::checkedValue()
 {
   QAbstractButton* button = this->ButtonGroup->checkedButton();
-  Q_ASSERT(button);
-  return button->text();
+  if (button) {
+    return button->text();
+  }
+  else {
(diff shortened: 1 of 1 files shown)


Base/QTCLI/qSlicerCLIModuleUIHelper.cxx
```diff
@@ -137,8 +137,12 @@ QButtonGroup* ButtonGroupWidgetWrapper::buttonGroup()const
 QString ButtonGroupWidgetWrapper::checkedValue()
 {
   QAbstractButton* button = this->ButtonGroup->checkedButton();
-  Q_ASSERT(button);
-  return button->text();
+  if (button) {
+    return button->text();
+  }
+  else {
```
jcfr: Thanks for the patch :+1: 

Nitpick: Could you add a comment like:

```
// Handle case where <string-enumeration> which has no <default> element
```

Thanks

michaelschwier: Done.

jcfr:
Thanks. Integrated in [r26161](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26161)

lassoan:
Thanks for the fix. It's yet another example that using ASSERT instead of error handling is a really bad practice. It's good that at least one more is hunted down.

pieper:
+1!  ASSERT causes a lot of problems.

michaelschwier:
Just a (probably silly) question, since I am new to git: I just updated my local master from the upstream. The  changes proposed in my pull request are all there. So I want to delete my fix-branch, but on this site as well as on my git command line it tells me that the branch has unmerged commits!?

pieper:
Yes, that's "normal" for Slicer - we still use svn as the master repository for mainly historical reasons.  We use git for everything else.

There's more info here:

https://www.slicer.org/wiki/Slicer:git-svn

michaelschwier:
Ah ok, thanks for the info.

PR #746, opened 2017-07-21
ENH: Update Numpy from 1.9.2 to 1.13.1

jcfr:
[ci skip]

Co-authored-by: Franklin King <franklin.king@queensu.ca>

### DIFF
--- a/SuperBuild/External_NUMPY.cmake
+++ b/SuperBuild/External_NUMPY.cmake
@@ -69,13 +69,13 @@ set(${proj}_WORKING_DIR \"${CMAKE_BINARY_DIR}/${proj}\")
 ExternalProject_Execute(${proj} \"install\" \"${PYTHON_EXECUTABLE}\" setup.py install)
 ")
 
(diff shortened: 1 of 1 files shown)


jcfr:
This is a follow-up of https://github.com/msmolens/Slicer/pull/1

We first update the version of Numpy. Later commit will take care of patching Numpy so that it can be built using VS2015 and above.

It was manually tested on Linux, macOS and Windows.

jcfr:
Integrated in [r26164](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26164)

fedorov:
This update causes runtime warning that is printed out during Slicer application startup, as discussed in https://discourse.slicer.org/t/runtime-warning-on-startup-in-numpy/757/16.

More importantly, the warning is interpreted as error in the extensions build, which prevents packaging of the extensions, see example here: http://slicer.cdash.org/viewBuildError.php?buildid=1079946

Can you guys please look into this?

PR #749, opened 2017-07-24
Initialize quantity/units attributes for scalar volumes loaded from DICOM

fedorov:
Scalar volumes loaded from various various flavors of MR and CT modalities will
be asigned DICOM.QuantityCode and DICOM.UnitsCode to describe the pixel values
stored in the volume. A function mapping SOPClassUID is introduced in the base
DICOM plugin class to perform the mapping.

See related discussion in

https://github.com/QIICR/QuantitativeReporting/issues/127#issuecomment-273925207

A similar approach is implemented in the PET and Parametric Map plugins
provided in the extensions.


fedorov:
Sorry for the conflicting branch, I am confused by my setup...

PR #752, opened 2017-07-27
ENH: Added saving of DICOM scalar volume voxel quantity&unit in volume node

lassoan:
It implements functionality originally implemented in this commit (but later reverted due to regressions):
"ENH: init quantity/units in scalar volume plugin" (4ac23a026eac73f471c53383c9cd73cf30eabc79).

This commit is functionally equivalent (except CT quantity incorrect code value has been fixed in
`mapSOPClassUIDToDICOMQuantityAndUnits`).

Quantity and unit of voxel values of scalar volumes can be accessed in `vtkMRMLScalarVolumeNode` by using these methods:
```
  void SetVoxelValueQuantity(vtkCodedEntry*);
  vtkGetObjectMacro(VoxelValueQuantity, vtkCodedEntry);
  void SetVoxelValueUnit(vtkCodedEntry*);
  vtkGetObjectMacro(VoxelValueUnit, vtkCodedEntry);
```

`vtkCodedEntry` class was added that simplifies storage, printing, serialization of code scheme/value/meaning triplets
so that it can be easily added to other node types in the future.

`DICOMScalarVolumePlugin` sets these values when the data is loaded from DICOM.

Added test of setting quantity and unit in scalar volume node during DICOM import (`py_DICOMReaders test`).

### DIFF
--- a/Applications/SlicerApp/Testing/Python/DICOMReaders.py
+++ b/Applications/SlicerApp/Testing/Python/DICOMReaders.py
@@ -97,15 +97,19 @@ class DICOMReadersTest(ScriptedLoadableModuleTest):
         "fileName": "Mouse-MR-example-where-GDCM_fails.zip",
         "name": "Mouse-MR-example-where-GDCM_fails",
         "seriesUID": "1.3.6.1.4.1.9590.100.1.2.366426457713813178933224342280246227461",
-        "expectedFailures": ["GDCM", "Archetype"]
+        "expectedFailures": ["GDCM", "Archetype"],
+        "voxelValueQuantity": "(DCM, 110852, MR signal intensity)",
+        "voxelValueUnit": "(UCUM, [hnsf'U], Hounsfield unit)"
       },
       { "url": "http://slicer.kitware.com/midas3/download/item/294857/deidentifiedMRHead-dcm-one-series.zip",
         "fileName": "deidentifiedMRHead-dcm-one-series.zip",
         "name": "deidentifiedMRHead-dcm-one-series",
         "seriesUID": "1.3.6.1.4.1.5962.99.1.3814087073.479799962.1489872804257.270.0",
-        "expectedFailures": []
+        "expectedFailures": [],
+        "voxelValueQuantity": "(DCM, 110852, MR signal intensity)",
+        "voxelValueUnit": "(UCUM, 1, no units)"
       }
-    ]''') 
+    ]''')
 
     # another dataset that could be added in the future - currently fails for all readers
     # due to invalid format - see https://issues.slicer.org/view.php?id=3569
@@ -113,7 +117,7 @@ class DICOMReadersTest(ScriptedLoadableModuleTest):
         #"fileName": "RIDER_bug.zip",
         #"name": "RIDER_bug",
         #"seriesUID": "1.3.6.1.4.1.9328.50.7.261772317324041365541450388603508531852",
-        #"expectedFailures": []
+        #"expectedFailures": [],
       #}
 
     loadingResult = {}
@@ -159,11 +163,11 @@ class DICOMReadersTest(ScriptedLoadableModuleTest):
         loadable = detailsPopup.getAllSelectedLoadables().keys()[0]
 
         #
-        # try loading using each of the selected readers, fail 
+        # try loading using each of the selected readers, fail
         # on enexpected load issue
         #
         scalarVolumePlugin = slicer.modules.dicomPlugins['DICOMScalarVolumePlugin']()
-        readerApproaches = scalarVolumePlugin.readerApproaches()   
+        readerApproaches = scalarVolumePlugin.readerApproaches()
         basename = loadable.name
         volumesByApproach = {}
         for readerApproach in readerApproaches:
@@ -176,6 +180,13 @@ class DICOMReadersTest(ScriptedLoadableModuleTest):
             raise Exception("Expected to NOT be able to read with %s, but could!" % readerApproach)
           if volumeNode:
             volumesByApproach[readerApproach] = volumeNode
+
+            # Test quantity and unit
+            if 'voxelValueQuantity' in dataset.keys():
+              self.assertEqual(volumeNode.GetVoxelValueQuantity().GetAsPrintableString(), dataset['voxelValueQuantity'])
+            if 'voxelValueUnit' in dataset.keys():
+              self.assertEqual(volumeNode.GetVoxelValueUnit().GetAsPrintableString(), dataset['voxelValueUnit'])
+
 
         #
         # for each approach that loaded as expected, compare the volumes
@@ -211,5 +222,5 @@ class DICOMReadersTest(ScriptedLoadableModuleTest):
     mainWindow.moduleSelector().selectModule('DICOMReaders')
 
     logging.info(loadingResult)
-  
+
     return testPass
--- a/Applications/SlicerApp/Testing/Python/DICOMReaders.py
+++ b/Applications/SlicerApp/Testing/Python/DICOMReaders.py
@@ -98,16 +98,16 @@ class DICOMReadersTest(ScriptedLoadableModuleTest):
         "name": "Mouse-MR-example-where-GDCM_fails",
         "seriesUID": "1.3.6.1.4.1.9590.100.1.2.366426457713813178933224342280246227461",
         "expectedFailures": ["GDCM", "Archetype"],
-        "voxelValueQuantity": "(DCM, 110852, MR signal intensity)",
-        "voxelValueUnit": "(UCUM, [hnsf'U], Hounsfield unit)"
+        "voxelValueQuantity": "(DCM, 110852, \\"MR signal intensity\\")",
+        "voxelValueUnits": "(UCUM, 1, \\"no units\\")"
       },
       { "url": "http://slicer.kitware.com/midas3/download/item/294857/deidentifiedMRHead-dcm-one-series.zip",
         "fileName": "deidentifiedMRHead-dcm-one-series.zip",
         "name": "deidentifiedMRHead-dcm-one-series",
         "seriesUID": "1.3.6.1.4.1.5962.99.1.3814087073.479799962.1489872804257.270.0",
         "expectedFailures": [],
-        "voxelValueQuantity": "(DCM, 110852, MR signal intensity)",
-        "voxelValueUnit": "(UCUM, 1, no units)"
+        "voxelValueQuantity": "(DCM, 110852, \\"MR signal intensity\\")",
+        "voxelValueUnits": "(UCUM, 1, \\"no units\\")"
       }
     ]''')
 
@@ -117,7 +117,7 @@ class DICOMReadersTest(ScriptedLoadableModuleTest):
         #"fileName": "RIDER_bug.zip",
         #"name": "RIDER_bug",
         #"seriesUID": "1.3.6.1.4.1.9328.50.7.261772317324041365541450388603508531852",
-        #"expectedFailures": [],
+        #"expectedFailures": []
       #}
 
     loadingResult = {}
@@ -181,11 +181,11 @@ class DICOMReadersTest(ScriptedLoadableModuleTest):
           if volumeNode:
             volumesByApproach[readerApproach] = volumeNode
 
-            # Test quantity and unit
+            self.delayDisplay('Test quantity and unit')
             if 'voxelValueQuantity' in dataset.keys():
               self.assertEqual(volumeNode.GetVoxelValueQuantity().GetAsPrintableString(), dataset['voxelValueQuantity'])
-            if 'voxelValueUnit' in dataset.keys():
-              self.assertEqual(volumeNode.GetVoxelValueUnit().GetAsPrintableString(), dataset['voxelValueUnit'])
+            if 'voxelValueUnits' in dataset.keys():
+              self.assertEqual(volumeNode.GetVoxelValueUnits().GetAsPrintableString(), dataset['voxelValueUnits'])
 
 
         #
--- a/Libs/MRML/Core/vtkCodedEntry.cxx
+++ b/Libs/MRML/Core/vtkCodedEntry.cxx
@@ -1,0 +1,156 @@
+/*=auto=========================================================================
+
+Portions (c) Copyright 2017 Brigham and Women's Hospital (BWH) All Rights Reserved.
+
+See COPYRIGHT.txt
+or http://www.slicer.org/copyright/copyright.txt for details.
+
+=========================================================================auto=*/
+
+#include "vtkCodedEntry.h"
+
+// VTK include
+#include <vtkObjectFactory.h>
+
+// STD include
+#include <sstream>
+
+vtkStandardNewMacro(vtkCodedEntry);
+
+//----------------------------------------------------------------------------
+vtkCodedEntry::vtkCodedEntry()
+: CodingSchemeDesignator(NULL)
+, CodeValue(NULL)
+, CodeMeaning(NULL)
+{
+}
+
+//----------------------------------------------------------------------------
+vtkCodedEntry::~vtkCodedEntry()
+{
+  this->Initialize();
+}
+
+//----------------------------------------------------------------------------
+void vtkCodedEntry::Initialize()
+{
+  this->SetCodingSchemeDesignator(NULL);
+  this->SetCodeValue(NULL);
+  this->SetCodeMeaning(NULL);
+}
+
+;
+
+//----------------------------------------------------------------------------
+void vtkCodedEntry::SetSchemeValueMeaning(const std::string& scheme,
+  const std::string& value, const std::string& meaning)
+{
+  this->SetCodingSchemeDesignator(scheme.c_str());
+  this->SetCodeValue(value.c_str());
+  this->SetCodeMeaning(meaning.c_str());
+}
+
+//----------------------------------------------------------------------------
+void vtkCodedEntry::PrintSelf(ostream& os, vtkIndent indent)
+{
+  Superclass::PrintSelf(os,indent);
+  os << indent << "CodingSchemeDesignator: " << (this->CodingSchemeDesignator?this->CodingSchemeDesignator:"(none)") << "\n";
+  os << indent << "CodeValue: " << (this->CodeValue?this->CodeValue:"(none)") << "\n";
+  os << indent << "CodeMeaning: " << (this->CodeMeaning?this->CodeMeaning:"(none)") << "\n";
+}
+
+//----------------------------------------------------------------------------
+void vtkCodedEntry::Copy(vtkCodedEntry* aType)
+{
+  if (!aType)
+    {
+    return;
+    }
+  this->SetCodingSchemeDesignator(aType->GetCodingSchemeDesignator());
+  this->SetCodeValue(aType->GetCodeValue());
+  this->SetCodeMeaning(aType->GetCodeMeaning());
+}
+
+//----------------------------------------------------------------------------
+std::string vtkCodedEntry::GetAsPrintableString()
+{
+  std::string printable = std::string("(")
+    + (this->CodingSchemeDesignator ? this->CodingSchemeDesignator : "(none)") + ", "
+    + (this->CodeValue ? this->CodeValue : "(none)") + ", "
+    + (this->CodeMeaning ? this->CodeMeaning : "(none)") + ")";
+  return printable;
+}
+
+//----------------------------------------------------------------------------
+std::string vtkCodedEntry::GetAsString()
+{
+  std::string str;
+  if (this->CodingSchemeDesignator)
+    {
+    str += "CodingSchemeDesignator:";
+    str += this->CodingSchemeDesignator;
+    }
+  if (this->CodeValue)
+    {
+    if (!str.empty())
+      {
+      str += "|";
+      }
+    str += "CodeValue:";
+    str += this->CodeValue;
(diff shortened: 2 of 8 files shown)


lassoan:
Reworked version of PR https://github.com/Slicer/Slicer/pull/750.

@fedorov please review.

fedorov:
Thank you Andras, I will review later today.

cpinter:
I read through the change and I agree with the decisions made.
Probably it would make sense to make the new vtkCodedEntry class the base class of [vtkSlicerTerminologyCategory](https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/Terminologies/Logic/vtkSlicerTerminologyCategory.h) and [vtkSlicerTerminologyType](https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/Terminologies/Logic/vtkSlicerTerminologyType.h), as vtkCodedEntry is exactly a subset of these classes, except for the name of the CodingScheme[Designator] member that would need fixing in the terminology logic. I can do this change once the PR is in.

lassoan:
> Probably it would make sense to make the new vtkCodedEntry class the base class of vtkSlicerTerminologyCategory

I completely agree. I've planned to do this later, once this branch is merged and everything works well.

Libs/MRML/Core/vtkCodedEntry.cxx
```diff
@@ -0,0 +1,156 @@
...
+  this->SetCodeMeaning(aType->GetCodeMeaning());
+}
+
+//----------------------------------------------------------------------------
+std::string vtkCodedEntry::GetAsPrintableString()
+{
+  std::string printable = std::string("(")
+    + (this->CodingSchemeDesignator ? this->CodingSchemeDesignator : "(none)") + ", "
+    + (this->CodeValue ? this->CodeValue : "(none)") + ", "
+    + (this->CodeMeaning ? this->CodeMeaning : "(none)") + ")";
```
fedorov: I would put code meaning into double quotes, to handle presence of a comma in the code meaning in a more readable way. I don't know how common it would be in practice, but looking over the code meanings for the codes included in DICOM, there are some examples that include comma, e.g. see http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_D.html#DCM_111239

Libs/MRML/Core/vtkCodedEntry.cxx
```diff
@@ -0,0 +1,156 @@
...
+  return str;
+}
+
+//----------------------------------------------------------------------------
+bool vtkCodedEntry::SetFromString(const std::string& content)
+{
+  this->Initialize();
+  bool success = true;
+  std::stringstream attributes(content);
+  std::string attribute;
```
fedorov: Maybe add a check that would fail if there is more than 2 `|`?

Libs/MRML/Core/vtkCodedEntry.h
```diff
@@ -0,0 +1,93 @@
...
+
+// VTK includes
+#include <vtkObject.h>
+
+/// \brief Simple class for storing standard coded entries (coding scheme, value, meaning triplets)
+///
+/// vtkCodedEntry is a lightweight class that stores standard coded entries consisting of
+/// CodingSchemeDesignator + CodeValue + CodeMeaning triplets.
+/// This is a commonly used concept in DICOM, see chapter 3: Encoding of Coded Entry Data
+/// (http://dicom.nema.org/dicom/2013/output/chtml/part03/sect_8.3.html).
```
lassoan: OK!

Modules/Scripted/DICOMLib/DICOMPlugin.py
```diff
@@ -277,3 +277,37 @@ def addSeriesInSubjectHierarchy(self,loadable,dataNode):
...
+    # http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_D.html
+    if classUID in MRname2UID.values():
+      quantity = slicer.vtkCodedEntry()
+      quantity.SetSchemeValueMeaning("DCM", "110852", "MR signal intensity")
+      units = slicer.vtkCodedEntry()
+      units.SetSchemeValueMeaning("UCUM", "1", "no units")
+
+    if classUID in CTname2UID.values():
+      quantity = slicer.vtkCodedEntry()
+      quantity.SetSchemeValueMeaning("DCM", "110851", "X-Ray Attenuation Coefficient")
```
lassoan: OK! (I just modified it because I've noticed that it was 110852 in the original pull request, probably copy-paste error)

Applications/SlicerApp/Testing/Python/DICOMReaders.py
```diff
@@ -97,23 +97,27 @@ def test_Part1DICOM(self):
         "fileName": "Mouse-MR-example-where-GDCM_fails.zip",
         "name": "Mouse-MR-example-where-GDCM_fails",
         "seriesUID": "1.3.6.1.4.1.9590.100.1.2.366426457713813178933224342280246227461",
-        "expectedFailures": ["GDCM", "Archetype"]
+        "expectedFailures": ["GDCM", "Archetype"],
+        "voxelValueQuantity": "(DCM, 110852, MR signal intensity)",
+        "voxelValueUnit": "(UCUM, [hnsf'U], Hounsfield unit)"
```
lassoan: Do you mean use units (as opposed to unit) everywhere or just here?
Do you know why in DICOM they use plural? In the standard it is stated that only one unit can be specified.

I'll fix the others.

Applications/SlicerApp/Testing/Python/DICOMReaders.py
```diff
@@ -97,23 +97,27 @@ def test_Part1DICOM(self):
...
+    ]''')
 
     # another dataset that could be added in the future - currently fails for all readers
     # due to invalid format - see https://issues.slicer.org/view.php?id=3569
       #{ "url": "http://slicer.kitware.com/midas3/download/item/293587/RIDER_bug.zip",
         #"fileName": "RIDER_bug.zip",
         #"name": "RIDER_bug",
         #"seriesUID": "1.3.6.1.4.1.9328.50.7.261772317324041365541450388603508531852",
-        #"expectedFailures": []
+        #"expectedFailures": [],
```
lassoan: Good catch!

Libs/MRML/Core/vtkCodedEntry.cxx
lassoan: I agree that when there is comma then it may be more readable to have a different separator. However, adding double quotes all the time would make things a bit harder to read in general.

What about using a different separator?
Or just adding double quotes when there is comma in the text?

Libs/MRML/Core/vtkCodedEntry.cxx
lassoan: Currently, if more fields are defined then a warning is logged for each unknown fields and the method returns with false. Should we change this behavior?

fedorov:
@ihnorton you responded with 👎  but didn't comment - was that a mistake, or do you have any reservations about this functionality?

Modules/Scripted/DICOMLib/DICOMPlugin.py
fedorov: Yes, I definitely messed it up by cut-and-paste! Thank you for catching that!

Applications/SlicerApp/Testing/Python/DICOMReaders.py
fedorov: I did a quick search, and it looks like "units of measurement" and "unit of measurement" are used interchangeably. 

Either way is fine with me, I just think that perhaps in this case it is better to be consistent with the standard to reduce confusion and the mental effort needed by the users typing these things while correlating Slicer implementation with DICOM.

Libs/MRML/Core/vtkCodedEntry.cxx
fedorov: The convention of putting code meaning in double quotes is always applied in DICOM (as an example, see [this section](http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.6.16.2.html#sect_C.7.6.16.2.11.1.2), so I would definitely suggest just putting it in double quotes. I do not think it makes things harder to read, quite the opposite. 

And as above, since we add support that makes things more consistent with the DICOM way of things, I think it is important to be follow the DICOM conventions.

pieper: 👍

Libs/MRML/Core/vtkCodedEntry.cxx
fedorov: You are right, I missed it! I agree it is good as is implemented now.

fedorov:
The following tests failed for me, but I really don't know if any of them are caused by the change! Anyone has any idea which ones can potentially be true positives?

As an example, the last one `py_LandmarkRegistration` failed because of failed download (although right now that URL works):

```
642: <b>Requesting download</b> <i>PreDentalSurgery.gipl.gz</i> from http://slicer.kitware.com/midas3/download/item/94510/Greyscale_presurg.gipl.gz...
642:
642: <b><font color="red">	Download failed: [Errno socket error] [Errno 8] nodename nor servname provided, or not known</font></b>
```

I am now going to rebuild and run the tests with the master branch.

```
The following tests FAILED:
	174 - qMRMLLayoutManagerTest1 (Failed)
	175 - qMRMLLayoutManagerTest2 (Failed)
	176 - qMRMLLayoutManagerTest3 (Failed)
	177 - qMRMLLayoutManagerTest4 (Failed)
	178 - qMRMLLayoutManagerVisibilityTest (Failed)
	179 - qMRMLLayoutManagerWithCustomFactoryTest (Failed)
	214 - qMRMLSliceWidgetTest1_vol_and_cube.mrml (Failed)
	215 - qMRMLSliceWidgetTest2_TestData/fixed.nrrd (Failed)
	225 - qMRMLVolumeThresholdWidgetTest2_vol_and_cube.mrml (Failed)
	246 - qMRMLSliceWidgetEventTranslatorPlayerTest1 (Failed)
	430 - py_VolumeRenderingThreeDOnlyLayout (Failed)
	439 - py_StandaloneEditorWidgetTest (Failed)
	503 - ModelToLabelMapTest (Failed)
	504 - ModelToLabelMapTestLabelValue (Failed)
	506 - N4ITKBiasFieldCorrectionTest (Failed)
	602 - py_AtlasTests (Timeout)
	604 - py_DICOMReaders (Failed)
	605 - py_KneeAtlasTest (Failed)
	610 - py_Slicer4Minute (Failed)
	611 - py_SlicerBoundsTest (Failed)
	617 - py_RSNAVisTutorial (Failed)
	618 - py_RSNAQuantTutorial (Failed)
	619 - py_slicerCloseCrashBug2590 (Failed)
	621 - py_SlicerTransformInteractionTest1 (Failed)
	623 - py_ViewControllersSliceInterpolationBug1926 (Failed)
	624 - py_RSNA2012ProstateDemo (Failed)
	625 - py_JRC2013Vis (Failed)
	627 - py_SubjectHierarchyGenericSelfTest (Failed)
	635 - py_SimpleFiltersModuleTest (Timeout)
	642 - py_LandmarkRegistration (Failed)
Errors while running CTest
```

Applications/SlicerApp/Testing/Python/DICOMReaders.py
lassoan: @pieper @ihnorton, as native speakers, what do you think: Do both volumeNode.SetVoxelValueUnit(...) and volumeNode.SetVoxelValueUnits(...) sound good to you for setting voxel unit, such as HU or MBq/g, etc.?

Libs/MRML/Core/vtkCodedEntry.cxx
lassoan: OK, if this is an established convention then it's good, I'll update the code accordingly.

jcfr:
> 642: <b>Requesting download</b> <i>PreDentalSurgery.gipl.gz</i> from http://slicer.kitware.com/midas3/download/item/94510/Greyscale_presurg.gipl.gz...
> 642:
> 642: <b><font color="red">	Download failed: [Errno socket error] [Errno 8] nodename nor servname provided, or not known</font></b>

I just asked our sysadmin if connectivity was interrupted last night. I will follow up.

Jc

Applications/SlicerApp/Testing/Python/DICOMReaders.py
pieper: What sounds right to me is that when there's only one thing being measured then you use a 'unit of measure' or a 'unit' for short.  It would sound right use 'units of measure' when there's more than one unit involved, like the units of measure in the metric system.  It's a little fuzzy when you have a compound unit, like meters per second, where there are multiple units involved.  Or when you have a plural quantity, like there are "three units of blood".  

Anyway, I think we should follow DICOM Data Dictionary conventions which uses Units in every case, even when there's only one involved.  For example DeviceDiameterUnits clearly has only one unit of measure specified (VM=1) and there are many similar cases and only a couple exceptions.

fedorov: Yes, that is my main argument too -- at this point, it is not that much about proper English, but about harmonization with the existing standard. No one (David Clunie included!) says it is perfect, but DICOM users life is already quite challenging to add further complications with incoherencies between the standard and implementation.

lassoan: I just wanted to avoid errors in the DICOM standard leak into Slicer. I highly value consistency but the DICOM standard is not consistent in itself, especially not among multiple domains (while Slicer covers all domains).

Anyway, if it's not that confusing and not terrible grammatically then I guess we can go with plural (units). In the documentation I'll describe why we chose plural and if it's not enough then we can add a singular version of the method name as well.

fedorov:
Here are the tests that are failing with the current master. Considering there might have been network connectivity problem in my last run, I will re-run the branch tests.

```
Total Test time (real) = 6487.60 sec

The following tests FAILED:
	174 - qMRMLLayoutManagerTest1 (Failed)
	175 - qMRMLLayoutManagerTest2 (Failed)
	176 - qMRMLLayoutManagerTest3 (Failed)
	177 - qMRMLLayoutManagerTest4 (Failed)
	178 - qMRMLLayoutManagerVisibilityTest (Failed)
	179 - qMRMLLayoutManagerWithCustomFactoryTest (Failed)
	214 - qMRMLSliceWidgetTest1_vol_and_cube.mrml (Failed)
	215 - qMRMLSliceWidgetTest2_TestData/fixed.nrrd (Failed)
	225 - qMRMLVolumeThresholdWidgetTest2_vol_and_cube.mrml (Failed)
	246 - qMRMLSliceWidgetEventTranslatorPlayerTest1 (Failed)
	430 - py_VolumeRenderingThreeDOnlyLayout (Failed)
	439 - py_StandaloneEditorWidgetTest (Failed)
	503 - ModelToLabelMapTest (Failed)
	504 - ModelToLabelMapTestLabelValue (Failed)
	506 - N4ITKBiasFieldCorrectionTest (Failed)
	616 - py_SlicerRestoreSceneViewCrashIssue3445 (Failed)
	623 - py_ViewControllersSliceInterpolationBug1926 (Failed)
Errors while running CTest
```

fedorov:
Here are the tests that are failing with the PR test, now that the network issue seems to have resolved itself:

```
Total Test time (real) = 9475.82 sec

The following tests FAILED:
	174 - qMRMLLayoutManagerTest1 (Failed)
	175 - qMRMLLayoutManagerTest2 (Failed)
	176 - qMRMLLayoutManagerTest3 (Failed)
	177 - qMRMLLayoutManagerTest4 (Failed)
	178 - qMRMLLayoutManagerVisibilityTest (Failed)
	179 - qMRMLLayoutManagerWithCustomFactoryTest (Failed)
	214 - qMRMLSliceWidgetTest1_vol_and_cube.mrml (Failed)
	215 - qMRMLSliceWidgetTest2_TestData/fixed.nrrd (Failed)
	225 - qMRMLVolumeThresholdWidgetTest2_vol_and_cube.mrml (Failed)
	243 - qMRMLROIWidgetEventTranslatorPlayerTest1 (Failed)
	246 - qMRMLSliceWidgetEventTranslatorPlayerTest1 (Failed)
	430 - py_VolumeRenderingThreeDOnlyLayout (Failed)
	439 - py_StandaloneEditorWidgetTest (Failed)
	503 - ModelToLabelMapTest (Failed)
	504 - ModelToLabelMapTestLabelValue (Failed)
	506 - N4ITKBiasFieldCorrectionTest (Failed)
	602 - py_AtlasTests (Timeout)
	623 - py_ViewControllersSliceInterpolationBug1926 (Failed)
Errors while running CTest
```

The non-overlapping tests (`qMRMLROIWidgetEventTranslatorPlayerTest1` and `py_AtlasTests`) passed when invoked manually one by one, so I say this PR does not deteriorate the test performance of the application on my platform.

lassoan:
@fedorov I cannot find your test results on the dashboard. Could you please upload them somewhere and share the link (or run the test with dashboard submission enabled)? All tests pass on all platforms (http://slicer.cdash.org/index.php?project=Slicer4) on factory computers, so it would be interesting to know what causes the failures on your computer.

fedorov:
@lassoan I can try later, if I can figure out how, but few points

* I don't do this often, and I didn't even think to enable the submission to the dashboard 
* I don't know how to enable submission to CDash, and I cannot find it quickly on the wiki
* It takes literally hours to complete the tests - I don't plan to stay in the office long enough to wait for them to complete (as an aside, is it really necessary to manipulate hundreds of items in the atlas hierarchy in the `py_AtlasTests`? that test alone contributes about 10 minutes!)
* running tests is disruptive, since the testing windows pop up and take mouse focus
* value of the tests is questionable, since some tests (as my experience described above shows) would sometime fail and sometime succeed on the same platform

All in all, there are many deterrents to running all tests manually for practical purposes. I also discussed with @jcfr and we thought that perhaps we should not require developers to do this for every single PR. It might be a lot more efficient to just check the dashboard the next day.

fedorov:
I can't find the instructions how to submit the test results to CDash. I actually don't even know if the test results are saved automatically into some file, or I would need to re-run all the tests again to submit to the dashboard. All I did was `ctest -V .`. Sorry.

lassoan:
Thank you for trying running the tests. You only need to send results to the dashboard if you want others to see the results (you can do that by running the test as ctest -D Experimental). I agree that if you don't have a computer that you can just leave there while tests are running, it is inconvenient to run the tests.

It would be good to run the test on at least one platform before integration (these tests do catch real errors), but the responsibility is ultimately up to the person who merges the change. Stakes are quite high nowadays, when the stable version is so old that many users have to use the nightly version. I'm in regular contact with only about 50-100 Slicer users, but if something is broken in the nightly build, even for a few days, I almost always got complaints and several hours of productivity is lost (especially when error is around scene saving or other core feature). So, right now we definitely should not rely on the nightly dashboard as a first line of defense to catch errors.

I think if we can deliver stable release in every 3 months then we can ask most users to stick to stable version. Then it will be OK to let some potential errors slip into the nightly builds and detect them by nightly dashboard tests.

fedorov:
> if something is broken in the nightly build, even for a few days, I almost always got complaints and several hours of productivity is lost 

I don't quite understand why this would be the case. You can always go back to yesterday's nightly. It is usually not the case that we have new and exciting features added to every nightly build. Are there really users who need a new Slicer every day?

lassoan:
If you want to share the test results privately, you can find the log in (SlicerSuperbuildBin)/Slicer-build/Testing/Temporary/LastTest.log.

fedorov:
> you can find the log in (SlicerSuperbuildBin)/Slicer-build/Testing/Temporary/LastTest.log.

Good to know, but unfortunately I re-ran some of the failing tests manually, so the log for the entire list of tests was overwritten :-(

lassoan:
> I don't quite understand why this would be the case. You can always go back to yesterday's nightly. It is usually not the case that we have new and exciting features added to every nightly build. Are there really users who need a new Slicer every day?

I think there are many users who update Slicer in every 1-2 weeks, probably on multiple computers (download counts, number of forum users, etc. suggest that -- we don't have direct data on this). If they happen to download Slicer on a day when there is something wrong, they may lose a few hours.

Downloading an older nightly is possible but most people don't realize if something is wrong with the version that they download (unless the installation fails or Slicer does not start) and don't know immediately what version they should retrieve and how.

Several fixes and improvements are added every week, which are important/exciting for certain groups of users. If you use a more rapidly evolving component (such as Segment editor) then probably you want to upgrade more frequently.

fedorov:
yes, I see your point

lassoan:
Thanks for the review. Committed r26185.

jcfr:
@fedorov  To follow up on the network issue, here is the HTTP response time for the night of July 27 to July 28th.

We can indeed see some lower response time (>= 400ms). 

That said, based on the error message it is most likely a name resolution error. See [1][2]. If you would like, we could run a script for few days on your workstation to monitor and log few metric related to slicer.kitware.com

unnamed

[1] https://github.com/python/cpython/blob/v2.7.13/Modules/getaddrinfo.c#L148
[2] https://github.com/python/cpython/blob/v2.7.13/Modules/getaddrinfo.c#L277

lassoan:
@fedorov FYI, we've set up a Windows continuous build, so if you commit anything relatively early during the day, you get a chance to see the automatic test results and make changes accordingly.

fedorov:
@lassoan thank you! Maybe it is a good idea to announce this on the forum? So if developers have a commit late at night, and it is not super-urgent, they should feel encouraged to commit it early morning to be safe.

PR #753, opened 2017-07-27
BUG: Fixed parent transform setting by drag-and-drop

lassoan:
In Data module / Transform hierarchy tab, transform parent can be set by drag-and-drop even if
the transform is being constantly modified (such as in a surgical navigation scenario, when
position tracker device continuously updates the transform through OpenIGTLink).

Fixes https://issues.slicer.org/view.php?id=4138

### DIFF
--- a/Libs/MRML/Widgets/qMRMLSceneModel.cxx
+++ b/Libs/MRML/Widgets/qMRMLSceneModel.cxx
@@ -681,7 +681,8 @@ bool qMRMLSceneModel::dropMimeData(const QMimeData *data, Qt::DropAction action,
   // random column.
   bool res = this->Superclass::dropMimeData(
     data, action, row, 0, parent.sibling(parent.row(), 0));
-  d->DraggedNodes.clear();
+  // Do not clear d->DraggedNodes yet, as node modification events may come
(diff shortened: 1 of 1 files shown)


pieper:
Sounds like a good fix to me. 👍

jcfr:
Ditto :+1: 

If feasible, do you think you could add a regression test ?

lassoan:
> If feasible, do you think you could add a regression test ?

Yes, it would be nice to have a test but seems hard to implement. Do you know how to simulate a drag-and-drop in a QTreeView? (QtTesting operates on the scene model, we would need a solution that manipulates the GUI widget)

pieper:
You could use the QtTesting infrastructure to record a test to xml, or you could write a script that emulates the mouse actions.  We use both approaches in various tests, but gui testing is still pretty fragile when the code changes.

lassoan:
I've checked QtTesting, but it only simulates the end result of the drag-and-drop operation (calls reparent on the model), while the problem was in how the GUI widget behaves between start dragging and dropping (before reparent is called).

lassoan:
Committed r26186

PR #754, opened 2017-07-27
ENH: Added clear() function to print 40 blank lines in console

moselhy:
Enhancement to issue: https://issues.slicer.org/view.php?id=4393

This would be useful for people developing an extension UI and want to clear the errors out from the console before reloading


lassoan:
Thanks for submitting this pull request. Printing some separators when a module is reloaded is a good idea (it helps in distinguishing new error messages from old), but I think we can do better than just adding a utility function.

Could you please look into automating printing some separators (maybe not 40 empty lines but a few empty lines just a and a separator like "------------") into the reload function of the scripted module? see https://github.com/Slicer/Slicer/blob/master/Base/Python/slicer/ScriptedLoadableModule.py#L140-L145

moselhy:
See https://github.com/Slicer/Slicer/pull/755

lassoan:
For next time, you can just push commits to the same branch on github. Github will update the pull request automatically. You don't need a separate pull request for each update in the proposed changes.

moselhy:
Oh, that makes sense. I will push next commits to my [most recent pull request, a minute before I saw this](https://github.com/Slicer/Slicer/pull/758).

PR #755, opened 2017-07-28
ENH: Print separators on module reload

moselhy:
Prints 10 empty lines and a dashed line when user clicks 'Reload', in order to separate error messages and other debugging info

### DIFF
--- a/Base/Python/slicer/ScriptedLoadableModule.py
+++ b/Base/Python/slicer/ScriptedLoadableModule.py
@@ -142,6 +142,8 @@ class ScriptedLoadableModuleWidget:
     ModuleWizard will substitute correct default moduleName.
     Generic reload method for any scripted module.
     """
+    for i in range(10): print('\n')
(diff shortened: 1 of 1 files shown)


Base/Python/slicer/ScriptedLoadableModule.py
```diff
@@ -142,6 +142,8 @@ def onReload(self):
     ModuleWizard will substitute correct default moduleName.
     Generic reload method for any scripted module.
     """
+    for i in range(10): print('\n')
+    print('------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------')
```
pieper: Nitpick: something like `print(('-'*60+'\n')*10)` would be cleaner.

pieper: Or I guess
```
print('\n'*10)
print('-'*60)
```

moselhy: Good idea

moselhy:
See https://github.com/Slicer/Slicer/pull/757

PR #756, opened 2017-07-28
BUG: remove false positive warning

pieper:
As reported by @fedorov, he was getting warning messages about
bad spacing even for data he thought was correct.  This was
confirmed with data from series 3 of this [1] dataset.

The problem was some faulty logic that to calculate which
loadable has the highest priority [2] but we can remove that
logic because the loadable are already flagged with a
'selected' attribute and we only need to generate the
dialog box when a selected loadable has a warning.

Tested with series 1 (with warning) and 3 of [1].

[1] http://slicer.kitware.com/midas3/folder/2182
[2] https://github.com/Slicer/Slicer/commit/c5a92bcdf471612d41cd41ff5418dca1bf994abb


pieper:
Committed as r26184

PR #758, opened 2017-07-28
STYLE: Changed color for Grow from Seeds error from green to red

moselhy:
previous:

image

proposed:

image


lassoan:
Thank you. Committed r26183.

PR #760, opened 2017-08-01
COMP: ensure application singletons are available

pieper:
Similar to an issue in the DataStore module [1],
a library that accesses qSlicerApplication::application()
or other singletons needs to be explicitly linked
to qSlicerBaseQTGUI on some platforms
(for example debian 9 with gcc6).

[1] https://github.com/Slicer/Slicer-DataStore/pull/2


lassoan:
Please also add a comment to that line in the code to make sure that it will not be accidentally removed in the future.

pieper:
Thanks for the reviews!

Committed as r26194

PR #762, opened 2017-08-03
COMP: Avoid symbol clash detecting system install of Qt. See #3573

jcfr:
If qmake is found in a system location, explicitly mark Qt as such. Doing so
will prevent system path from being prepended to PATH or (DY)LD_LIBRARY_PATH
when generating the launcher settings and avoid system libraries symbols from
conflicting with Slicer version of these libraries.


jcfr:
Cc: @dzenanz

jcfr:
Thanks for the review

jcfr:
Integrated in [r26193](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26193)

PR #763, opened 2017-08-03
BUG: ExtensionWizard: Provide informative msg if an option is not supported

jcfr:
If Python is not built with SSL support, the variable _haveGit
is set to False. In that case, user should still provide the argument
but the wizard should fail providing an informative message.
is True.

For example, in case the option "--publish" is not supported the following
message will be displayed:

```
/path/to/bin/slicerExtensionWizard --publish ./Foo
Option '--publish' is not available.

Consider re-building Slicer with SSL support or downloading
Slicer from http://download.slicer.org
```

See #3935

Co-authored-by: Francois Budin <francois.budin@kitware.com>
Suggested-by: Andriy Fedorov <fedorov@bwh.harvard.edu>

### DIFF
--- a/Utilities/Scripts/SlicerWizard/ExtensionWizard.py
+++ b/Utilities/Scripts/SlicerWizard/ExtensionWizard.py
@@ -590,20 +590,20 @@ class ExtensionWizard(object):
                         help="print the extension description (s4ext)"
                              " to standard output")
 
-    if _haveGit:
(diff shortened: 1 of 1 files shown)


jcfr:
Thanks for the review.

jcfr:
Integrated in [r26197](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26197)

PR #765, opened 2017-08-03
Update minimum required cmake

jcfr:
<h3>NOTE: The topic include two commits to facilitate review. I will squash the two commits before integration.</h3>


CMake/SlicerExtensionCPack.cmake
```diff
@@ -187,7 +187,7 @@ if(APPLE)
     # for SuperBuild extensions.
 
     file(WRITE ${slicer_extension_cpack_bundle_fixup_directory}/CMakeLists.txt
-    "cmake_minimum_required(VERSION 3.0)
+    "cmake_minimum_required(VERSION 3.3)
```
lassoan: Is it intentionally just 3.3 and not 3.5?

jcfr: Good catch. This one should be updated.

jcfr:
> Looks good. Do we need to update required CMake version in all extensions?

The statement `cmake_minimum_required(VERSION X.Y)` does **NOT** need to be updated in the extension source tree.

That said:
* whenever possible, extension developer are encouraged to update it in the top level `CMakeLists.txt`. This will most likely avoid confusion ... 
* the version of CMake used to build the extension would **need** to be 3.5. To facilitate the overall maintenance, I made this is a hard requirement by adding explicit statement `cmake_minimum_required(VERSION 3.5)`  to `SlicerConfig.cmake` and `UseSlicer.cmake`

dzenanz:
Oops, I forgot there is the second commit which changes 3.3 -> 3.5.

jcfr:
Thanks everyone for the review.

Integrated in [r26201](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26201), [r26202](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26202) and [r26203](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26203)

Associated forum topic: https://discourse.slicer.org/t/minimum-required-version-of-cmake-updated-to-3-5-0/828

PR #766, opened 2017-08-04
ENH: Make continuous build driver more configurable

lassoan:
These changes allow running a single continuous build, which does not delete the build tree and can be launched at arbitrary time intervals using a task scheduler. To run a single continuous build, CTEST_CONTINUOUS_DURATION variable has to be set to 0. If the value is not set or >0 then the previous behavior is kept.


lassoan:
@jcfr FYI, this script is now used on factory.perklab to run continuous Windows builds during the day, reusing the nightly build tree.

lassoan:
Committed r26200

pieper:
👍

PR #767, opened 2017-08-09
ENH: Added option to disable DICOM reference check

cpinter:
- Added checkbox in the DICOM references dialog to not check for references again. Selection is saved regardless of pressed button
- Added generic options group in Application Settings / DICOM where this setting can be changed
- Added optional constructor argument to DICOMLoadable for easy conversion from qSlicerDICOMLoadable (useful for debugging when the otherwise deprecated python DICOMLoadable class is used internally)

### DIFF
--- a/Modules/Scripted/DICOM/DICOM.py
+++ b/Modules/Scripted/DICOM/DICOM.py
@@ -110,6 +110,24 @@ This work is supported by NA-MIC, NAC, BIRN, NCIGT, and the Slicer Community. Se
 class _ui_DICOMSettingsPanel(object):
   def __init__(self, parent):
     vBoxLayout = qt.QVBoxLayout(parent)
+    # Add generic settings
+    genericGroupBox = ctk.ctkCollapsibleGroupBox()
+    genericGroupBox.title = "Generic DICOM settings"
+    genericGroupBoxFormLayout = qt.QFormLayout(genericGroupBox)
+    loadReferencesComboBox = ctk.ctkComboBox()
+    loadReferencesComboBox.toolTip = "Determines whether referenced DICOM series are " \
+      "offered when loading DICOM, or the automatic behavior if interaction is disabled. " \
+      "Interactive selection of referenced series is the default selection"
+    loadReferencesComboBox.addItem("Ask user", qt.QMessageBox.InvalidRole)
+    loadReferencesComboBox.addItem("Always", qt.QMessageBox.Yes)
+    loadReferencesComboBox.addItem("Never", qt.QMessageBox.No)
+    loadReferencesComboBox.currentIndex = 0
+    genericGroupBoxFormLayout.addRow("Load referenced series:", loadReferencesComboBox)
+    parent.registerProperty(
+      "DICOM/automaticallyLoadReferences", loadReferencesComboBox, "currentUserDataAsString", qt.SIGNAL("currentIndexChanged (int)"))
+    vBoxLayout.addWidget(genericGroupBox)
+
+    # Add settings panel for the plugins
     plugins = slicer.modules.dicomPlugins
     for pluginName in plugins.keys():
       if hasattr(plugins[pluginName], 'settingsPanelEntry'):
--- a/Modules/Scripted/DICOMLib/DICOMPlugin.py
+++ b/Modules/Scripted/DICOMLib/DICOMPlugin.py
@@ -30,27 +30,36 @@ class DICOMLoadable(object):
   instead.
   """
 
-  def __init__(self):
-    # the file list of the data to be loaded
-    self.files = []
-    # name exposed to the user for the node
-    self.name = "Unknown"
-    # extra information the user sees on mouse over of the thing
-    self.tooltip = "No further information available"
-    # things the user should know before loading this data
-    self.warning = ""
-    # is the object checked for loading by default
-    self.selected = False
-    # confidence - from 0 to 1 where 0 means low chance
-    # that the user actually wants to load their data this
-    # way up to 1, which means that the plugin is very confident
-    # that this is the best way to load the data.
-    # When more than one plugin marks the same series as
-    # selected, the one with the highest confidence is
-    # actually selected by default.  In the case of a tie,
-    # both series are selected for loading.
-    self.confidence = 0.5
-
+  def __init__(self, qLoadable=None):
+    if qLoadable is None:
+      # the file list of the data to be loaded
+      self.files = []
+      # name exposed to the user for the node
+      self.name = "Unknown"
+      # extra information the user sees on mouse over of the thing
+      self.tooltip = "No further information available"
+      # things the user should know before loading this data
+      self.warning = ""
+      # is the object checked for loading by default
+      self.selected = False
+      # confidence - from 0 to 1 where 0 means low chance
+      # that the user actually wants to load their data this
+      # way up to 1, which means that the plugin is very confident
+      # that this is the best way to load the data.
+      # When more than one plugin marks the same series as
+      # selected, the one with the highest confidence is
+      # actually selected by default.  In the case of a tie,
+      # both series are selected for loading.
+      self.confidence = 0.5
+    else:
+      self.name = qLoadable.name
+      self.tooltip = qLoadable.tooltip
+      self.warning = qLoadable.warning
+      self.files = []
+      for file in qLoadable.files:
+        self.files.append(file)
+      self.selected = qLoadable.selected
+      self.confidence = qLoadable.confidence
 
 #
 # DICOMPlugin
--- a/Modules/Scripted/DICOMLib/DICOMWidgets.py
+++ b/Modules/Scripted/DICOMLib/DICOMWidgets.py
@@ -768,8 +768,11 @@ class DICOMDetailsBase(VTKObservationMixin, SizePositionSettingsMixin):
     if len(referencedFileLists):
       (self.referencedLoadables, loadEnabled) = self.getLoadablesFromFileLists(referencedFileLists)
 
-    if loadEnabled:
+    automaticallyLoadReferences = int(slicer.util.settingsValue('DICOM/automaticallyLoadReferences', qt.QMessageBox.InvalidRole))
+    if loadEnabled and automaticallyLoadReferences == qt.QMessageBox.InvalidRole:
       self.showReferenceDialogAndProceed()
+    elif loadEnabled and automaticallyLoadReferences == qt.QMessageBox.Yes:
+      self.addReferencesAndProceed()
     else:
       self.proceedWithReferencedLoadablesSelection()
 
@@ -777,7 +780,13 @@ class DICOMDetailsBase(VTKObservationMixin, SizePositionSettingsMixin):
 
   def showReferenceDialogAndProceed(self):
     referencesDialog = DICOMReferencesDialog(self, loadables=self.referencedLoadables)
-    if referencesDialog.exec_() == qt.QMessageBox.Ok:
+    answer = referencesDialog.exec_()
+    if referencesDialog.rememberChoiceAndStopAskingCheckbox.checked == True:
+      if answer == qt.QMessageBox.Yes:
+        qt.QSettings().setValue('DICOM/automaticallyLoadReferences', qt.QMessageBox.Yes)
+      if answer == qt.QMessageBox.No:
+        qt.QSettings().setValue('DICOM/automaticallyLoadReferences', qt.QMessageBox.No)
+    if answer == qt.QMessageBox.Yes:
       # each check box corresponds to a referenced loadable that was selected by examine;
       # if the user confirmed that reference should be loaded, add it to the self.loadablesByPlugin dictionary
       for plugin in self.referencedLoadables:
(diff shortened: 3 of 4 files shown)


cpinter:
image

cpinter:
image

pieper:
Makes a lot of sense - 👍

lassoan:
Thank you, this will be useful. Just two small comments:

1. In general, negative statements are sometimes difficult to interpret ("uncheck Disable reference check checkbox"). It is not really a check, but offering automatic loading. So, it could be "Automatically load referenced items" (or "Automatically load referenced volumes" if it's only for volumes).
3. In the application settings, it may be better to be consistent with how ctkMessageBox handles "don't show again" checboxes. See ctkMessageBox::dontShowAgain. It may even worth switching to ctkMessageBox from QMessageBox base class to make implementation simpler and more consistent.

cpinter:
@lassoan Yes your avatar appeared above my shoulder when I entered the negative statement :) The reason I left it was that
1. The meaning of the checkbox in the dialog and the application settings would be opposing (do not show <-> automatically check)
2. There was already a disabledPlugins variable in the DICOM settings
I can change it everybody thinks it's necessary.

I don't quite understand the second comment. What do you mean more consistent with ctkMessageBox _in the application settings_?

lassoan:
ctkMessageBox handles bypassing of popup dialogs already (storing user selection in application settings and not showing it if user clicked that checkbox). If you use it, it is less code overall and it is more consistent with how these kind of popups are handled elsewhere in Slicer.

cpinter:
This part I understand. What I don't understand is what does it have to do with the application settings? 
"In the application settings, it may be better to be consistent with how ctkMessageBox..."

lassoan:
disabledPlugins is different. It's an exception, mainly for debugging. Not a good practice to follow (what if you have a list of enabled plugins and disabled plugins - how do you know which list is stronger, etc.

Current name and title is not clear enough anyway, as "checking" is not specific enough. "Loading" is more clear. Also, most of the "don't show again" popups remember the use selection (all ctkMessageBox based popups do), so disabling is just half of the story. User may want to always enable loading. So, overall I don't see any advantage in using negative logic.

I just spend this much time arguing about this seemingly irrelevant detail because application setting keys and values are really difficult to change (without breaking backward compatibility and/or losing data) so suboptimal decisions haunt us for years. Moreover, through copy/paste and striving for consistency, bad choices often spread further.

lassoan:
There is a way how ctkMessageBox stores user's selection in application settings (code of standard button). It makes code more consistent if we store all "don't show again" choice the same way.

jcfr:
> Yes your avatar appeared above my shoulder when I entered the negative statement :)

Fun fact: I almost gave the same comment but ... I ended up doing a compromise.  That said, I agree with Andras, we should avoid negative logic.

Thanks @cpinter for dealing with us :smile:

cpinter:
OK I'll do another round.

cpinter:
Unfortunately the Don't show again button is added incorrectly to the layout:
https://github.com/commontk/CTK/blob/master/Libs/Widgets/ctkMessageBox.cpp#L212
It is assumed it is always second row second column, no span, and there is no way from python to get the widget and re-add it to the layout again.

It seems that ctkMessageBox was not intended to be subclassed and added more widgets to it in a more complex layout, but just showing a message with some standard buttons, and optionally this checkbox. If I am to use ctkMessageBox for this popup, I'll need to change things in CTK. Thoughts?

pieper:
if you have time to improve CTK that would be great of course. 😄

cpinter:
Well, of course. However:
* ctkMessageBox has not changed for six years, so apparently it has perfectly fulfilled its purpose. It just does not seem to be that useful to work on this, if I only fulfill this one case, and maybe another one in six more years
* This change will probably take an unreasonably long time comparing to how uesful this feature will be. With "claiming" this class, and going through all the integration and future maintenance work, I'm looking at many more hours of work than it now seems

While trying not to sound big-headed, to me personally it seems that my time is worth more than spending potentially a week in grand total on a message box feature that may never even be used, while this particular job I just did will not look better to the user whatsoever.

pieper:
@cpinter  It's funny I wrote a sentence saying that I didn't think changing CTK was needed just for this particular use case but deleted it from the buffer and replaced it with a smiley instead.  I guess in retrospect the meaning was unclear so let me say in words that I agree your time is extremely valuable and you should allocate it in the way that you consider most productive!

cpinter:
I added a [Mantis issue](https://issues.slicer.org/view.php?id=4412) about this ctkMessageBox flaw, so that we can record if this comes up again and make changes based on what we know if it becomes needed.

I'll change the name of the settings variable and make the stored values conform with those used by ctkMessageBox if we decide to make the change.

One more, perhaps philosophical note to this topic: I'm not sure if a message box is the proper container for such a dialog, which contains a table with checkboxes. A message box's purpose is to show a simple message, and allow the user to make a basic decision. I think a QDialog would be a better fit to this particular widget. That said I'd not change it now, because it works as intended :)

pieper:
+1

cpinter:
Another hurdle when doing the requested changes: ctkSettingsPanel::registerProperty only allows defining a property of a Qt object to be set as value of the settings key, see "checked" here https://github.com/cpinter/Slicer/blob/disable-dicom-loadable-suggestions/Modules/Scripted/DICOM/DICOM.py#L121-L123

If we want to ensure that the current implementation will have the same values on each selection in the settings as ctkMessageBox will if (!) we make a change, then I see two options (other than not using ctkSettingsPanel::registerProperty, which would mean even more refactoring):
1. Subclass QCheckBox now and add a new property that returns QMessageBox::Ok when checked and QMessageBox::InvalidRole when unchecked (because these are the ctkMessageBox values)
2. Subclass ctkMessageBox later when (if!) we change it so that it sets true/false instead of the above values to the settings

But again, I consider this whole topic (preserving a settings value for the future of a not-so-important flag) so marginal that even typing this seems to be an unnecessary overhead.

I changed the settings key to DICOM/automaticallyLoadReferences and I'm ready to amend the pull request once we reach a decision. Thanks!

lassoan:
Settings key name and value are important because difficult to change in the future and bad practices are copied for sake of consistency.

Even more important: A yes/no flag it's not enough. We need ask/always yes/always no options (maybe you implemented like that already).

cpinter:
So 1 or 2?

cpinter:
So solution 1 then?

cpinter:
@lassoan We cannot offer always yes. The found loadables are offered by multiple plugins, and it doesn't happen that somebody always wants to load the loadables using all of the plugins that have a confidence >0.

jcfr:
> ctkSettingsPanel::registerProperty only allows defining a property of a Qt object to be set as value of the settings key, see "checked"

The [ctkBooleanMapper](https://github.com/commontk/CTK/blob/master/Libs/Core/ctkBooleanMapper.h) may help here.

See [here](https://github.com/commontk/CTK/blob/cdb75ce742f618112b8a0f299a484d00baa4e8ff/Libs/Widgets/ctkSettingsPanel.h#L70-L76) for an example of usage with `ctkSettingsPanel`

lassoan:
I think ctk message box saves the button code (yes/no or ok/cancel) so that it can simulate what button is clicked.

lassoan:
I always want to load all referenced data sets and I would prefer not to see the popup each time.

lassoan:
I always prefer to see what's linked by showing them in the viewers and then delete what I'm not interested in.

fedorov:
> I always want to load all referenced data sets

Sorry to add to this long discussion, but just another use case - sometime it happens that we have multiple segmentation series corresponding to the same image series. In this case it only makes sense to load referenced series for the first segmentation, and not load for the subsequent ones. This use case is no obvious, and I wonder if you guys have a mechanism to help user "change mind" or somehow deal with this situation after preference was indicated with some "always" option...

cpinter:
@lassoan Sorry but I have a hard time believing that when you want to load an RTSS and the referenced volume is offered by the Scalar Volume and the DWI plugins, you want both of them to load the volume. I know you will tell me to disable the DWI plugin, but there are other use cases such as Andrey's where loading all of them is undesired.

@fedorov Yes the option is added in Application Settings, hence most of the difficulty.

I however am on the verge of abandoning the whole pull request and the 8ish hours of work I did on this insignificant feature and keep unchecking the offered loadables and click OK every time.

lassoan:
I like how the non-advanced DICOM loading works: By default I don't have to make a decision but can rely on reasonable defaults that work most of the time (and I can override if I have any problem). So, any solution that is close to this (or can evolve to this in the future, without any disruption) is good.

The time spent on this discussion and learning how application settings and ctk popup dialog are all useful for the future (even if you abandon the PR, which I hope will not happen).

cpinter:
@jcfr ctkBooleanMapper seemed very promising; unfortunately it's not python wrapped. I need to look for another solution. My best bet so far is to subclass QCheckBox to give -1 (InvalidRole) and 1024 (Ok) instead ot false and true, respectively.

jcfr:
> ctkBooleanMapper seemed very promising; unfortunately it's not python wrapped.

Let me look into that. It shouldn't be too hard to expose it

cpinter:
Thanks, @jcfr! If you point me to where to look to fix the wrapping, I can give it a go.

jcfr:
@cpinter This patch should allow you to instantiate the ctkBooleanMapper from python.

See https://github.com/commontk/CTK/compare/master...jcfr:wrap-ctkBooleanMapper

If it works out, let me know and I will integrate to CTK and update Slicer.

cpinter:
Thanks a lot!

cpinter:
I replaced that one file, built CTK (first inner then outer), build Slicer superbuild, still no ctk.ctkBooleanMapper(). Should I build something else, or rebuild something?

jcfr:
Hi,

In fact the full signature "ctk.ctkBooleanMapper(arg1, arg2, . ..)" Should
be available.

cpinter:
I understand you say it should be available. But again, I built CTK with the modifications, and even if I rebuild Slicer, the following error appears.

```
>>> import ctk
>>> ctk.ctkBooleanMapper()
AttributeError: 'module' object has no attribute 'ctkBooleanMapper'
```

So my question is the same: What should I do to access the wrapped class?

jcfr:
> What should I do to access the wrapped class?

The following should work:

* apply patch to CTK
* rebuild target CTKWidgetPythonQt

cpinter:
I'm not entirely sure how to put it in a different way:
* I applied the patch
* Built CTK
* ctkBooleanMapper is not accessible from python
Have you tried yourself?

cpinter:
Correction: I applied most of the patch... I forgot to add the register line.

lassoan:
CTK uses superbuild, so you might need to build the inner CTK project. Maybe it helps.
________________________________

cpinter:
I got it now, I missed one line in the patch. My mistake!

jcfr:
No problem :)

cpinter:
I believe I solved all the issues. I amended the commit, and created a CTK topic branch with my changes:
https://github.com/cpinter/CTK/tree/settings-related-improvements

Here's what is different from what I had before:
* Instead of an OK and Cancel button, now there is a Yes, No, and Cancel buttons in the popup. Yes loads the selected loadables, No does not load any loadables, Cancel aborts the loading. I also changed the wording, see screenshot
* In the Application Settings window now there is a combobox instead of a checkbox, to accommodate for the more granular selection, also see screenshot
* In Slicer.ini, the property name is automaticallyLoadReferences, with the possible values qt.QMessageBox.InvalidRole (interactive, popup appears - this is the default), qt.QMessageBox.Yes (do not ask and load referenced loadables), qt.QMessageBox.No (do not ask and don't load references)
* To support this combobox way, I added an extra property in ctkComboBox so that these custom values can be saved for each selection (see CTK branch). I kept Jc's improvements too (with an additional change of the constructor signature so that it can be properly instantiated from python) because they are still useful. I'll submit a pull request to CTK when there is agreement.

image
image

lassoan:
Thank you Csaba, it looks perfect!

Just one minor thing to consider: maybe combobox options could be simply "Ask user", "Always", "Never" (instead of "Interactive selection of referenced series" etc).

pieper:
Nice work!

cpinter:
@lassoan Makes sense, I amended the commit.

cpinter:
I also submitted the pull request to CTK
https://github.com/commontk/CTK/pull/743

Thanks all for participating in the discussion and push me to finish this :)

lassoan:
This work will serve as a good example for saving combobox settings to MRML or application settings in a robust way. Instead of saving user-displayed (potentially translated, may contain bspecial characters, etc) strings directly to settings, combobox user data string (machine readable code string) is written to/read from settings.

cpinter:
I integrated the PR in this commit
https://github.com/Slicer/Slicer/commit/ee5c8703642cf8f42e9594bc780cb0801623b39d

fedorov:
I hope it doesn't break anything new. I've been waiting for quite some time for a working nightly + extensions we need for MICCAI ....

cpinter:
This has been sitting here for almost a week, after thorough testing. I also hope it doesn't break anything!!

I'll not integrate the segment editor buttons one, as you requested (although I thought you needed it for MICCAI, that's why I implemented it relatively fast).

fedorov:
I need to finalize tutorial materials (finalize binaries to distribute, prepare handouts, make screenshots, finalize prerequisite instructions so people can come with software installed in advance), and I can't leave this to the last day (unless I have no choice). Also, the experience of dry-run last Friday showed that testing by a live crowd uncovers unexpected situations, and I won't have another opportunity for a crowd-testing before the conference, so I need to minimize the changes that take place in the application. As much as I can anyway....

cpinter:
I completely understand! Which day is the tutorial? First or last?

fedorov:
First - Sept 10, and Sept 9 I am on the road all day.

fedorov:
The good news is that the bug in the today's Slicer nightly does not seem to affect the tutorial workflow, so I may just use the today's set of binaries+extensions, to make everyone's life easier (including my own).

cpinter:
I really think this PR will not cause any issues, it should be safe to use tomorrow's nightly. Those python errors send a bad message if anyone sees that in the tutorial.

PR #770, opened 2017-08-11
ENH: Adding  VTKPlots in Slicer for Interactive manipulation

Punzo:
Hi @jcfr @pieper  @lassoan ,

this pull request is to add VTK Plots in Slicer as discussed during the Summer Project week. 

Here a video showing how I designed it:
https://www.dropbox.com/s/g08jzgacjxy54dm/videoPlot.mkv?dl=0

Comparing with the Charting with jqPlot:
A) vtkMRMLPlotLayoutNode has the similar role of vtkMRMLChartNode.
B) vtkMRMLPlotNode has the role of vtkMRMLDoubleArrayNode.

Once it is plotted, the table and the plot are coupled (i.e. modifying the table will update the plot etc...):
The panning and zooming is done with middle-click;
The selection with right-click ('S' change the type of selection). The selection IDs are stored in the PlotLayoutNode.
Click and Drag is done with left-click (and it's disabled for the BarPLot).

I tested up to 10^6 points and It was fully interactive 

Few notes:
1) Click and Drag will work after this patch to VTK https://github.com/Punzo/VTK/commit/1d5a9219a57f2f60512de069f6b32fa4ecb50f83

So at the moment you can't build and test unless you download the modified classes (vtkChart and vtkChartXY) from my fork of VTK. Or you can disable the click and drag just commenting this line
https://github.com/Punzo/Slicer/commit/29316ce388dadfcfe92996162fd48c791823a869#diff-2230803b602f286c1352621a5cddef3fR125
 
Shall I submit to VTK the patch and then try to update the VTK git tag of Slicer? @jcfr what is the status of  updating Slicer to the current VTK(8)? Ref: https://github.com/Kitware/VTK/pull/44

2) @lassoan I had to change the Type from string to float of the column of the table (in the table Module).  vtkplot does not accept values as vtkStringArray. I don't like this too much, but converting on the fly to me seems even worst. I am open to any suggestion here.
https://github.com/Punzo/Slicer/commit/29316ce388dadfcfe92996162fd48c791823a869#diff-101bce1e10d0c03a223a2f738a774488R315 

3) at the moment, it is not possible to plot errors in vtk.


lassoan:
Wow. This is outstanding! I'll review it as soon as I can.

While columns that you create on the GUI are now all set to string by default, internally all column types are supported in Tables module (column type is saved in a schema .csv file). I'll add GUI for setting column type.

Send a pull request with your VTK changes to https://github.com/Slicer/VTK.

fedorov:
Thanks for working on this, a general purpose wrapper for VTK plots would be very nice to have in Slicer! 👍

pieper:
👍 

I'd want the pan/zoom mouse buttons to match the sliceView and threeDView behavior: right drag up/down to zoom and middle mouse drag to pan.  That would seem most natural within the context of Slicer.

Then Selection in the plot could be left mouse which seems intuitive to me. 

Also middle click  that you have now for zoom is not easy to generate for mac laptops; for the other views we have shift-left-drag for pan as a workaround.

Libs/MRML/Logic/vtkMRMLPlotLayoutLogic.h
```diff
@@ -0,0 +1,34 @@
...
+///  of PlotLayout Nodes to Plot Nodes
+
+#ifndef __vtkMRMLPlotLayoutLogic_h
+#define __vtkMRMLPlotLayoutLogic_h
+
+// MRMLLogic includes
+#include "vtkMRMLAbstractLogic.h"
+
+
+class VTK_MRML_LOGIC_EXPORT vtkMRMLPlotLayoutLogic : public vtkMRMLAbstractLogic
```
jcfr: Reading this short introduction:  https://www.slicer.org/wiki/Documentation/Nightly/Developers/MRML#MRML_Node_References

As well as the more  exhaustive documentation could be useful: https://www.slicer.org/wiki/Documentation/Nightly/Developers/MRML/NodeReferences

Punzo:
Hi all, thanks for the feedback! 

@pieper, I used the default click configuration from CTK and added the click and drag for left-click. But I agree it will be good to use Slicer standard for click interactions. I am not sure that I can use combination of keys and click at VTK level or if I can handle them at Qt level and then pass them to the VTK context (most likley yes, but I have to check, I just don't know).

@lassoan thank for the review. 
1) Reference Observations: yeah you are right. I was using workarounds because I think I was not using the reference infrastracture properly ( @jcfr thanks for the links!).
2) test and documentation: I wanted to add them before pushing this PR, but I didn't make in time for those. I plan to add them next week before going to vacation (I pushed the PR anyway to get early feedback).
3) I'll reply at the various comments in detail asap.  
4) @jcfr I saw your commit regarding vtk8 and qt5. Is it better that I'll wait their intragation before pushing https://github.com/Punzo/VTK/commit/1d5a9219a57f2f60512de069f6b32fa4ecb50f83 to the https://github.com/Slicer/VTK as suggested by Andras?

lassoan:
Regarding documentation: Explanation of general, high-level concepts in class overviews would help in understanding and reviewing the code. No need to refine method documentation yet, because methods may still change.

CMake/itkCheckCXXAcceptsFlags.cmake
```diff
@@ -46,7 +46,7 @@ macro(itkCHECK_CXX_ACCEPTS_FLAGS FLAGS VAR)
        "Source file was:\n${_SOURCE}\n")
    else(${VAR})
      message(STATUS "Checking to see if CXX compiler accepts flag ${FLAGS} - No")
-     file(APPEND ${CMAKE_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/CMakeError.log
+     file(APPEND ${CMAKE_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/CMakeerror.log
```
Punzo: yeap, thanks for catching it!

Libs/MRML/Core/vtkMRMLPlotLayoutNode.h
```diff
@@ -0,0 +1,261 @@
...
+
+  //----------------------------------------------------------------
+  /// Access methods
+  //----------------------------------------------------------------
+
+  ///
+  /// Add and Observe a Plot to the PlotLayout. Parameter "name" is used for
+  /// referencing the Plot when setting properties for plotting the
+  /// Plot or for removing the Plot from the PlotLayout.
+  virtual void AddAndObservePlot(const char *name, const char * id);
```
Punzo: I wrote the API lookig at the ChartNode where the the plot ID are saved in a map containing both the name and the ID of the plot. I'll see if the name can be taken out and store only the ID. The issue is that in the controller widget the Plot are identified by name. I'll revise it and we can review this at the new iteration.

Libs/MRML/Core/vtkMRMLPlotLayoutNode.h
```diff
@@ -0,0 +1,261 @@
...
+  ///
+  /// Remove a Plot and Observation from the PlotLayout by a particular name
+  virtual void RemovePlotAndObservationByName(const char *name);
+
+  ///
+  /// Remove a Plot and Observation from the PlotLayout by ID
+  virtual void RemovePlotAndObservationByID(const char *id);
+
+  ///
+  /// Update all missing Observations
```
Punzo: This most likley will disapper after I use properly the reference infrastracture. Essentially this method check if all the PlotNode are observeing vtkplot.

Libs/MRML/Core/vtkMRMLPlotLayoutNode.h
```diff
@@ -0,0 +1,261 @@
...
+  virtual void UpdateObservations();
+
+  ///
+  /// Update a Plot Observation if is missing
+  virtual void UpdateObservation(const char * id);
+
+  ///
+  /// Remove all the Plots
```
Punzo: OK

Libs/MRML/Core/vtkMRMLPlotLayoutNode.h
```diff
@@ -0,0 +1,261 @@
...
+  virtual void ClearPlotIDs();
+
+  ///
+  /// Get the Plot id referenced by a particular name
+  virtual const char* GetPlotID(const char *name);
+
+  ///
+  /// Get the Plot referenced by a particular name
```
Punzo: If we remove the map with name and IDs this will disappear as well. However this is not the same of vtkMRMLScene (in which you give the node for getting the ID). Here you give the name (fro example from the controller widget) and get the ID mapped. But again I guess can simply store the name in vtkMRMLPlotNode and just using IDs.

Libs/MRML/Core/vtkMRMLPlotLayoutNode.h
```diff
@@ -0,0 +1,261 @@
...
+  virtual vtkMRMLPlotNode* GetPlotbyName(const char *name);
+
+  ///
+  /// Get the Plot referenced by a particular id
```
Punzo: ok

Libs/MRML/Core/vtkMRMLPlotLayoutNode.h
```diff
@@ -0,0 +1,261 @@
...
+  virtual vtkMRMLPlotNode* GetPlotByID(const char *id);
+
+  ///
+  /// Get the list of Plot names
+  virtual vtkStringArray* GetPlotNames();
```
Punzo: Yeah, I'll change to GetPlotNames(vtkStringArray* arr). 

There is also GetPlotNames(std::vector<std::string> &plotNodeNames) which I'll remove in favor of GetPlotNames(vtkStringArray* arr).

Libs/MRML/Core/vtkMRMLPlotLayoutNode.h
```diff
@@ -0,0 +1,261 @@
...
+  virtual int GetPlotNames(std::vector<std::string> &plotNodeNames);
+
+  ///
+  /// Get the list of Plot ids
+  virtual vtkStringArray* GetPlotIDs();
```
Punzo: same up

Libs/MRML/Core/vtkMRMLPlotLayoutNode.h
```diff
@@ -0,0 +1,261 @@
...
+  virtual int GetPlotIDs(std::vector<std::string> &plotNodeIDs);
+
+  ///
+  /// Get the collection of PlotNodes
+  virtual vtkCollection* GetPlotNodes();
```
Punzo: same up

Libs/MRML/Core/vtkMRMLPlotLayoutNode.h
```diff
@@ -0,0 +1,261 @@
...
+
+  ///
+  /// Set/Get the selected PlotNodeIDs
```
Punzo: you are right (: I'll change to Selected

Libs/MRML/Core/vtkMRMLPlotLayoutNode.h
```diff
@@ -0,0 +1,261 @@
...
+  virtual vtkStringArray* GetSelectionPlotNodeIDs();
+
+  ///
+  /// Set/Get the collection of selectedIDArrays
+  virtual void SetSelectionArrays(vtkCollection* selectedArrays);
+  virtual vtkCollection *GetSelectionArrays();
+
+  ///
+  /// Set the PlotNodeIDs and selectedArrays
+  virtual void SetSelection(vtkStringArray* selectedPlotNodeIDs,
```
Punzo: ok I'll describe this better.

Libs/MRML/Core/vtkMRMLPlotLayoutNode.h
```diff
@@ -0,0 +1,261 @@
...
+                            vtkCollection* selectedArrays);
+
+  ///
+  /// Get the index of a PlotNode
+  virtual vtkIdType GetPlotNodeIndex(vtkMRMLPlotNode* plotNode);
+
+  ///
+  /// Set/Get a property for a specific Plot to control how it will
+  /// appear in the PlotLayout. If the Plot name is "default", then the property
```
Punzo: Sorry, that was a description of the old ChartNode (in which a property colud be of the overall Chart or of the single plots). Here the property are only of the overall Chart (in this case PlotLayout). Thanks for finding this I'll remove this description and I'll explain better.

Libs/MRML/Core/vtkMRMLPlotLayoutNode.h
```diff
@@ -0,0 +1,261 @@
...
+  virtual void RemovePlotObservation(const char * id);
+
+ protected:
+  //----------------------------------------------------------------
+  /// Data
+  //----------------------------------------------------------------
+
+  ///
+  /// This map stores the names and the IDs of the Plot Node to observe
+  PlotIDMap *MapPlotNamesIDs;
```
Punzo: yea I think yes. As I commetted up, I'll revise this.

Libs/MRML/Core/vtkMRMLPlotLayoutNode.h
```diff
@@ -0,0 +1,261 @@
...
+
+  ///
+  /// Utility array for get methods (they are instantiate at the get call)
+  vtkStringArray *PlotIDs;
```
Punzo: Yeap, I'll revise all the ID and Name reference issues.

Libs/MRML/Core/vtkMRMLPlotLayoutNode.h
```diff
@@ -0,0 +1,261 @@
...
+  ///
+  /// Both PlotLayout and Plots Properties are stored
+  PlotLayoutPropertyMap *Properties;
+
+  ///
+  /// A collection of the pointers of the PlotNodes
+  /// to observe. If the PlotNodeId is stored in MapPlotNamesIDs,
+  /// but it is missing the relative Plot in plotNodes then it is needed to
+  /// call UpdateObservations() to restore the observations.
+  vtkSmartPointer<vtkCollection> plotNodes;
```
Punzo: Ok

Libs/MRML/Core/vtkMRMLPlotLayoutNode.h
```diff
@@ -0,0 +1,261 @@
...
+
+  ///
+  ///
+  /// Two collections of the pointers of the vktMRMLPlotNodes and vtkIdTypeArrays.
+  /// They keep track of the user-selections in the chart View.
+  vtkSmartPointer<vtkStringArray> selectionPlotNodeIDs;
```
Punzo: yaep I'll update this as well

Libs/MRML/Core/vtkMRMLPlotNode.h
```diff
@@ -0,0 +1,249 @@
...
+
+class vtkMRMLStorageNode;
+class vtkMRMLTableNode;
+
+// VTK Includes
+class vtkColor4ub;
+class vtkPlot;
+class vtkTable;
+
+/// \brief MRML node to represent a Plot object
```
Punzo: Ok

Punzo: No. a vtkMRMLPlotNode is only one vtkPlot which it can handle only two coloumns (x and y) from one Table. So if you want to plot two tables, you need to create two vtkMRMLPlotNode referencing different tables and then you need to save their IDs in a vtkMRMLPlotLayoutNode. The view node will be able to overlay the two  vtkMRMLPlotNode.

Libs/MRML/Core/vtkMRMLPlotNode.h
```diff
@@ -0,0 +1,249 @@
...
+///
+class VTK_MRML_EXPORT vtkMRMLPlotNode : public vtkMRMLNode
```
Punzo: Ok

Libs/MRML/Core/vtkMRMLPlotNode.h
```diff
@@ -0,0 +1,249 @@
...
+
+  ///
+  /// Get/Set the status of the Plot
+  vtkGetMacro(Dirty, bool);
+  vtkSetMacro(Dirty, bool);
+
+  ///
+  /// Get the index of the XColumn
+  /// \brief vtkGetMacro
+  vtkGetMacro(XColumn, vtkIdType);
```
Punzo: Ok

Libs/MRML/Core/vtkMRMLPlotNode.h
```diff
@@ -0,0 +1,249 @@
...
+
+  ///
+  /// Set the index of the XColumn and assure the data connection
+  /// \brief vtkSetMacro
+  /// \sa SetInputData
+  virtual void SetXColumn(vtkIdType xColumn);
```
Punzo: OK

Libs/MRML/Core/vtkMRMLPlotNode.h
```diff
@@ -0,0 +1,249 @@
...
+
+  ///
+  /// Utility method to get the name of the columnX.
+  /// The name is automatically set internally
+  virtual std::string GetXColumnName();
+
+  ///
+  /// Get the index of the YColumn
+  /// \brief vtkGetMacro
+  vtkGetMacro(YColumn, vtkIdType);
```
Punzo: OK

Libs/MRML/Core/vtkMRMLPlotNode.h
```diff
@@ -0,0 +1,249 @@
...
+  ///
+  /// String ID of the table MRML node.
+  /// Note that anytime the Table node is modified, the observing display node
+  /// fires a Modified event.
+  /// No Table node by default.
+  /// \sa SetTableNodeID(), GetTableNodeID(),
+  /// TableNode
+  char *TableNodeID;
+
+  vtkMRMLTableNode *TableNode;
```
Punzo: OK

Libs/MRML/Core/vtkMRMLSelectionNode.cxx
```diff
@@ -149,6 +156,7 @@ void vtkMRMLSelectionNode::WriteXML(ostream& of, int nIndent)
   of << " activeTableID=\"" << (this->ActiveTableID ? this->ActiveTableID : "NULL") << "\"";
   of << " activeViewID=\"" << (this->ActiveViewID ? this->ActiveViewID : "NULL") << "\"";
   of << " activeLayoutID=\"" << (this->ActiveLayoutID ? this->ActiveLayoutID : "NULL") << "\"";
+  of << " activePlotLayoutID=\"" << (this->ActivePlotLayoutID ? this->ActivePlotLayoutID : "NULL") << "\"";
```
Punzo: here I mean the vtkMRMLPlotLayoutNode. I think just calling Plot can be misleading to a vtkMRMLPlotNode, no?

Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
```diff
@@ -405,9 +411,31 @@ void vtkMRMLApplicationLogic::PropagateTableSelection()
       continue;
       }
     tnode->SetTableNodeID( tableId );
-    }
+  }
 }
 
+//----------------------------------------------------------------------------
+void vtkMRMLApplicationLogic::PropagatePlotLayoutSelection()
```
Punzo: here I mean the vtkMRMLPlotLayoutNode. I think just calling Plot can be misleading to a vtkMRMLPlotNode, no?

Libs/MRML/Logic/vtkMRMLColorLogic.cxx
```diff
@@ -377,6 +377,12 @@ const char *vtkMRMLColorLogic::GetDefaultChartColorNodeID()
 }
 
 //----------------------------------------------------------------------------
+const char *vtkMRMLColorLogic::GetDefaultPlotColorNodeID()
```
Punzo: OK

Libs/MRML/Logic/vtkMRMLPlotLayoutLogic.h
Punzo: OK

Libs/MRML/Core/vtkMRMLPlotNode.h
```diff
@@ -0,0 +1,249 @@
...
+  /// \sa SetTypeAndColor
+  vtkGetMacro(Type, int);
+
+  ///
+  /// Set the type of the plot (line, scatter, bar)
+  virtual void SetType(int type);
+
+  ///
+  /// Get/Set the status of the Plot
+  vtkGetMacro(Dirty, bool);
```
Punzo: It is needed here: https://github.com/Punzo/Slicer/commit/29316ce388dadfcfe92996162fd48c791823a869#diff-2230803b602f286c1352621a5cddef3fR685

essentially when you change the vtkPlot pointer stored in vtkMRMLPlotNode, it is needed to remove the old vtkplot from the view "manually"

Libs/MRML/Core/vtkMRMLPlotNode.h
```diff
@@ -0,0 +1,249 @@
...
+  /// Write this node's information to a MRML file in XML format.
+  virtual void WriteXML(ostream& of, int indent);
+
+  ///
+  /// Copy the node's attributes to this object
+  virtual void Copy(vtkMRMLNode *node);
+
+  ///
+  /// Copy the node's attributes to this object but changing the name and type
+  virtual void CopyAndSetNameAndType(vtkMRMLNode *node, const char *name, int Type);
```
Punzo: yes. I just thought taht it could be useful one method to do everything in one line, instead of calling:
-Copy.
-SetName.
-SetType.

Libs/MRML/Core/vtkMRMLPlotLayoutNode.h
```diff
@@ -0,0 +1,261 @@
...
+  virtual vtkCollection *GetSelectionArrays();
+
+  ///
+  /// Set the PlotNodeIDs and selectedArrays
+  virtual void SetSelection(vtkStringArray* selectedPlotNodeIDs,
+                            vtkCollection* selectedArrays);
+
+  ///
+  /// Get the index of a PlotNode
+  virtual vtkIdType GetPlotNodeIndex(vtkMRMLPlotNode* plotNode);
```
Punzo: is the "index" of teh plot in plotNodes collection. But plotNodes should go away as well updating to the reference infrastructure

Punzo:
Ok I have seen in detail the comments and replied. There is some work relative to use properly the reference infrastracture and removing pointers not needed in vtkMRMLPlotLayoutNode. Next week I'll revise it and I'll submit a commit for this (+ tests). I'll also make a LAB page on the Slicer wiki for a skeleton documentation that it may be used also later on for the final docs.

Libs/MRML/Core/vtkMRMLPlotNode.h
lassoan: I don't think this combination would be such a common operation that it would warrant a convenience function.

Libs/MRML/Core/vtkMRMLPlotNode.h
lassoan: Instead of flags like this, all changes should be managed by observations, checking ModifiedTime (change can be detected by comparing ModifiedTime to a stored ModifiedTime), etc.

Libs/MRML/Core/vtkMRMLPlotNode.h
lassoan: This seems to be unnecessarily complicated. A MRML node should be a higher-level concept that a user can visualize, edit, etc. A vtkPlot object is lower level: you cannot display a vtkPlot object - you need a vtkChart (and scene, etc) for that.

So, instead of having a separate vtkMRMLPlotNode and vtkMRMLPlotLayoutNode, I would recommend to store all data in a single vtkMRMLPlotNode class. The implementation would be simpler than the current one, because you would not need to manage so many node references, you would store all the data in a single class (you would still need to use node references for table nodes, but not for plot nodes).

You can have a look at how the [sequence browser node](https://github.com/SlicerRt/Sequences/blob/master/SequenceBrowser/MRML/vtkMRMLSequenceBrowserNode.cxx) refers to multiple nodes and associated metadata for each node: It stores a list of "postfix" strings (usually "0", "1", "2", ...) uses that to generate unique role names and unique attribute names.

```
<SequenceBrowser id="vtkMRMLSequenceBrowserNode1" name="SequenceBrowser"
 virtualNodePostfixes="0 1 2" references="dataNodeRef0:vtkMRMLMarkupsFiducialNode1;dataNodeRef1:vtkMRMLMarkupsFiducialNode2;dataNodeRef2:vtkMRMLMarkupsFiducialNode3;sequenceNodeRef0:vtkMRMLSequenceNode2;sequenceNodeRef1:vtkMRMLSequenceNode3;sequenceNodeRef2:vtkMRMLSequenceNode4;"
 SynchronizationPropertiesMap0="playback true recording true overwriteProxyName false saveChanges false "  
 SynchronizationPropertiesMap1="playback true recording true overwriteProxyName true saveChanges true "  
 SynchronizationPropertiesMap2="playback false recording false overwriteProxyName false saveChanges false " >
</SequenceBrowser>
```
For plots, it would look something like this:
```
<Plot id="vtkMRMLSequenceBrowserNode1" name="SequenceBrowser"
 ...
 plotPostfixes="0 1 3"
 references="tableNodeRef0:vtkMRMLTableNode1;tableNodeRef1:vtkMRMLTableNode2;tableNodeRef3:vtkMRMLTableNode5"
 plotProperties0="name:My plot 1|type:line|xColumnName:time|yColumnName:distance"
 plotProperties1="name:My plot 2|type:line|xColumnName:time|yColumnName:distance"
 plotProperties3="name:My plot 3|type:line|xColumnName:time|yColumnName:distance"
</Plot>
```

jcfr:
>  I saw your commit regarding vtk8 and qt5 (which is great by the way!). Is it better that I'll wait their > 
> integration before pushing Punzo/VTK@1d5a921 to the https://github.com/Slicer/VTK as suggested by 
> Andras?

After submitting a [merge request](https://gitlab.kitware.com/vtk/vtk/blob/master/CONTRIBUTING.md) to upstream VTK:
* we will cherry pick the change and integrate to [Slicer/VTK@slicer-v8.0.0-2017-08-07-88c80af](https://github.com/Kitware/VTK/compare/master...Slicer:slicer-v8.0.0-2017-08-07-88c80af) - Branch used when building with Qt4 + VTK7.1+ OpenGL backend
* and backport it on [Slicer/VTK@slicer-v7.1.0-2016-09-23-3b13ad9](https://github.com/Kitware/VTK/compare/master...Slicer:slicer-v7.1.0-2016-09-23-3b13ad9)- Branch used when building with Qt5 + VTK8 + OpenGL2 backend

Libs/MRML/Core/vtkMRMLPlotNode.h
Punzo: Hi @lassoan, sorry for the late reply. I understand the call for a simpler MRML nodes structure, however I still prefer the two nodes structure for the following reasons: 
1) I actually like the double structure because of the information regarding one MRMLPlotNode is stored and always accessible by the scene. For example, if I remove the ID of a MRMLPlotNode from MRMLPlotLayoutNode, I can always readd it anytime. This is useful in the controller widget. See example in this video:
https://www.dropbox.com/s/08lubmg2bx7wcbz/videoPlot1.mkv?dl=0

while in your proposed case, the user needs to recreate the plot or I have to add something like a visibility parameter (of course then the widget needs some work).

2) Having two classes make a clear distinction between the Layout and its properties (e.g., type of the font, size of the font, etc..) and of the properties of the single plot (e.g., width of the line of the plot).

3) having a MRMLPlotLayoutNode with only IDs of MRMLPlotNodes (instead of storing the pointers of vtkplots directly) avoids duplication of memory for creating the same vtkPlot in more "Layout" nodes.

Please let me know. If you really think a simpler MRML infrastructure in this case is much better anyway, then I'll see how to rearrange everything.

Libs/MRML/Core/vtkMRMLPlotNode.h
Punzo: I am not sure anymore about this. Do you mean that we should allocate display nodes as well?

Libs/MRML/Core/vtkMRMLPlotNode.h
Punzo: well the problem is more that we a new vtkplot pointer and the old one has to be manually removed from the vtkChartXY. I guess the best is just to fire something like vtkPlotRemoved from MRMLPlotNode everytime a vtkplot is deleted and put a qvtkconnect at the widget level.

Punzo:
VTK Click and Drag action ref: https://gitlab.kitware.com/vtk/vtk/merge_requests/3160

Libs/MRML/Core/vtkMRMLPlotNode.h
lassoan: In VTK, an actor cannot be in two renderers at the same time. Are you sure you can reuse a vtkPlot on multiple vtkCharts? Are interaction, selection, zoom, etc settings of the same plot synchronized in different charts?

Punzo: It seems working:
https://www.dropbox.com/s/7g0ui4ysi6b90el/videoPlot2.mkv?dl=0

Interactions are not handled by vtkplots, but by vtkChartXY and vtkChartContext.

lassoan: OK. We can try to keep plots as separate nodes then.

After studying the code a bit more, it seems that removing vtkMRMLPlotLayoutNode (by moving all responsibilities to vtkMRMLPlotViewNode) would simplify things. vtkMRMLPlotLayoutNode is a very lightweight class, so if you need to display the same chart in several views then you could easily clone the vtkMRMLPlotViewNode (plot properties would be still shared). What do you think?

Punzo: It makes sense. Also the name will fit because the "Layout" properties are part of the View in a certain way. I like it and in this way we can remove one MRML class. I'll try and see if everything will work. Maybe for tomorrow afternoon I'll have the new version.

Libs/MRML/Core/vtkMRMLPlotNode.h
lassoan: You can add an observer to node reference change events (node reference added/removed/modified). If that changes then you update the vtkPlot list (add new plots, remove orphan plot objects, etc).

Libs/MRML/Core/vtkMRMLPlotNode.h
lassoan: I've checked what properties are stored in vtkMRMLDisplayNode and almost all of them are irrelevant. So, we can keep this as is.

lassoan:
@Punzo Thank you, the code is much nicer now.

vtkMRMLPlotLayout node name is still very confusing, as we have layout manager, layout node, we can switch between layouts - but these all have nothing to do with vtkMRMLPlotLayout. vtkMRMLPlotViewNode is basically empty and vtkMRMLPlotLayout stores information about how to view plots, so I think the best would be to move all contents of vtkMRMLPlotLayout into vtkMRMLPlotViewNode and delete vtkMRMLPlotLayout class.

jcfr:
Ditto. Thanks for working on this :smile: 

>  so I think the best would be to move all contents of vtkMRMLPlotLayout into vtkMRMLPlotViewNode and delete vtkMRMLPlotLayout class.

:+1: 

A Layout is what bring one or multiple "view" together.

In that context, a layout is describes using a simple XML string.

Punzo:
Hi thanks, I was writing the Lab page with docuemntation: https://www.slicer.org/wiki/Documentation/Labs/VTKPLOT#Constructing_a_plot

Before to merge we have to wait https://gitlab.kitware.com/vtk/vtk/merge_requests/3160

Punzo:
OK I tried to move PlotLayoutNode into the PlotViewNode. I run into some issues. Let me check again a final time the code and I still I don't succeed, I'll try to explain my issue (I agree that it will be better to remove one class).

Punzo:
Fixed few bugs. I'll check how to remove PlotLayoutNode later on next week. I am already in vacation so it can be that I'll reply at the end of the week (: .

lassoan:
No problem, have a nice vacation. We are in preparation for releasing a new stable version, so most likely the integration would not happen anyway within 1-2 weeks.

Punzo:
Hi @lassoan, @jcfr . I tried to remove the vtkMRMLPlotLayoutNode, but that it is not possible:

1) vtkMRMLPlotViewNode is child of vtkMRMLAbstractViewNode. This means that only one node can created for each PlotWidget (they need to be coupled, see: 
https://github.com/Slicer/Slicer/blob/master/Libs/MRML/Widgets/qMRMLLayoutManager.cxx).  

2) Having only one "LayoutNode" for each Widget does not make sense: in fact in this case the user will lose all the settings relative to the "PlotLayout" if he/she wants to generate another plot. 

3) Therefore, I propose to leave the mrml structure as it is 
     (https://www.slicer.org/wiki/Documentation/Labs/VTKPLOT#MRML_Nodes), but we can rename 
     vtkMRMLLayoutNode in a way that reflects more its role: a node that keeps track of which plot to 
     display in a PlotView.
     Few example: vtkMRMLPlotFrameNode, vtkMRMLPlotContextNode, vtkMRMLMapPlotsNode.

Punzo:
P.S: @jcfr  do you think that https://gitlab.kitware.com/vtk/vtk/merge_requests/3160 is ready? and do you have the rights to merge it?

lassoan:
> Having only one "LayoutNode" for each Widget does not make sense: the user has to overwrite all the Attributes relative to the "PlotLayout" if he/she wants to generate another plot (including title, axes, etc...) in the same widget.

This is how it works in Paraview, but I agree that it may be better have the option of storing multiple charts independently from views.

> I propose to leave the mrml structure as it is ... but we can rename vtkMRMLLayoutNode

OK. What do you think about vtkMRMLPlotChartNode? We could also rename vtkMRMLPlotNode to be more specific and so use "Plot" as a prefix.

Class names would be then:
* vtkMRMLPlotSeriesNode (or vtkMRMLPlotDataNode)
* vtkMRMLPlotChartNode
* vtkMRMLPlotViewNode

Punzo:
ok, perfect. I also like the names (for the first one to me sound better vtkMRMLPlotDataNode, but for no reason). @fedorov @pieper @jcfr  may you also give your preference for the names?

pieper:
+1 to the proposed name change

jcfr:
VTKv7 and VTKv8 used in Slicer have been updated to support the integration of this PR.

I suggest to update the topic squashing the changes together into one or few commits.

jcfr:
@Punzo  I rebased the topic and squashed the commit. See https://github.com/Slicer/Slicer/compare/master...jcfr:edf4a8e3176af3cf921dd49be5d21eda586c8bfa

Could you review ? 

also it would be nice to add a short description to the commit. May be a summary of the PR description ?

lassoan:
Is "European Research Consil" in the file headers a typo (do you mean Council)?

Punzo:
@jcfr  Thanks for rebasing and squashing. I'll check right now. @lassoan yes a typo, thanks!

Punzo:
I'll also think a comment for the commit. 
P.S.:
link LAB page: https://www.slicer.org/wiki/Documentation/Labs/VTKPLOT
Shall this be moved into the documentation?

Punzo:
ok perfect I was fixing the typo, you were faster (: 

https://www.slicer.org/wiki/Documentation/Labs/VTKPLOT is already similar to https://www.slicer.org/wiki/Documentation/Nightly/Developers/Charts

If it is fine for you I'll move https://www.slicer.org/wiki/Documentation/Labs/VTKPLOT into 
https://www.slicer.org/wiki/Documentation/Nightly/Developers/Plots

jcfr:
I missed you previous comment.

> link LAB page: https://www.slicer.org/wiki/Documentation/Labs/VTKPLOT
> Shall this be moved into the documentation?

Yes

Punzo:
ok, here is a comment for the commit, let me know if it fine : 

"
Integration of VTK plotting capabilities in Slicer. 

This provides Plotting facilities that include a Plot View that can be packed in the layout, similar to the Slice Views and 3D Views. The architecture also includes nodes that represent the data to be displayed in the plot as well as nodes to represent the display properties of the data and the plot itself. In Addition, the PlotView is connected with the Table Module and Views allowing the plotting of arrays with a simple click.

references:
https://www.slicer.org/wiki/Documentation/Nightly/Developers/Plots
https://www.na-mic.org/wiki/Project_Week_25/Interactive_Manipulation_of_Plots_and_Graphs
https://www.slicer.org/wiki/Developer_Meetings/20170425
"

Punzo:
documentation moved from https://www.slicer.org/wiki/Documentation/Labs/VTKPLOT to https://www.slicer.org/wiki/Documentation/Nightly/Developers/Plots

Punzo:
@jcfr https://github.com/Slicer/Slicer/compare/master...jcfr:PlottingVTK-rebased for me is good to go, thanks!

jcfr:
Great.

If there are no objection, I will integrate the PR this afternoon.

Regarding the wiki pages, would it be possible to explain why one approach should be preferred vs the other ?

May a page named https://www.slicer.org/wiki/Documentation/Nightly/Developers/Plots-vs-Charts could be linked of the the description of both Plots and Charts section on the front developer page ...

Punzo:
ok. Is a entire page too much? What about just a short (two/three lines) exaplanation at the beginning of both pages?

jcfr:
> What about just a short (two/three lines) explanation at the beginning of both pages?

That would work too.

In that case, consider creating a "Charting" category in https://www.slicer.org/wiki/Documentation/Nightly/Developers/FAQ by adding an entry like "What is the difference between Slicer Plot and Chart ?"

Punzo:
Ok

jcfr:
Thanks

pieper:
Nice work!

jcfr:
Integrated in [r26380](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26380)

Remaining documentation items:
* [ ] update [developer FAQ](https://www.slicer.org/wiki/Documentation/Nightly/Developers/FAQ) and reference the entry from the front page
* [ ] on the [plots wiki page](https://www.slicer.org/wiki/Documentation/Nightly/Developers/Plots#Constructing_a_plot), could you remove the hardcoded reference to layout `39` and use the corresponding enum name (e.g `slicer.vtkMRMLLayoutNode.NameOfCorrespondingLayout`)

Punzo:
Documentation updated:
1) Added link to developer FAQ in 
https://www.slicer.org/wiki/Documentation/Nightly/Developers/Charts#Chart_capabilities
https://www.slicer.org/wiki/Documentation/Nightly/Developers/Plots#Plot_capabilities

2) developer FAQ extended (please check it) 
https://www.slicer.org/wiki/Documentation/Nightly/Developers/FAQ#What_is_the_difference_between_Slicer_Plot_and_Chart_.3F

3) fixed wiki Plot with enum instead of interger

lassoan:
@Punzo I've entered two issuesin mantis:
https://issues.slicer.org/view.php?id=4435
https://issues.slicer.org/view.php?id=4436

Have you received notification?
Is "Davide" your username?
Could you fix them?

Punzo:
Hi @lassoan, sorry didn't see. Davide is my username on mantis (Although I didn't recerived any notification and I actually I have issues for login). Anyway, I'll look into them.

PR #771, opened 2017-08-11
ENH: Support Qt5 in dashboard script

msmolens:
When building with Qt4, QT_QMAKE_EXECUTABLE is required.
When building with Qt5, Qt5_DIR is required.


jcfr:
Integrated in [26268](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26268)

PR #773, opened 2017-08-15
Update segment statistics

chribaue:
We added a plugin mechanism for the SegmentStatistics module with configuration via a parameter node. In addition, the plugins provide DICOM related information for the measurements as required for Quantitative Reporting and DICOM SR export.

### DIFF
--- a/Modules/Scripted/SegmentStatistics/CMakeLists.txt
+++ b/Modules/Scripted/SegmentStatistics/CMakeLists.txt
@@ -2,6 +2,7 @@
 set(MODULE_NAME SegmentStatistics)
 
 add_subdirectory(SubjectHierarchyPlugins)
+add_subdirectory(SegmentStatisticsPlugins)
 
 #-----------------------------------------------------------------------------
 set(MODULE_PYTHON_SCRIPTS
--- a/Modules/Scripted/SegmentStatistics/SegmentStatistics.py
+++ b/Modules/Scripted/SegmentStatistics/SegmentStatistics.py
@@ -1,12 +1,8 @@
-import os
-import unittest
 import vtk, qt, ctk, slicer
 from slicer.ScriptedLoadableModule import *
 import logging
+from SegmentStatisticsPlugins import *
 
-#
-# SegmentStatistics
-#
 
 class SegmentStatistics(ScriptedLoadableModule):
   """Uses ScriptedLoadableModule base class, available at:
@@ -14,37 +10,34 @@ class SegmentStatistics(ScriptedLoadableModule):
   """
 
   def __init__(self, parent):
-    import string
     ScriptedLoadableModule.__init__(self, parent)
     self.parent.title = "Segment Statistics"
     self.parent.categories = ["Quantification"]
     self.parent.dependencies = ["SubjectHierarchy"]
-    self.parent.contributors = ["Andras Lasso (PerkLab), Steve Pieper (Isomics)"]
+    self.parent.contributors = ["Andras Lasso (PerkLab), Christian Bauer (University of Iowa), Steve Pieper (Isomics)"]
     self.parent.helpText = """
 Use this module to calculate counts and volumes for segments plus statistics on the grayscale background volume.
 Computed fields:
-Segment labelmap stastistics (LM): voxel count, volume mm3, volume cc.
+Segment labelmap statistics (LM): voxel count, volume mm3, volume cm3.
 Requires segment labelmap representation.
-Grayscale volume statistics (GS): voxel count, volume mm3, volume cc (where segments overlap grayscale volume),
+Scalar volume statistics (SV): voxel count, volume mm3, volume cm3 (where segments overlap scalar volume),
 min, max, mean, stdev (intensity statistics).
-Requires segment labelmap representation and selection of a grayscale volume
-Closed surface statistics (CS): surface mm2, volume mm3, volume cc (computed from closed surface).
+Requires segment labelmap representation and selection of a scalar volume
+Closed surface statistics (CS): surface mm2, volume mm3, volume cm3 (computed from closed surface).
 Requires segment closed surface representation.
 """
     self.parent.helpText += self.getDefaultModuleDocumentationLink()
     self.parent.acknowledgementText = """
 Supported by NA-MIC, NAC, BIRN, NCIGT, and the Slicer Community. See http://www.slicer.org for details.
 """
-
   def setup(self):
     # Register subject hierarchy plugin
     import SubjectHierarchyPlugins
     scriptedPlugin = slicer.qSlicerSubjectHierarchyScriptedPlugin(None)
     scriptedPlugin.setPythonSource(SubjectHierarchyPlugins.SegmentStatisticsSubjectHierarchyPlugin.filePath)
+    
+    import SegmentStatisticsPlugins
 
-#
-# SegmentStatisticsWidget
-#
 
 class SegmentStatisticsWidget(ScriptedLoadableModuleWidget):
   """Uses ScriptedLoadableModuleWidget base class, available at:
@@ -57,8 +50,8 @@ class SegmentStatisticsWidget(ScriptedLoadableModuleWidget):
     self.logic = SegmentStatisticsLogic()
     self.grayscaleNode = None
     self.labelNode = None
-    self.fileName = None
-    self.fileDialog = None
+    self.parameterNode = None
+    self.parameterNodeObserver = None
 
     # Instantiate and connect widgets ...
     #
@@ -79,17 +72,17 @@ class SegmentStatisticsWidget(ScriptedLoadableModuleWidget):
     self.segmentationSelector.setToolTip( "Pick the segmentation to compute statistics for" )
     inputsFormLayout.addRow("Segmentation:", self.segmentationSelector)
 
-    # Grayscale volume selector
-    self.grayscaleSelector = slicer.qMRMLNodeComboBox()
-    self.grayscaleSelector.nodeTypes = ["vtkMRMLScalarVolumeNode"]
-    self.grayscaleSelector.addEnabled = False
-    self.grayscaleSelector.removeEnabled = True
-    self.grayscaleSelector.renameEnabled = True
-    self.grayscaleSelector.noneEnabled = True
-    self.grayscaleSelector.showChildNodeTypes = False
-    self.grayscaleSelector.setMRMLScene( slicer.mrmlScene )
-    self.grayscaleSelector.setToolTip( "Select the grayscale volume for intensity statistics calculations")
-    inputsFormLayout.addRow("Grayscale volume:", self.grayscaleSelector)
+    # Scalar volume selector
+    self.scalarSelector = slicer.qMRMLNodeComboBox()
+    self.scalarSelector.nodeTypes = ["vtkMRMLScalarVolumeNode"]
+    self.scalarSelector.addEnabled = False
+    self.scalarSelector.removeEnabled = True
+    self.scalarSelector.renameEnabled = True
+    self.scalarSelector.noneEnabled = True
+    self.scalarSelector.showChildNodeTypes = False
+    self.scalarSelector.setMRMLScene( slicer.mrmlScene )
+    self.scalarSelector.setToolTip( "Select the scalar volume for intensity statistics calculations")
+    inputsFormLayout.addRow("Scalar volume:", self.scalarSelector)
 
     # Output table selector
     outputCollapsibleButton = ctk.ctkCollapsibleButton()
@@ -114,19 +107,61 @@ class SegmentStatisticsWidget(ScriptedLoadableModuleWidget):
     self.applyButton.enabled = False
     self.parent.layout().addWidget(self.applyButton)
 
+    # Parameter set
+    parametersCollapsibleButton = ctk.ctkCollapsibleButton()
+    parametersCollapsibleButton.text = "Parameters"
+    #parametersCollapsibleButton.collapsed = True
+    self.layout.addWidget(parametersCollapsibleButton)
+    self.parametersLayout = qt.QFormLayout(parametersCollapsibleButton)
+
+    # Parameter set selector
+    self.parameterNodeSelector = slicer.qMRMLNodeComboBox()
+    self.parameterNodeSelector.nodeTypes = ( ("vtkMRMLScriptedModuleNode"), "" )
+    self.parameterNodeSelector.addAttribute( "vtkMRMLScriptedModuleNode", "ModuleName", "SegmentStatistics" )
+    self.parameterNodeSelector.selectNodeUponCreation = True
+    self.parameterNodeSelector.addEnabled = True
+    self.parameterNodeSelector.renameEnabled = True
+    self.parameterNodeSelector.removeEnabled = True
+    self.parameterNodeSelector.noneEnabled = False
+    self.parameterNodeSelector.showHidden = True
+    self.parameterNodeSelector.showChildNodeTypes = False
+    self.parameterNodeSelector.baseName = "SegmentStatistics"
+    self.parameterNodeSelector.setMRMLScene( slicer.mrmlScene )
+    self.parameterNodeSelector.setToolTip( "Pick parameter set" )
+    self.parametersLayout.addRow("Parameter set: ", self.parameterNodeSelector)
+
+    # Edit parameter set button to open SegmentStatisticsParameterEditorDialog
+    # Note: we add the plugins' option widgets to the module widget instead of using the editor dialog
+    #self.editParametersButton = qt.QPushButton("Edit Parameter Set")
+    #self.editParametersButton.toolTip = "Editor Statistics Plugin Parameter Set."
+    #self.parametersLayout.addRow(self.editParametersButton)
+    #self.editParametersButton.connect('clicked()', self.onEditParameters)
+    # add caclulator's option widgets
+    self.addPluginOptionWidgets()
+
     # Add vertical spacer
     self.parent.layout().addStretch(1)
 
     # connections
     self.applyButton.connect('clicked()', self.onApply)
-    self.grayscaleSelector.connect('currentNodeChanged(vtkMRMLNode*)', self.onNodeSelectionChanged)
+    self.scalarSelector.connect('currentNodeChanged(vtkMRMLNode*)', self.onNodeSelectionChanged)
     self.segmentationSelector.connect('currentNodeChanged(vtkMRMLNode*)', self.onNodeSelectionChanged)
     self.outputTableSelector.connect('currentNodeChanged(vtkMRMLNode*)', self.onNodeSelectionChanged)
+    self.parameterNodeSelector.connect('currentNodeChanged(vtkMRMLNode*)', self.onNodeSelectionChanged)
+    self.parameterNodeSelector.connect('currentNodeChanged(vtkMRMLNode*)', self.onParameterSetSelected)
 
+    self.parameterNodeSelector.setCurrentNode(self.logic.getParameterNode())
     self.onNodeSelectionChanged()
+    self.onParameterSetSelected()
+
+  def cleanup(self):
+    if self.parameterNode and self.parameterNodeObserver:
+      self.parameterNode.RemoveObserver(self.parameterNodeObserver)
 
   def onNodeSelectionChanged(self):
-    self.applyButton.enabled = (self.segmentationSelector.currentNode() is not None) and (self.outputTableSelector.currentNode() is not None)
+    self.applyButton.enabled = (self.segmentationSelector.currentNode() is not None and
+                                self.outputTableSelector.currentNode() is not None and
+                                self.parameterNodeSelector.currentNode() is not None)
     if self.segmentationSelector.currentNode():
       self.outputTableSelector.baseName = self.segmentationSelector.currentNode().GetName() + ' statistics'
 
@@ -137,8 +172,15 @@ class SegmentStatisticsWidget(ScriptedLoadableModuleWidget):
     self.applyButton.text = "Working..."
     self.applyButton.setEnabled(False)
     slicer.app.processEvents()
+    # set up parameters for computation
+    self.logic.getParameterNode().SetParameter("Segmentation", self.segmentationSelector.currentNode().GetID())
+    if self.scalarSelector.currentNode():
+      self.logic.getParameterNode().SetParameter("ScalarVolume", self.scalarSelector.currentNode().GetID())
+    else:
+      self.logic.getParameterNode().UnsetParameter("ScalarVolume")
+    self.logic.getParameterNode().SetParameter("MeasurementsTable", self.outputTableSelector.currentNode().GetID())
     # Compute statistics
-    self.logic.computeStatistics(self.segmentationSelector.currentNode(), self.grayscaleSelector.currentNode())
+    self.logic.computeStatistics()
     self.logic.exportToTable(self.outputTableSelector.currentNode())
     # Unlock GUI
     self.applyButton.setEnabled(True)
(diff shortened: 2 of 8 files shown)


Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/ScalarVolumeSegmentStatisticsCalculator.py
```diff
@@ -0,0 +1,167 @@
...
+    return stats
+
+  def getMeasurementInfo(self, key):
+
+    scalarVolumeNode = slicer.mrmlScene.GetNodeByID(self.getParameterNode().GetParameter("ScalarVolume"))
+
+    try:
+      scalarVolumeQuantity = scalarVolumeNode.GetVoxelValueQuantity()
+      scalarVolumeUnits = scalarVolumeNode.GetVoxelValueUnits()
+    except AttributeError:
```
lassoan: Add a comment here why this is needed (if needed at all). I guess it is here for backward compatibility? But do we want to use this module with older Slicer version that does not have scalar volume unit support?

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/ScalarVolumeSegmentStatisticsCalculator.py
```diff
@@ -0,0 +1,167 @@
...
+      scalarVolumeQuantity = self.initCodedEntry("", "", "")
+      scalarVolumeUnits = self.initCodedEntry("", "", "")
+    if not scalarVolumeQuantity:
+      scalarVolumeQuantity = self.initCodedEntry("", "", "")
```
lassoan: change initCodedEntry to createCodedEntry to match actual behavior

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/SegmentStatisticsCalculatorBase.py
```diff
@@ -0,0 +1,201 @@
...
+  """
+
+  @staticmethod
+  def initCodedEntry(codeValue, codingScheme, codeMeaning, returnAsString=False):
+    entry = slicer.vtkCodedEntry()
+    entry.SetValueSchemeMeaning(codeValue, codingScheme, codeMeaning)
+    return entry if not returnAsString else entry.GetAsString()
+
+  @staticmethod
+  def getDICOMTriplet(codeValue, codingSchemeDesignator, codeMeaning):
```
lassoan: Why not use a single method for getting a coded entry in any type? (merging initCodedEntry and getDICOMTriplet)

    def createCodedEntry(codeValue, codingScheme, codeMeaning, type='vtk')

type could be 'vtk', 'json', and 'string'

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/SegmentStatisticsCalculatorBase.py
```diff
@@ -0,0 +1,201 @@
...
+    """Utility method to form DICOM triplets in getMeasurementInfo"""
+    return {'CodeValue':codeValue,
+            'CodingSchemeDesignator': codingSchemeDesignator,
+            'CodeMeaning': codeMeaning}
+
+  @staticmethod
+  def generateMeasurementInfo(name, description, units, quantityCode, unitsCode, measurementMethodCode=None,
```
lassoan: What type quantityCode and unitsCode is expected in? VTK, string, or DICOM?
At least the argument name should reflect it, but preferably method documentation should specify it.

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/SegmentStatisticsCalculatorBase.py
```diff
@@ -0,0 +1,201 @@
...
+      "DICOM.UnitsCode": unitsCode
+      }
+    if measurementMethodCode:
+      info["DICOM.MeasurementMethodCode"] = measurementMethodCode
+    if derivationCode:
+      info["DICOM.DerivationCode"] = derivationCode
+    return info
+
+  def __init__(self):
+    self.name = "" # name of the statistics calculator
```
lassoan: Documentation of base plugin class may be used by developers, so it would be useful if you could document it in a way to facilitate automatic documentation generation later.
Sphinx expects member variable documentation in the line above the assignment, prefixed by ```# :``` (see http://www.sphinx-doc.org/en/stable/ext/autodoc.html#directive-autoattribute)

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/SegmentStatisticsCalculatorBase.py
```diff
@@ -0,0 +1,201 @@
...
+    self.parameterNodeObserver = None
+
+  def __del__(self):
+    if self.parameterNode and self.parameterNodeObserver:
+      self.parameterNode.RemoveObserver(self.parameterNodeObserver)
+
+  def computeStatistics(self, segmentID):
+    """Compute measurements for requested keys on the given segment"""
+    stats = {}
+    #segmentationNode = slicer.mrmlScene.GetNodeByID(self.getParameterNode().GetParameter("Segmentation"))
```
lassoan: can this be removed?

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/SegmentStatisticsCalculatorBase.py
```diff
@@ -0,0 +1,201 @@
...
+  def updateGuiFromParameterNode(self, caller=None, event=None):
+    if not self.parameterNode:
+      return
+    isEnabled = self.parameterNode.GetParameter(self.name+'.enabled')!='False'
+    self.calculatorCheckbox.checked = isEnabled
+    for (key, checkbox) in self.requestedKeysCheckboxes.iteritems():
+      parameter = key+'.enabled'
+      value = self.parameterNode.GetParameter(parameter)=='True'
+      if checkbox.checked!=value:
+        checkbox.blockSignals(True)
```
lassoan: Always store return value of blockSignals and restore that value. If you hardcode blockSignals(False) then you may unblock a widget that was already blocked before.

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/ClosedSurfaceSegmentStatisticsCalculator.py
```diff
@@ -0,0 +1,72 @@
...
+    massProperties = vtk.vtkMassProperties()
+    massProperties.SetInputData(segmentClosedSurface)
+
+    # Add data to statistics list
+    ccPerCubicMM = 0.001
+    stats = {}
+    if "Closed Surface.surface_mm2" in requestedKeys:
+      stats["Closed Surface.surface_mm2"] = massProperties.GetSurfaceArea()
+    if "Closed Surface.volume_mm3" in requestedKeys:
+      stats["Closed Surface.volume_mm3"] = massProperties.GetVolume()
```
lassoan: You are mixing human-readable string (Closed Surface) that may need to be translated, contain international characters, etc., and machine code (volume_mm3) that uses restricted character set (it shouldn't contain your separator character, or other special characters).
Either use the class name to distinguish plugins; or have a separate plugin name and title attributes (name: machine-readable, such as "ClosedSurface" or "closed_surface"; title: human-readable, such as "Closed surface").

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/LabelmapSegmentStatisticsCalculator.py
```diff
@@ -0,0 +1,94 @@
...
+      stats["Labelmap.volume_cc"] = stat.GetVoxelCount() * cubicMMPerVoxel * ccPerCubicMM
+    return stats
+
+  def getMeasurementInfo(self, key):
+    """Get information (name, description, units, ...) about the measurement for the given key"""
+    info = {}
+
+    # @fedorov could not find any suitable code.
+    # DCM has "Number of needles" etc., so probably "Number of voxels"
+    # should be added too. Need to discuss with @dclunie. For now, a
```
fedorov: This is my omission. I think it should be `("{voxels}", "UCUM", "voxels")`, extrapolating from the reference to needles is in row 11 of [TID4202](http://dicom.nema.org/medical/dicom/current/output/chtml/part16/sect_TID_4204.html).

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/LabelmapSegmentStatisticsCalculator.py
```diff
@@ -0,0 +1,94 @@
...
+    # QIICR private scheme placeholder.
+
+    info["Labelmap.voxel_count"] = \
+      self.generateMeasurementInfo(name="Voxel count", description="Number of voxels", units="voxels",
+                                   quantityCode=self.initCodedEntry("nvoxels", "99QIICR", "Number of voxels", True),
```
fedorov: The idea was to use this as a temporary code (coding scheme 99*) while adding it to the standard, but we should instead use the code mentioned in my response above - `("{voxels}", "UCUM", "voxels")`.

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/ScalarVolumeSegmentStatisticsCalculator.py
```diff
@@ -0,0 +1,167 @@
...
+      stats["Scalar Volume.volume_mm3"] = stat.GetVoxelCount() * cubicMMPerVoxel
+    if "Scalar Volume.volume_cc" in requestedKeys:
+      stats["Scalar Volume.volume_cc"] = stat.GetVoxelCount() * cubicMMPerVoxel * ccPerCubicMM
+    if stat.GetVoxelCount()>0:
+      if "Scalar Volume.min" in requestedKeys:
+        stats["Scalar Volume.min"] = stat.GetMin()[0]
+      if "Scalar Volume.max" in requestedKeys:
+        stats["Scalar Volume.max"] = stat.GetMax()[0]
+      if "Scalar Volume.mean" in requestedKeys:
+        stats["Scalar Volume.mean"] = stat.GetMean()[0]
```
fedorov: This specific plugin is ScalarVolume plugin, so the quantity of the mean is defined by the scalar type. The size of the island you refer to in your example would be computed by the Labelmap plugin or similar.

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/LabelmapSegmentStatisticsCalculator.py
fedorov: @chribaue let me know if you want me to fix this and make a PR to your branch.

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+set(${MODULE_NAME}Calculators_PYTHON_SCRIPTS
+  #${MODULE_NAME}Calculators
+  __init__
+  ${MODULE_NAME}CalculatorBase
+  Labelmap${MODULE_NAME}Calculator
```
chribaue: Hi @lassoan! Thank you for your constructive review. I made the requested changed. A point by point response to your comments follows below.

> use SegmentStatistics instead of ${MODULE_NAME

done

> replace "Calculator" by "Plugin"

done

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/ClosedSurfaceSegmentStatisticsCalculator.py
```diff
@@ -0,0 +1,72 @@
...
+from SegmentStatisticsCalculators import SegmentStatisticsCalculatorBase
+
+
+class ClosedSurfaceSegmentStatisticsCalculator(SegmentStatisticsCalculatorBase):
+  """Statistical calculator for closed surfaces"""
+
+  def __init__(self):
+    super(ClosedSurfaceSegmentStatisticsCalculator,self).__init__()
+    self.name = "Closed Surface"
+    self.keys = tuple(self.name+'.'+m for m in ("surface_mm2", "volume_mm3", "volume_cc"))
```
chribaue: I simplified how keys are specified in plugins. It's now simply self.keys=["surface_mm2", "volume_mm3", "volume_cc"]. SegmentStatisticsLogic and the parameters in the ParameterNode now add the Plugin's class name+'.' as prefix to the keys where needed. Handling of this prefix is done by SegmentStatisticsLogic and SegmentsStatisticsPluginBase. Developers of plugins do not have to deal wit it.

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/ClosedSurfaceSegmentStatisticsCalculator.py
```diff
@@ -0,0 +1,72 @@
...
+    segmentClosedSurface = segment.GetRepresentation(closedSurfaceName)
+
+    # Compute statistics
+    massProperties = vtk.vtkMassProperties()
+    massProperties.SetInputData(segmentClosedSurface)
+
+    # Add data to statistics list
+    ccPerCubicMM = 0.001
+    stats = {}
+    if "Closed Surface.surface_mm2" in requestedKeys:
```
chribaue: see comment above; it's now simply `if "surface_mm2" in requestedKeys:`

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/ClosedSurfaceSegmentStatisticsCalculator.py
chribaue: same as above

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/ClosedSurfaceSegmentStatisticsCalculator.py
```diff
@@ -0,0 +1,72 @@
...
+
+  def getMeasurementInfo(self, key):
+    """Get information (name, description, units, ...) about the measurement for the given key"""
+    info = dict()
+
+    # I searched BioPortal, and found seemingly most suitable code.
+    # Prefixed with "99" since CHEMINF is not a recognized DICOM coding scheme.
+    # See https://bioportal.bioontology.org/ontologies/CHEMINF?p=classes&conceptid=http%3A%2F%2Fsemanticscience.org%2Fresource%2FCHEMINF_000247
+    #
+    info["Closed Surface.surface_mm2"] = \
```
chribaue: same as above

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/ClosedSurfaceSegmentStatisticsCalculator.py
```diff
@@ -0,0 +1,72 @@
...
+                                   quantityCode=self.initCodedEntry("000247", "99CHEMINF", "Surface area", True),
+                                   unitsCode=self.initCodedEntry("mm2", "UCUM", "squared millimeters", True))
+
+    info["Closed Surface.volume_mm3"] = \
+      self.generateMeasurementInfo(name="Volume mm3", description="Volume in mm3", units="mm3",
+                                   quantityCode=self.initCodedEntry("G-D705", "SRT", "Volume", True),
+                                   unitsCode=self.initCodedEntry("mm3", "UCUM", "cubic millimeter", True))
+
+    info["Closed Surface.volume_cc"] = \
+      self.generateMeasurementInfo(name="Volume cm3", description="Volume in cm3", units="cm3",
```
chribaue: maybe we can add this functionality later

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/LabelmapSegmentStatisticsCalculator.py
chribaue: I made the changes according to @fedorov; the comment is not necessary anymore

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/LabelmapSegmentStatisticsCalculator.py
chribaue: I made the changes according to @fedorov

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/ScalarVolumeSegmentStatisticsCalculator.py
chribaue: What 'mean' refers to depends on the plugin and the description of the measurement info should provide information about the quantity.

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/ScalarVolumeSegmentStatisticsCalculator.py
chribaue: @che85 could you comment on this?

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/ScalarVolumeSegmentStatisticsCalculator.py
chribaue: done

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/ScalarVolumeSegmentStatisticsCalculator.py
```diff
@@ -0,0 +1,167 @@
...
+      self.generateMeasurementInfo(name="Voxel count", description="Number of voxels", units="voxels",
+                                   quantityCode=self.initCodedEntry("nvoxels", "99QIICR", "Number of voxels", True),
+                                   unitsCode=self.initCodedEntry("voxels", "UCUM", "voxels", True))
+
+    info["Scalar Volume.volume_mm3"] = \
+      self.generateMeasurementInfo(name="Volume mm3", description="Volume in mm3", units="mm3",
+                                   quantityCode=self.initCodedEntry("G-D705", "SRT", "Volume", True),
+                                   unitsCode=self.initCodedEntry("mm3", "UCUM", "cubic millimeter", True))
+
+    info["Scalar Volume.volume_cc"] = \
```
chribaue: I changed `cc` to `cm3`

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/SegmentStatisticsCalculatorBase.py
chribaue: `getDICOMTriplet` is obsolete and I removed it

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/SegmentStatisticsCalculatorBase.py
chribaue: I changed `quantityCode` to `quantityDicomCode` etc.

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/SegmentStatisticsCalculatorBase.py
chribaue: done

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/SegmentStatisticsCalculatorBase.py
chribaue: I removed it. I had it as a reference for plugin developers but the plugins themselves are probably better references.

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/SegmentStatisticsCalculatorBase.py
chribaue: done

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/ScalarVolumeSegmentStatisticsCalculator.py
che85: @fedorov Could you answer this question?

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/LabelmapSegmentStatisticsCalculator.py
fedorov: Ok guys, sorry I must admit I messed up! I was in a rush yesterday, and didn't read the comment carefully. I was confused and thought the discussion is about the units. However, the issue in question is the quantity defining the measurement.

Indeed, the quantity should be "Number of voxels" , there is no such quantity defined in the standard, and that's why I referred to the "Number of needles" as an example. I think I was confused because Andras mistakenly thought this code defines the units. So `quantityCode` should remain the same private code (I am fine if the CodeValue is replaced with something other than `nvoxels`, but now that I clarified the meaning, maybe it is not necessary). We cannot replace it with a UCUM units code here.

@chribaue I am really sorry for the confusion. I can make a PR to clean this up.

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/ScalarVolumeSegmentStatisticsCalculator.py
fedorov: I agree this check is not needed. Probably @che85 added this check, since he mistakenly was testing this branch with an older version of Slicer that didn't have units support.

chribaue: Then, I'll remove the try/except block

chribaue: @che85: the title of your PR that introduced this try/except block was:
> ENH: fixed issues when ScalarVolume is uninitialized

Should we leave it?

chribaue: @che85 I think I understand the reason for your change. Two lines above `scalarVolumeNode` is set to
`slicer.mrmlScene.GetNodeByID(self.getParameterNode().GetParameter("ScalarVolume"))`
which might return none `None` or a node not having the method `GetVoxelValueUnits()`
I'll add a comment instead of removing the try/except block

fedorov: I think it should be just a check for `None` and return if it's not initialized. We don't need to check for the presence of the method, since this code does not need to be backward compatible with earlier versions of Slicer.

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/LabelmapSegmentStatisticsCalculator.py
lassoan: Yes, sorry, you are right, quantity is not used in the column header, so "nvoxels" would not be visible to the users.

What is the standard code value for "number of needles": nneedles? It would be good to follow that naming convention, even if it's a private coding scheme now, to make mapping to standard fields in the future more straightforward.

fedorov: The specific code I referred to is from the DCM coding scheme, and there the convention is to assign a number: `(111439, DCM, "Number of needles around target")` (used here: http://dicom.nema.org/medical/dicom/current/output/chtml/part16/sect_TID_4204.html). I chose to assign a somewhat readable code value instead. I think going forward it makes sense to define a private coding scheme in Slicer (or dcmqi), and keep extending/following it, since updating DICOM has a huuuuuge latency. We could just submit proposals for new codes in batches periodically, and then phase out private codes when they get into the standard. But that is an action item for the future, separate from this PR.

lassoan: OK, 111439 does not look like a convention that it would make sense to follow. "nvoxels" is not bad, as it is not displayed to users and it's very short, but it may be easier to maintain the coding scheme if code values are a bit more explicit, such as "NumberOfVoxels", "number_of_voxels", "voxel_count", ...

I just spend time with discussing this minor detail, as these codes may be very difficult to change in the future, so it is important to choose conventions that we can consistently adhere to. If we choose a very simplistic naming scheme (just lowercase letters, no separators) then it may be more difficult to find consistent and expressive code values for longer, more complex items.

fedorov: CodeValue is SH VR, so it can be at most 16 characters (see http://dicom.nema.org/dicom/2013/output/chtml/part05/sect_6.2.html). Inevitably, we will have constraints about how expressive we could be. If we choose a convention like you suggest, we eventually will run into a problem when the approach will hit the 16 character limit, and then the question will become how to abbreviate. I personally think it doesn't matter,  since CodeValue is not expected to be readable. As long as it is unique, it is not a big deal. 

The easiest and least contentious I guess is to just assign a number, and have a central location for contributing new terms to eliminate the possibility of clashes.

chribaue: I would like to refresh that it's optional for the plugins to provide DICOM codes for measurements. If the plugin developer doesn't know the DICOM code or there is no DICOM code for it, the developer does not have to specify it. The SegmentStatistics Module and their plugins should still be fully functional.

Only exporting the measurement to DICOM SR would not work. And I think there should be a warning in the Quantitative Reporting Extension informing the user in such a situation.

lassoan: OK, if there is a character limit then we will not be able to use full words and "nvoxels" is a reasonable choice. For me, "nvoxels" is better than 111222, as there is no need for strict coordination on assignment of numbers, there is less chance of copy-paste errors or typos, similar items can have similar codes, etc.

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/ScalarVolumeSegmentStatisticsCalculator.py
chribaue: I changed it to `if scalarVolumeNode`...

Modules/Scripted/SegmentStatistics/SegmentStatisticsCalculators/LabelmapSegmentStatisticsCalculator.py
chribaue: So, we'll keep it as is?
`quantityCode=self.initCodedEntry("nvoxels", "99QIICR", "Number of voxels", True),`

lassoan: Yes.

fedorov:
Are there any other concerns regarding this PR? Can it be merged?

lassoan:
Looks good to me. Squash commits before merge. I'm traveling, so I cannot merge today.

jcfr:
I will take care of merging

jcfr:
I rebased on top of master, squashed commit together and improve the commit message.

@fedorov @chribaue @che85  Could you review the commit message ?

fedorov:
I would add another item to "Key contributors": 

`@chribaue @lassoan @fedorov: Concept development`

Otherwise (with or without this addition) LGTM!

fedorov:
Maybe also include reference to https://github.com/QIICR/QuantitativeReporting/issues/168. Many discussions there!

jcfr:
@fedorov Thanks for the feedback.

Integrated in [r26293](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26293)

PR #774, opened 2017-08-16
BUG: Fix module updating mainwindow at load time when using Qt5

jcfr:
With Qt5, relying on "Qtimer::singleShow" with a time of 0ms is not
sufficient anymore to ensure the associated slot is called after the
main window is shown.

The following errors were reported at startup time:

```
Failed to obtain reference to 'FileMenu'
Failed to obtain reference to 'qSlicerAppMainWindow'
No Data Probe frame - cannot create DataProbe
Failed to obtain reference to 'qSlicerAppMainWindow'
Failed to obtain reference to 'FileMenu'
```

To ensure module can update the main window once it is shown, this
commit introduces the signal "qSlicerApplication::mainWindowShown()" that
can be connected when a scripted or loadable module is initialized.

Once the main window is shown, the signal "mainWindowShown()" ends up being
triggered.

### DIFF
--- a/Applications/SlicerApp/Main.cxx
+++ b/Applications/SlicerApp/Main.cxx
@@ -209,6 +209,8 @@ int SlicerAppMain(int argc, char* argv[])
     {
     window.reset(new qSlicerAppMainWindow);
     window->setWindowTitle(window->windowTitle()+ " " + Slicer_VERSION_FULL);
+
+    QObject::connect(window.data(), SIGNAL(windowShown()), &app, SIGNAL(mainWindowShown()));
     }
   else if (app.commandOptions()->showPythonInteractor()
     && !app.commandOptions()->runPythonAndExit())
--- a/Applications/SlicerApp/Main.cxx
+++ b/Applications/SlicerApp/Main.cxx
@@ -209,8 +209,6 @@ int SlicerAppMain(int argc, char* argv[])
     {
     window.reset(new qSlicerAppMainWindow);
     window->setWindowTitle(window->windowTitle()+ " " + Slicer_VERSION_FULL);
-
-    QObject::connect(window.data(), SIGNAL(windowShown()), &app, SIGNAL(mainWindowShown()));
     }
   else if (app.commandOptions()->showPythonInteractor()
     && !app.commandOptions()->runPythonAndExit())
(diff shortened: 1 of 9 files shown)


lassoan:
There are a number of scripted modules that do various initialization steps after "Slicer is started". Currently, they mostly use the single shot timer method. Slicer app emitting a signal would be a nicer solution. However, this method should work when there is no main window. So, I'm wondering if we should have a different name, such as StartupCompleted. We could then document when this signal is emitted, e.g., after all modules are initialized and (if main window is requested) after main window is fully set up and shown.

lassoan:
For scripted module initialization, we should probably add a method applicationStarted() or applicationInitialized() or something similar. It would be easier to override that than observing a signal. Also, we could call these methods in an order that respects module dependency, while we cannot control order of signal processing.

jcfr:
@lassoan  Thanks for the review.

Followup commit introduces `qSlicerApplication::startupCompleted()` signal and add corresponding tests.

```
    Start 605: py_startupcompleted_signal_emitted_with_mainwindow_test
1/2 Test #605: py_startupcompleted_signal_emitted_with_mainwindow_test ......   Passed    1.77 sec
    Start 606: py_startupcompleted_signal_emitted_without_mainwindow_test
2/2 Test #606: py_startupcompleted_signal_emitted_without_mainwindow_test ...   Passed    1.36 sec
```


> For scripted module initialization, we should probably add a method applicationStarted() or 
> applicationInitialized() or something similar. It would be easier to override that than observing a signal. 

Good idea. I suggest we do that a latter time. I think the current topic address the issue while improving the existing API and also addressing the problem with Qt5.

> Also, we could call these methods in an order that respects module dependency, while we cannot 
> control order of signal processing.

Based on the Qt documentation, "If several slots are connected to one signal, the slots will be executed one after the other, in the order they have been connected, when the signal is emitted."

This means that slot connected during the module discovery will be executed in the expected order.

lassoan:
I agree that it is already as good as it was before.

Module initialization is done in random order (only module logic initialization is performed in the order based on module dependencies), so these startupCompleted() signals will be executed in "random" order, but it was like this before, so it will not introduce new problems.

jcfr:
Thanks for the review. Integrated in [r26279](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26279) to [r26280](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26280)

> Module initialization is done in random order

We should definitively fix this. I couldn't find an issue in mantis. Do you recall if one was already created ?

pieper:
👍

PR #779, opened 2017-08-21
PERF: don't update markups while scene is batch processing

pieper:
This avoids unacceptable delays when performing batch
operations on fiducial lists.

### DIFF
--- a/Modules/Loadable/Markups/MRMLDM/vtkMRMLMarkupsDisplayableManager2D.cxx
+++ b/Modules/Loadable/Markups/MRMLDM/vtkMRMLMarkupsDisplayableManager2D.cxx
@@ -310,6 +310,11 @@ void vtkMRMLMarkupsDisplayableManager2D
 ::ProcessMRMLNodesEvents(vtkObject *caller,unsigned long event,void *callData)
 {
 
+  if (this->GetMRMLScene()->IsBatchProcessing())
+    {
+    return;
+    }
+
   vtkMRMLMarkupsNode * markupsNode = vtkMRMLMarkupsNode::SafeDownCast(caller);
   vtkMRMLMarkupsDisplayNode *displayNode = vtkMRMLMarkupsDisplayNode::SafeDownCast(caller);
   vtkMRMLInteractionNode * interactionNode = vtkMRMLInteractionNode::SafeDownCast(caller);
--- a/Modules/Loadable/Markups/MRMLDM/vtkMRMLMarkupsDisplayableManager3D.cxx
+++ b/Modules/Loadable/Markups/MRMLDM/vtkMRMLMarkupsDisplayableManager3D.cxx
@@ -303,6 +303,11 @@ void vtkMRMLMarkupsDisplayableManager3D
 ::ProcessMRMLNodesEvents(vtkObject *caller,unsigned long event,void *callData)
 {
 
+  if (this->GetMRMLScene()->IsBatchProcessing())
+    {
+    return;
+    }
+
   vtkMRMLMarkupsNode * markupsNode = vtkMRMLMarkupsNode::SafeDownCast(caller);
   vtkMRMLMarkupsDisplayNode *displayNode = vtkMRMLMarkupsDisplayNode::SafeDownCast(caller);
   vtkMRMLInteractionNode * interactionNode = vtkMRMLInteractionNode::SafeDownCast(caller);


pieper:
This is a simple change and I don't see any side effects (except speedup).

I'd like to get this into the nightly as soon as possible because it's a blocker
for some of our research.  Any objections?

Modules/Loadable/Markups/MRMLDM/vtkMRMLMarkupsDisplayableManager2D.cxx
```diff
@@ -310,6 +310,11 @@ void vtkMRMLMarkupsDisplayableManager2D
 ::ProcessMRMLNodesEvents(vtkObject *caller,unsigned long event,void *callData)
 {
 
+  if (this->GetMRMLScene()->IsBatchProcessing())
+    {
+    return;
+    }
```
jcfr: For sake of consistency, should this also apply to the 3D displayable manager ?

jcfr: A greater speedup would be obtained then.

Do you have benchmark: Load time with/without this change ?

pieper: Sure, good idea - I can add that.  I was tempted to apply this across the board to all displayable managers but didn't want to make a sweeping change.

jcfr:
Nitpick: As rule of thump, in the commit message it is good to include the domain concerned with the update, and provide some numbers:

```
PERF: don't update markups while scene is batch processing
```
->
```
PERF: Markups: Skip node event processing while batch processing

This commit speeds up the loading of scene with N nodes by Y % (in that
particular case, load time was reduced from ?? s to ?? s
```

Modules/Loadable/Markups/MRMLDM/vtkMRMLMarkupsDisplayableManager2D.cxx
pieper: added and updated PR

lassoan:
1. It would be better to manage this with StartModify/EndModify.

* In general, batch processing is much more efficient with StartModify/EndModify. A lot of extra processing is done when you enter&exit batch processing mode (several logic classes refreshes all their cached actors, pipelines, etc; all models are re-rendered, ...).
* Batch processing affects all nodes of certain classes in the scene. Would you really need to update all the markup lists? Wouldn't it be better to temporarily block update of the affected markup list(s)?
* The change would make enter&exit batch processing a more heavyweight operation. So, for example when you show/hide a branch in the model tree then as a side-effect now it would also re-create all actors of all markup lists.

2. (If we choose to use batch processing to handle this) Is there a mechanism that refreshes all the fiducial lists are automatically when batch processing ends? (probably there is, it was just not obvious from reading the code; anyway, it should be checked to make sure there is a robust mechanism, not just some complex chain of events that might be accidentally changed in the future)

lassoan:
Stopping updates in batch mode could definitely improve performance, but wiping out of all cached actors in all displayable managers would have the side effect of making enter/exit batch processing much slower operation. It could be OK, we would just need to review that batch processing is activated appropriately and if needed convert some uses to StartModify/EndModify. For example, heavyweight Start/EndBatchProcessing is not a problem for scene loading, but may be too slow when showing/hiding model hierarchy trees.

pieper:
@lassoan I don't see this change having any impact on the things you mentioned.  Start/End batch processing is something that is used in situations where many nodes are known to be changing in transient ways and the displayable managers and other classes should ignore these changes.  Without this change there are many many useless vtk objects and widgets being created and destroyed and updated on every mrml event. 

This change doesn't make any changes as to when batch processing is or isn't used, only how the displayable manager behaves when the mode is active.

It also doesn't change anything about what the displayable managers do when batch processing ends.

pieper:
By the way @lassoan I agree we should be careful about introducing performance improvements that might have unintended consequences, which is why I made this PR rather than committing directly even though I'm pretty certain this has no side effects.  EndBatchProcessing is something all the code needs to handle correctly because by definition the scene may be completely different than it was before StartBatchProcessing.

lassoan:
I just wanted to make clear that there may be impact of this change.

Most of the displayable managers refresh their actors on start/end batch processing, so this change makes the behavior more consistent, which is good. The impact of the change is that it makes existing Start/End batch processing calls a longer operation - it depends on the number and size of markup lists if this is significant or not. Most often the overhead would be probably insignificant and if Start/End batch processing was used appropriately then even significant overhead should not cause problems.

1. Could you check if there are `StartState` BatchProcessing calls that may be used relatively frequently and would be negatively affected by the extra processing time?

2. Have you checked why there are many modifications to markup nodes that are not protected by Start/EndModify mechanism? This would be useful regardless of how displayable managers handle Start/End batch processing.

jcfr:
> Is there a mechanism that refreshes all the fiducial lists are automatically when batch processing ends?

Here is what I found ... 

```
/// Called everytime the scene has been significantly changed.
  /// If the scene BatchProcessState events are observed (in
  /// SetMRMLSceneInternal() ), UpdateFromMRMLScene is called after each
  /// batch process (Close, Import, Restore...). It is also being called by
  /// default when a new scene is set (SetMRMLScene).
  /// \sa SetMRMLSceneInternal, UnobserveMRMLScene, ObserveMRMLScene
  virtual void UpdateFromMRMLScene();
```

The function `UpdateFromMRMLScene()` is then called on the displayable manager side at the end of batch processing.

In case of Markups, it is implemented in `vtkMRMLMarkupsDisplayableManager2D::UpdateFromMRMLScene()` which end up calling [vtkMRMLMarkupsDisplayableManager2D::UpdateFromMRML](https://github.com/Slicer/Slicer/blob/468e2d0c6ca230084b12b2cab9c3479da32e45f3/Modules/Loadable/Markups/MRMLDM/vtkMRMLMarkupsDisplayableManager2D.cxx#L238-L285)

the function `UpdateFromMRML` then call [CreateWidget](https://github.com/Slicer/Slicer/blob/468e2d0c6ca230084b12b2cab9c3479da32e45f3/Modules/Loadable/Markups/MRMLDM/vtkMRMLMarkupsFiducialDisplayableManager2D.cxx#L261-L358) for each fiducial markup node.

jcfr:
And on the widget side, in addition of the MRMLcombo box managing the batch event, the function `qSlicerMarkupsModuleWidget::onMRMLSceneEndBatchProcessEvent` is called, which ends up calling `qSlicerMarkupsModuleWidget::updateWidgetFromMRML`

pieper:
I just found that there are other bugs with the markups and batch processing mode so I'm going to close this pull request.

It turns out that in this branch in some cases deleting a fiducial in BatchProcessing mode leaves a seed widget the view.  So there is something incorrect in the way EndBatchProcess is handled in the markups.  I'll see if I can narrow down that issue with a small test case but in the mean time optimize the bulk processing script a different way.

pieper:
Also thank you for the quick and thorough review!

lassoan:
It may help if you put bulk modifications between markupnode->StartModify/EndModify calls.

pieper:
As another data point on this topic, on the current master the lines below work correctly, with the widget being removed from the sliceViews at the end of the batch process.  With the change in this PR branch there is a leftover seed widget in the slice views that is not connected to the scene.


```
markupsLogic = slicer.modules.markups.logic()
fiducialListNodeID = markupsLogic.AddNewFiducialNode('testNode',slicer.mrmlScene)
list = getNode(fiducialListNodeID)
list.AddFiducial(0,0,0)
slicer.mrmlScene.StartState(slicer.mrmlScene.BatchProcessState)
list.RemoveMarkup(0)
slicer.mrmlScene.EndState(slicer.mrmlScene.BatchProcessState)
```

PR #780, opened 2017-08-24
COMP: Fix extension packaging setting CMAKE_MACOSX_RPATH to 0  **WIP**

jcfr:
## DO NOT INTEGRATE - CTK pull request need to be integrated 

Following the update of the minimum required version of CMake to 3.5,
the policy CMP0042 started to default to NEW leading to CMAKE_MACOSX_RPATH
being initialized to 1 by default.

This had the side-effect of including "@rpath/..." into libraries
installed during the packaing on macOS and prior the fixup script had
a chance to recursively collect, copy and update (aka fixup) the libraries.

This commit also updates the external project removing patches explicitly
setting CMP0042 to OLD. Instead, the variable CMAKE_MACOSX_RPATH is passed
to all of them.

Related commits:
* 25180 (BUG: Extension packaging: Ensure @rpath is not applied at build time)


pieper:
Whew - that must have been fun to track down!  👍

fedorov:
🎆 👍

jcfr:
Integrated in [r26301](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26301)

PR #782, opened 2017-08-26
WIP: check for uninitialized DICOM db location

fedorov:
DatabaseDirectory setting is not available on first start of Slicer. As
described in https://issues.slicer.org/view.php?id=4424, this can lead to the
default location be non-writeable, without user being notified. This commit
aims to fix this problem, but not fully resolves the issue. Current runtime
error:

```
File
"/Applications/Slicer.app/Contents/lib/Slicer-4.7/qt-scripted-modules/DICOM.py",
line 333, in setup
      slicer.dicomDatabase.connect('databaseChanged()', self.onDatabaseChanged)
      AttributeError: 'NoneType' object has no attribute 'connect'
```

I would use help from someone more familiar with this code!

### DIFF
--- a/Modules/Scripted/DICOMLib/DICOMWidgets.py
+++ b/Modules/Scripted/DICOMLib/DICOMWidgets.py
@@ -79,14 +79,14 @@ class DICOMDetailsBase(VTKObservationMixin, SizePositionSettingsMixin):
     VTKObservationMixin.__init__(self)
     self.settings = qt.QSettings()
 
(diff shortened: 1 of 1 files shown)


lassoan:
Thank you. I've fixed the remaining error and committed the change:
https://github.com/Slicer/Slicer/commit/5f013eefb1698aeb0c2c60db60cc0018620fd694

@fedorov It would be great if you could test it.

fedorov:
@lassoan tested, no errors - thank you!

PR #783, opened 2017-08-26
ENH: Overhaul of Segment Editor button row

cpinter:
Renamed Create surface button to Show 3D, and added convenience action to set smoothing factor

Added Properties button that switches to Segmentations module. Added convenience actions to export to labelmap and model hierarchy

### DIFF
--- a/Modules/Loadable/Segmentations/Resources/UI/qSlicerSegmentationsModule.ui
+++ b/Modules/Loadable/Segmentations/Resources/UI/qSlicerSegmentationsModule.ui
@@ -83,7 +83,7 @@
            <string>Add empty segment</string>
           </property>
           <property name="text">
-           <string>Add segment</string>
+           <string>Add</string>
           </property>
           <property name="icon">
            <iconset resource="../../Widgets/Resources/qSlicerSegmentationsModuleWidgets.qrc">
@@ -94,10 +94,10 @@
         <item>
          <widget class="QPushButton" name="pushButton_RemoveSelected">
           <property name="toolTip">
-           <string>Remove selected segment</string>
+           <string>Remove selected segment(s)</string>
           </property>
           <property name="text">
-           <string>Remove selected</string>
+           <string>Remove</string>
           </property>
           <property name="icon">
            <iconset resource="../../Widgets/Resources/qSlicerSegmentationsModuleWidgets.qrc">
@@ -107,8 +107,11 @@
         </item>
         <item>
          <widget class="QPushButton" name="pushButton_EditSelected">
+          <property name="toolTip">
+           <string>Edit segments in the Segment Editor module</string>
+          </property>
           <property name="text">
-           <string>Edit selected</string>
+           <string>Segment Editor</string>
           </property>
           <property name="icon">
            <iconset resource="../../Widgets/Resources/qSlicerSegmentationsModuleWidgets.qrc">
--- a/Modules/Loadable/Segmentations/Widgets/Resources/UI/qMRMLSegmentEditorWidget.ui
+++ b/Modules/Loadable/Segmentations/Widgets/Resources/UI/qMRMLSegmentEditorWidget.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>758</width>
-    <height>873</height>
+    <width>406</width>
+    <height>426</height>
    </rect>
   </property>
   <property name="windowTitle">
@@ -89,8 +89,11 @@
          <height>0</height>
         </size>
        </property>
+       <property name="toolTip">
+        <string>Add new empty segment</string>
+       </property>
        <property name="text">
-        <string>Add segment</string>
+        <string>Add</string>
        </property>
        <property name="icon">
         <iconset resource="../qSlicerSegmentationsModuleWidgets.qrc">
@@ -102,12 +105,15 @@
       <widget class="QPushButton" name="RemoveSegmentButton">
        <property name="minimumSize">
         <size>
-         <width>116</width>
+         <width>96</width>
          <height>0</height>
         </size>
        </property>
+       <property name="toolTip">
+        <string>Remove selected segment(s)</string>
+       </property>
        <property name="text">
-        <string>Remove selected</string>
+        <string>Remove</string>
        </property>
        <property name="icon">
         <iconset resource="../qSlicerSegmentationsModuleWidgets.qrc">
@@ -116,15 +122,15 @@
       </widget>
      </item>
      <item>
-      <widget class="QPushButton" name="CreateSurfaceButton">
+      <widget class="ctkMenuButton" name="Show3DButton">
        <property name="minimumSize">
         <size>
-         <width>104</width>
+         <width>96</width>
          <height>0</height>
         </size>
        </property>
        <property name="text">
-        <string>Create surface</string>
+        <string> Show 3D</string>
        </property>
        <property name="icon">
         <iconset resource="../qSlicerSegmentationsModuleWidgets.qrc">
@@ -132,6 +138,22 @@
        </property>
        <property name="checkable">
         <bool>true</bool>
+       </property>
+      </widget>
+     </item>
+     <item>
+      <widget class="ctkMenuButton" name="SwitchToSegmentationsButton">
+       <property name="minimumSize">
+        <size>
+         <width>104</width>
+         <height>0</height>
+        </size>
+       </property>
+       <property name="toolTip">
+        <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Edit segmentation properties in Segmentations module:&lt;br/&gt;- Import from volume/models (e.g. from nrrd)&lt;br/&gt;- Export to volume/models (e.g. save to STL)&lt;br/&gt;- Advanced visibility options&lt;br/&gt;- Conversion: representations, converter options (e.g. smoothing)&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
+       </property>
+       <property name="text">
+        <string> Segmentations...</string>
        </property>
       </widget>
      </item>
@@ -444,6 +466,11 @@
    <header>ctkFittedTextBrowser.h</header>
   </customwidget>
   <customwidget>
+   <class>ctkMenuButton</class>
+   <extends>QPushButton</extends>
+   <header>ctkMenuButton.h</header>
+  </customwidget>
+  <customwidget>
    <class>ctkRangeWidget</class>
    <extends>QWidget</extends>
    <header>ctkRangeWidget.h</header>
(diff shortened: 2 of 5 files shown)


cpinter:
image

image

image

fedorov:
Looks great!

Did you test with QuantitativeReporting?

Is it possible to hide the Properties button programmatically? I think we could confuse user with it in the context of the QuantitativeReporting workflow.

fedorov:
@che85 is away for a few weeks, so I would at least like to have a complete set of Slicer binaries with working QuantitativeReporting for MICCAI before integrating this (we did quite a few fixes today), since I won't have Christian to help me debug any issues that could potentially come up.

cpinter:
I only have access to a Mac in the lab and I'm only going there on Monday. If you have access to a Windows (any Windows!) then I can send the packages to you today. Please let me know.

(He chose a strange time to leave...)

cpinter:
> Is it possible to hide the Properties button programmatically?

It is now!

fedorov:
> (He chose a strange time to leave...)

He is going to an event with fixed dates, he planned this vacation well ahead, and it is the first time we do this tutorial at MICCAI, so he didn't even think I could need him in this time.

I only have access to Windows at work. Let me confirm next week I have working setup on all 3 systems, then I can finalize Slicer packages, and test your branch.

cpinter:
@fedorov I prepared Windows release packages, hopefully with all the extensions you need. Please find them here
https://www.dropbox.com/sh/a45d0xu3yvlk5zb/AADxlURP-1dX6d933tzXJYNda?dl=0

fedorov:
If I just install the extensions you included in the package, I get this error on startup, which I don't get in the nightly:

```
  File "<string>", line 1, in <module>
  File "C:/Users/fedorov/AppData/Roaming/NA-MIC/Extensions-26295/QuantitativeReporting/lib/Slicer-4.7/qt-scripted-modules/QuantitativeReporting.py", line 16, in <module>
    from SlicerDevelopmentToolboxUtils.widgets import DICOMBasedInformationWatchBox, ImportLabelMapIntoSegmentationWidget
ImportError: cannot import name ImportLabelMapIntoSegmentationWidget
```

Looks like there is some version mismatch in the extensions?

cpinter:
I updated all of them to latest. I'll check if some of them use fixed hashes.

fedorov:
I think it is more likely that some of them changed after you checked out. Did you do it Friday or Saturday?

I tried pointing to the latest version of SlicerDevelopmentToolkit, but it didn't help.

cpinter:
This morning

fedorov:
Then I am really puzzled. Today's nightly just works.

cpinter:
Can you use the features to test with my changes? I'll keep looking nevertheless.

cpinter:
Try this
https://www.dropbox.com/s/du0lrukkqmob6ax/26311-win-amd64-SlicerDevelopmentToolbox-git13f6af9-2017-08-25.zip?dl=0

fedorov:
Something is still not right, since PET DICOM plugin does not show up.

The features look good to me. The only thing that I find confusing is the label of the "Properties" button. I would rename it to "Segmentations". I know that can also be confusing, but at least it is consistent.

I would like to get a set of working extensions and the Slicer binaries for the 3 platforms for MICCAI, before merging this, to be on the safe side.

cpinter:
OK, no rush.

We were thinking with @lassoan about the label of the button and we decided for Properties, because the users are often not aware of the exact names of the modules. They in this case sound similar too, and the difference between them is that Segmentations manages the node properties, and Segment Editor edits the contents. As other than import/export - which is explained in the tooltip and are exposed too as options - Segmentations allow changing properties such as segment metadata, display, representation configuration, this name choice sounded more informative.

lassoan:
I don't feel fully comfortable with either "Segmentations" or "Properties" as button name. Can we think some more about what button name would work? Should we try and ask on the Slicer Forum?

fedorov:
> I don't feel fully comfortable with either "Segmentations" or "Properties" as button name. 

Indeed, me too!

I still think that "Segmentations" is better, since it is a shortcut to another module. It does not expose a panel or new functionality within the present module. Maybe we could add some icon representing shortcut in the button, before "Segmentations", and establish some convention across the Slicer application for introducing shortcuts to other modules?

cpinter:
> Maybe we could add some icon representing shortcut in the button, before "Segmentations"

The icon in the Properties button is the icon of the Segmentations module.

I agree that neither is much better than the other, and I'm OK with Segmentations too.

@lassoan I think the wide community won't be able to help as much.

@pieper @jcfr What do you think is a good label for a shortcut button from Segment Editor to Segmentations?

fedorov:
> The icon in the Properties button is the icon of the Segmentations module.

Yes, I noticed that. However, I am not sure how well-known that icon is (it didn't ring the bell to me), and what I was suggesting is to explicitly highlight the fact that the button is a shortcut. It is not uncommon for buttons/modules in Slicer to use duplicate icons, so I am not sure how much attention users pay to them, especially for those that do not stand out.

fedorov:
Also, to be clear, I am OK either way. I just thought I mention this observation.

pieper:
I agree, neither name really describes what that widget is used for.

How about calling it "More..." and putting it at the right side (swap it with Show 3D) and make it just a menu.  To start it can have the two operations and an option "Go to Segmentations module".  This way you could add other handy operations here in the future.

lassoan:
> putting it at the right side (swap it with Show 3D)
👍 +1

> make it just a menu
Require 2 clicks to switch to Segmentations would be a bit less convenient, so I think it would be better to keep the button with the menu.
Icon could be the arrow icon that is used in module switcher toolbar.

image

I'm not sure if it is good idea to put "Export to label map" and "Export to model hierarchy" shortcuts there, as:
* they are needed only once, after completing the segmentation
* already available only one click away (in Segmentations)
* there is no way to specify advanced parameters (reference volume, etc)
* after finishing a segmentation and there is no shortcut to undo the operation (hide/delete exported data), therefore, people would click on it and not understand why they cannot modify the surface anymore (because it's a model) - maybe if the exported labelmap/models are not shown then this is not a big problem

cpinter:
I have no strong preference for the name/location/icon of the buttons, so any solution we agree on is fine with me. I'd just like to have a consensus so that I can implement it and integrate it (when @fedorov is ready for it).

Compared to the current PR, I propose to do the following changes:
* Move the shortcut button to the right, rename it Segmentations, and change the icon to the one Andras suggested
* Change the two menu actions under the Segmentations button to one, called Import/Export, which would - in addition to switching to Segmentations - open up that section and scroll down to it. I agree with Andras that it may be confusing (especially that the result of the export is not immediately visible, and it doesn't make sense to do it mid-segmentation), but I think it's useful to have an action saying "export", because it's more explicit, and as many users ask about it nowadays, maybe this will make it easier to discover

cpinter:
That said we can add other similar actions to Segmentations, such as "Advanced visibility", etc.

Also I want to say that even if we don't have a clear consensus now (none of the above options are clearly better than the others), time and user feedback will decide which is the best, and I can change it easily any time.
However it would be nice to have these features in rather sooner than later. Smoothing option is called for, and the shortcut to Segmentations is also very useful.

jcfr:
>  time and user feedback will decide which is the best, and I can change it easily any time.

Agreed :+1:  Perfect is the enemy of the good.

Modules/Loadable/Segmentations/Resources/UI/qSlicerSegmentationsModule.ui
```diff
@@ -119,7 +122,7 @@
        </layout>
       </item>
       <item>
-       <widget class="qMRMLSegmentsTableView" name="SegmentsTableView">
+       <widget class="qMRMLSegmentsTableView" name="SegmentsTableView" native="true">
```
jcfr: Nitpick: Could you consider excluding this type of change.

cpinter: I don't even know what it means, apparently Qt Designer added it at some point. I'll change it back, thanks!

cpinter: What about
156: <widget class="qMRMLSegmentationDisplayNodeWidget" name="SegmentationDisplayNodeWidget" native="true"/>
Remove the native attribute here as well?

jcfr: this one too

jcfr: Thanks

jcfr: At some point, we will do a "one shot" pass on all ui files to update them.

fedorov:
> when @fedorov is ready for it

I was out of office yesterday, and only have access to Windows/Linux in office, so I could not test last night binary. I will finalize today.

fedorov:
Hey guys - you can proceed with this PR as you see appropriate. I am having technical difficulties unrelated to Slicer, but from the platforms I could test so far, I have a working set of Slicer binaries + extensions. I don't want to hold your work without reason. Thank you for your patience!

cpinter:
Excellent, thanks @fedorov! I hope those unrelated difficulties will not be hard to resolve!

Unless there are objections, I'd go forward with the plan I outlined above. I know that jumping to Export section thing is new, so if it's not to your liking, please let me know.

jcfr:
I am building locally, will report back in few mins

jcfr:
@cpinter I update the topic tweaking the UI file.

Otherwise, after playing with the functionality, here is a suggested improvement:
* go to "Segment Editor"
* Add a segment
* paint
* Click on "Properties (it goes to "Segmentations" module)
* -> It would be nice to select the same segment that was selected in the "Segment Editor"

This can definitively be done at some later time

jcfr:
This was observed in Qt5 build and the following message is reported in the console:

```
void qSlicerSegmentationsModuleWidget::onEditSelectedSegment() : MRMLNodeComboBox_Segmentation is not found in Segment Editor module

```

jcfr:
@cpinter 154ba35 addressed my last comment about the error message (feel free to squash before integration)

cpinter:
I amended the PR according to plan above. I'll integrate tomorrow if there are no objections.

jcfr:
@cpinter Well done.

Could you also squash in this commit: https://github.com/jcfr/Slicer/commit/c81ec5f9953aafe01cb444dfcf7d8ab42424f53c

cpinter:
@jcfr Done (I think)

jcfr:
:+1:  Thanks

cpinter:
Integrated in https://github.com/Slicer/Slicer/commit/36fe64b16734c962156c5f574280a6fa7b9cccb6

PR #784, opened 2017-08-28
ENH: Build system: Update FindPatch module

jcfr:
This commit updates the module to match latest CMake version.
See https://gitlab.kitware.com/cmake/cmake/merge_requests/1184

### DIFF
--- a/CMake/FindPatch.cmake
+++ b/CMake/FindPatch.cmake
@@ -7,10 +7,15 @@
 #
 # The module defines the following variables:
 #
-# ``PATCH_EXECUTABLE``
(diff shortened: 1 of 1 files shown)


jcfr:
@lassoan This version of `FindPatch.cmake` is more robust and thanks to the use of [PATH_SUFFIXES](https://cmake.org/cmake/help/v3.5/command/find_program.html) it should address your initial concerns.

jcfr:
Integrated in [r26327](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26327)

PR #785, opened 2017-08-30
ENH: Extend scripted module API adding support for cleanup() method

jcfr:
This allows to do some cleanup before a module is effectively unloaded. It
is particularly useful to update objects or widgets and avoid memory leaks
due to complex interdependencies between C++ application and python layer.

### DIFF
--- a/Applications/SlicerApp/Testing/Python/CMakeLists.txt
+++ b/Applications/SlicerApp/Testing/Python/CMakeLists.txt
@@ -178,10 +178,14 @@ endif()
 # Check if scripted module import works as expected
 #
 
+set(testname ScriptedModuleDiscoveryTest)
 slicer_add_python_test(
-  SCRIPT ScriptedModuleDiscoveryTest.py
+  SCRIPT ${testname}.py
   SLICER_ARGS --no-main-window --disable-builtin-modules --additional-module-path ${CMAKE_CURRENT_SOURCE_DIR}/ScriptedModuleDiscoveryTest
   TESTNAME_PREFIX nomainwindow_
+  )
+set_tests_properties(py_nomainwindow_${testname} PROPERTIES
+  PASS_REGULAR_EXPRESSION "ModuleAWidget finalized"
   )
 
 #
--- a/Applications/SlicerApp/Testing/Python/ScriptedModuleDiscoveryTest/ModuleA.py
+++ b/Applications/SlicerApp/Testing/Python/ScriptedModuleDiscoveryTest/ModuleA.py
@@ -20,5 +20,9 @@ class ModuleA(ScriptedLoadableModule):
     return SOMEVAR
 
 class ModuleAWidget(ScriptedLoadableModuleWidget):
+  def __init__(self, parent=None):
+    ScriptedLoadableModuleWidget.__init__(self, parent)
   def setup(self):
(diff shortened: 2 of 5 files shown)


jcfr:
Cc: @lassoan @pieper

pieper:
LGTM, but is there an example use case where this is needed in the core?

jcfr:
I just realized there was a "cleanup()" function that was added to support reloading. See [here](https://github.com/Slicer/Slicer/blob/ee5c8703642cf8f42e9594bc780cb0801623b39d/Base/Python/slicer/util.py#L532-L536)

I will unify both by renaming "finalize" into "cleanup"

> an example use case where this is needed in the core?

Not in the core but for a scripted extension having memory leaks. In that case, it is `SlicerCIP`.

It can also be useful to simplify the code that used be in the core and is using [saferef](https://github.com/Slicer/Slicer/blob/master/Base/Python/saferef.py) module to address the circular dependency issue. See [here](https://github.com/SlicerDMRI/SlicerDMRI/blob/a780e4b939ddfcbbf7da3567733a79387805ecc9/Modules/Scripted/DICOM2FullBrainTractography/DICOM2FullBrainTractographyLib/workflow_support.py#L134-L141)

fedorov:
Can this PR wait until tomorrow, unless it is really urgent? I would like to get a working set of binaries for the MICCAI tutorial.

jcfr:
It can definitively wait and it needs some more work

jcfr:
@pieper Now that I consolidated with the existing `cleanup` function. Do you have any additional feedback ?

pieper:
As far as I can see it looks good 👍

jcfr:
Thanks @pieper

jcfr:
Integrated in [r26328](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26328)

Applications/SlicerApp/Testing/Python/CMakeLists.txt
```diff
@@ -178,11 +178,15 @@ endif()
...
 
+set(testname ScriptedModuleDiscoveryTest)
 slicer_add_python_test(
-  SCRIPT ScriptedModuleDiscoveryTest.py
+  SCRIPT ${testname}.py
   SLICER_ARGS --no-main-window --disable-builtin-modules --additional-module-path ${CMAKE_CURRENT_SOURCE_DIR}/ScriptedModuleDiscoveryTest
   TESTNAME_PREFIX nomainwindow_
   )
+set_tests_properties(py_nomainwindow_${testname} PROPERTIES
+  PASS_REGULAR_EXPRESSION "ModuleAWidget finalized"
```
lassoan: This test will fail on Windows with console disabled. Instead of using stdout to communicate with the tester process, use output files instead, the same way as it is done for slicerrc and startupCompleted event testing.

jcfr: Good catch. I will fix it before the nightly kick in

>  create some common utilities to make it easier to implement module tests that write output files.

Indeed. I will look into that

jcfr: Waiting I implement a proper fix. I am excluding the text on windows

PR #787, opened 2017-09-05
ENH: Extend application API adding environment()/startupEnvironment()

jcfr:
With the upcoming feature, it will be possible to easily invoke processes by excluding the Slicer environment.:
* from c++ (using `app->startupEnvironment()`), or;
* from python (using `slicer.util.startupEnvironment()` 

Here are two examples:

Without using the `startup` environment, this first example fails (as expected):

```
>>> from subprocess import check_output
>>> check_call(["/usr/bin/python3", "-c", "print('hola')"])
  File "/home/jcfr/Projects/Slicer-2-build/python-install/lib/python2.7/subprocess.py", line 186, in check_call
    raise CalledProcessError(retcode, cmd)
CalledProcessError: Command '['/usr/bin/python3', '-c', "print('hola')"]' returned non-zero exit status -6
```

Now, using the sanitized environment, this one succeeds:

```
>>> from subprocess import check_output
>>> check_output(
  ["/usr/bin/python3", "-c", "print('hola')"], 
  env=slicer.util.startupEnvironment())
'hola\n'
```

### DIFF
--- a/Applications/SlicerApp/Testing/Python/CMakeLists.txt
+++ b/Applications/SlicerApp/Testing/Python/CMakeLists.txt
@@ -274,6 +274,12 @@ slicer_add_python_unittest(
   TESTNAME_PREFIX nomainwindow_
   )
 
+slicer_add_python_unittest(
+  SCRIPT ${Slicer_SOURCE_DIR}/Base/Python/slicer/tests/test_slicer_environment.py
+  SLICER_ARGS --no-main-window --disable-modules
+  TESTNAME_PREFIX nomainwindow_
+  )
+
 # DCMTKPrivateDictionary test
 slicer_add_python_test(
   SCRIPT DCMTKPrivateDictTest.py
--- a/Base/Python/slicer/tests/test_slicer_environment.py
+++ b/Base/Python/slicer/tests/test_slicer_environment.py
@@ -1,0 +1,25 @@
+
+import qt
+import slicer
+import unittest
+
+
+class SlicerEnvironmentTests(unittest.TestCase):
+
+  def setUp(self):
+    pass
+
+  def test_slicer_util_startupEnvironment(self):
+    startupEnv = slicer.util.startupEnvironment()
+    assert type(startupEnv) is dict
+    assert "Slicer-build" not in startupEnv["PATH"]
+
+  def test_slicer_app_startupEnvironment(self):
+    startupEnv = slicer.app.startupEnvironment()
+    assert type(startupEnv) is qt.QProcessEnvironment
+    assert "Slicer-build" not in startupEnv.value("PATH")
+
+  def test_slicer_app_environment(self):
+    env = slicer.app.environment()
+    assert type(env) is qt.QProcessEnvironment
+    assert "Slicer-build" in env.value("PATH")
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -49,6 +49,24 @@ def sourceDir():
   """
   return _readCMakeCache('Slicer_SOURCE_DIR')
 
+def startupEnvironment():
+  """Returns the environment without the Slicer specific values.
+
+  Path environment variables like `PATH`, `LD_LIBRARY_PATH` or `PYTHONPATH`
+  will not contain values found in the launcher settings.
+
+  Similarly `key=value` environment variables also found in the launcher
+  settings are excluded. Note that if a value was associated with a key prior
+  starting Slicer, it will not be set in the environment returned by this
+  function.
+
+  The function excludes both the Slicer launcher settings and the revision
+  specific launcher settings.
+  """
+  import slicer
+  startupEnv = slicer.app.startupEnvironment()
+  return {varname: startupEnv.value(varname) for varname in startupEnv.keys()}
+
 #
 # Custom Import
 #
--- a/Base/QTCore/qSlicerCoreApplication.cxx
+++ b/Base/QTCore/qSlicerCoreApplication.cxx
@@ -663,6 +663,50 @@ bool qSlicerCoreApplication::testAttribute(qSlicerCoreApplication::ApplicationAt
 }
 
 //-----------------------------------------------------------------------------
+QProcessEnvironment qSlicerCoreApplication::startupEnvironment() const
+{
+  Q_D(const qSlicerCoreApplication);
+
+  ctkAppLauncherSettings appLauncherSettings;
+  appLauncherSettings.setLauncherName(this->applicationName().replace("-tmp", ""));
+  appLauncherSettings.setLauncherDir(d->SlicerHome);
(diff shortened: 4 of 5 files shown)


lassoan:
It looks OK, but it is not very clean and it may fail in unexpected ways.
What do you think about saving the original values of the modified/added/removed environment variables in other environment variables?
For example:

```
PATH=...new path with additional values...
PYTHONPATH=...modified by this applauncher...
LD_LIBRARY_PATH=...modified by this applauncher...
APPLAUNCHER_LEVEL=0
APPLAUNCHER_0_PATH=...original path before the applauncher modified it...
APPLAUNCHER_0_SOME_DELETED_VAR=...stores value of a deleted env var SOMETHING_DELETED...
APPLAUNCHER_0_SOME_ADDED_VAR=  #...when a new env var is added then the previous value would be stored as empty
```

To restore previous level's environment, you would just need to go through all the APPLAUNCHER_<LEVEL>_* variables and set them. Set to empty value deletes an env var, so this can be used for deleting added env vars.

APPLAUNCHER_LEVEL can be used for cleanly managing multiple levels off app launching (applauncher library running an application that uses applauncher). For example:

```
PATH=...new path with additional values...
PYTHONPATH=...modified by this applauncher...
LD_LIBRARY_PATH=...modified by this applauncher...
APPLAUNCHER_LEVEL=1
APPLAUNCHER_0_PATH=...original path before the applauncher modified it...
APPLAUNCHER_0_SOME_DELETED_VAR=...stores value of a deleted env var SOMETHING_DELETED...
APPLAUNCHER_0_SOME_ADDED_VAR=  #...when a new env var is added then the previous value would 
APPLAUNCHER_1_PATH=...original path before the applauncher modified it...
APPLAUNCHER_1_SOME_DELETED_VAR=...stores value of a deleted env var SOMETHING_DELETED...
APPLAUNCHER_1_SOME_MORE_DELETED_VAR=...stores value of a deleted env var SOMETHING_DELETED...
APPLAUNCHER_1_SOME_ADDED_VAR=
APPLAUNCHER_1_SOME_MORE_ADDED_VAR=
```

If a launcher did not modify any env vars then it would just increment APPLAUNCHER_LEVEL.

AppLauncher library could be responsible for creating environments from previous levels.

jcfr:
Thanks for the review.

That said, I like the simplicity of the current approach.  

Since modifying the environment from a child process does not affect the parent process, I am not sure it makes sense to keep track of modified/added/removed env vars.

Since environment can be modified using `setenv/putenv` in c++, `os.environ` in python and also the `qSlicerCoreApplication::setEnvironmentVariable` which takes care keeping `os.environ` in sync, I am not sure we can have  a systematic way of keeping track of modified/added/removed env variables.

> not very clean and it may fail in unexpected ways.

Are you thinking about something specific ?

jcfr:
Since we agree on the public facing API, I will go ahead and integrate. We can definitively revisit later on.

jcfr:
Integrated in [r26351](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26351)

lassoan:
>> not very clean and it may fail in unexpected ways.
>
>Are you thinking about something specific ?

Yes, there are many potential issues. Modified variables are deleted instead of reverted, paths that were already in the list are removed, etc. - all these can very easily break things and developers would have a hard time figuring out where and how certain values and/or variables disappear.

What I described guarantees that exactly the caller's environment variables are preserved, no heuristics needed, no information is lost. Also, all the saving and restoring of env vars is implemented in one place, the applauncher library, which makes it easy to keep env modification/restore in sync (while with the current design, in the application we try to undo some of the changes that the applauncher performed, so when applauncher is modified then the application has to be updated as well). Implementation in applauncher also make this feature available in all applications that uses applauncher, not just in Slicer.

I agree that having this new method in the Slicer application is already a step forward, because all modules can use it so we prevent proliferation of various hacks. When the current solution breaks or some refactoring is done in applauncher then we can implement this feature properly there, and replace the current method by a call to the applauncher library.

lassoan:
> I am not sure we can have a systematic way of keeping track of modified/added/removed env variables.

What changes happen to environment variables at the application level, we should be able to control (revert) in the application. For example, we could require developers to always use qSlicerCoreApplication::setEnvironmentVariable, and this method could keep track of all changes so that when we request a clean env then those changes would be reverted.

jcfr:
>Modified variables are deleted instead of reverted, paths that were already in the list are removed, 

To clarify, a copy of Environment object is created:

https://github.com/Slicer/Slicer/blob/2b46072d9aa21d3a979285718d3f2cb1b686b108/Base/QTCore/qSlicerCoreApplication.cxx#L678

And since (1) the copied object is a [QProcessEnvironment](http://doc.qt.io/qt-4.8/qprocessenvironment.html) and (2) modifing such object does **NOT** update the current environment, I think the issue you described is caduc.

Indeed, we can read from the documentation

> The QProcessEnvironment class wraps that concept and allows easy manipulation of those variables. It's meant to be used along with QProcess, to set the environment for child processes. It cannot be used to change the current process's environment.

lassoan:
Of course, Slicer would not be impacted. Simply the launched processes could fail in various ways.

jcfr:
To conclude, there are few limitations of the current approaches:

* As outlined in the [doxygen documentation](https://github.com/Slicer/Slicer/blob/2b46072d9aa21d3a979285718d3f2cb1b686b108/Base/QTCore/qSlicerCoreApplication.h#L119-L121): 

> if a value was associated with a key prior starting Slicer, it will not be set in the environment returned by this function.

If that turns to be a issue in practice, it would indeed make sense to address the problem at the launcher library level.

* It is not possible for CLIs to get the startup environment and start a process with the Slicer env removed.

Beside of these two issues, I think what we have is fairly reasonable.

jcfr:
> Simply the launched processes could fail in various ways.

This is the part I do not understand. 

* Variables of the form `FOO=bar` that are explicitly set by Slicer are not available to the launched process.

* Similarly, variables of the type *path environment variables* (e.g PYTHONPATH, PATHS, ...) are stripped from any path explicitly added by the Slicer launcher.

We could also discuss further during the developer hangout on Sept 19th (I will be travelling next week)

lassoan:
Variables that the applauncher modified might have contained useful values. Removing them is not the same as restoring their original values.

If applauncher added a directory to a PATHS variable that was already there, then it will be removed instead of keeping it there. You may get failures eve if you just change the order of directories in the path.

We could further complicate the heuristics to try to patch these holes, but it would be much simpler and more robust to just store previous values of modified env vars and restore them when needed. But I agree that it is not urgent, we can implement when needed.

jcfr:
Totally agree with your last statements. Given the resources at hand, I purposely implemented a simpler solution your suggestion of using `APPLAUNCHER_LEVEL` makes sense.

This is tracked here: https://github.com/commontk/AppLauncher/issues/80

PR #788, opened 2017-09-06
STYLE: prevent __version__.py being marked as modified on Windows

dzenanz:
Building Slicer regenerates `__version__.py`, and if the file was checked out with autocrlf it will be marked as modified. This creates an annoyance for many versioning-related operations, such as committing, re-basing, cherry-picking etc.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # Ignore all dotfiles...
 .*
+!.gitattributes
 
 # Ignore all back-up files...
 *~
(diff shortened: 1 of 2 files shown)


dzenanz:
@jcfr I think I have the solution

jcfr:
Thanks @dzenanz .  Makes sense

https://github.com/Slicer/Slicer/blob/32cedd392701a9fbfd7cc533b1d4404567608c19/Utilities/Scripts/SlicerWizard/CMakeLists.txt#L7-L16

jcfr:
Waiting sysadmin re-enable the access to dot file on the svn server, I tweaked the commit message.

We will commit as soon possible.

jcfr:
Integrated in [r26354](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26354)

PR #789, opened 2017-09-07
ENH: Update EMSegment removing register

hjmjohnson:
The use of the register key word is deprecated:
Slicer-bld/EMSegment/AMF/vtkMinHeap.h:235:5: warning: 'register' storage class specifier is deprecated and incompatible with C++1z
      [-Wdeprecated-register]
    register int  down;
    ^~~~~~~~~


jcfr:
Thanks @hjmjohnson  :+1: 

Integrated in [r26356](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26356)

PR #791, opened 2017-09-08
BUG: Fixed incorrect value ending up in composite view's volume node IDs

cpinter:
By not checking the compared object by contained values, the string 'keep-current' could end up in the volume node IDs (background, foreground, label) stored in vtkMRMLSliceCompositeNode, making it hard to check if a valid volume node in shown in a slice view.

For some reason the string 'NULL' can also end up in these char* members that should only contain valid node ID strings or null pointer, hence the fix in qSlicerSubjectHierarchyVolumesPlugin. The source of the string 'NULL' is still unknown.


cpinter:
Unfortunately, no (I included this in the commit message as well). I spent a few minutes trying to track it down, but I couldn't find anything decisive. My suspicion would be an incorrectly implemented ReadXMLAttributes

cpinter:
Should I investigate the origin of NULL, or is it good to go?

lassoan:
I don't think it's important to know where NULL comes from. I'll merge this now.

lassoan:
Committed r26369

cpinter:
Thanks!

PR #792, opened 2017-09-18
ENH: Import vtkSegmentationCore to python by default. WIP

cpinter:
Attempt to implicitly import vtkSegmentationCorePython, based on what was in the code for vtkAddon. Unfortunately it does not work, so I'd appreciate any help. Thanks!

### DIFF
--- a/Base/Python/CMakeLists.txt
+++ b/Base/Python/CMakeLists.txt
@@ -12,6 +12,7 @@ set(Slicer_PYTHON_SCRIPTS
   teem
   vtkAddon
   vtkITK
+  vtkSegmentationCore
   )
 
 set(Slicer_PYTHON_MODULES_CONFIG "
--- a/Base/Python/slicer/slicerqt-with-tcl.py
+++ b/Base/Python/slicer/slicerqt-with-tcl.py
@@ -12,7 +12,7 @@ def tcl(cmd):
     import tpycl
     _tpycl = tpycl.tpycl()
 
-    packages = ['freesurfer', 'mrml', 'teem', 'vtk', 'vtkAddon', 'vtkITK']
+    packages = ['freesurfer', 'mrml', 'teem', 'vtk', 'vtkAddon', 'vtkITK', 'vtkSegmentationCore']
     for p in packages:
       _tpycl.py_package(p)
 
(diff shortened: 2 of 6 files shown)


cpinter:
To test whether importing implicitly was successful:
`segNode = slicer.vtkMRMLSegmentationNode()
slicer.mrmlScene.AddNode(segNode)
segNode.GetSegmentation()`

If the output is similar to this
`(vtkCommonCorePython.vtkObject)0000017BA6862EE8`
then it was unsuccessful. If it is like this
`(vtkCommonCorePython.vtkMRMLSegmentationNode)000002215A36E0A8`
then it was successful.

cpinter:
@jcfr @pieper Do you know what's missing from this so that vtkSegmentationCorePython is imported when Slicer is started? Thanks!

jcfr:
@cpinter Does it work when building against Qt5 + VTK8 ? 

Is yes, this it is related to VTK hierarchy files .. that became mandatory in VTK8 and not enable in the current build to avoid breaking anything.

Or you could remove the condition in `vtkMacroKitPythonWrap.cmake` and try with your current build .. 

See https://github.com/Slicer/Slicer/blob/ac3a3faba26e9536ce58a5832d5bf67cabf2d93a/CMake/vtkMacroKitPythonWrap.cmake#L112-L130

cpinter:
@jcfr I haven't tried with Qt5 or VTK8, I just changed the current master.

Thanks, I'll try what you suggest. Although I don't understand how vtkAddon and the others work, as what you pointed to is not specific to any of these libraries. What makes vtkSegmentationCore different to vtkAddon?

ihnorton:
@cpinter were you testing this on Windows? I just tried this PR (rebased to current trunk), and it seems to work correctly on Mac/Qt4/VTK7. If so, maybe that helps narrow down the issue...

```
>>> segNode = slicer.vtkMRMLSegmentationNode()
>>> slicer.mrmlScene.AddNode(segNode)
(vtkCommonCorePython.vtkMRMLSegmentationNode)0x14b257db8
>>> segNode.GetSegmentation()
(vtkCommonCorePython.vtkSegmentation)0x14b257ef0
>>> 
```

lassoan:
I've found the missing piece, it works now on Windows as well.

https://github.com/Slicer/Slicer/commit/02f0812df1615de427aef5267d0d36a954aa7c28

jcfr:
👍

cpinter:
Thanks a lot to you both! I'll integrate this tomorrow or Friday if there are no objections.

cpinter:
Oh I just see @lassoan integrated it. Thanks!

PR #795, opened 2017-09-21
Additional features for vtkPlotting

Punzo:
Addressing Mantis 0004435, 0004436 and 0004437. @lassoan

### DIFF
--- a/Libs/MRML/Core/Testing/vtkMRMLPlotChartNodeTest1.cxx
+++ b/Libs/MRML/Core/Testing/vtkMRMLPlotChartNodeTest1.cxx
@@ -76,8 +76,11 @@ int vtkMRMLPlotChartNodeTest1(int , char * [] )
 
   // Set and Observe the MRMLTableNode
   plotDataNode1->SetAndObserveTableNodeID(tableNode->GetID());
+  plotDataNode1->SetXColumnName(tableNode->GetColumnName(0));
+  plotDataNode1->SetYColumnName(tableNode->GetColumnName(1));
   plotDataNode2->SetAndObserveTableNodeID(tableNode->GetID());
-  plotDataNode2->SetYColumnIndex(2);
+  plotDataNode2->SetXColumnName(tableNode->GetColumnName(0));
+  plotDataNode2->SetYColumnName(tableNode->GetColumnName(2));
 
   // Add and Observe plots IDs in PlotChart
   node->AddAndObservePlotDataNodeID(plotDataNode1->GetID());
--- a/Libs/MRML/Core/vtkMRMLPlotChartNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLPlotChartNode.cxx
@@ -73,6 +73,7 @@ vtkMRMLPlotChartNode::vtkMRMLPlotChartNode()
   this->SetAttribute("AxisLabelFontSize", "12");
 
   this->SetAttribute("LookupTable", "");
+  this->SetAttribute("fitPlotToAxes", "off");
 
   vtkIntArray  *events = vtkIntArray::New();
   events->InsertNextValue(vtkCommand::ModifiedEvent);
--- a/Libs/MRML/Core/vtkMRMLPlotDataNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLPlotDataNode.cxx
@@ -79,8 +79,6 @@ vtkMRMLPlotDataNode::vtkMRMLPlotDataNode()
 {
   this->Plot = NULL;
   this->Type = -1;
-  this->XColumnIndex = 0;
-  this->YColumnIndex = 1;
   this->XColumnName = "(none)";
   this->YColumnName = "(none)";
   this->HideFromEditorsOff();
@@ -124,8 +122,8 @@ void vtkMRMLPlotDataNode::WriteXML(ostream& of, int nIndent)
   // Start by having the superclass write its information
   Superclass::WriteXML(of, nIndent);
   of << " Type=\"" << this->GetType() << "\"";
-  of << " XColumnIndex=\"" << this->GetXColumnIndex() << "\"";
-  of << " YColumnIndex=\"" << this->GetYColumnIndex() << "\"";
+  of << " XColumnName=\"" << this->GetXColumnName() << "\"";
+  of << " YColumnName=\"" << this->GetYColumnName() << "\"";
   of << " ";
 }
 
@@ -146,13 +144,13 @@ void vtkMRMLPlotDataNode::ReadXMLAttributes(const char** atts)
       {
       this->SetType(StringToInt(attValue));
       }
-    else if (!strcmp(attName, "XColumnIndex"))
+    else if (!strcmp(attName, "XColumnName"))
       {
-      this->SetXColumnIndex(StringToInt(attValue));
+      this->SetXColumnName(attValue);
       }
-    else if (!strcmp(attName, "YColumnIndex"))
+    else if (!strcmp(attName, "YColumnName"))
       {
-      this->SetYColumnIndex(StringToInt(attValue));
+      this->SetYColumnName(attValue);
       }
     }
   this->EndModify(disabledModify);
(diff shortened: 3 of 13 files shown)


lassoan:
Thanks a lot Davide, I'll test and merge it.

lassoan:
Thanks a lot! Committed r26385-r26391

I'm working on improving Tables module GUI to be able to change column types.

Punzo:
OK perfect!! then maybe next or next next we can discuss if and how we'd like to implement an extended PlotViewController GUI for the module ViewControllers.

lassoan:
I've added GUI to table module to set default column type and edit existing column types.
Many things are still not clear:
* why non-numeric columns appear in the series list? (if you select them then the chart is not updated anymore)
* can I have multiple plots in a single chart?
* can display columns from different tables in a single chart?
* why certain columns
* why do we need to create an "indexes" column?
* why "scatter" points shown up in the series list?

I don't think we need to make the GUI look similar to the Slicer Chart GUI. Something like the Paraview GUI would be much better.

Punzo:
Hi @lassoan, thanks for the additional GUI for the table. I'll test asap. All the other comments are good points. I'll reply and address them tomorrow. For the GUI, do you mean an extend version to put in the ViewController (or another module)? I don't think that the Paraview like GUI will be good in the PlotView.

Punzo:
1) you are right, I should check the column type and not add them if not numeric. I'll fix this. Thanks!
2) from the GUI is not possible, I think also this one should go on the extended GUI. If we put a way to change the type for each PlotData, the GUI will not even fit the PlotView.
3) You are completly right, I didn't see that method. I'll fix this. Thanks!
4) Let me check again, for the marker I was creating a "fake" plotData. I know, I also don't like this. I'll try to remove this.
5) We can add a Marker button as in the Chart GUI controller. I'll do the change if you think it is better.
6) OK, I good to know. I never did such conversion, but you are right.

lassoan:
We should have GUI to add/remove/edit plot data nodes at one place. Removing and (limited) editing is already in the ChartViewController, so it makes sense to try to add plot data adding and better editing there. We should probably add a custom MRML widget, which can be put in the ChartViewController but can be also added to custom modules. If we find that ChartViewController gets too busy if we add all the necessary plot editing options, then we could add a plot module. Plot module GUI could contain all the details, while ChartViewController could contain most frequently used options (similar to slice view controller widget and Reformat module).

Maybe before implementing anything, you could draw some GUI mockup that is easier to change. Either a powerpoint slide or online GUI mockup tools would work.

Punzo:
OK, I'll see and maybe the best is to further discuss this at the next developer hangout. However, tomorrow I can't (sorry).

Punzo:
By the way what about adding in the setting some GUI for changing the default value for the type of Column of the table GUI? something like the type of rendering (CPU or GPU). I don't know if it complicated and worth to do it, but it will avoid that users have every time to set the preferred default type of the column.

lassoan:
A clean separation could be to keep only Chart editing in ChartViewController (list of plots to show, title, axis labels, grid, ...) and move all plot data node editing (plot type, marker type, input table, columns) to a module GUI.

Punzo:
This last solution is what already I was thinking. At the moment, in fact, I didn't put anything to edit PlotData. The GUI at the moment allow to modify only PlotChart (i.e., the "global prop" of the Plot).

lassoan:
> setting some GUI for changing the default value for the type of Column of the table GUI

You can set the default column type already and it is saved into the scene (set separately for each new table).

For new tables, I would keep the default column type as "string", as that way values are not changed when copy-pasted into the table or between columns. If somebody does not know or does not care about column types, it would always work as expected.

Note that you can change the default column type by defining a custom default node for vtkMRMLTableNode in the scene (any module or extension can do it at application startup, even from Python - see http://apidocs.slicer.org/master/classvtkMRMLScene.html#ae302c5ed4aabb2910bc35dcc9aa2513f).

lassoan:
> I didn't put anything to edit PlotData

Isn't plot type (line, bar, ...) stored in PlotData?

Punzo:
Ok thanks for the info regarding the Table. Yes each PlotData has a pointer to a vtkplot object (which can be Line, scatter or Bar) and the setType method takes care to set the connection with the right vtkPlot object. My point is more: in mind the type is also a PlotChart property (at the moment is designed like this) and the GUI Type button is connected to the PlotChart Type property. So the user with only one click can change the Type of all the plotDataNodes selected. Of course, in the extended version we can add also GUI to change the Type of the PlotData one by one.

lassoan:
I think it's better to keep plot type in plotdata node:
1. I may need to mix different plot types in a single figure, e.g., show a bar chart overlaid with lines.
2. It is stored like that in VTK anyway, so it would be much more complicated to work around that (e.g., when you add the same plotdata to multiple charts, then you would need to duplicate the plot object to allow different types, ...)

It would be also nice to add the option for using labels. Especially for bar charts, we need to be able to give names of the columns based on a string array. Labels are already supported by VTK classes, we just have to expose that property in Slicer.

Punzo:
1) Ok.
2) Sorry maybe I was not clear. The type attribute and all the memory (vtkObject and so on) are all stored in PlotData. So it will work as you expect: you  can use the method SetType on a PlotData Object and change the type of each PlotData one by one (and for the extended GUI we can expose this). 

However, what I mean is that PlotChart has a string attribute ( https://github.com/Punzo/Slicer/blob/master/Libs/MRML/Core/vtkMRMLPlotChartNode.h#L193 ) connected to the Type button in the GUI of the current PlotViewController. If the user changes the value of the button, then the widget will call SetType for all the selected PlotDataNodes selected in the current PlotChartNode. I definitively want to keep this features due that I don't want that the user to change the type of the PlotDataNodes needs to do one by one with tens of clicks.

3) Labels and Bar. I know, I agree will nice to have, but at the moment I don't have time. It was already a point in the Future Work section in the LAB and now in the documentation https://www.slicer.org/wiki/Documentation/Nightly/Developers/Plots 
Sincerely I don't know how much work this will need, I'll try, but I can't guarantee this feature.

lassoan:
> I definitively want to keep this features due that I don't want that the user to change the type of the PlotDataNodes needs to do one by one with tens of clicks.

There are two requirements then, which are not contradicting, so we should be able to find a good GUI:
* On the GUI it must be very clear the plot type is not a chart property. The current solution is not good, because what I see is that if I click on the type button then various charts are randomly changed (the problem of course that the same PlotData nodes are used in multiple charts).
* It should be easy to change plot type from the chart view controller.

One solution could be to add a few (or all?) plotdata editing to chart view controller. To change properties in multiple plotdata nodes at once, user could multi-select them.

It is not clear for me what the plot button in the table module is supposed to do. Always create new plotdata nodes? Does it add it to all chart nodes? Also, are all the plotdata nodes shown in all chart view controller widgets?

You showed plot editing (moving points in the chart by drag-and-drop). Is that available, too? I could not figure out how to achieve that.

> Labels and bar

It is not urgent, we don't have to implement it now.

Punzo:
Ok, let's do this:
1) tomorrow I'll fix the minor issues that you reported earlier
2) I'll think a little bit more about the GUI and solutions. Then we can discuss agian by hangout meeting (which it can be earlier respect to next next Tuesday), which is faster and clearer. 
3) Plot button in the Table Module: it create PlotDataNodes for each Column (after the first one which is used as X-Axis) and it add automatically the references of the "selected" ones in the "selected" PlotChatNode. If one or more PlotData exist already, then it will simply readd the reference (and therefore, reselecting) to the PlotChartNode (so without duplicate PlotDataNodes). The "selected" PlotChartNode is the one in the widget (and its ID is stored in the selection Singleton). If there is no PlotChartNode selected, it will create it and set as the "selected" one. I don't know if I have been enough clear, I can explain faster and better at the hangout call.
4) as written in the documentation: KEY_S is to change the type of selection (twice for the Click and Drag) https://www.slicer.org/wiki/Documentation/Nightly/Developers/Plots

Punzo:
P.S.: I don't think that the current implmenetation of the Type button is not good. Because in a cetrain way it is exactly what you like to have: change the type of some selected plots. The problem is that you are using the same plotDataNode in more PlotChartNodes. You will have the same issue even if you have an interface to edit all teh Properties of PlotDataNode one by one. I think that the issue is more taht you would like to create more PlotDataNode with the same input (which at the moment from the GUI is not possible, because the plot button will not create a duplicate)

PR #798, opened 2017-09-27
ENH: Write coordinate system to model files

lassoan:
https://issues.slicer.org/view.php?id=4408

In recent years, LPS coordinate system has become much more widely used than RAS.
When Slicer saves a node to file, it already uses RAS for images and transforms and can use LPS for markups.
However, model node have been still saved as RAS. This caused a problem, because when models and images are
used or created by external software, models and images are not displayed consistently.

Unfortunately, generic mesh file formats have no standard way of storing coordinate system (axis orientation)
information, therefore switching to LPS could not be performed in a single step without breaking backward compatibility.

Proposed solution: Switch to LPS-based model files in multiple steps.

Step 1: Write coordinate system information (SPACE=RAS) to model files. Use comments in STL, PLY, and OBJ files; as field data in VTK files.
Keep writing data as RAS.

Step 2: Allow reading/writing models that use LPS coordinate system (add an option in Add data dialog; add an option in storage node).
Use coordinate system that is stored in the file header. If no coordinate system information is found in the file then assume RAS.

Step 3: Write models to file in LPS coordinate system by default.

Step 4: If no coordinate system information is found in the file then assume LPS.

---

Step 3-4 somewhat break backward compatibility, therefore there should be either sufficient amount of time (1-2 years) has to be
left after Step 1, or these steps should be done in a new major release.

This commit implements Step 1 (with the limitation that coordinate system info cannot be saved into OBJ files and ascii STL files
due to limitation of VTK API - but these formats are rarely used anyway).

### DIFF
--- a/Libs/MRML/Core/vtkMRMLModelStorageNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLModelStorageNode.cxx
@@ -22,6 +22,7 @@
 #include <vtkBYUReader.h>
 #include <vtkCellArray.h>
 #include <vtkDataSetSurfaceFilter.h>
+#include <vtkFieldData.h>
 #include <vtkNew.h>
 #include <vtkObjectFactory.h>
 #include <vtkOBJReader.h>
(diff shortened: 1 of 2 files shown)


pieper:
Steps 1 and 2 make a lot of sense and there's no reason not to do that.

Steps 3 and 4 sound risky to me, but if there's no better alternative that can be the default plan.

What I'd much rather see that if we are changing the default storage of models we pick a completely different format to avoid ambiguities.  DICOM surface segmentations could be a good choice but we could see if there are others that would make more sense.  We could maintain VTK export, with the SPACE declaration, as a backwards compatibility mode along with STL and OBJ.

lassoan:
Yes, step 3 and 4 may cause troubles so probably a few year transition time is preferable.

I agree that it would be nice to go towards using DICOM as a primary format for saving data, and in general save everything into DICOM. Currently, saving is mixed with exporting, so it is all somewhat complicated and error-prone, but hopefully we can separate them a bit better in the coming years.

lassoan:
Merged in r26398. Future steps will be discussed in https://issues.slicer.org/view.php?id=4408

jcfr:
Closing. Integrated in [r26398](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26398)

PR #799, opened 2017-09-28
ENH: Added projection option to model slice display

lassoan:
Models consisting of thin structures (such as catheters, coils, stents), then are hardly visible
in slice views, as their intersection with the slice is just a few points or small circles.

Projection option is already available for markups and annotations to show points in slice views that are out
of the current plane and a common technique (maximum intensity projection) for images as well.

Now projection option is also added to model nodes. User interface is not yet added for this feature
(it would require more experimentation to see how it is the best to expose this option), but can
be activated from the Python console for a model node:

    slicer.util.getNode('MyModel').GetDisplayNode().SetSliceProjectionModeToFull()


lassoan:
Coil displayed in slice views as intersection:

image

Coil displayed in slice views as projection:

image

fedorov:
Nice feature!

pieper:
👍

lassoan:
Any suggestion where to put display the projection option in the Models module GUI? Add one more line below "Slice intersections visible? Or create a new section for all slice intersection related options (slice intersection thickness, slice intersections opacity, slice intersection projection)?

Current Models module GUI:

image

pieper:
Yes, a collapsable box called Slice Intersections makes a lot of sense. 

It looks like you have it set it up to support showing both the Intersections and the Projection - does that work?  It seems like the most valuable view.  

A check box and (and maybe opacity slider) for the projections makes sense.

lassoan:
Currently it is either projection or intersection (there is one actor that either uses the cut or a flattened output).

I've tried loading the same model twice and show both intersection and projection but it's not really useful (intersections look like some artifacts or speckles on the image).

To show what is on each side of the slice, I think we should cut the model in half and show it with different colors (similarly how markup points or lines are rendered with different style depending on whether in front or behind the slice plane). Or, we could add a slice thickness parameter and only project those parts of the model that are within that thickness (it can be a hard cut or a gradual fadeout - farther the point, the more transparent it becomes).

I've made projection an enum so that we can add this more sophisticated display modes in addition to the simple "full projection" mode (when the entire object is projected). Maybe "uniform" would be a better name (uniform projection of the entire model, regardless of position).

pieper:
Okay, makes sense.

In that case maybe a radio button to select between projection and intersection would be the most logical.  The group could be Slice Display and they could share the opacity slider and the width option would only be enabled in Intersection mode.

lassoan:
I've committed the changes in MRML node and displayable manager, but GUI had to be modified quite a bit, so it would be good to get some feedback.

Models module's Display section was quite inconsistent and difficult to use (see for example https://issues.slicer.org/view.php?id=4336 and related ticket for a good summary).

Here is the proposed GUI (as it is shown by default if Models module is opened):

image


With 3D display / Advanced option exposed:

image

@pieper @jcfr @fedorov @cpinter @agirault - What do you think?

pieper:
LGTM - yes, the models module gui needed improvement.  Thanks!

fedorov:
Looks like a good improvement to me!

jcfr:
Closing. Integrated in [r26406](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26406)

lassoan:
Merged Models module GUI update in [r26408](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26408)

PR #800, opened 2017-09-29
ENH: Added show option to Add data dialog for volume loading

lassoan:
'Show' is enabled by default and so volume that is loaded is displayed.
However, user can disable 'Show' option and then current view settings will be preserved.

Also removed switching to batch processing mode during volume loading, as this is a very expensive operation
(takes time, screen flickers, etc) and it is not necessary anymore.

The option is also accessible from Python:

    slicer.util.loadVolume('fixed.nrrd',{'show':False})

### DIFF
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -363,6 +363,16 @@ def loadLabelVolume(filename, properties={}, returnNode=False):
   return loadNodeFromFile(filename, filetype, properties, returnNode)
 
 def loadVolume(filename, properties={}, returnNode=False):
+  """Properties:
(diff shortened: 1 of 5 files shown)


lassoan:
Add data dialog with 'Show' option:

image

jcfr:
Thanks.

LGTM :+1

cpinter:
Excellent, this is very useful and I bet the problem this solves has bothered many users and developers. I wonder if the lack of batch processing will have any side effect. Do you know why it was there in the first place?

pieper:
👍

lassoan:
Integrated in rev 26407

PR #803, opened 2017-10-03
COMP: Update Slicer_Remote_Add to account for full set of dependent options

jcfr:
This commit ensures that configuring with -DSlicer_BUILD_CLI set to OFF
will effectively disable Slicer_BUILD_EMSegment

Fixes https://issues.slicer.org/view.php?id=3701


jcfr:
Integrated in [r26413](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26413)

PR #808, opened 2017-10-04
BUG: fix PlotInfo GUI MarkerComboBox unEnabling for Box type

Punzo:
a very minor fix for the PlotInformation GUI. @jcfr or @lassoan may you merge this as well? Thanks!


jcfr:
Should a similar fix be applied for `Type` and `XAxis` attributes ?

Punzo:
No Type and XAxis do not need the fix. It is necessary only for the MarkerComboBox because of this block https://github.com/Slicer/Slicer/pull/808/files#diff-64a2077f94e8c587f82e315716131dd0L435 which comes before the Custom checks

jcfr:
Integrated in [r26424](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26424)

PR #812, opened 2017-10-12
ENH: model DICOM acquisition geometry (addresses #4409)

pieper:
Add a step to the DICOMScalarVolumePlugin to check that the loaded
volume node's slice geometry matches what is expected from
the position and orientation information of the individual
slices in the DICOM headers.

If needed, a grid transform is created to reposition each slice
to the appropriate position in patient space.

The DICOMReaders self test has been extended to test this in the
case where slices are missing from the middle of the volume.  The
grid transform results in an interpolated band of pixels using
information from adjacent slices to 'fill in' for the missing
data.  The test confirms that the last slice of the reconstructed
volume is in the expected location in patient space.  This correction
means that calculations like ruler or segment statistics can be
robust even with missing slices.

Currently the only available testing data is for the case of missing
slices.  However this code should allow proper loading and display of
several not-uncommon imaging scenarios:

1) Missing slices due to network failure or accidental file deletion.

2) Gantry tilted image acquisitons (see #4409) which is
used, for example, to minimize radiation exposure to the retinas
especially in pediatric imaging.

3) Variable table speed during CT acquisition, which is sometimes
used in abdominal or musculoskeletal imaging to provide high resolution
in selected body parts and lower resolution between (e.g. high res
at the ankle, knee and hip, but low res in the shin and thigh).

Consderations:

By creating a transform we preserve the original data and leave it
to the user to resample if desired.  The option of resampling
during load was rejected since it is unclear what sampling grid
would be best in all cases.

Although the transform should correct the geometry, a warning dialog is
still generated during loading since the issue could be correctable
(such as missing slices) or may be triggered by other data issues
not forseen by this code.

### DIFF
--- a/Applications/SlicerApp/Testing/Python/DICOMReaders.py
+++ b/Applications/SlicerApp/Testing/Python/DICOMReaders.py
@@ -83,9 +83,11 @@ class DICOMReadersTest(ScriptedLoadableModuleTest):
     """Run as few or as many tests as needed here.
     """
     self.setUp()
-    self.test_Part1DICOM()
+    self.test_AlternateReaders()
+    self.setUp()
+    self.test_MissingSlices()
 
-  def test_Part1DICOM(self):
+  def test_AlternateReaders(self):
     """ Test the DICOM loading of sample testing data
     """
     testPass = True
@@ -153,7 +155,7 @@ class DICOMReadersTest(ScriptedLoadableModuleTest):
         indexer.waitForImportFinished()
 
         #
-        # selct the series
+        # select the series
         #
         detailsPopup = slicer.modules.DICOMWidget.detailsPopup
         detailsPopup.open()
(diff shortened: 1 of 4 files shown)


pieper:
This is a work in progress.  It works for the case I tried, which is embodied in the associated self test for the case of missing slices.  I'm hoping to collect some additional datasets that would allow testing of the other conditions.

For the self test, this is the image when loaded with the missing slices:

image

And here is what it looks like after the fix has been applied:
image

lassoan:
Steve, this is awesome! With that it required minimum amount of change, avoided data loss, and handled gantry tilt/missing slices/varying table speed/potentially other unusual geometries all at once, in a generic way - it is such an elegant solution that I haven't seen elsewhere. I'm just a bit sad that this nice design was not my idea. I'll give it a test tomorrow!

pieper:
Thank you Andras!

😃

pieper:
By the way of course, this approach is only possible due to all the great nonlinear transform support you guys have added so thanks for that too!

fedorov:
Conceptually, this looks great to me! I have several deadlines next week, but I can spend more time going over the code and testing towards the end of the next week.

Is this going to work passing these volumes to CLIs?

pieper:
@fedorov  In the last developer hangout we discussed the issue of CLIs and transforms and decided that we'll leave that up to the user for now (create a resampled volume in whatever space is needed).

Essentially the change in this pull request just adds a transform to the loaded volume, so it's no different than any other case where you might use a CLI on a transformed volume.

Because of that I think the right thing could be to add options to the node selectors defining the desired sampling space of the volume (e.g. by selecting another volume node from a submenu).  Then a volume in that space can be created on the fly and sent to the CLI.

pieper:
I was not able to get a gantry tilted volume for testing, but I did some more testing of this with different volumes and didn't see any issues.  So I think this is ready to commit.  It seems unlikely to break anything as far as I can see.

lassoan:
@pieper It worked well for me for a gantry-tilted CT! The loaded image has correct shape. 

However, there were a few problems with other series:
- scout images looked collapsed and there were many warnings printed that transform inversion failed; probably these images (where slices intersect each other)  should be loaded as sequence nodes
- it seems that now all volume is placed under a transform, not just where there is a non-rigid transformation needed
- loading of some volume hangs or takes really long time (I did not have time to investigate)

I'll try to anonymity the images that I have and send it to you.

pieper:
Thanks for testing!  Those all sound fixable.  Any data fort testing would
be great.

fedorov:
> @fedorov In the last developer hangout we discussed the issue of CLIs and transforms and decided that we'll leave that up to the user for now (create a resampled volume in whatever space is needed).
>
> Essentially the change in this pull request just adds a transform to the loaded volume, so it's no different than any other case where you might use a CLI on a transformed volume.

Makes sense, except for a non-savvy user, it is indeed quite different. In other situations transform is explicitly created by the user in the process of applying other modules. In this specific case it sounds like the transform would be added quietly by the DICOM Browser. Even advanced users may not think this transform is present. 

I have not tested, so I don't know if there is any indication to the user that would help them understand that the volume they see in the slice viewers is a result of a transformation. This potentially can lead to situations that might be tricky to troubleshoot.

lassoan:
We could display a warning during import the same way as before, but the text could explain how to resample the volume.

Or, we could make CLI execution fail if an input volume is under non-linear transform (error display is quite nice in CLI module widgets, so users would easily see the message). 

More user-friendly solutions would break CLI backward compatibility (XML file should describe for each volume in which coordinate system the volume is required to be in; and CLIs could choose to get the transform or get the transformed volume). I would not make the change before Slicer5.

fedorov:
> We could display a warning during import

I doubt these warnings on import are effective. Basic users have enough challenges using Slicer to remember (or even read and understand!) this warning, which would only apply in certain situations.

> we could make CLI execution fail if an input volume is under non-linear transform

I really like this idea! I would not fail, because you never know what users may actually want to do - maybe better to always give warning? Or is it possible to force users to accept something before CLI is run? If it is a warning, if processing gives unexpected results, they know how to troubleshoot. Warning will not help if processing results are not above some threshold of "unexpectedness", so explicit accept may be better, but I don't know if it is possible.

> More user-friendly solutions would break CLI backward compatibility

pieper:
How about something like the following:

If the user selects a volume as input to a CLI in the node selector combo box we add a yellow exclamation point (or other icon) with a tooltip suggesting hardening or resampling.

Since this is independent of the gantry tilt issue I've started a new issue to track it:

https://issues.slicer.org/view.php?id=4459

lassoan:
Silencing inverse computation failures could be problematic, as currently there is no other mechanism to report inverse computation errors. Could you add a flag that switches between reporting first/all inversion errors (it can be "report first occurrence only" by default) and/or invoke some custom event each time when the error occurs?

By the way, neither missing frames, nor gantry tilt should create these errors. Does it happen only for scout scans - where you have intersecting planes?

pieper:
@lassoan this fix: https://github.com/Slicer/Slicer/pull/812/commits/9999a975c65c82e69a531827eb61d58be95e2977 should address the app hanging issue for certain transforms.  Could you review the logic and see if you agree?


Regarding the extra transforms where they probably aren't needed, that seems to be a tolerance issue with numpy.allclose [here](https://github.com/Slicer/Slicer/pull/812/files#diff-5c81585867008dffacf021ea0c9d23d4R710).  I'll try to come up with something rational.

Regarding the scout/localizer this looks like an independent issue.  Even when there's a series that has two or three images it is being loaded as a single slice.  This behavior is consistent with 4.6.2.

pieper:
@lassoan 

> Silencing inverse computation failures could be problematic, as currently there is no other mechanism to report inverse computation errors. Could you add a flag that switches between reporting first/all inversion errors (it can be "report first occurrence only" by default) and/or invoke some custom event each time when the error occurs?

I find it hard to imagine that anyone would want all those messages except maybe for debugging.  When it happens it seems to happen for every pixel and that would almost always be excessive.

How about a warning macro on the first occurrence and then debug macros afterwards?  Or yes, a custom event would also be fine and probably that's the cleaner solution.

> 
> By the way, neither missing frames, nor gantry tilt should create these errors. Does it happen only for scout scans - where you have intersecting planes?

It was also happening for DWI scans where there are multiple slices at the same physical location.   These are not scalar volumes, and SlicerDMRI should detect them as DWI to load them as diffusion images (loadable with a higher priority).

I also found that these convergence error messages came up a lot while I was debugging the acquisition transform code, even for transforms that I thought should be valid, but of course it was hard to debug when the application would crash.

lassoan:
Often you evaluate the inverse transform only at particular locations and it may be important to not miss any error messages. Typical example is that you use registration to get resampling transform (no need to compute inverse transform) but you need to invert the transform to propagate target point positions (just a couple of points, it is useful to see which points were not transformed accurately).

pieper:
In those cases your program will want to know the error so invoking an observable event sounds like a good solution.  I'll add that.

jcfr:
@pieper Is this ready for integration ?

pieper:
@jcfr in my opinion yes, it should work fine in the cases where it is
needed and it has no known failure modes.  The only question in my mind is
whether it should go in the 4.8 release or go into the Nightly for a
while.  Let's decide during out hangout today.

pieper:
After discussion we determined at the hangout that we'll hold off committing this until after the 4.8 release is out later this week.

pieper:
I made another couple tweaks committed as 26500 and 26501.

Thanks for the reviews!

PR #813, opened 2017-10-13
Improve subject hierarchy combobox

agirault:
- [x] Option for new vertical alignment of the popup treeview
- [x] Disable selections of parent items when using attribute filter
- [x] Do not hide popup when no item was selected
- [x] Implement qMRMLSubjectHierarchyComboBox::clearCurrentItem 
- [x] Update combobox label on SubjectHierarchyItemModifiedEvent

In a future pr:
- [ ] Scene item shown when root item != scene and filters returns no item
- [ ] combobox does not stretch in layout like qMRMLNodeComboBox
- [ ] popup wider that combobox due to margins

### DIFF
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSortFilterSubjectHierarchyProxyModel.cxx
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSortFilterSubjectHierarchyProxyModel.cxx
@@ -261,7 +261,8 @@ bool qMRMLSortFilterSubjectHierarchyProxyModel::filterAcceptsRow(int sourceRow, 
 }
 
 //------------------------------------------------------------------------------
-bool qMRMLSortFilterSubjectHierarchyProxyModel::filterAcceptsItem(vtkIdType itemID)const
+bool qMRMLSortFilterSubjectHierarchyProxyModel::filterAcceptsItem(vtkIdType itemID,
+                                                                  bool canAcceptIfAnyChildIsAccepted/*=true*/)const
 {
   Q_D(const qMRMLSortFilterSubjectHierarchyProxyModel);
 
@@ -319,8 +320,15 @@ bool qMRMLSortFilterSubjectHierarchyProxyModel::filterAcceptsItem(vtkIdType item
     std::string levelFilterStr(d->LevelFilter.toLatin1().constData());
     if (!shNode->IsItemLevel(itemID, levelFilterStr))
       {
-      // If level was requested but different, then only show if any of its children are shown
-      onlyAcceptIfAnyChildIsAccepted = true;
+      if (canAcceptIfAnyChildIsAccepted)
+        {
+        // If level was requested but different, then only show if any of its children are shown
+        onlyAcceptIfAnyChildIsAccepted = true;
+        }
+      else
+        {
+        return false;
+        }
       }
     }
 
@@ -330,8 +338,15 @@ bool qMRMLSortFilterSubjectHierarchyProxyModel::filterAcceptsItem(vtkIdType item
     std::string attributeNameFilterStr(d->AttributeNameFilter.toLatin1().constData());
     if (!shNode->HasItemAttribute(itemID, attributeNameFilterStr))
       {
-      // If attribute was requested but missing, then only show if any of its children are shown
-      onlyAcceptIfAnyChildIsAccepted = true;
+      if (canAcceptIfAnyChildIsAccepted)
+        {
+        // If attribute was requested but missing, then only show if any of its children are shown
+        onlyAcceptIfAnyChildIsAccepted = true;
+        }
+      else
+        {
+        return false;
+        }
       }
     else if (!d->AttributeValueFilter.isEmpty())
       {
@@ -339,8 +354,15 @@ bool qMRMLSortFilterSubjectHierarchyProxyModel::filterAcceptsItem(vtkIdType item
       std::string attributeValue = shNode->GetItemAttribute(itemID, attributeNameFilterStr);
       if (attributeValue.compare(attributeValueFilterStr))
         {
-        // If attribute value check was requested but failed, then only show if any of its children are shown
-        onlyAcceptIfAnyChildIsAccepted = true;
+        if (canAcceptIfAnyChildIsAccepted)
+          {
+          // If attribute value check was requested but failed, then only show if any of its children are shown
+          onlyAcceptIfAnyChildIsAccepted = true;
+          }
+        else
+          {
+          return false;
+          }
         }
       }
     }
@@ -351,7 +373,15 @@ bool qMRMLSortFilterSubjectHierarchyProxyModel::filterAcceptsItem(vtkIdType item
     QString itemName(shNode->GetItemName(itemID).c_str());
     if (!itemName.contains(d->NameFilter, Qt::CaseInsensitive))
       {
-      onlyAcceptIfAnyChildIsAccepted = true;
+      if (canAcceptIfAnyChildIsAccepted)
+        {
+        // If item name was requested but different, then only show if any of its children are shown
+        onlyAcceptIfAnyChildIsAccepted = true;
+        }
+      else
+        {
+        return false;
+        }
       }
     }
 
@@ -377,4 +407,26 @@ bool qMRMLSortFilterSubjectHierarchyProxyModel::filterAcceptsItem(vtkIdType item
 
   // All criteria were met
   return true;
+}
+
+//------------------------------------------------------------------------------
+Qt::ItemFlags qMRMLSortFilterSubjectHierarchyProxyModel::flags(const QModelIndex & index)const 
+{
+  vtkIdType itemID = this->subjectHierarchyItemFromIndex(index);
+  bool isSelectable = this->filterAcceptsItem(itemID, false);
+  qMRMLSubjectHierarchyModel* sceneModel = qobject_cast<qMRMLSubjectHierarchyModel*>(this->sourceModel());
+  QStandardItem* item = sceneModel->itemFromSubjectHierarchyItem(itemID);
+  if (!item)
+    {
+    return Qt::ItemFlags();
+    }
+
+  if (isSelectable)
+    {
+    return item->flags() | Qt::ItemIsSelectable;
+    }
+  else
+    {
+    return item->flags() & ~Qt::ItemIsSelectable;
+    }
 }
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.cxx
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.cxx
@@ -96,8 +96,12 @@ void qMRMLSubjectHierarchyComboBoxPrivate::init()
   // Make connections
   QObject::connect(this->TreeView, SIGNAL(currentItemChanged(vtkIdType)),
                    q, SLOT(updateComboBoxTitleAndIcon(vtkIdType)));
+  QObject::connect(this->TreeView, SIGNAL(currentItemModified(vtkIdType)),
+                   q, SLOT(updateComboBoxTitleAndIcon(vtkIdType)));
   QObject::connect(this->TreeView, SIGNAL(currentItemChanged(vtkIdType)),
                    q, SIGNAL(currentItemChanged(vtkIdType)));
+  QObject::connect(this->TreeView, SIGNAL(currentItemModified(vtkIdType)),
+                   q, SIGNAL(currentItemModified(vtkIdType)));
 }
 
 // --------------------------------------------------------------------------
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.cxx
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.cxx
@@ -155,6 +155,13 @@ void qMRMLSubjectHierarchyComboBox::setMRMLScene(vtkMRMLScene* scene)
 }
 
 //------------------------------------------------------------------------------
+void qMRMLSubjectHierarchyComboBox::clearSelection()
+{
+  Q_D(const qMRMLSubjectHierarchyComboBox);
+  d->TreeView->clearSelection();
+}
+
+//------------------------------------------------------------------------------
 vtkIdType qMRMLSubjectHierarchyComboBox::currentItem()const
 {
   Q_D(const qMRMLSubjectHierarchyComboBox);
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.cxx
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.cxx
@@ -56,6 +56,7 @@ public:
 
 public:
   int MaximumNumberOfShownItems;
+  bool AlignPopupVertically;
 
   qMRMLSubjectHierarchyTreeView* TreeView;
 };
@@ -64,6 +65,7 @@ public:
 qMRMLSubjectHierarchyComboBoxPrivate::qMRMLSubjectHierarchyComboBoxPrivate(qMRMLSubjectHierarchyComboBox& object)
   : q_ptr(&object)
   , MaximumNumberOfShownItems(20)
+  , AlignPopupVertically(true)
   , TreeView(NULL)
 {
 }
@@ -259,6 +261,20 @@ void qMRMLSubjectHierarchyComboBox::setMaximumNumberOfShownItems(int maxNumberOf
   d->MaximumNumberOfShownItems = maxNumberOfShownItems;
 }
 
+//--------------------------------------------------------------------------
+bool qMRMLSubjectHierarchyComboBox::alignPopupVertically()const
+{
+  Q_D(const qMRMLSubjectHierarchyComboBox);
+  return d->AlignPopupVertically;
+}
+
+//--------------------------------------------------------------------------
+void qMRMLSubjectHierarchyComboBox::setAlignPopupVertically(bool align)
+{
+  Q_D(qMRMLSubjectHierarchyComboBox);
+  d->AlignPopupVertically = align;
+}
+
 //-----------------------------------------------------------------------------
 void qMRMLSubjectHierarchyComboBox::showPopup()
 {
@@ -289,28 +305,36 @@ void qMRMLSubjectHierarchyComboBox::showPopup()
   listRect.setWidth( listRect.width() + container->frameWidth());
   listRect.setHeight( listRect.height() + marginTop + marginBottom + tvMarginTop + tvMarginBottom);
 
-  // Position horizontally
-  listRect.moveLeft(above.x());
+  if(d->AlignPopupVertically)
+    {
+    // Position horizontally
+    listRect.moveLeft(above.x());
 
-  // Position vertically so the currently selected item lines up with the combo box
-  const QRect currentItemRect = d->TreeView->visualRect(d->TreeView->currentIndex());
-  const int offset = listRect.top() - currentItemRect.top();
-  listRect.moveTop(above.y() + offset - listRect.top());
+    // Position vertically so the currently selected item lines up with the combo box
+    const QRect currentItemRect = d->TreeView->visualRect(d->TreeView->currentIndex());
+    const int offset = listRect.top() - currentItemRect.top();
+    listRect.moveTop(above.y() + offset - listRect.top());
 
-  if (listRect.width() > screen.width() )
-    {
-    listRect.setWidth(screen.width());
+    if (listRect.width() > screen.width() )
+      {
+      listRect.setWidth(screen.width());
+      }
+    if (mapToGlobal(listRect.bottomRight()).x() > screen.right())
+      {
+      below.setX(screen.x() + screen.width() - listRect.width());
+      above.setX(screen.x() + screen.width() - listRect.width());
+      }
+    if (mapToGlobal(listRect.topLeft()).x() < screen.x() )
+      {
+      below.setX(screen.x());
+      above.setX(screen.x());
+      }
     }
-  if (mapToGlobal(listRect.bottomRight()).x() > screen.right())
+  else
     {
-    below.setX(screen.x() + screen.width() - listRect.width());
-    above.setX(screen.x() + screen.width() - listRect.width());
+    listRect.moveTo(below);
     }
-  if (mapToGlobal(listRect.topLeft()).x() < screen.x() )
-    {
-    below.setX(screen.x());
-    above.setX(screen.x());
-    }
+
   container->setGeometry(listRect);
   container->raise();
   container->show();
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.cxx
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.cxx
@@ -98,8 +98,6 @@ void qMRMLSubjectHierarchyComboBoxPrivate::init()
                    q, SLOT(updateComboBoxTitleAndIcon(vtkIdType)));
   QObject::connect(this->TreeView, SIGNAL(currentItemChanged(vtkIdType)),
                    q, SIGNAL(currentItemChanged(vtkIdType)));
-  QObject::connect(this->TreeView, SIGNAL(currentItemChanged(vtkIdType)),
-                   container, SLOT(hide()));
 }
 
 // --------------------------------------------------------------------------
@@ -378,6 +376,10 @@ void qMRMLSubjectHierarchyComboBox::updateComboBoxTitleAndIcon(vtkIdType selecte
     this->setDefaultIcon(QIcon());
     return;
     }
+
+  // Hide popup
+  QFrame* container = qobject_cast<QFrame*>(this->view()->parentWidget());
+  container->hide();
 
   // Assemble title for selected item
   QString titleText(shNode->GetItemName(selectedShItemID).c_str());
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
@@ -311,6 +311,9 @@ void qMRMLSubjectHierarchyTreeView::setSubjectHierarchyNode(vtkMRMLSubjectHierar
 
   d->SubjectHierarchyNode = shNode;
 
+  qvtkReconnect( shNode, vtkMRMLSubjectHierarchyNode::SubjectHierarchyItemModifiedEvent,
+                 this, SLOT( onSubjectHierarchyItemModified(vtkObject*,void*) ) );
+
   if (!shNode)
     {
     d->Model->setMRMLScene(NULL);
@@ -1295,6 +1298,64 @@ void qMRMLSubjectHierarchyTreeView::setMultiSelection(bool multiSelectionOn)
 bool qMRMLSubjectHierarchyTreeView::multiSelection()
 {
   return (this->selectionMode() == QAbstractItemView::ExtendedSelection);
+}
+
+//-----------------------------------------------------------------------------
+vtkIdType qMRMLSubjectHierarchyTreeView::firstSelectedSubjectHierarchyItemInBranch(vtkIdType itemID)
+{
+  // Check if item itself is selected
+  Q_D(qMRMLSubjectHierarchyTreeView);
+  if (d->SelectedItems.contains(itemID))
+    {
+    return itemID;
+    }
+
+  // Look for selected item in children recursively
+  std::vector<vtkIdType> childItemIDs;
+  d->SubjectHierarchyNode->GetItemChildren(itemID, childItemIDs, true);
+  for (std::vector<vtkIdType>::iterator childIt=childItemIDs.begin(); childIt!=childItemIDs.end(); ++childIt)
+    {
+    vtkIdType selectedId = this->firstSelectedSubjectHierarchyItemInBranch(*childIt);
+    if (selectedId != vtkMRMLSubjectHierarchyNode::INVALID_ITEM_ID)
+      {
+      return selectedId;
+      }
+    }
+
+  // That item is not selected and does not have
+  // any children items selected
+  return vtkMRMLSubjectHierarchyNode::INVALID_ITEM_ID;
+}
+
+//-----------------------------------------------------------------------------
+void qMRMLSubjectHierarchyTreeView::onSubjectHierarchyItemModified(vtkObject *caller, void *callData)
+{
+  vtkMRMLSubjectHierarchyNode* shNode = vtkMRMLSubjectHierarchyNode::SafeDownCast(caller);
+  if (!shNode)
+    {
+    qCritical() << Q_FUNC_INFO << ": Failed to access subject hierarchy node";
+    return;
+    }
+
+  // Get item ID
+  vtkIdType itemID = vtkMRMLSubjectHierarchyNode::INVALID_ITEM_ID;
+  if (callData)
+    {
+    vtkIdType* itemIdPtr = reinterpret_cast<vtkIdType*>(callData);
+    if (itemIdPtr)
+      {
+      itemID = *itemIdPtr;
+      }
+    }
+
+  // Forward `currentItemModified` if the modified item or one of
+  // its children was selected, to adequately update other widgets
+  // that use that modified item such as qMRMLSubjectHierarchyComboBox
+  vtkIdType selectedId = this->firstSelectedSubjectHierarchyItemInBranch(itemID);
+  if (selectedId != vtkMRMLSubjectHierarchyNode::INVALID_ITEM_ID)
+    {
+    emit currentItemModified(selectedId);
+    }
 }
 
 //-----------------------------------------------------------------------------
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
@@ -415,7 +415,7 @@ void qMRMLSubjectHierarchyTreeView::setCurrentItems(QList<vtkIdType> items)
     return;
     }
 
-  this->selectionModel()->clearSelection();
+  this->clearSelection();
 
   foreach (long itemID, items)
     {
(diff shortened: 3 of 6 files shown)


agirault:
Related to [Disable selections of parent items when using attribute filter](52419ff):
1. In [`onSelectionChanged`](https://github.com/Slicer/Slicer/pull/813/commits/52419ffd9082a037a1381eafb1ef9bc5d737f61d#diff-784aa1951ed8cd51bfe8a1bb9fe2ee5bR848) of `qMRMLSubjectHierarchyTreeView`, I catch the items that should not have been selected (to do this, I had to make [`filterAcceptsItem`](https://github.com/Slicer/Slicer/pull/813/commits/52419ffd9082a037a1381eafb1ef9bc5d737f61d#diff-43b6ab175e21e11004c1b5f3760efd7dR97) in `qMRMLSortFilterSubjectHierarchyProxyModel` public, and add a parameter `acceptIfAnyChildIsAccepted`, defaulting to true to keep the previous behavior, but set to false in this case). Then, if nothing is selected, I select the scene item if nothing was blocked (previous default behavior), but if something was blocked, I think we should reselect the deselected items: for example, if I had a study selected, and then select the patient, the popover menu should not close and the study should be re-selected. Right now I do not yet know how to do this using the `QItemSelection deselected`).

2. In [`qMRMLSubjectHierarchyTreeView::mousePressEvent`](https://github.com/Slicer/Slicer/pull/813/commits/52419ffd9082a037a1381eafb1ef9bc5d737f61d#diff-784aa1951ed8cd51bfe8a1bb9fe2ee5bR779) I only forward the event to the base class `QTreeView` if `filterAcceptsItem` (with `acceptIfAnyChildIsAccepted=false`). This avoid the item from even getting selected when pressed, but due to this, the branch control ([-] or [+] on the left of the item to contract/expand the branch) is not working anymore. Could investigate the `Qt::ItemIsSelectable` flag instead?

agirault:
Related to [Disable selections of parent items when using attribute filter](52419ff):

In `onSelectionChanged` of [`qMRMLSubjectHierarchyTreeView`]((https://github.com/Slicer/Slicer/pull/813/commits/02df4a6baf86b7c6c0bbc6627df34b2696093fbb#diff-784aa1951ed8cd51bfe8a1bb9fe2ee5b)), I catch the items that should not have been selected (to do this, I had to make `filterAcceptsItem` in [`qMRMLSortFilterSubjectHierarchyProxyModel`](https://github.com/Slicer/Slicer/pull/813/files#diff-43b6ab175e21e11004c1b5f3760efd7d) public, and add a parameter `acceptIfAnyChildIsAccepted`, defaulting to true to keep the previous behavior, but set to false in this case). Then, if nothing is selected, I select the scene item if nothing was blocked (previous default behavior), but if something was blocked, I think we should reselect the deselected items: for example, if I had a study selected, and then select the patient, the popover menu should not close and the study should be re-selected. Right now I do not yet know how to do this using the `QItemSelection deselected`).

I thought about only forwarding the event to the base class `QTreeView` in [`qMRMLSubjectHierarchyTreeView::mousePressEvent`](https://github.com/Slicer/Slicer/blob/cb1c7208464d13fb337659dadc62af7ee6fc25f4/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx#L776) if `filterAcceptsItem` (with `acceptIfAnyChildIsAccepted=false`), to avoid the item from even getting selected when pressed, but due to this, the branch control ([-] or [+] on the left of the item to contract/expand the branch) would not be working anymore. Could investigate the `Qt::ItemIsSelectable` flag instead?

cpinter:
After a chat with @agirault we have a better plan for the "Disable selections of parent items when using attribute filter" part. The setXFilter methods would call a function called updateItemSelectableFlags that goes through each item and sets the selectable flag to false if the filter rejects it (ignoring the parents, as it is already in this PR). To make sure that the flags are correct after a full update, the model needs to be connected to the tree view (in TreeViewPrivate::init) like this
QObject::connect( this->Model, SIGNAL(subjectHierarchyUpdated()), q, SLOT(updateItemSelectableFlags ()) );

This way the user won't have the chance to select items not accepted by the filter in the first place.

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
```diff
@@ -603,6 +606,37 @@ void qMRMLSubjectHierarchyTreeView::updateRootItem()
...
+  qMRMLSubjectHierarchyModel* sceneModel = qobject_cast<qMRMLSubjectHierarchyModel*>(this->model());
+
+  // Collect selected subject hierarchy items
+  std::vector<vtkIdType> childItemIDs;
+  vtkIdType sceneId = d->SubjectHierarchyNode->GetSceneItemID();
+  d->SubjectHierarchyNode->GetItemChildren(sceneId, childItemIDs, true);
+  foreach(vtkIdType itemID, childItemIDs)
+    {
+    bool isSelectable = this->sortFilterProxyModel()->filterAcceptsItem(itemID, false);
+    QStandardItem* item = sceneModel->itemFromSubjectHierarchyItem(itemID); // or proxy model?
```
agirault: - it seems that the `QStandardItems` are shared between treeviews,
so changing their selection impacts all the other treeviews
- is it ok to use `itemFromSubjectHierarchyItem` from the scenemodel?
Should there be a mapping with the sortFilterProxyModel? The latter
however does not have a `itemFromSubjectHierarchyItem` method.

cpinter: * Cannot be possible. QStandardItems are the view items that are displayed in each treeview separately. Please read http://doc.qt.io/qt-4.8/model-view-programming.html and http://doc.qt.io/qt-4.8/modelview.html
* Using that function from the scene model when there is a proxy filter will probably make the application crash. Use the proxy model's converters instead.

cpinter: What is strange that the tree views create their own model and proxymodel in qMRMLSubjectHierarchyTreeViewPrivate::init. I don't see how it is possible that the filters are applied to all of the views. Also I cannot reproduce this behavior, the filters are correctly applied on the separate widgets individually

dzenanz:
@agirault This does not seem to be working correctly. We can talk on Monday if you need me to show what is broken.

cpinter:
@dzenanz This is WIP. Please wait a bit until @agirault implements what we have discussed and I explained above.

dzenanz:
He requested a review, that's why I tested this - even though it looked like it was not finished. I could not contribute constructively towards code changes, so I tried compiling and running it.

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
agirault: > Use the proxy model's converters instead.

What do you mean by converter here? Are you saying I should implement `itemFromSubjectHierarchyItem` in the proxy model class, and properly use the mapping within it?

agirault:
@dzenanz, thanks. I do not expect it to work yet, the commit and pr are both marked WIP. I added you as a reviewer along with Csaba so you could give your inputs if you had any, given that this was for your project.

I'll work with Csaba's help to put this together properly, and I'll keep you posted then. 

Thank you @cpinter

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
cpinter: If it's not there yet then yes please. You only need to call the same mapping function as the others and call the scene model's converter.

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSortFilterSubjectHierarchyProxyModel.cxx
```diff
@@ -198,6 +198,14 @@ QModelIndex qMRMLSortFilterSubjectHierarchyProxyModel::indexFromSubjectHierarchy
   return this->mapFromSource(sceneModel->indexFromSubjectHierarchyItem(itemID, column));
 }
 
+//-----------------------------------------------------------------------------
+QStandardItem* qMRMLSortFilterSubjectHierarchyProxyModel::itemFromSubjectHierarchyItem(vtkIdType itemId, int column)const
+{
+  qMRMLSubjectHierarchyModel* sceneModel = qobject_cast<qMRMLSubjectHierarchyModel*>(this->sourceModel());
+  QModelIndex index = this->mapFromSource(sceneModel->indexFromSubjectHierarchyItem(itemId, column));
+  return sceneModel->itemFromIndex(index);
```
agirault: @cpinter is that what you were envisioning? This currently returns a null pointer in my use case for every item [here](https://github.com/Slicer/Slicer/pull/813/files#diff-784aa1951ed8cd51bfe8a1bb9fe2ee5bR628), so i'll keep investigating.

cpinter: First of all, please don't reimplement anything that is there and works, i.e.
https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSortFilterSubjectHierarchyProxyModel.cxx#L195

About the actual issue, I'm not sure if using the scene model's itemFromIndex function on an index coming form the proxy model would work. I'll take a deeper look.

cpinter: This call works for the highlights
https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx#L1247
I wonder why it didn't work for you. Apparently the scene model can be used to get the items of the columns from SH ID. I'd start from here. Thanks!

agirault:
The issue was with the way I was setting the flag: I needed `|` when adding `selectable`, and `&` to remove it ([here](https://github.com/Slicer/Slicer/pull/813/files#diff-784aa1951ed8cd51bfe8a1bb9fe2ee5bR635)). The initial way I retrieved the model index and item was correct, there was indeed no need for any mapping.

Still putting WIP since the combobox closes when selecting an unselectable item (but does not show it as the current item, so that's a good sign, almost there).

cpinter:
Sounds great, thanks for the update!
Yes, getting the item is apparently fine from the scene model, but the index needs to be retrieved through the proxy model. Sorry about the confusion, it was a long time since I implemented these, and the inner feelings of Qt are quite complex when it comes to view/models!

agirault:
No problem at all, thanks for the directions. Learning some of that too is good for me.

Following up on the combobox closing but empty: even if the selectable flag is set to false, clicking on an unselectable item will trigger the `selectionChanged`, since it seems to deselect the previously selected items (but the list of selected items is empty). Right now, this will result in the scene item being selected in `onSelectionChanged`

agirault:
@cpinter I found a cleaner and more robust way to do this by subclassing [`QAbstractItemModel::flags`](http://doc.qt.io/qt-4.8/qabstractitemmodel.html#flags) which ensure all items in a same row have the proper flags.

As adviced in Qt doc here: http://doc.qt.io/qt-4.8/qabstractitemmodel.html#subclassing
> To enable editing in your model, you must also implement setData(), and reimplement flags() to ensure that `ItemIsEditable` is returned. 

Two things remain up to discussion:
1. Do you think there is any reason to call `Superclass::flags` in [here](https://github.com/Slicer/Slicer/pull/813/files) ?
2. Because of [this line](https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx#L867), the scene is set as the selected item when nothing is selected. Returning directly instead would make my code work (combobox stays open, unselectable row does not get selected), but that could be an issue when we try to unselect an item in a treeview, wouldn't it?

cpinter:
Selected and deselected are unused because it was easier to get the selection and set it directly instead of doing it "incrementally" as selection happens. It could be used as well, but it was not how I implemented the function.

OK, if subclassing works, then I don't have any objections going that way.

> 1. Do you think there is any reason to call Superclass::flags in here ?
Where? The link points to the whole diff

> 2.
Actually I don't know why the scene item ID needs to be the selected one instead of the invalid item ID. In any case, `emit currentItemChanged(newCurrentItemID);` needs to be called, and the highlight needs to be refreshed (cleared). You may try changing it to invalid item ID, but then it needs to be tested extensively. Do you know exactly which line makes the popup close?

agirault:
Well, `emit currentItemChanged(newCurrentItemID);` is what makes the popup close, because it acknowledged a change of selection, after updating it with `
  qSlicerSubjectHierarchyPluginHandler::instance()->setCurrentItems(selectedShItems);` that's why I wanted to return before selecting the scene as the current item and emitting that signal.

cpinter:
It the selection has not changed, that is nothing has been selected or deselected, then it does not need to be emitted. So if an unselectable item has been clicked, then as far as I understand it's not needed to be hit.

agirault:
> It the selection has not changed, that is nothing has been selected or deselected, then it does not need to be emitted. So if an unselectable item has been clicked, then as far as I understand it's not needed to be hit.

Agreed, what I'm afraid of is that `onSelectionChanged` is still called when I click on a non-selectable item, and I'm not sure why, or how to check if a selection has changed intentionally (clearSelection), or not (clicking on an unselectable item).

I'll investigate more.

cpinter:
You said that if an unselectable item was clicked, then the `selected` and `deselected` containers are both empty. Maybe that can be used for deciding this.

agirault:
That was incorrect: `deselected` will have the previous id of the item that was previously selected. 

To summarize, with my change, `selected` is now empty (flag helped not selecting the parent item), but `deselected` is the same (clicking on the parent item will still deselect the previous item). What I could do would be to reselect the item from `deselected`, but then I assume that would break any calls to `clearSelection`.

The best thing I can think of would be to avoid unselecting that initial item when clicking on the parent item, to address the problem at its source, I'm just not sure yet how to do this but I'll look more.

cpinter:
Sounds reasonable, although it's hard to imagine what it will be like when actually using it.

Maybe we're (almost) back where we started: re-adding the selection enclosed in this->blockSignals(true/false); statements when only deselection happens and nothing else is selected.
Please test the regular tree view as well (in Data module) - with single and multi-selection too - to make sure nothing unexpected happens when clicking around. Thanks!

agirault:
>  when only deselection happens and nothing else is selected.

My issue is that I don't think this is a strong enough condition since that it will be valid too when calling `clearSelection`, which happens when programmatically setting items:
- https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx#L418
- https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx#L446

cpinter:
When that condition is fulfilled, you can clear the cache container (SelectedItems) to be safe, but when the selection is made in setCurrentItem, then onSelectionChanged will run as expected and populate the container again.

agirault:
Here is to illustrate the issue I am trying to bring across:



What you can see here, is that - with the change we discussed - the shnodes do not get unselected when clicking outside of an item, since it reselects the deselected items if there are no selected items at the end of the selectionChanged.

That's bad if we want to reset a combobox for example, so to clear any current item (to show again the "Select subject hierarchy item")

lassoan:
I think in a list or tree widget you cannot remove selection in a list by clicking in the empty area of a widget. Have you seen any widget where you could remove item selection like that?

jcfr:
I  do not have the full picture and did not thoroughly read the entire thread , but to programmatically unselect items,  I call `clearSelection`. See [here](https://github.com/commontk/CTK/blob/89071902e2746fca879e4ed8bde51567261eeeae/Libs/Widgets/ctkFileDialog.cpp#L218) for an example.

Depending on the selection mode, it is possible to have no item selected. For example, using the CTRL key it is possible to unselect all tiems. See [QAbstractItemView::SelectionMode](http://doc.qt.io/qt-4.8/qabstractitemview.html#SelectionMode-enum) and [QItemSelectionModel::Clear](http://doc.qt.io/qt-4.8/qitemselectionmodel.html#SelectionFlag-enum).

lassoan:
Yes, unselect with Control key is the method usually available in widgets. Clicking in empty area usually just changes the focus.

cpinter:
@agirault A combobox always have to ahve a selected item. This is why the MRML node comboboxes have a None option.

agirault:
@jcfr @lassoan thanks a lot for your input.

In the current master, Tree view DOES unselect its selected items when pressing something else than an item:


To be specific, what it does is catching a `deselect` in `onSelectionChanged` [here](https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx#L837), then selects the scene item id to ensure something is selected, [here](https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx#L867-L874), then emits a current item changed signal for INVALID_ITEM_ID [here](https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx#L885-L890).

In this PR, it used to do the same when tapping on a parent item (which was disabled for selection because it doesn't pass the filters). You can see on the following gif that studies are selectable and that selecting patients (disabled for selection) will select the scene and emit a change, therefore closing the treeview popup:


It defeats the purpose of maintaining the previous selection when clicking on something that we want to disable for selection. This is what I'd want to achieve:



However, doing this implies reselecting all deselected items when no selected items are shown, which results in the first gif I shared of the data module, where items can never be deselected:



... which also means that we would not be able to reset a subject hierarchy combobox to the message "Select subject hierarchy item" (could do so by setting an invalid item id), which I need to do in my own use case: when I change a study in the top combobox (`Study`), I want the comboboxes below (`Image1` and `Image21) to set their root item to that study, and to clear their previous selection, which can not happen anymore. You can see that in the third gif of this comment, `Image2` combobox does not reset to "Select subject hierarchy item" when I select a different study in the `Study` combobox.

cpinter:
> which also means that we would not be able to reset a subject hierarchy combobox to the message "Select subject hierarchy item"

This is why I said above that comboboxes do not support "none" selection. In Slicer when it is desirable it is achieved by adding a None item.
In your case you want a selection to determine another combobox's root item. I assume you set the root item in a slot that handles the currentItemChanged signal of the study combobox. If the new study does not contain the previously selected volume of the Image combobox, then the text it shows does not change? If that's the case I think it isn't that the selection didn't change, but simply that the text shown by the combobox didn't change, which is then a bug in this function
https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.cxx#L340

Furthermore, if handling the selection should be different in the tree view and the combobox, then the combobox could implement its own onSelectionChanged function, and connect it to the tree view's signal instead of the tree view's slot.

agirault:
I updated, cleaned up and squashes the commits. Now selecting a disabled item will deselect the previously selected item (like default), but the popup containing the treeview will not hide, to encourage the user to select a selectable item.

This allows to clear a combobox (which will set the scene as the current item, and emit an item changed signal to an invalid id to reset the combobox).

Also added a fix to ensure that the combobox text would get updated when a subject hierarchy item would get modified.

I could still work on one thing: it seems that a treeview with a root item different than the scene (using `setRootItem`) and with filters filtering out all items (using `setAttributeFilter`) will not display the proper items (which should be no items, or the root item not selectable if `showRootItem`). Instead, it displays the `Scene` item, and the app will crash when selecting it.

cpinter:
@agirault Thank you so much for working on this! I'll build your branch and try it out.

The root item issue won't be easy to fix I think; I struggled with the root item a lot already. If this is something that is present in the master, then I suggest adding a Mantis issue about it for now, and you and/or I can look at it after this PR is merged.

agirault:
We're going to need that fixed for our work,since it makes Slicer crash when there is a filter and a different root item than the scene, so I don't mind spending more time on this PR to ship all of it together.

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
```diff
@@ -1298,6 +1301,63 @@ bool qMRMLSubjectHierarchyTreeView::multiSelection()
 }
 
 //-----------------------------------------------------------------------------
+vtkIdType qMRMLSubjectHierarchyTreeView::firstSelectedInHierarchyFrom(vtkIdType itemID)
```
cpinter: I think the method name is not descriptive enough, it doesn't specify first of what is looked for. Also, the convention in SH is that we call "sub-hierarchies" branches. I'd suggest calling this firstSelectedSubjectHierarchyItemInBranch

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
```diff
@@ -1298,6 +1301,63 @@ bool qMRMLSubjectHierarchyTreeView::multiSelection()
...
+{
+  // Check self
```
cpinter: Again, not descriptive enough. I'd say
Check if item itself is selected

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
```diff
@@ -1298,6 +1301,63 @@ bool qMRMLSubjectHierarchyTreeView::multiSelection()
...
+  Q_D(qMRMLSubjectHierarchyTreeView);
+  if (d->SelectedItems.contains(itemID))
+    {
+    return itemID;
+    }
+
+  // Check children
```
cpinter: Let's make the code more readble. If you add a comment, make it actually explain something
"Look for item in children, recursively"

cpinter:
> We're going to need that fixed for our work

That's fine, I just say that this PR contains one well-defined set of features, and considering that the root item issue will probably be hard to fix, I'd make it a logically separate piece of work. Please let's get this PR done first, then we can move on to that bug.

agirault:
@cpinter would not showing a popup when the number of items in the treeview is nul be an acceptable workaround?

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
agirault: great name :+1: I was not happy about the current one.

cpinter:
I'm afraid no. If nothing shows up, the user will think the combobox doesn't work. It needs to show something, and not crash. I'm investigating the crash right now, at the least want to see what makes it crash. It's also not very nice that it shows the scene. I think the best would be to show a popup saying "No items matching the filter" if there is a filter and that's why it's empty. It should be easy to add to the popup in qMRMLSubjectHierarchyComboBox::showPopup, but it still won't fix the crash in the tree view I think. Can we pleease add a bug for that and worry about it once the PR is integrated?

cpinter:
(Can be a minute after, I'd just like to see it separated. Doing everything in one go will make it harder to scan the history later and seeing what has what effect)

cpinter:
The crash comes from here
https://github.com/agirault/Slicer/blob/improve-subjecthierarchy-combobox/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSortFilterSubjectHierarchyProxyModel.cxx#L421

agirault:
Since the crash comes from my changes (flags) I should probably fix it in this pr then? And keep the issue you are talking about (hiding Scene) for a next pr ?

I rebased the commits accordingly so they should not need to be squashed.

Two other issues I get:

1. The popup view is wider than the combobox (see below, on the right side). It seems to be due to [these two lines](
https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.cxx#L288-L289). Any reason they are here? Could I remove them?

screen shot 2017-10-18 at 11 04 16 am

screen shot 2017-10-18 at 1 01 28 pm

2. The combobox does not seem to stretch all the way horizontally (this is in a form layout):

screen shot 2017-10-18 at 10 59 57 am

If I replace any of them by a MRMLNode combobox instead, it will stretch properly without changing anything else:

screen shot 2017-10-18 at 1 19 26 pm

So it seems to be only for the subject hierarchy combo box

cpinter:
> Since the crash comes from my changes (flags) I should probably fix it in this pr then? And keep the issue you are talking about (hiding Scene) for a next pr ?

Yes, please at least add a check if item is not NULL. That is usually a good idea anyway.

> 1. The popup view is wider than the combobox (see below, on the right side). It seems to be due to these two lines. Any reason they are here? Could I remove them?

They seem to be margins, so if you remove them then at least on some platforms some of the tree view content will be clipped. Please keep them as is for now. This is another issue that can be niely addressed differently. Let's not do everything in one PR! It would be like those "salad" laws that contain unrelated stuff just so that a few wobbly ones can pass as well.

> 2. The combobox does not seem to stretch all the way horizontally

Would it be possible to keep these separate issues in mind and just do whatever was the purpose of this PR? I promise they will be integrated as well :)

agirault:
>  It would be like those "salad" laws that contain unrelated stuff just so that a few wobbly ones can pass as well.

Well, if we keep all the commits self-contained and separated like I have, I'm not sure why it's an issue to put all those changes regarding the subject hierarchy combobox in the same pr, but I'll do that.

> Yes, please at least add a check if item is not NULL. That is usually a good idea anyway.

Done, ready for final review/merge then :+1:

cpinter:
> Well, if we keep all the commits self-contained and separated like I have, I'm not sure why it's an issue to put all those changes regarding the subject hierarchy combobox in the same pr, but I'll do that.

Existing bugs should be recorded and addressed in Mantis, that's why we have an issue tracker.

That said, I don't think that it makes sense to keep all your commits separate, as some of them didn't even yield a functional revision, and were just later reversed and changed. Please squash at least those commits together that are incremental changes for the same thing (for example fixing the crash: it doesn't make sense to be a separate commit if the crash was caused by another commit within the same PR).

agirault:
> for example fixing the crash: it doesn't make sense to be a separate commit if the crash was caused by another commit within the same PR

Yes, I did that when I fixed it: I did an interactive rebase and put it in the `Disable selections of parent items with filters` commit (c8b9ca0)

Let me know if I need to squash more commits together.

cpinter:
OK thanks!

agirault:
@cpinter @jcfr looks like the doc was properly generated but not published. Any idea why? Is it from my pr? If not can this be merged?

jcfr:
>  Any idea why? 

yes - too much data pushed to https://github.com/Slicer/preview.apidocs.slicer.org/tree/gh-pages

I will setup an automatic reset (every Sunday)

> Is it from my pr? 

no 

> If not can this be merged?

Yes - Development of Slicer 4.9.0 is now open.

agirault:
@cpinter are you happy with the current state of the pr? Do you want me to merge or should I?

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.cxx
```diff
@@ -64,6 +65,7 @@ class qMRMLSubjectHierarchyComboBoxPrivate
 qMRMLSubjectHierarchyComboBoxPrivate::qMRMLSubjectHierarchyComboBoxPrivate(qMRMLSubjectHierarchyComboBox& object)
   : q_ptr(&object)
   , MaximumNumberOfShownItems(20)
+  , AlignPopupVertically(true)
```
cpinter: Does this mean that the default is now changed?

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.cxx
```diff
@@ -155,6 +159,13 @@ void qMRMLSubjectHierarchyComboBox::setMRMLScene(vtkMRMLScene* scene)
 }
 
 //------------------------------------------------------------------------------
+void qMRMLSubjectHierarchyComboBox::clearCurrentItem()
```
cpinter: Please call this function also clearSelection. I don't see a reason for this to have a name different from the one in the tree view when we only call that.

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
```diff
@@ -415,7 +418,7 @@ void qMRMLSubjectHierarchyTreeView::setCurrentItems(QList<vtkIdType> items)
     return;
     }
 
-  this->selectionModel()->clearSelection();
+  this->clearSelection();
```
cpinter: Can you please add a comment why calling this is necessary?

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
```diff
@@ -443,7 +446,7 @@ void qMRMLSubjectHierarchyTreeView::setCurrentItems(vtkIdList* items)
...
```
cpinter: Can you please add a comment why calling this is necessary?

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.h
```diff
@@ -176,6 +177,9 @@ protected slots:
   /// (e.g. after tree was updated in the model from the subject hierarchy)
   virtual void updateRootItem();
 
+  /// Called when a subject hierarchy item is modified
```
cpinter: This is evident. Please explain what it does with the modified item.

cpinter:
I added some minor comments. After addressing those you can integrate if you have the privileges to do that.

Then we can work on the bugs you caught. It's good to see what change has what effect, even better if there is a nightly between the PR and the bugfix so that we can use those later if necessary.

Thanks again!

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.cxx
agirault: No, the default would align it vertically (selected item overlayed on top of combobox)

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.cxx
agirault: Because treeviews, models etc. deal with **selected** items/indexes : http://doc.qt.io/qt-4.8/qtreeview.html#selectedIndexes

While a combobox only deals with **current** item/index, not "selections" : 
http://doc.qt.io/qt-4.8/qcombobox.html#currentIndex-prop

Thoughts?

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
agirault: I do not know what motivates the clear before the change of items, this was here prior to my changes, I just replaced the `model::clearSelection()` by the wrapped version of that function in that class.

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
agirault: I do not know what motivates the clear before the change of items, this was here prior to my changes, I just replaced the `model::clearSelection()` by the wrapped version of that function in that class.

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.cxx
cpinter: OK, let's see which one seems better in the long run. Maybe the way you implemented it is more desirable to the users. I don't really have a preference right now, just wanted to know.

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.cxx
cpinter: This function right now is the most basic wrapper imaginable for TreeView::clearSelection, as the only thing it does is calls that. Having a different name for such a wrapper is I think misleading.

Also I think the name clearCurrentItem is not straightforward, as the first thing that occurs to me is that this function clears the _content_ of the current item.

I really think clearSelection is better in every way.

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
cpinter: Oh, OK then. Thanks for the explanation! I just thought based on the GitHub diff as I saw it that this is a newly added line, and in that case there must have been a reason for adding it. As it's not a new line, I don't expect you to figure out the reason.

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
cpinter: [See above] Thanks!

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.cxx
agirault: sounds good, changing that.

agirault:
Thank you for the reviews.

Integrated in:
- [r26495](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26495)
- [r26496](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26496)
- [r26497](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26497)
- [r26498](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26498)
- [r26499](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26499)

cpinter:
Great! Thanks for working on this and dealing with my nitpicks!

PR #814, opened 2017-10-13
ENH: ExtensionWizard: Support layout with main CMakeLists in subdirectory

jcfr:
This allows to contribute description file update for the CarreraSlice
extension.

In that particular case, the top-level CMakeLists.txt doesn't contain
all variables of the form EXTENSION_*, instead they are found in
src/CMakeLists.txt.


jcfr:
Integrated in [r26461](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26461)

PR #816, opened 2017-10-14
ENH: Added qSlicerMarkupsPlaceWidget to designer plugin list

tavaughan:


jcfr:
Thanks for the patch 👍 

Integrated in [r26457](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26457)

PR #820, opened 2017-10-16
COMP: Qt5: Require Multimedia to support extension depending on QSound

jcfr:


jcfr:
Integrated in [r26482](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26482)

PR #823, opened 2017-10-20
BUG: Fix crash ThresholdSlider in segmentEditor when changing volume

Punzo:
Hi @cpinter another fix in the Segmentation editor relative to Qt5.

The crash happens when the minimum of the scalar range is greater then the current maximum of the ctkRangeWidget (at least in linux).

The fix is simply to use ctkRangeWidget::setRange.
P.S.: in the python wrapping setRange is not available.

### DIFF
--- a/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorThresholdEffect.py
+++ b/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorThresholdEffect.py
@@ -109,7 +109,7 @@ class SegmentEditorThresholdEffect(AbstractScriptedSegmentEditorEffect):
     masterImageData = self.scriptedEffect.masterVolumeImageData()
     if masterImageData:
       lo, hi = masterImageData.GetScalarRange()
-      self.thresholdSlider.minimum, self.thresholdSlider.maximum = lo, hi
+      self.thresholdSlider.setRange(lo, hi)
       self.thresholdSlider.singleStep = (hi - lo) / 1000.
       if (self.scriptedEffect.doubleParameter("MinimumThreshold") == self.scriptedEffect.doubleParameter("MaximumThreshold")):
         # has not been initialized yet
--- a/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
+++ b/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
@@ -1380,8 +1380,7 @@ void qMRMLSegmentEditorWidget::onMasterVolumeImageDataModified()
     {
     double range[2] = { 0.0, 0.0 };
     d->MasterVolumeNode->GetImageData()->GetScalarRange(range);
-    d->MasterVolumeIntensityMaskRangeWidget->setMinimum(range[0]);
-    d->MasterVolumeIntensityMaskRangeWidget->setMaximum(range[1]);
+    d->MasterVolumeIntensityMaskRangeWidget->setRange(range[0], range[1]);
     d->MasterVolumeIntensityMaskRangeWidget->setEnabled(true);
     }
   else
--- a/SuperBuild/External_CTK.cmake
+++ b/SuperBuild/External_CTK.cmake
@@ -82,7 +82,7 @@ if(NOT DEFINED CTK_DIR AND NOT ${CMAKE_PROJECT_NAME}_USE_SYSTEM_${proj})
 
   ExternalProject_SetIfNotDefined(
     ${CMAKE_PROJECT_NAME}_${proj}_GIT_TAG
-    "134532b5eda5323d81a0c848b31a21ed19dcc732"
+    "4911acd884b6ccf897fcd116df646b7fd7f7b2bc"
     QUIET
     )
 


Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorThresholdEffect.py
```diff
@@ -109,7 +110,13 @@ def masterVolumeNodeChanged(self):
     masterImageData = self.scriptedEffect.masterVolumeImageData()
     if masterImageData:
       lo, hi = masterImageData.GetScalarRange()
-      self.thresholdSlider.minimum, self.thresholdSlider.maximum = lo, hi
+      if ((lo - self.thresholdSlider.maximum) > math.pow(10, -self.thresholdSlider.decimals)) and ((hi - lo) > math.pow(10, -self.thresholdSlider.decimals)):
+        self.thresholdSlider.maximum = hi
+        self.thresholdSlider.minimum = lo
+      else :
+        self.thresholdSlider.minimum = lo
+        self.thresholdSlider.maximum = hi
```
cpinter: Exactly the same code is executed in the two branches, only in reversed order. Does it change anything?

Punzo: yeah the point is exactly that.  In the first case the minimum is major than the current self.thresholdSlider.maximum and therefore you have to set first the maximum with hi and then you can set the minimum (with the other order will crush).  In the else case is viceversa.

Ideally the best will be to have the setRange method also in Python that handle this automatically...
However, I don't understand why the setRange method is not available in Python.

P.S..: this is a bug in Qt5, I didn't experienced in Qt4.

jcfr:
Thanks for helping moving forward with Qt5

Nitpick: `crush` -> `crash`

Otherwise, may be worth looking at https://github.com/commontk/CTK/issues/570

Punzo:
yeap sorry, always messing around with crash (: 

The bug is similar to the the one described in commontk/CTK#570, however it is not the same one. I mean using setRange already fix this particular issue. In general will be good to fix also the singlestep bug, but the segmentation widgets do not set it.

Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorThresholdEffect.py
cpinter: Thanks for the explanation! In that case I think this change is fine. Can you, however, add a comment that nicely explains this? Others looking at the code will have the exact same question as I did, and someone may just change it back silently, saying that "it does the same anyway". Thanks!

lassoan: Only Q_INVOKABLE method, property get/set methods, and slots are python-wrapped. setRange(double min, double max) is none of them. To make it available from Python, you just have to add Q_INVOKABLE.

@Punzo Could you try if it all works well if you call setRange from Python?

Punzo: @lassoan  thanks for the explanation, I didn't know. I checked and if using setRange in python will work as well. I changed the commit and I'll make a pull request to CTK repository as well.

Punzo:
Please merge this only after https://github.com/commontk/CTK/pull/760

lassoan:
Merged https://github.com/commontk/CTK/pull/760. Please update your pull request with the new CTK hash.

Punzo:
@lassoan  I don't see the commit in CTK https://github.com/commontk/CTK/commits/master

jcfr:
@Punzo Et voila,  I just merged it.

jcfr:
@Punzo  Could you rebase your topic on top of current master and squash them into one ? I had issue cherry-picking the sequences of commits.

Thanks

Punzo:
@jcfr Ok no problem, but I'll do tomorrow.

agirault:
@jcfr @lassoan : we know get those errors in circleci build due to CTK:
https://circleci.com/gh/Slicer/Slicer/1910?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

```
/usr/src/Slicer/Libs/MRML/Widgets/qMRMLSceneColorTableModel.cxx
/usr/src/Slicer/Libs/MRML/Widgets/qMRMLSceneColorTableModel.cxx: In member function 'void qMRMLSceneColorTableModelPrivate::ColorGradient::updatePixmap(vtkScalarsToColors*)':
/usr/src/Slicer/Libs/MRML/Widgets/qMRMLSceneColorTableModel.cxx:67:37: error: 'scalarsToColorsImage' is not a member of 'ctk'
   this->Pixmap = QPixmap::fromImage(ctk::scalarsToColorsImage( scalarsToColors, this->Pixmap.size() ));
```

Punzo:
@jcfr done!

Punzo:
@jcfr  I removed the includes as well, now should be good to go.

cpinter:
Integrated in https://github.com/Slicer/Slicer/commit/4cc81bae9b7450b30597382b060f2775c18885ea

Thanks Davide for working on this :)

Punzo:
Thanks for integrating!

cpinter:
I get the following with a Qt4 build:

  File "C:\d\S4R\Slicer-build\lib\Slicer-4.9\qt-scripted-modules\SegmentEditorEffects\SegmentEditorThresholdEffect.py", line 112, in masterVolumeNodeChanged
    self.thresholdSlider.setRange(lo, hi)
AttributeError: ctkRangeWidget has no attribute named 'setRange'

It seems it is related to this PR. @Punzo can you please look at this? Thanks!

cpinter:
(Maybe I just need to rebuild CTK. Let me check first)

cpinter:
Nevermind, my hunch was right. The superbuild didn't build CTK after the update. It's all good.

jcfr:
> The superbuild didn't build CTK after the update.

That is an issue.

Did you have locally modified files? If it is the case, the build system will not update.

lassoan:
I often have to rebuild CTK manually when there is some change (but maybe only if there are local modifications).

cpinter:
@jcfr No I didn't have any local modifications. For me this happens quite frequently, not only with CTK, but sometimes with VTK as well. Building of updated dependencies is not very reliable on Windows. I didn't consider this an issue because the factory builds from scratch, and we can deal with it once or twice a week as it happens.

cpinter:
(Well, this is an issue, but there are many things much higher on the priority list)

jcfr:
@cpinter After updating to [r26545](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26545), it would be interesting to rebuild and redirect into a log file.

```
cd Slicer-Superbuild
cmake --build . --config Release > build-log.txt 2>&1
```

Then, we could check if the CTK inner build kicked in.

cpinter:
The cmake command didn't build CTK, but it seems it didn't do the build properly
[build-log.txt](https://github.com/Slicer/Slicer/files/1441737/build-log.txt)
When I ran the superbuild my usual way this build log was created, it doesn't show CTK build
[build-log.txt](https://github.com/Slicer/Slicer/files/1442165/build-log.txt)
The CTK build log also doesn't show that it was actually built
[CTK.log](https://github.com/Slicer/Slicer/files/1442166/CTK.log)

PR #826, opened 2017-10-23
Change maximum number of frames to capture

hherhold:
Added edit box in "Advanced" region to change maximum number of frames to capture.

Default is still 600, so if user doesn't do anything, behavior is unchanged.


jcfr:
Thanks for the contribution 👍 

After few comments, we should be able to integrate.

Modules/Scripted/ScreenCapture/ScreenCapture.py
```diff
@@ -264,6 +264,14 @@ def setup(self):
     self.fileNamePatternWidget.text = "image_%05d.png"
     advancedFormLayout.addRow("Image file name pattern:", self.fileNamePatternWidget)
 
+    self.maxFramesValidator = qt.QIntValidator(1, 9999)
+    self.maxFramesWidget = qt.QLineEdit()
+    self.maxFramesWidget.setToolTip(
+      "Maximum number of frames to be captured. Default: 600, maximum: 9999.")
+    self.maxFramesWidget.setValidator(self.maxFramesValidator)
+    self.maxFramesWidget.text = "600"
```
jcfr: I suggest to use `qt.QSpinBox` instead

Modules/Scripted/ScreenCapture/ScreenCapture.py
```diff
@@ -304,10 +312,20 @@ def setup(self):
     self.videoExportCheckBox.connect('toggled(bool)', self.videoFormatWidget, 'setEnabled(bool)')
     self.videoFormatWidget.connect("currentIndexChanged(int)", self.updateVideoFormat)
     self.singleStepButton.connect('toggled(bool)', self.numberOfStepsSliderWidget, 'setDisabled(bool)')
+    self.maxFramesWidget.connect('textChanged(QString)', self.maxFramesChanged)
 
     self.updateVideoFormat(0)
     self.updateViewOptions()
 
+  def maxFramesChanged(self):
```
jcfr: With QSpinBox you would then be sure to always have an `int` as parameter of the `valueChanged` signal

Modules/Scripted/ScreenCapture/ScreenCapture.py
```diff
@@ -304,10 +312,20 @@ def setup(self):
...
+    maxFramesText = self.maxFramesWidget.text
+    # An exception is raised if maxFramesText is empty. Catch it and set it to 1 instead.
+    try:
+      maxFrames = int(maxFramesText)
+    except Exception as e:
```
jcfr: Generally speaking, it is recommended to catch more specific exception (in that case, that would be `ValueError`). That said, by using a spinbox, you wouldn't have to worry about this

hherhold: OK, I redid it with a QSpinBox. For some reason I thought using a QLineEdit would be easier, but you're absolutely right - the QSpinBox was simpler and shorter.

What's the best way to proceed - submit a new pull request?

jcfr:
Thanks for following up with this 👍 

> What's the best way to proceed - submit a new pull request?

You can for example push force on your existing topic . That said, if it is easier, publish a new topic works well too.

hherhold:
Redone using QSpinBox. Opened new pull request (827).

PR #830, opened 2017-10-27
COMP: Fix NUMPY rebuild error ensuring patch are re-applied if needed

jcfr:
If DOWNLOAD_STAMP is modified after the patch has been applied, this mean
that the source have been re-downloaded and that the patch must be re-applied.

Fixes https://issues.slicer.org/view.php?id=4452

Reported-by: Andras Lasso <lasso@queensu.ca>
Reported-by: Steve Pieper <pieper@bwh.harvard.edu>


jcfr:
Integrated in [r26517](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26517)

PR #831, opened 2017-10-31
RFC: BUG: slicer.util.exit should use first-called exit status

ihnorton:
Fixes https://issues.slicer.org/view.php?id=4470

Ref where sys.exit from Python was required to correctly return status:
  https://github.com/pnlbwh/ukftractography/commit/bcd856906d1dbd5eeee608937e62aa3f156c0463

Explanation: QApplication::exit() terminates the event loop and then returns to caller -- so the given PythonQt `eval` runs to completion and multiple calls in a single block/file will only honor the last-called exit status.

### DIFF
--- a/Applications/SlicerApp/Testing/Python/CMakeLists.txt
+++ b/Applications/SlicerApp/Testing/Python/CMakeLists.txt
@@ -31,9 +31,16 @@ add_test(
   --testing --disable-modules --no-main-window
   --python-code "exit(slicer.util.EXIT_FAILURE)"
   )
+add_test(
+  NAME py_nomainwindow_SlicerPythonCodeDoubleExitFailureTest
+  COMMAND ${Slicer_LAUNCH_COMMAND} $<TARGET_FILE:${APP_TARGET_NAME}>
+  --disable-modules --no-main-window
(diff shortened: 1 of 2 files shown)


ihnorton:
This is a simple Python-only fix since the fundamental issue is mismatch between QApplication::exit() semantic and PythonQt.

cc @pieper

pieper:
LGTM 👍

jcfr:
Thanks @ihnorton  

Could you extend the test suite accordingly ?

https://github.com/Slicer/Slicer/blob/e77630416ec79c7cb6a8145c6286b39a25ebaec1/Applications/SlicerApp/Testing/Python/CMakeLists.txt#L9-L61

ihnorton:
Ok, test added.

I looked at little bit at doing this more cleanly by interrupting python interpretation, but I don't see a simple way to do that in PythonQt. Related:

- https://sourceforge.net/p/pythonqt/discussion/631393/thread/8d49df35/
- https://sourceforge.net/p/pythonqt/discussion/631393/thread/6a851191/

ihnorton:
@lassoan @jcfr should we integrate this or close it? I think there was some side discussion in a hangout that led to my last comment, and this fizzled out. But this PR does fix the problem.

ihnorton:
Nevermind, I see the comment on the issue.

PR #833, opened 2017-11-03
ENH: add MeanDiffusivity calculation to DTMathematics

ihnorton:
Note this is tested in https://github.com/SlicerDMRI/SlicerDMRI/pull/86

(also fixes a bug where the COLOR_ORIENTATION_[middle, min]_EIGENVEC options were not available, because the macro-setter was clamped to an older enum range)

### DIFF
--- a/Libs/vtkTeem/Testing/vtkDiffusionTensorMathematicsTest1.cxx
+++ b/Libs/vtkTeem/Testing/vtkDiffusionTensorMathematicsTest1.cxx
@@ -96,7 +96,7 @@ int vtkDiffusionTensorMathematicsTest1(int vtkNotUsed(argc), char* vtkNotUsed(ar
   filter->SetMaskLabelValue(0);  // mask all the labels different from 0
   filter->SetMaskWithScalars(1); // turn on masking
   for (int i = vtkDiffusionTensorMathematics::VTK_TENS_TRACE;
-       i <=vtkDiffusionTensorMathematics::VTK_TENS_PERPENDICULAR_DIFFUSIVITY;
+       i <=vtkDiffusionTensorMathematics::VTK_TENS_MEAN_DIFFUSIVITY;
        ++i)
     {
     filter->SetOperation(i);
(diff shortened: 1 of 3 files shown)


pieper:
Should we finally also add ADC (defined as Trace divided by 3)?  It's very common clinical scans and I know people have been confused that it's not there.

fedorov:
> ADC (defined as Trace divided by 3)

@pieper where is this definition of ADC coming from?

The ones I've seen (and that we implemented in [DWModeling](https://www.slicer.org/wiki/Documentation/Nightly/Modules/DWModeling)) are the following:

image

Source: Toivonen J., Merisaari H., Pesola M., Taimen P., Boström PJ., Pahikkala T., Aronen HJ., Jambor I. 2015. Mathematical models for diffusion-weighted imaging of prostate cancer using b values up to 2000 s/mm2: Correlation with Gleason score and repeatability of region of interest analysis. Magnetic resonance in medicine: official journal of the Society of Magnetic Resonance in Medicine / Society of Magnetic Resonance in Medicine 74:1116–1124. DOI: 10.1002/mrm.25482. http://onlinelibrary.wiley.com/doi/10.1002/mrm.25482/full

pieper:
@fedorov we're talking about the same thing - the diffusion tensor already includes an exponential model.

http://mriquestions.com/trace-vs-adc-map.html

"In a prior Q&A we explained that biological tissues are anisotropic. They have multiple diffusion coefficients that vary by direction and are represented by the diffusion tensor, a 3x3 array of numbers. The term "trace" comes from matrix algebra where it means the sum of diagonal elements of such an array. The trace of the diffusion tensor (Dtrace) equals (Dxx + Dyy + Dzz). Using the average value of the trace, (Dxx + Dyy + Dzz)/3, reduces the multi-directional diffusivity at each point into a single number that can be considered a consolidated apparent diffusion coefficient (ADC). Thus the terms average trace and ADC are often used interchangeably."

ihnorton:
@pieper this is trace / 3. We'll add documentation mentioning "ADC" so people can find that if they want it, but see:

http://lists.mristudio.org/pipermail/mristudio-users/2013/002592.html

(and, e.g., the last paragraph here: http://mriquestions.com/apparent-diffusion.html)

ihnorton:
cc @ljod

pieper:
Addressing it in the documentation should be fine.

ihnorton:
merged in http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26553

PR #834, opened 2017-11-04
ENH: Create user-specific temporary directory on unix systems

lassoan:
Slicer temporary directory is owned by the user who started Slicer first and is not writeable. Issue identified by Artem Mamonov.

Fixes https://issues.slicer.org/view.php?id=4340


ihnorton:
Is there a reason to use a persistent name here instead of process-unique? (e.g. [mkdtemp](https://linux.die.net/man/3/mkdtemp) / QTemporaryDir). Is this where Slicer puts the sample data, and we want that to persist?

lassoan:
Yes, exactly. Slicer keeps cached data in temp dir, so we need to preserve it.

lassoan:
Integrated into r26549.

PR #835, opened 2017-11-08
ENH: Added python utilities that download, unzip, import, and load DI…

cpinter:
…COM data

New context called OnlineDICOMData was added in DICOMUtils that does the following:
- Download zip file from URL
- Unzip file to specified directory
- Import the unzipped DICOM files into the database
- Load only just imported files into Slicer, with the option of verifying the selected DICOM plugins and the number of loaded nodes of given types

SubjectHierarchyGenericSelfTest updated to use this new context

### DIFF
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -1091,3 +1091,106 @@ interactor.AddObserver(vtk.vtkCommand.LeftButtonPressEvent, onClick)
   up()
   interactor.SetShiftKey(0)
   interactor.SetControlKey(0)
+
+def downloadFile(url, targetFilePath):
+  """ Download ``url`` to local storage as ``targetFilePath``
+
+  Target file path needs to indicate the file name and extension as well
+  """
+  import os
+  import logging
+  if not os.path.exists(targetFilePath) or os.stat(targetFilePath).st_size == 0:
+    logging.info('Downloading from\n  %s\nas file\n  %s\nIt may take a few minutes...' % (url,targetFilePath))
+    try:
+      import urllib
+      urllib.urlretrieve(url, targetFilePath)
+    except Exception, e:
+      import traceback
+      traceback.print_exc()
+      logging.error('Failed to download file from ' + url)
+      return False
+  else:
+    logging.info('Requested file has been found: ' + targetFilePath)
+  return True
+
+def extractArchive(archiveFilePath, outputDir, expectedNumberOfExtractedFiles=None):
+  """ Extract file ``archiveFilePath`` into folder ``outputDir``.
+
+  Number of expected files unzipped may be specified in ``expectedNumberOfExtractedFiles``.
+  If folder contains the same number of files as expected (if specified), then it will be
+  assumed that unzipping has been successfully done earlier.
+  """
+  import os
+  import logging
+  from slicer import app
+  if not os.path.exists(archiveFilePath):
+    logging.error('Specified file %s does not exist' % (archiveFilePath))
+    return False
+  fileName, fileExtension = os.path.splitext(archiveFilePath)
+  if fileExtension.lower() != '.zip':
+    #TODO: Support other archive types
+    logging.error('Only zip archives are supported now, got ' + fileExtension)
+    return False
+
+  numOfFilesInOutputDir = len(getFilesInDirectory(outputDir, False))
+  if expectedNumberOfExtractedFiles is not None \
+      and numOfFilesInOutputDir == expectedNumberOfExtractedFiles:
+    logging.info('File %s already unzipped into %s' % (archiveFilePath, outputDir))
+    return True
+
+  extractSuccessful = app.applicationLogic().Unzip(archiveFilePath, outputDir)
+  numOfFilesInOutputDirTest = len(getFilesInDirectory(outputDir, False))
+  if extractSuccessful is False or (expectedNumberOfExtractedFiles is not None \
+      and numOfFilesInOutputDirTest != expectedNumberOfExtractedFiles):
+    logging.error('Unzipping %s into %s failed' % (archiveFilePath, outputDir))
+    return False
+  logging.info('Unzipping %s into %s successful' % (archiveFilePath, outputDir))
+  return True
+
+def downloadAndExtractArchive(url, archiveFilePath, outputDir, \
+                              expectedNumberOfExtractedFiles=None, numberOfTrials=3):
+  """ Downloads an archive from ``url`` as ``archiveFilePath``, and extracts it to ``outputDir``.
+
+  This combined function tests the success of the download by the extraction step,
+  and re-downloads if extraction failed.
+  """
+  import os
+  import shutil
+  import logging
+
+  maxNumberOfTrials = numberOfTrials
+
+  def _cleanup():
+    # If there was a failure, delete downloaded file and empty output folder
+    logging.warning('Download and extract failed, removing archive and destination folder and retrying. Attempt #%d...' % (maxNumberOfTrials - numberOfTrials))
+    os.remove(archiveFilePath)
+    shutil.rmtree(outputDir)
+    os.mkdir(outputDir)
+
+  while numberOfTrials:
+    if not downloadFile(url, archiveFilePath):
+      numberOfTrials -= 1
+      _cleanup()
+      continue
+    if not extractArchive(archiveFilePath, outputDir, expectedNumberOfExtractedFiles):
+      numberOfTrials -= 1
+      _cleanup()
+      continue
+    return True
+
+  _cleanup()
+  return False
+
+def getFilesInDirectory(directory, absolutePath=True):
+  """ Collect all files in a directory and its subdirectories in a list
+  """
+  import os
+  allFiles=[]
+  for root, subdirs, files in os.walk(directory):
+    for fileName in files:
+      if absolutePath:
+        fileAbsolutePath = os.path.abspath(os.path.join(root, fileName)).replace('\\','/')
+        allFiles.append(fileAbsolutePath)
+      else:
+        allFiles.append(fileName)
+  return allFiles
--- a/Modules/Scripted/DICOMLib/DICOMUtils.py
+++ b/Modules/Scripted/DICOMLib/DICOMUtils.py
@@ -16,10 +16,37 @@ comment = """
 ######################################################### 
 
 #------------------------------------------------------------------------------
-# Get patient UID by patient name for easy loading of a patient
-def getPatientUIDByPatientName(name):
-  if not hasattr(slicer,'dicomDatabase'):
-    logging.error('DICOM database cannot be accessed!')
+def loadPatientByUID(patientUID):
+  """ Load patient by patient UID from DICOM database
+  """
+  if not hasattr(slicer, 'dicomDatabase') or not hasattr(slicer.modules, 'dicom'):
+    logging.error('DICOM module or database cannot be accessed')
+    return False
+
+  if not patientUID in slicer.dicomDatabase.patients():
+    logging.error('No patient found with DICOM database UID %s' % patientUID)
+    return False
+
+  # Select all series in selected patient
+  studies = slicer.dicomDatabase.studiesForPatient(patientUID)
+  if len(studies) == 0:
+    logging.warning('No studies found in patient with DICOM database UID ' + str(patientUID))
+    return False
+
+  series = [slicer.dicomDatabase.seriesForStudy(study) for study in studies]
+  seriesUIDs = [uid for uidList in series for uid in uidList]
+  if len(seriesUIDs) == 0:
+    logging.warning('No series found in patient with DICOM database UID ' + str(patientUID))
+    return False
+
+  return loadSeriesByUID(seriesUIDs)
+
+#------------------------------------------------------------------------------
+def getDatabasePatientUIDByPatientName(name):
+  """ Get patient UID by patient name for easy loading of a patient
+  """
+  if not hasattr(slicer, 'dicomDatabase'):
+    logging.error('DICOM database cannot be accessed')
     return None
   patients = slicer.dicomDatabase.patients()
   for patientUID in patients:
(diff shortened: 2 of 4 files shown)


Modules/Scripted/DICOMLib/DICOMUtils.py
```diff
@@ -156,3 +156,180 @@ def __enter__(self):
...
+def seriesUIDsForFiles(files):
+  seriesUIDs = set()
+  for file in files:
+    seriesUID = slicer.dicomDatabase.seriesForFile(file)
+    if seriesUID != '':
+      seriesUIDs.add(seriesUID)
+  return seriesUIDs
+
+#------------------------------------------------------------------------------
+class OnlineDICOMData:
```
lassoan: I'm wondering if we could find a more accurate name for this class.

I think usually there is a verb after `with` statement, such as:

    with open("x.txt") as f:
        ...

Also, OnlineDICOMData emphasizes that the data is online (vs offline), which is not that much important (we may enhance the class to be able to work from local files).

The name of the class does not give a hint about what it does (download, unzip, import, load, verify).

Maybe something like "LoadDICOMFilesToDatabase" would be a bit better (not emphasizing online/offline and would tell them ultimately the data is loaded).

Modules/Scripted/DICOMLib/DICOMUtils.py
```diff
@@ -156,3 +156,180 @@ def __enter__(self):
...
+  def __enter__(self):
+    if downloadFile(self.url, self.zipFilePath):
+      if unzipFile(self.zipFilePath, self.dicomDataDir, self.expectedNumberOfUnzippedFiles):
+        import os
+        dicomFiles = slicer.util.getFilesInDirectory(self.dicomDataDir)
+        if importDicom(self.dicomDataDir):
+          seriesUIDs = seriesUIDsForFiles(dicomFiles)
+          return loadSeriesWithVerification(seriesUIDs, self.selectedPlugins, self.loadedNodes)
+    return False
+  def __exit__(self, type, value, traceback):
```
lassoan: If we don't do any cleanup then why this is implemented as a context manager?

Base/Python/slicer/util.py
```diff
@@ -1091,3 +1091,17 @@ def onClick(caller,event):
...
   interactor.SetShiftKey(0)
   interactor.SetControlKey(0)
+
+def getFilesInDirectory(directory, absolutePath=True):
+  """ Collect all files in a directory and its subdirectories in a list
+  """
+  import os
+  allFiles=[]
+  for root, subdirs, files in os.walk(directory):
+    for file in files:
```
jcfr: Since `file`is a python built-in function, consider using `fileName`. See https://docs.python.org/2/library/functions.html#file

Modules/Scripted/DICOMLib/DICOMUtils.py
```diff
@@ -78,16 +78,16 @@ def openDatabase(databaseDir):
 
 #------------------------------------------------------------------------------
 # Open temporary DICOM database, return current database directory
-def openTemporaryDatabase(directory=None, absolutePath=False):
+def openTemporaryDatabase(directory=None, isAbsolutePath=False):
   if not hasattr(slicer,'dicomDatabase') or not hasattr(slicer.modules,'dicom'):
```
jcfr: I wonder if we could remove this extra parameter and use https://docs.python.org/2/library/os.path.html#os.path.isabs

Modules/Scripted/DICOMLib/DICOMUtils.py
```diff
@@ -156,3 +156,180 @@ def __enter__(self):
     return slicer.dicomDatabase
   def __exit__(self, type, value, traceback):
     closeTemporaryDatabase(self.originalDatabaseDir)
+
+#------------------------------------------------------------------------------
+# Download file to local storage.
+# Target file path needs to indicate the file name and extension as well
+def downloadFile(url, targetFilePath):
```
jcfr: Consider adding the description of the method as a python docstring:

```
def downloadFile(url, targetFilePath):
  """Download ``url`` to local storage as ``targetFilePath``.
  """
```

If ``targetFilePath`` extension is required, we could check for it using https://docs.python.org/2/library/os.path.html#os.path.splitext and return an exception. 

If this function is general enough, may it could be added to ``slicer.util`` instead.

Modules/Scripted/DICOMLib/DICOMUtils.py
```diff
@@ -156,3 +156,180 @@ def __enter__(self):
...
+      return False
+  else:
+    logging.info('Requested file has been found: ' + targetFilePath)
+  return True
+
+#------------------------------------------------------------------------------
+# Unzip file into folder. Number of expected files unzipped may be specified.
+# If folder contains the same number of files as expected (if specified), then
+#   it will be assumed that unzipping has been successfully done earlier.
+def unzipFile(zipFilePath, unzipDir, expectedNumberOfUnzippedFiles=0):
```
jcfr: consider setting `expectedNumberOfUnzippedFiles=None` if you want to ignore the value.

jcfr: ``unzipFile`` -> ``unzipArchive``

Modules/Scripted/DICOMLib/DICOMUtils.py
```diff
@@ -156,3 +156,180 @@ def __enter__(self):
...
+    seriesUID = slicer.dicomDatabase.seriesForFile(file)
+    if seriesUID != '':
+      seriesUIDs.add(seriesUID)
+  return seriesUIDs
+
+#------------------------------------------------------------------------------
+class OnlineDICOMData:
+  """Context manager to conveniently load DICOM files downloaded zipped from the internet
+  """
+  def __init__(self, url, zipFilePath=None, dicomDataDir=None, expectedNumOfFiles=0, selectedPlugins={}, loadedNodes={}):
```
jcfr: `expectedNumOfFiles=None` if it is unspecified

jcfr: I also suggest to add a `database` parameter, if `None` a new database could be initialized.

Modules/Scripted/DICOMLib/DICOMUtils.py
```diff
@@ -156,3 +156,180 @@ def __enter__(self):
...
+    if zipFilePath is None:
+      fileName = strftime("%Y%m%d_%H%M%S_", gmtime()) + 'OnlineDICOMData.zip'
+      self.zipFilePath = slicer.app.temporaryPath + '/' + fileName
+    else:
+      self.zipFilePath = zipFilePath
+    if dicomDataDir is None:
+      directoryName = strftime("%Y%m%d_%H%M%S_", gmtime()) + 'OnlineDICOMData'
+      self.dicomDataDir = slicer.app.temporaryPath + '/' + directoryName
+    else:
+      self.dicomDataDir = dicomDataDir
```
jcfr: I  suggest to set default values first, and update object properties at the end.

```
[...]
+    if zipFilePath is None:
+      fileName = strftime("%Y%m%d_%H%M%S_", gmtime()) + 'OnlineDICOMData.zip'
+      self.zipFilePath = slicer.app.temporaryPath + '/' + fileName
+
+    if dicomDataDir is None:
+      directoryName = strftime("%Y%m%d_%H%M%S_", gmtime()) + 'OnlineDICOMData'
+      dicomDataDir = slicer.app.temporaryPath + '/' + directoryName
+ 
+    self.url = url
+    self.zipFilePath = zipFilePath
+    self.dicomDataDir = dicomDataDir
+    self.expectedNumberOfUnzippedFiles = expectedNumOfFiles
+    self.selectedPlugins = selectedPlugins
+    self.loadedNodes = loadedNodes
```

jcfr:
Thanks @cpinter for moving forward with this 👍

Modules/Scripted/DICOMLib/DICOMUtils.py
cpinter: Sounds reasonable. But then the other context class in the utils called TemporaryDICOMDatabase should be renamed too, and it may be used outside Slicer core by now.

Modules/Scripted/DICOMLib/DICOMUtils.py
cpinter: Because Jc requested it...

cpinter: Do you think any cleanup would be necessary? Re-using already downloaded and unzipped data is I think valuable, so deleting those wouldn't improve things I think

cpinter: Maybe I can add a flag for the context in case somebody wants to clean up. Do you think this is something that could be useful?

Modules/Scripted/DICOMLib/DICOMUtils.py
cpinter: 0 seemed to make sense because nobody wants to extract 0 files. Is None a convention to use for the abscence of an argument value?

Modules/Scripted/DICOMLib/DICOMUtils.py
cpinter: These are the context managers in Slicer
* NodeModify
* TemporaryDICOMDatabase
* ExtensionProject

cpinter:
I believe I fixed everything and amended the commit. Please check it out. Thx

Modules/Scripted/DICOMLib/DICOMUtils.py
jcfr: > Re-using already downloaded and unzipped data is I think valuable

Agreed.

That said, cache should be centralized. Directory and database used for testing should be specific to the context and cleared up at the end.

Base/Python/slicer/util.py
```diff
@@ -1091,3 +1091,97 @@ def onClick(caller,event):
   up()
   interactor.SetShiftKey(0)
   interactor.SetControlKey(0)
+
+def downloadFile(url, targetFilePath):
+  """ Download ``url`` to local storage as ``targetFilePath``
+      Target file path needs to indicate the file name and extension as well
+  """
```
jcfr: No need to align or add extra space. See https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

```
"""Download ``url`` to local storage as ``targetFilePath``.
"""
```

Also I don't think this implementation require the extension to be specified in the target file.

Base/Python/slicer/util.py
```diff
@@ -1091,3 +1091,97 @@ def onClick(caller,event):
...
+  else:
+    logging.info('Requested file has been found: ' + targetFilePath)
+  return True
+
+def extractArchive(archiveFilePath, outputDir, expectedNumberOfExtractedFiles=None):
+  """ Extract file ``archiveFilePath`` into folder ``outputDir``. Number of expected files
+      unzipped may be specified in ``expectedNumberOfExtractedFiles``.
+      If folder contains the same number of files as expected (if specified), then
+        it will be assumed that unzipping has been successfully done earlier.
+  """
```
jcfr: """Extract file ``archiveFilePath`` into folder ``outputDir``.

If ``expectedNumberOfExtractedFiles`` is specified and the number of files contained
in ``outputDir`` matches, the function returns early indicating that the archive has already
been extracted.
 """

Base/Python/slicer/util.py
```diff
@@ -1091,3 +1091,97 @@ def onClick(caller,event):
...
+    except Exception, e:
+      import traceback
+      traceback.print_exc()
+      logging.error('Failed to download file from ' + url)
+      return False
+  else:
+    logging.info('Requested file has been found: ' + targetFilePath)
+  return True
+
+def extractArchive(archiveFilePath, outputDir, expectedNumberOfExtractedFiles=None):
```
jcfr: What about adding an `archiveType="zip"` parameter.

Base/Python/slicer/util.py
```diff
@@ -1091,3 +1091,97 @@ def onClick(caller,event):
...
+  numOfFilesInOutputDirTest = len(getFilesInDirectory(outputDir, False))
+  if extractSuccessful is False or (expectedNumberOfExtractedFiles is not None \
+      and numOfFilesInOutputDirTest != expectedNumberOfExtractedFiles):
+    logging.error('Unzipping %s into %s failed' % (archiveFilePath, outputDir))
+    return False
+  logging.info('Unzipping %s into %s successful' % (archiveFilePath, outputDir))
+  return True
+
+def downloadAndExtractArchive(url, archiveFilePath, outputDir, \
+                              expectedNumberOfExtractedFiles=None, numberOfTrials=10):
```
jcfr: ``numberOfTrials=3`` ?

Base/Python/slicer/util.py
```diff
@@ -1091,3 +1091,97 @@ def onClick(caller,event):
...
+  """ Downloads an archive from ``url`` as ``archiveFilePath``, and extracts
+      it to ``outputDir``. This combined function tests the success of the
+      download by the extraction step, and re-downloads if extraction failed.
+  """
+  success = False
+  currentNumberOfTrials = 0
+  while currentNumberOfTrials < numberOfTrials and success is False:
```
jcfr: while numberOfTrials:

Base/Python/slicer/util.py
```diff
@@ -1091,3 +1091,97 @@ def onClick(caller,event):
...
+    return False
+  logging.info('Unzipping %s into %s successful' % (archiveFilePath, outputDir))
+  return True
+
+def downloadAndExtractArchive(url, archiveFilePath, outputDir, \
+                              expectedNumberOfExtractedFiles=None, numberOfTrials=10):
+  """ Downloads an archive from ``url`` as ``archiveFilePath``, and extracts
+      it to ``outputDir``. This combined function tests the success of the
+      download by the extraction step, and re-downloads if extraction failed.
+  """
```
jcfr: Suggest (more pythonic implementation)

```python
def downloadAndExtractArchive(
  url, archiveFilePath, outputDir, expectedNumberOfExtractedFiles=None, numberOfTrials=10):
  """
  """
  import os
  import shutil
  import logging

  maxNumberOfTrials = numberOfTrials

  def _cleanup():
    # If there was a failure, delete downloaded file and empty output folder
    logging.warning('Download and extract failed, removing archive and destination folder and retrying. Attempt #%d...' % (maxNumberOfTrials - numberOfTrials))
    os.remove(archiveFilePath)
    shutil.rmtree(outputDir)
    os.mkdir(outputDir)

  while numberOfTrials:
    if not downloadFile(url, archiveFilePath):
      numberOfTrials -= 1
      _cleanup()
      continue
    if not extractArchive(archiveFilePath, outputDir, expectedNumberOfExtractedFiles):
      numberOfTrials -= 1
      _cleanup()
      continue
    return True

  _cleanup()
  return False
```

Modules/Scripted/DICOMLib/DICOMUtils.py
```diff
@@ -16,10 +16,11 @@
 ######################################################### 
 
 #------------------------------------------------------------------------------
-# Get patient UID by patient name for easy loading of a patient
 def getPatientUIDByPatientName(name):
+  """ Get patient UID by patient name for easy loading of a patient
+  """
```
jcfr: `if not hasattr(slicer,'dicomDatabase'):` -> `if not hasattr(slicer, 'dicomDatabase'):`

Argument should have a space after ",".

(I will setup python style check shortly)

Modules/Scripted/DICOMLib/DICOMUtils.py
```diff
@@ -29,45 +30,73 @@ def getPatientUIDByPatientName(name):
...
+  """ Load patient by patient UID from DICOM database
+  """
   if not hasattr(slicer,'dicomDatabase') or not hasattr(slicer.modules,'dicom'):
-    logging.error('DICOM module or database cannot be accessed!')
-    return
+    logging.error('DICOM module or database cannot be accessed')
+    return False
+
+  if not patientUID in slicer.dicomDatabase.patients():
+    logging.error('No patient found with DICOM database UID ' + str(patientUID))
```
jcfr:
```
logging.error('No patient found with DICOM database UID %s' % patientUID)
```

jcfr: Here is a suggestion of improved message:
```
logging.error('Patient with UID [%s] not found in DICOM database' % patientUID)
```

The previous message seemed to indicated that the patient was not found the database having UID xyz

jcfr: Or if we want to have the UID at the end (to facilitate review of logs where multiple patient can not be found:

```
logging.error('No patient found with DICOM UID [%s]' % patientUID)
```

Base/Python/slicer/util.py
cpinter: There can be many levels of this. Specify both file name and extension, only specify file name, only specify path. I don't think any is better than the other, but we need to choose. As one is already implemented, I think it makes sense to choose that one.

Modules/Scripted/DICOMLib/DICOMUtils.py
cpinter: The wording of the message is intentional to indicate that the UID is not DICOM PatientID, but DICOM database UID.

cpinter:
Another round done, please check it out.

cpinter:
Please let me know if there are any more issues with this PR, otherwise I'll integrate it once I'm back in Kingston. Thanks!

cpinter:
Integrated in https://github.com/Slicer/Slicer/commit/fce63b47161baecc320672e93b879c0e08d3ed9a

PR #837, opened 2017-11-10
Improve slicer util message boxes

lassoan:
Made messageBox show actual application name. "Slicer" application name was hardcoded before.

Also switched to using ctkMessageBox instead of QMessageBox to be able to have "don't show again" checkbox.


lassoan:
Thanks for the review, merged in r26560.

PR #840, opened 2017-11-16
COMP: Make it easier to build on Debian 9 systems

gregsharp:
This makes it easier for developers to build Slicer on Debian >= 9 systems, and possibly other linux systems, which have gcc >= 6 as default.
1) It allows to explicitly enable -DCMAKE_CXX_STANDARD=11 on the cmake command line instead of only allowing this based on choice of VTK/Qt version
2) It allows to use system cmake <= 3.8.2 if -DSlicer_USE_SimpleITK:BOOL=OFF


jcfr:
Closing. Corresponding changes have been integrated in [r26905](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26905) and [r26841](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26841)

PR #841, opened 2017-11-16
COMP: Update OpenSSL version for UNIX

gregsharp:


jcfr:
@gregsharp I squashed the commit and updated the commit message.

Would it be possible to summarize:
* Which Operating System exactly ? 
* What was the stack trace associated with the crash (if possible).
* Is it to address issue with Qt4 or Qt5 build ? 

I just want to make sure we properly capture the context.

also wonder if we should use the newer OpenSSL only with Qt5

gregsharp:
@jcfr Sorry, just noticed this.  
#. Debian 9 and later
#. See mantis ticket: https://issues.slicer.org/view.php?id=4125 (the second stack trace)
#. Both Qt4 and Qt5 are affected

jcfr:
OpenSSL version used on Unix has been updated to `1.0.2n`. See [r26841](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26841)

@gregsharp  As a side note, your authorship will be restored when we transition to GitHub. See https://www.slicer.org/wiki/Documentation/Labs/TransitionToGit#Authorship

jcfr:
Also, the [qt-easy-build](https://github.com/jcfr/qt-easy-build)  scripts for linux and macOS have been updated to use the updated version

PR #844, opened 2017-11-20
BUG: getLoadablesFromFileLists was expecting a list of tuples only

che85:
* it could also be a list of lists

Here is a list of tuples produced:
https://github.com/Slicer/Slicer/blob/fce63b47161baecc320672e93b879c0e08d3ed9a/Modules/Scripted/DICOMLib/DICOMWidgets.py#L640-L659

Here is a list of lists produced:
https://github.com/Slicer/Slicer/blob/fce63b47161baecc320672e93b879c0e08d3ed9a/Modules/Scripted/DICOMLib/DICOMWidgets.py#L766-L781

### DIFF
--- a/Modules/Scripted/DICOMLib/DICOMWidgets.py
+++ b/Modules/Scripted/DICOMLib/DICOMWidgets.py
@@ -680,8 +680,8 @@ class DICOMDetailsBase(VTKObservationMixin, SizePositionSettingsMixin):
     """
     loadablesByPlugin = {}
     loadEnabled = False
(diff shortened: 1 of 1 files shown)


che85:
Any concerns regarding this fix? Please merge if not. It broke our extension `QuantitativeReporting` and it probably broke other DICOMPlugins that are working with referencedInstanceUIDs

cpinter:
Integrated in https://github.com/Slicer/Slicer/commit/39d125d51fe074a986afb56660603de5edceb5ad
Sorry for the delay, I wasn't sure if others want to comment.

PR #846, opened 2017-11-22
BUG: Fixed new segment color after last segment has been removed

cpinter:
If the last segment has been removed (and it was Segment_1), the display node still contains the properties of Segment_1 until an update is triggered (for example by slice controller widget, but it depends on what the user does). But the segment ID generator index is also reset to 0, so the new segment will be Segment_1, which simply "claims" those display properties. However, a new segment color is then not generated, so it will be gray. The implemented solution explicitly calls the update function when a segment is removed or added.


cpinter:
Another solution could be not to reset the segment ID generator index at
https://github.com/Slicer/Slicer/blob/master/Libs/vtkSegmentationCore/vtkSegmentation.cxx#L531
However I haven't tested this option, and we must have had a reason to reset it.

PR #847, opened 2017-11-23
ENH: Move edited segmentation to the study of the master volume

cpinter:
In addition, a feature is added to warn the user about issues with DICOM export. Many users have asked why they could not export a segmentation, and the answer always was that because it was not under a patient+study. A check is now done when clicking Export to DICOM in subject hierarchy, and a warning is shown that briefly describes the issue and the solution if the exported node did not have the proper parents.

Fixes https://issues.slicer.org/view.php?id=4313

### DIFF
--- a/Modules/Scripted/DICOMLib/SubjectHierarchyPlugins/qSlicerSubjectHierarchyDICOMPlugin.cxx
+++ b/Modules/Scripted/DICOMLib/SubjectHierarchyPlugins/qSlicerSubjectHierarchyDICOMPlugin.cxx
@@ -35,6 +35,7 @@
 #include <QDebug>
 #include <QAction>
 #include <QStandardItem>
+#include <QMessageBox>
 
 // MRML includes
 #include <vtkMRMLScene.h>
@@ -42,6 +43,10 @@
 // VTK includes
 #include <vtkObjectFactory.h>
 #include <vtkSmartPointer.h>
+#include <vtkCollection.h>
+
+// STD includes
+#include <algorithm>
 
 //-----------------------------------------------------------------------------
 /// \ingroup Slicer_QtModules_SubjectHierarchy_Widgets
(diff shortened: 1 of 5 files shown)


cpinter:
I don't think it would be better than this. If the user right-clicked the segmentation and clicked OK to create the default items, then the segmentation alone is moved under the default patient/study, and not with the volume. In that case for example, RT export would not be possible. It should be the user who puts all the data to export under a study.

lassoan:
Users are not interested in doing this properly. They just want things to be simple, even if they are not. We may achieve that by having a default behavior that works well most of the time.

It's a good point that moving only the segmentation is not enough. Instead, we should also move the master volume under the same study. Or, maybe even better: move all nodes in the scene that are not already under a study.

cpinter:
Yes, if we move all the nodes, then the DICOM plugins can find those that they can handle and ignore the rest.

pieper:
I like the idea of auto-moving the segmentation and master volumes into a default subject if they aren't already in it.  Not so sure about moving all nodes in the scene since we don't really know what they all are and we might end up mixing data from different patients.

cpinter:
My problem with moving the segmentation with the volume is that it's a special case for segmentations, while the DICOM export is handled by the DICOM SH plugin. I'd hesitate to add type-specific features in that plugin, but I agree that it would be desirable from a user point of view.

The reason I don't see moving all the nodes problematic is that most people do very simple workflows, and whoever work with multiple patients use DICOM. I may be completely mistaken here though.

pieper:
But the SH should have a way to handle derivation relationships --
segmentations are one kind but also we could/should encode these
relationships, for example, when a filtered image is made from an original
or when a transform is calculated between two volumes.

cpinter:
That's true, we could use the references. What if the node and all the nodes referenced by either MRML or DICOM are moved to the new subject/study?

pieper:
Yes, I like that.

fedorov:
Please let us know if this change in behavior will affect modules that care about assignment of the segmentations. From what I understand, nothing should change in behavior for the segmentations created from master volumes that are loaded from DICOM, is this right?

cc: @che85

cpinter:
Right. This change only affect organization of nodes in subject hierarchy. The segmentation will be moved to the same branch as the master volume, and also makes sure on DICOM export (from SH only) that the SH is correctly set up for export. No changes in MRML whatsoever

cpinter:
I amended the commit, please check out. Now DICOM and MRML references are moved with the exported node.

Modules/Scripted/DICOMLib/SubjectHierarchyPlugins/qSlicerSubjectHierarchyDICOMPlugin.cxx
```diff
@@ -429,6 +433,99 @@ void qSlicerSubjectHierarchyDICOMPlugin::openDICOMExportDialog()
...
+        {
+        // Move existing study under new patient
+        shNode->SetItemParent(currentItemID, patientItemID);
+        }
+      else
+        {
+        // Collect referenced items for the selected data item
+        // DICOM references
+        std::vector<vtkIdType> referencedItems = shNode->GetItemsReferencedFromItemByDICOM(currentItemID);
+        // MRML references available in SH
```
lassoan: Do we only add the nodes that are referenced directly? Shouldn't we add all referenced nodes recursively? (you can get that list of nodes using vtkMRMLScene::GetReferencedSubScene)

cpinter:
> Also, if referenced nodes are already under a valid study, shouldn't we just move the node under that study?

I don't support this, this would take us down a bumpy road. What if two referenced nodes are under two different studies?

@pieper what do you think about recursively add all references? I haven't seen a case like this (data references I've seen were usually quite simple), so don't have a strong opinion.

lassoan:
> What if two referenced nodes are under two different studies?

I guess we can just pick one randomly. For example, when you have a spatial registration object that registers two studies, then it does not matter which one you pick - one is just as good as the other. Probably in other cases there are preferable choices, but the user can always correct the automatic assignment later. If this does not prove to work well enough then we can prompt the user to choose an existing study / create a new study&patient.

> recursively add all references? I haven't seen a case like this (data references I've seen were usually quite simple)

Yes, for basic data nodes (such as volume and model nodes), only display and storage nodes are referenced, so if we add all referenced nodes recursively then most likely we'll not get much more than with single references. If we added both referenced and referring nodes, then we can find more relationships, but the result will be less predictable (associations may be accidentally established by choosing two nodes in a CLI module parameter node as inputs; or using the same color node). Maybe it is better to use direct references only; and only consider adding referred nodes (ignore referencing nodes).

lassoan:
I just wanted to have a little bit of discussion about this, to see if we can enhance things without significant extra effort. The current implementation is already a big improvement compared to what we had before, so if there is no obvious way to make it better then we can just integrate this PR as is, and improve in the future as needed.

cpinter:
I changed the implementation based on Andras' suggestion. If a referenced item is in a study, then the exported item is moved to the same study. If multiple are found, then the first one is used, and warning logged for the subsequent ones.

One question. Now that we use MRML references in an SH feature, would it make sense to highlight the MRML-referenced items in yellow, similarly to how we highlight the DICOM references? It would be very easy to add.

cpinter:
I went for highlighting also the MRML references, and also added highlight of also the referencing nodes with a slightly different color.

cpinter:
Integrated in https://github.com/Slicer/Slicer/commit/9ccdf7b0a50519f7b4e350bf9c423c06d8b6ccce

PR #849, opened 2017-11-30
COMP: Update build system introducing Slicer_BUILD_SEGMENTATION_SUPPORT option

jcfr:
This allows to build Slicer without segmentation support. This is
particularly useful when building custom application based on Slicer.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -296,6 +296,10 @@ option(Slicer_USE_QtTesting    "Integrate QtTesting framework into Slicer." ON)
 mark_as_advanced(Slicer_USE_QtTesting)
 mark_as_superbuild(Slicer_USE_QtTesting)
 
+option(Slicer_BUILD_SEGMENTATION_SUPPORT "Build Slicer with Segmentation support." ON)
+mark_as_advanced(Slicer_BUILD_SEGMENTATION_SUPPORT)
+mark_as_superbuild(Slicer_BUILD_SEGMENTATION_SUPPORT)
+
 option(Slicer_USE_SimpleITK "Build Slicer with SimpleITK support" ON)
 mark_as_superbuild(Slicer_USE_SimpleITK)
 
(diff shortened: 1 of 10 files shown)


jcfr:
Cc: @lassoan @pieper

lassoan:
I don't really mind making it optional other than this is yet another thing to keep in mind when writing Slicer core code: we cannot assume that Segmentation classes are available but have to add #ifdefs around each usage. It is particularly error-prone in Python code, where you don't get compile-time errors.

What was the main motivation? Is Segmentation that big, takes too long to build, causes slowdown anywhere when it is not used, or it is just nicely separated from the rest of the application and so it is easy to make it optional?

We still have unused/rarely used modules and features in the Slicer core. To make Slicer leaner, we could remove those.

pieper:
Yes, this strikes me as a bit odd - with this patch if I turn of segmentation support I'll still build classes like Libs/MRML/Widgets/qMRMLSegmentSelectorWidget.cxx but with methods like void qMRMLSegmentSelectorWidget::setSelectedSegmentIDs(QStringList segmentIDList)
that are now noops?  What's the point of that?

If we really want to be able to build without segments it seems we'd want to move these classes into a different library and just exclude it from the build.

Speaking a bit from my experience the past few days at RSNA I can say that the segment editor is very popular and very general purpose so I don't see a lot of motivation for excluding it from a build.  Things like FreeSurfer support or even diffusion volume support are much less "core" in comparison.

jcfr:
> or it is just nicely separated from the rest of the application and so it is easy to make it optional?

yes. 

> qMRMLSegmentSelectorWidget::setSelectedSegmentIDs(QStringList segmentIDList)
that are now noops? 

This was the simplest way to get a PR working for review. Without having to update the UI file associated with the Slice controller.

>  segment editor is very popular

Agreed. When the custom Slicer application is used for segmentation, it is definitively relevant to have the corresponding module available.

I may streamline the Pr and simply exclude the modules without doing anything to MRML

For now, I am marking the Pr was "work in progress"

PR #854, opened 2017-12-06
BUG: Added Latin1 encoding specification to saved MRML file

cpinter:
In case special characters (e.g. in the patient name) ended up being saved in the MRML file, loading the MRML failed due to an XML reading error. These special characters were valid in Slicer, as they were produced by slicer.util.toVTKString. If the character set is specified in the MRML file, then loading is successful.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLScene.cxx
+++ b/Libs/MRML/Core/vtkMRMLScene.cxx
@@ -989,14 +989,18 @@ int vtkMRMLScene::Commit(const char* url)
 
   int indent=0;
 
-    // this event is being detected by GUI to provide feedback during load
-    // of data. But,
-    // commented out for now because CLI modules are using MRML to write
-    // data in another thread, causing GUI to crash.
-//  this->InvokeEvent (vtkMRMLScene::SaveProgressFeedbackEvent );
+  // this event is being detected by GUI to provide feedback during load
+  // of data. But, commented out for now because CLI modules are using MRML
+  // to write data in another thread, causing GUI to crash.
+  //this->InvokeEvent (vtkMRMLScene::SaveProgressFeedbackEvent );
 
   //file << "<?xml version=\"1.0\" standalone='no'?>\n";
   //file << "<!DOCTYPE MRML SYSTEM \"mrml20.dtd\">\n";
+
+  // Add XML encoding specification. Since Slicer uses the Latin1 (ISO-8859-1) character set,
+  // but the MRML file did not specify it, the extra characters made XML loading fail with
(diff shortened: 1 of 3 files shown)


cpinter:
For example the following characters caused XML read failure, which are now correctly loaded: ³, ¼, ±

Of course they don't represent the patient name correctly, but this is what comes from ctkDICOMDatabase, and ensuring that a valid MRML can be loaded is important.

cpinter:
Cc: @pieper @jcfr

lassoan:
Thanks Csaba, this is useful!

I've been wondering why we don't have this in our XML files, but did not have a specific reason to change this. It'll be good to have it.

lassoan:
Have you tried to run tests? As I remember, there are a few tests that compare scene files as text file and those may require update of baseline data.

cpinter:
Indeed I didn't. I'll do that tomorrow and update baselines. Thanks!

jcfr:
Can this wait for after 4.8.1 or should it be included ?

jcfr:
The change is small enough, if I have a chance .. I will run the test today.

cpinter:
Considering the growing Asian user base it is potentially important. Without this change if a DICOM patient name contains Asian characters, and a scene is saved with it, it cannot be loaded back.

That said, as this has not been reported yet, it is not super critical to have it in 4.8.1, so it's up to you.

In any case, I'll start the tests now, and will let you know if updating baselines is needed (which I haven't done yet)

pieper:
Thanks for the fix!  I agree it is good to be cautious with changes in the core unless there's a known show stopper bug.

lassoan:
As expected, these tests fail after the change - all due to raw scene file comparison:
1>  	157 - vtkMRMLCameraDisplayableManagerTest1 (Failed)
1>  	648 - EMSeg_MRML_DiffDefaultNodes (Failed)
1>  	649 - EMSeg_MRML_DiffReadWriteNodes (Failed)

Otherwise it all looks good.

cpinter:
I'll fix these three tests, and will integrate after the release is done.

cpinter:
I amended the PR. I also fixed the EMSegment tests, for which the baselines need to be committed to the EMSegment Slicer4 branch. I'm not sure if I have commit right, but there is the patch
[EMSegment_patch20171207.zip](https://github.com/Slicer/Slicer/files/1540248/EMSegment_patch20171207.zip)

lassoan:
I'll take care of this - merging both EMSegment and Slicer master.

lassoan:
Merged into r26687.

PR #855, opened 2017-12-06
Bug in cli.py, fix to runSync method

mschumak:
The call to run that is made in cli.runSync does not pass along the update_display parameter, even though runSync accepts it as a parameter. This fix passes the value of update_display to the run method.


lassoan:
Good catch, thank you, I'll merge this patch today.

lassoan:
Committed r26680.

jcfr:
👍  

For reference, this fix a  "regression" introduced in [r24741](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24741)

PR #856, opened 2017-12-09
COMP: don't exclude Qt as system library when packaging

ihnorton:
This allows to package Qt from e.g. Homebrew in a local build, even though it lives in `/usr/local/lib/...`.


ihnorton:
d983be2e8f65ae thanks.

PR #857, opened 2017-12-09
COMP: ensure copied items writable in cpack staging dir

ihnorton:
Should fix errors like below when packaging with Homebrew Qt.

```
error: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/install_name_tool: can't open input file: /Users/inorton/git/slcr/r4nj/Slicer-build/_CPack_Packages/macosx-amd64/DragNDrop/Slicer-4.9.0-2016-10-03-macosx-amd64/Slicer.app/Contents/Frameworks/QtTest.framework/Versions/4/QtTest for writing (Permission denied)
```

### DIFF
--- a/CMake/BundleUtilitiesWithRPath.cmake
+++ b/CMake/BundleUtilitiesWithRPath.cmake
@@ -604,6 +604,12 @@ function(copy_resolved_item_into_bundle resolved_item resolved_embedded_item)
   else()
     #message(STATUS "copying COMMAND ${CMAKE_COMMAND} -E copy ${resolved_item} ${resolved_embedded_item}")
     execute_process(COMMAND ${CMAKE_COMMAND} -E copy "${resolved_item}" "${resolved_embedded_item}")
+
+    if (UNIX)
+      # fix permissions in staging dir
+      execute_process(COMMAND chmod -R u+w "${resolved_embedded_item}")
+    endif()
+
     if(UNIX AND NOT APPLE)
       file(RPATH_REMOVE FILE "${resolved_embedded_item}")
     endif()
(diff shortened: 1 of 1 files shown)


ihnorton:
Also a small optimization proposal in db1e1f9. No benchmark but "should be a little faster"...

ihnorton:
First try didn't work because of cmake semantics. (`file(INSTALL` always assumes target is directory 😑 ). Now we just set the permissions for copy bundles.

(removed the optimization commit and will make separate PR)

ihnorton:
With this PR I was able to create a working package with Qt4 from Homebrew.

CMake/BundleUtilitiesWithRPath.cmake
```diff
@@ -604,6 +604,12 @@ function(copy_resolved_item_into_bundle resolved_item resolved_embedded_item)
   else()
     #message(STATUS "copying COMMAND ${CMAKE_COMMAND} -E copy ${resolved_item} ${resolved_embedded_item}")
     execute_process(COMMAND ${CMAKE_COMMAND} -E copy "${resolved_item}" "${resolved_embedded_item}")
+
+    if (UNIX)
```
jcfr: Nitpick: `if (UNIX)` -> `if(UNIX)`

jcfr:
Thanks @ihnorton

CMake/BundleUtilitiesWithRPath.cmake
ihnorton: ✔️

ihnorton:
merged: a19a982

PR #859, opened 2017-12-14
ENH: Made DICOM loading of scalar volume more configurable

lassoan:
Added options to application settings:

- Acquisition geometry regularization: enable/disable adding regularization transform.
  A default option is added for those who don't know/don't care. For now, it is disabled by default.
  When an issue is found then Slicer displays a warning and the user may enable regularization.
  In the future we may add a "resample" option as well, to not just apply a transform but also use
  it to resample the volume.

- Allow loading subseries by time: if disabled (default) then loadables are not offered by content or trigger time.
  This reduces clutter and increases chance that users try alternative loadables, grouped for example by image orientation.

Minor improvements:
- Simpler node names are generated for subseries. Instead of including actual image orientation and timestamps (which are long, cryptic
strings), we use just indexes (imageOrientationPatient 1, imageOrientationPatient 2, ...).
- Acquisition transform is added under the same subject hierarchy folder as the volume.

### DIFF
--- a/Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
+++ b/Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
@@ -56,17 +56,44 @@ class DICOMScalarVolumePluginClass(DICOMPlugin):
     by the DICOM module.
     """
     formLayout = qt.QFormLayout(parent)
+
     readersComboBox = qt.QComboBox()
     for approach in DICOMScalarVolumePluginClass.readerApproaches():
       readersComboBox.addItem(approach)
-
-    readersComboBox.toolTip = "Preferred back end.  Archetype was used by default in Slicer before June of 2017.  Change this setting if data that previously loaded stops working (and report an issue)."
-
+    readersComboBox.toolTip = ("Preferred back end.  Archetype was used by default in Slicer before June of 2017."
+      "Change this setting if data that previously loaded stops working (and report an issue).")
     formLayout.addRow("DICOM reader approach:", readersComboBox)
-
     panel.registerProperty(
       "DICOM/ScalarVolume/ReaderApproach", readersComboBox,
       "currentIndex", str(qt.SIGNAL("currentIndexChanged(int)")))
+
+    importFormatsComboBox = ctk.ctkComboBox()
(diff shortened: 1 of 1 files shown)


lassoan:
Current master version of Slicer creates acquisition transform during DICOM loading in some cases when it is not desirable. This commit makes it an application settings (turned off by default). It would be good to have this integrated into 4.8.1.

Later, as we get more experience and refine tolerances, etc. we can make acquisition transform creation enabled by default.

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
```diff
@@ -56,18 +56,45 @@ def settingsPanelEntry(panel, parent):
...
+      "currentUserDataAsString", str(qt.SIGNAL("currentIndexChanged(int)")),
+      "DICOM examination settings", ctk.ctkSettingsPanel.OptionRequireRestart)
+    # DICOM examination settings are cached so we need to restart to make sure changes take effect
+
+    allowLoadingByTimeCheckBox = qt.QCheckBox()
+    allowLoadingByTimeCheckBox.toolTip = ("Offer loading of individual slices or group of slices"
+      " that were acquired at a specific time (content or trigger time)."
+      " If this option is enabled then a large number of loadable items may be displayed in the Advanced section of DICOM browser.")
+    formLayout.addRow("Allow loading subseries by time:", allowLoadingByTimeCheckBox)
+    exitMapper = ctk.ctkBooleanMapper(allowLoadingByTimeCheckBox, "checked", str(qt.SIGNAL("toggled(bool)")))
```
cpinter: I would rename the mapper instance variable to allowLoadingByTimeBoolMapper or similar.

lassoan: Good catch, I'll update it.

cpinter:
Removing the typically useless per-content-time loadables will make the loadables list more usable and easier to handle. Thanks!

fedorov:
These all look like very nice improvements, thank you!

pieper:
LGTM - thanks for getting this together for the 4.8.1 release!

lassoan:
Thanks for the feedback. Merged into r26720.

PR #860, opened 2017-12-24
BUG: fix crash in CTK on bad scalar range

pieper:
This bug fix should be merged into Slicer but it's not urgent.  There are a few other prior commits about color transfer widget have been verified (I haven't looked at those).  If anyone knows those are okay then this update to CTK should be merged.

https://github.com/commontk/CTK/pull/779

https://discourse.slicer.org/t/crash-when-using-volumes-module-on-windows-7-after-4-9-version/1711


jcfr:
Closing. Integrated in [r26862](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26862)

PR #862, opened 2018-01-03
ENH: observers of undo/redo in legacy editor

pieper:
Editor extensions that needed to hook into the undo/redo
functionality needed to connect to button signals, which
is fragile when the editor is embedded in another module.

See discussion here:
https://github.com/QIICR/PETTumorSegmentation/issues/10

With this change the extension effect directly observes the
undo/redo class.  Multiple observers can be used.

The older callback API is left in place for backwards compatibility.

The PETTumorSegmentationEffect is being updated to use these new calls.

### DIFF
--- a/Modules/Scripted/EditorLib/EditUtil.py
+++ b/Modules/Scripted/EditorLib/EditUtil.py
@@ -401,6 +401,8 @@ class UndoRedo(object):
     self.undoSize = undoSize
     self.undoList = []
     self.redoList = []
+    self.undoObservers = []
+    self.redoObservers = []
     self.stateChangedCallback = self.defaultStateChangedCallback
 
   def defaultStateChangedCallback(self):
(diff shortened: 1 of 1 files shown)


pieper:
Merged via svn: http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26818

lassoan:
At some point it would be probably worth transitioning to Segment Editor and its reusable segment editor widget.

pieper:
@lassoan 

> At some point it would be probably worth transitioning to Segment Editor and its reusable segment editor widget.

Yes, we discussed this (at some length! see linked issue report) but decided this was the most expedient solution for now.

fedorov:
@lassoan the effect in question is already available in Segment Editor. However, one of the legacy extensions - LongitudinalPETCT - is using legacy Editor, and we were looking for a quick way to make this functionality available there.

We don't know if anyone is using LongitudinalPETCT, and the original developer is no longer maintaining that extension, so it is not clear the effort to transition that extension to Segment Editor is worthwhile.

lassoan:
Maybe we can have a look at this at the project week.

PR #863, opened 2018-01-09
Add ::retrieveExtensionMetadataByName; update DMRIInstall helper

ihnorton:
- 2e2fcef adds a utility function `qSlicerExtensionsManagerModel::retrieveExtensionMetadataByName`
- dcc57af uses this utility to query metadata, by name, from Python and then run an installer task directly (plus other misc updates).

### DIFF
--- a/Modules/Scripted/DMRIInstall/DMRIInstall.py
+++ b/Modules/Scripted/DMRIInstall/DMRIInstall.py
@@ -16,8 +16,7 @@ class DMRIInstall(ScriptedLoadableModule):
 
   helpText = textwrap.dedent(
   """
-  Please use the Extension Manager to install the "SlicerDMRI" extension for
-  diffusion-related tools including:
+  The SlicerDMRI extension provides diffusion-related tools including:
 
   <ul>
     <li> Diffusion Tensor Estimation</li>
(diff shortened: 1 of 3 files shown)


Modules/Scripted/DMRIInstall/DMRIInstall.py
```diff
@@ -58,20 +85,43 @@ def setup(self):
...
   def onApply(self):
-    slicer.app.openExtensionsManagerDialog()
+    emm = slicer.app.extensionsManagerModel()
+
+    if emm.isExtensionInstalled("SlicerDMRI"):
+      self.textBox.setHtml("<h4>SlicerDMRI is already installed.<h4>")
+      self.applyButton.enabled = False
+      return
+
+    md = emm.retrieveExtensionMetadataByName("SlicerDMRIFoo")
```
jcfr: Is `SlicerDMRIFoo` the expected name ?

Modules/Scripted/DMRIInstall/DMRIInstall.py
```diff
@@ -58,20 +85,43 @@ def setup(self):
...
+
+    if md == {} or not md.has_key('extension_id'):
```
jcfr: `if not md or not md.has_key('extension_id'):`

Modules/Scripted/DMRIInstall/DMRIInstall.py
ihnorton: Hah, oops. I forgot to tack out the failure-mode test. Thanks!

ihnorton:
Updated

lassoan:
Thank you, looks good to me!

jcfr:
Integrated in [r26820](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26820) and [r26821](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26821)

ihnorton:
I noticed that `qRestAPI::Parameters` is just a `QMap<string,string>` typedef, so this could be reworked for more general use as in the following commits:

https://github.com/Slicer/Slicer/compare/master...ihnorton:update_extsmgr_and_dmriinstallutil

Then the input can be auto-converted from a Python dict, and we could use it from Python to query arbitrary metadata combinations instead of only extension-by-name (for example to simplify the Midas query implementation in [ExtensionStats](https://github.com/fbudin69500/SlicerDeveloperToolsForExtensions/blob/master/ExtensionStats/ExtensionStats.py)).

One question though, should we add an API to return an array if there are multiple results? Right now there are internal checks whether more than one result was returned from the query, in which case the function just returns an empty dict.

PR #864, opened 2018-01-10
Qt5 misc fixes

vovythevov:
This PR addresses the following issues for the Qt5 transition:
 - In Volume Rending module, the items of the ‘View’ combobox are no longer checkable.
 - Styles are broken (cannot select Slicer Dark Style)
 - QComboBox becomes empty after selection changed (https://issues.slicer.org/view.php?id=4456)

Note that the base style has changed to QFusionStyle. This creates slight differences in the Slicer's appearance. As an example, the landing page in fusion (Qt5):

image

And with Qt4:

image

CC: @cpinter, @jcfr


cpinter:
This is excellent, Johan! I'll make sure to test it with Qt5 on Windows.

cpinter:
I built and tested it on Windows with Qt 5.9.1:
* QComboBox disappearance -> Fixed
* Dark style -> Mostly fixed
* Checkable combobox -> Fixed

One thing I found is that the combobox for the style in application settings seems to have an invalid selection by default (without selecting anything), need to change selection so that it shows in the combobox. The options also changed a bit, they are now all lower case, and they used to be camel case.

Another thing I found but was not among the goals of this PR is that the search box in extension manager looks strange, the text is rendered twice:
image

vovythevov:
@cpinter: Thanks for the review and the feedback !  I'll look into these issues.

vovythevov:
@jcfr , @cpinter: I added commits to:
 - remove lingering QCleanlooksStyle that are is longer in use
 - Fix the styles display in the settings

If it's ok with you, I'll commit to slicer.

vovythevov:
(Note, I am looking at the issue with the extension manager but I think it would be better as a separate PR).

vovythevov:
Merged

Punzo:
Hi @vovythevov I tested right now on linux. It is very nice!!! 

Few nitpicks which I don't know if they are shown only in linux (or my build):

1) some & still appear in the comboboxes and icons in modules and views seems much smaller
screenshot from 2018-01-12 09-35-20

2) the settings menu have some disalignments 
screenshot from 2018-01-12 09-37-16

PR #698, opened 2017-04-04
ENH: quickly reinstall previously installed extensions v2

mneugeba:
This pull request is related to this PR (https://github.com/Slicer/Slicer/pull/632 - could be closed as PR includes all the features). It includes all the feedback from the hangout with JC as well as feedback from Ron from January (check for extensions on startup, allow for completely automatic hands-off installation of extensions if a new slicer build is installed and started for the first time). Details: https://www.slicer.org/wiki/Documentation/Labs/AutomaticUpdateAndInstallationFramework#Adaptations_.28March.29_according_to_input_from_JC_.28December.29_and_Ron_.28January.29

As it relies on tracking previous installations of extensions, the Slicer.ini needs to be adapted to emulate previous tracking.
Example:

[ExtensionsHistory]
Revisions\25844=ChangeTracker, SlicerToKiwiExporter, CleaverExtension
Revisions\25843=AnglePlanesExtension, ABC, CurveMaker, SlicerToKiwiExporter
Revisions\25842=AnglePlanesExtension, CurveMaker, CMFreg
ScheduledForRemoval=

Adding this setting to Slicer.ini would lead to identifying ChangeTracker, SlicerToKiwiExporter, CleaverExtensions as previously installed extensions (prompt on startup, preselection in the qSlicerExtensionRestoreWidget that would also show details about the other extensions installed in older versions of slicer).

PR #865, opened 2018-01-11
ENH: quickly reinstall previously installed extensions v3

hmeine:
This is a successor of #698 (since I cannot push to @mneugeba's branch directly).

In addition to that PR, this fixes compilation on Unix, adds translation features and some UI polish.

### DIFF
--- a/Applications/SlicerApp/qSlicerAppMainWindow.cxx
+++ b/Applications/SlicerApp/qSlicerAppMainWindow.cxx
@@ -1165,6 +1165,7 @@ void qSlicerAppMainWindow::showEvent(QShowEvent *event)
     d->WindowInitialShowCompleted = true;
     this->disclaimer();
     this->pythonConsoleInitialDisplay();
+    qSlicerApplication::application()->checkExtensionHistory();
     emit initialWindowShown();
     }
 }
--- a/Applications/SlicerApp/qSlicerAppMainWindow.cxx
+++ b/Applications/SlicerApp/qSlicerAppMainWindow.cxx
@@ -1165,7 +1165,7 @@ void qSlicerAppMainWindow::showEvent(QShowEvent *event)
     d->WindowInitialShowCompleted = true;
     this->disclaimer();
     this->pythonConsoleInitialDisplay();
-    qSlicerApplication::application()->checkExtensionHistory();
+    qSlicerApplication::application()->gatherExtensionsHistoryInformationOnStartup();
     emit initialWindowShown();
     }
 }
--- a/Base/QTCore/qSlicerCoreApplication.cxx
+++ b/Base/QTCore/qSlicerCoreApplication.cxx
@@ -355,6 +355,7 @@ void qSlicerCoreApplicationPrivate::init()
 
   qSlicerExtensionsManagerModel * model = new qSlicerExtensionsManagerModel(q);
   model->setExtensionsSettingsFilePath(q->slicerRevisionUserSettingsFilePath());
+  model->setExtensionsHistorySettingsFilePath(q->slicerUserSettingsFilePath());
   model->setSlicerRequirements(q->repositoryRevision(), q->os(), q->arch());
   q->setExtensionsManagerModel(model);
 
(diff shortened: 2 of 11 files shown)


pieper:
@jcfr I went over the logic for this with @hmeine and it sounds good.  We expect there may be some tweaks later but we don't expect any harm from merging this now.

hmeine:
@pieper This branch is ready to be merged, I think. And the superceded #698 can then be closed.

mneugeba:
@jcfr @hmeine very good :). This is is a bit behind the pull request it was derived from but includes most of the changes. I also learned, that at least the assert issue (see: https://github.com/Slicer/Slicer/pull/698#issuecomment-352260024) is also fixed now. Is the timeout issue described in the comment still a problem (or a problem at all, maybe it was just local)? Because it would hinder the start of Slicer, if the user is offline.

jcfr:
> If no internet connectivity is available, the timeout seems not to trigge

Few options:
* Make sure the timeout is triggered
* Or ensure internet connection is available before calling the function

Last the new functionality should be documented here: https://www.slicer.org/wiki/Documentation/Nightly/SlicerApplication/ExtensionsManager

hmeine:
I now changed the default such that the startup check is off.  In fact, I renamed the settings key, so that when qRestAPI is fixed, we can just revert this commit and people will get the startup check enabled again.  With this commit, this PR can be merged, and people can explicitly enable the check in the extension manager and finally test this feature.

finetjul:
Hi Hans,

As an answer to your email:

> Hi Julien,
> 
> I am at project week and trying to get https://github.com/Slicer/Slicer/pull/865 merged, which should benefit all Slicer powerusers (like Ron) who often install nightly builds, since they get an easy way to reinstall all their extensions.  There’s one remaining problem, which Mathias described in https://github.com/Slicer/Slicer/pull/698#issuecomment-352260024 , namely that qMidasAPI::synchronousQuery() does not properly time out.
> 
> qMidasAPI::synchronousQuery() uses qRestAPI::setTimeOut(), which is apparently all code from you, from https://github.com/commontk/qRestAPI/commit/07854251e13728e91abec18fa7e78c19aa2553e2
> 
> From reading the code, when the timer fires, QNetworkReply::abort() is called.  OTOH, qMidasAPI::synchronousQuery() relies on SIGNAL(errorReceived(QUuid,QString)) being fired, which does not seem to be the case.  From looking at QNetworkReply::NetworkError (http://doc.qt.io/archives/qt-4.8/qnetworkreply.html), there’s a OperationCanceledError that should be triggered (as well as a TimeoutError and various other errors that might be triggered if no network connection is available).
> 
> Can you help with this?
> 
> We could publicly discuss it in the above PR, for instance: https://github.com/Slicer/Slicer/pull/865
> 
> Thanks
>   Hans
> 
> PS: I wonder why https://github.com/commontk/qRestAPI/commit/07854251e13728e91abec18fa7e78c19aa2553e2#diff-21c1ab43b794715485cc278bad3d3bf4R245 is commented out, or whether that made it work previously?  It was apparently committed like this, and never active in the official sources.`

It seems that abort() does not trigger any error. Can you try to add a breakpoint in the abort() method and see why the signal is not triggered ? Maybe the behavior changed with newer Qt version.

Hth,
Julien.

hmeine:
@pieper @jcfr This should be finally ready to go:
* (Optionally) This could be merged as-is.
* After https://github.com/commontk/qRestAPI/pull/17 and https://github.com/commontk/qRestAPI/pull/16 have been reviewed and merged (I'd propose to merge https://github.com/commontk/qRestAPI/pull/18 as well, while you're at it), and
* qRestAPI has been updated in the SuperBuild,
* https://github.com/Slicer/Slicer/pull/865/commits/66779ae28f2fcf33a50f992f6de0f632df4911fe can be reverted or stripped of this branch, and it can be merged.  (I am happy to prepare this in my repo/this branch, so that we don't need yet another pull request.)

jcfr:
@hmeine  Thanks for updating the PR and fixing qRestAPI. 

Now that qRestAPI has been updated, 66779ae should not be needed ?

Also, would it be possible to squash together the commits that make a coherent unit ?

jcfr:
Closing. This is superseded by https://github.com/Slicer/Slicer/pull/867

PR #866, opened 2018-01-16
COMP: Fixing generation when using VTK9

adamrankin:
Wrap hierarchy stamp file removed in VTK9, so use base hierarchy file instead


jcfr:
Thanks Adam. Will review and merge tomorrow morning.
Jc

jcfr:
Thanks for the patch @adamrankin 

It has been integrated in [r26834](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26834)

PR #632, opened 2016-12-08
ENH: quickly reinstall previously installed extensions

hmeine:
See https://www.slicer.org/wiki/Documentation/Labs/AutomaticUpdateAndInstallationFramework
and http://www.na-mic.org/Bug/view.php?id=2779
Implementation by Mathias Neugebauer
(I only cleaned up some „mneugebauer“ comments and fixed indentation.)

PR #867, opened 2018-01-16
ENH: Add Restore Tab to the extension manager widget

jcfr:
This commit improves the extension manager so that it can keep track of
the installed extensions across version. The goal is to simplify the number
of steps allowing people to reinstall extensions that were used in a previous
installation.

(1) After installing a new version of Slicer, the user can select from all previously
    installed extensions and batch-install them.
(2) Extensions installed in the last (previous) Slicer versions are marked and preselected
    (one click-solution for getting the previous setup)
(3) Extensions that where previously installed but are not compatible for some reason, are
    shown but not selectable
(4) UI: a third tab ("Restore Extensions") is added to the Extension Manager and provide at
    least a list view for selection / a button button for installing the selected extensions.

More specifically, this topic implements the following changes:

- add setter/getter for extension history settings file path
- set path to Slicer.ini as extension model extension history path
- add function to write extension id to settings file after installation
- installed (and scheduled for removal) extensions are reflected in ini file
- add check functionality for previous extensions
- add headless mode to restore widget for initial extension restore
- add messagebox to advice restart after extension restore
- extension check on startup coupled with settings
- user-visible strings are translatable

References:
- https://www.slicer.org/wiki/Documentation/Labs/AutomaticUpdateAndInstallationFramework
- https://github.com/Slicer/Slicer/pull/632
- https://github.com/Slicer/Slicer/pull/865
- https://github.com/Slicer/Slicer/pull/698

Co-authored-by: Hans Meine <hans_meine@gmx.net>

Reviewed-by: Andras Lasso <lasso@queensu.ca>
Reviewed-by: Jean-Christophe Fillion-Robin <jchris.fillionr@kitware.com>
Reviewed-by: Steve Pieper <pieper@bwh.harvard.edu>

### DIFF
--- a/Applications/SlicerApp/qSlicerAppMainWindow.cxx
+++ b/Applications/SlicerApp/qSlicerAppMainWindow.cxx
@@ -1165,6 +1165,7 @@ void qSlicerAppMainWindow::showEvent(QShowEvent *event)
     d->WindowInitialShowCompleted = true;
     this->disclaimer();
     this->pythonConsoleInitialDisplay();
+    qSlicerApplication::application()->gatherExtensionsHistoryInformationOnStartup();
     emit initialWindowShown();
     }
 }
--- a/Base/QTCore/qSlicerCoreApplication.cxx
+++ b/Base/QTCore/qSlicerCoreApplication.cxx
@@ -355,6 +355,7 @@ void qSlicerCoreApplicationPrivate::init()
 
   qSlicerExtensionsManagerModel * model = new qSlicerExtensionsManagerModel(q);
   model->setExtensionsSettingsFilePath(q->slicerRevisionUserSettingsFilePath());
+  model->setExtensionsHistorySettingsFilePath(q->slicerUserSettingsFilePath());
   model->setSlicerRequirements(q->repositoryRevision(), q->os(), q->arch());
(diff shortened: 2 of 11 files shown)


jcfr:
@hmeine @mneugeba  This updated topic should be identical to #865 (with the exception that it does not include 66779ae28f2fcf33a50f992f6de0f632df4911fe `ENH: temporarily change default for check on startup `)

If this looks good to you, we will integrate.

jcfr:
After starting the application, it reports a progress bar that is never closed ... 

screenshot from 2018-01-16 08-23-46

hmeine:
So, I just tested again a clean build of this PR with Qt4, and it seems to work fine.  No hanging progress dialog.  I get the expected popup at start.

jcfr:
@hmeine  @mneugeba  The topic was integrated in [r26847](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26847) 🎆 

That said, it would be great if  you could build against Qt5 and look into the popup not being closed. See https://github.com/Slicer/Slicer/pull/867#issuecomment-357958333

cpinter:
Added issue to track the popup problem
https://issues.slicer.org/view.php?id=4503

PR #868, opened 2018-01-17
ENH: Throw an exception if getNode does not find a node

lassoan:
Novice developers were often confused by why their Python code did not work when they have a failed MRML node lookup using getNode function.
The problem was that if a node was not found then no errors were reported immediately and the unexpected None value in the variable caused errors later.

Fixed it by throwing a ValueError exception when getNode does not find any nodes.

### DIFF
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -596,6 +596,7 @@ def getNodes(pattern="*", scene=None, useLists=False):
   If multiple node share the same name, using ``useLists=False`` (default behavior)
   returns only the last node with that name. If ``useLists=True``, it returns
   a dictionary of lists of nodes.
+  Throws ValueError exception if node is not found that matches the specified pattern.
   """
   import slicer, collections, fnmatch
   nodes = collections.OrderedDict()
@@ -612,6 +613,8 @@ def getNodes(pattern="*", scene=None, useLists=False):
         nodes.setdefault(node.GetName(), []).append(node)
       else:
         nodes[node.GetName()] = node
+  if not nodes:
+    raise ValueError("could not find nodes in the scene by name or id '%s'" % pattern)
   return nodes
 
 def getNode(pattern="*", index=0, scene=None):


lassoan:
Example:

Old behavior:

>>> r=getNode('Red')
>>> r=getNode('Red2')
>>> 

New behavior:

>>> r=getNode('Red')
>>> r=getNode('Red2')
  File "C:\D\S4R\Slicer-build\bin\Python\slicer\util.py", line 624, in getNode
    nodes = getNodes(pattern, scene)
  File "C:\D\S4R\Slicer-build\bin\Python\slicer\util.py", line 616, in getNodes
    raise ValueError("could not find nodes in the scene by name or id '%s'" % pattern)
ValueError: could not find nodes in the scene by name or id 'Red2'
>>>

pieper:
Yes, makes sense.

jcfr:
Few comments:
* I would prefer a specific exception is used instead of the general `ValueError`. What about `MRMLNodeNotFound` ? 
* Also, I do **not** think plural version of `getNodesXXXX()` should throw an exception. It is reasonable to return an empty list.
* If we decide to throw an exception for the `getNodesXXXX()` ... I think we should be consistent and also update `getNodesByClass`, `getFirstNodeByClassByName`, `getFirstNodeByName`, ... and also `findChildren`, ... 

If the user is confused, an other approach would be to update `getNode` to print a message if no node is found ?

lassoan:
> I would prefer a specific exception is used instead of the general ValueError. What about MRMLNodeNotFound ?

OK

> Also, I do not think plural version of getNodesXXXX() should throw an exception. It is reasonable to return an empty list.

Good point, don't throw for plural versions. Reason: users need to iterate through result of getNodes anyway; trying to access first element without checking number of elements would generate exception (when nothing is found).

I would keep all other functions as is, because getFirstNodeByClassByName and getFirstNodeByName  are similar to plurals in that we are searching for a node and we are not certain that we'll find something.

Also note that getNodes is unique in that it is mainly a convenience function for debugging, due to the uncertainty in its behavior (you don't know if you have found a node by name or ID, you may specify wildcards).

jcfr:
> Good point, don't throw for plural versions.
> would keep all other functions as is

👍  

`MRMLNodeNotFound` (or similar name) (could be added to a file named `../slicer/exceptions.py` .. [here is an example](https://github.com/scikit-build/scikit-build/blob/master/skbuild/exceptions.py) of custom exceptions)

lassoan:
@jcfr I've updated the PR based on your feedback, please check.

jcfr:
Thanks @lassoan 

Please, consider updating/consolidating the following page: https://www.slicer.org/wiki/Documentation/Nightly/Developers/Tutorials/MigrationGuide

lassoan:
Perfect, I didn't know the page exists already. I move the content there.

jcfr:
> I move the content there.

👍 

I guess it could be added to this page: https://www.slicer.org/wiki/Documentation/Nightly/Developers/Tutorials/MigrationGuide/Slicer

PR #869, opened 2018-01-17
COMP: pass CMAKE_EXPORT_COMPILE_COMMANDS to slicer child build

ihnorton:
Allows to work with various tooling that uses `compile_commands.json`.

### DIFF
--- a/SuperBuild.cmake
+++ b/SuperBuild.cmake
@@ -416,6 +416,11 @@ foreach(name IN ITEMS
   endif()
 endforeach()
 
+# Respect top-level setting to request creation of compile_commands.json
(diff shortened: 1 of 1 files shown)


jcfr:
So that the change can more easily apply to other projects (Slicer extension, CTK, etc ...) using Artichoke, here is a PR that will shortly be integrated:

https://github.com/commontk/Artichoke/pull/20

jcfr:
The feature has been added to Artichoke, CTK has been updated.

Closing. Integrated in [r26863](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26863)

PR #873, opened 2018-01-19
EHN: move OpenIGTLinkIF to extension

leochan2009:


lassoan:
I'll test this on Windows now.

jcfr:
In the mean time, I only disabled the option in [r26842](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26842). This should allow the extension nightly build to work without chocking on duplicated targets.

jcfr:
Consider also removing https://github.com/Slicer/Slicer/blob/master/SuperBuild/External_OpenIGTLink.cmake 

Please, update https://github.com/openigtlink/SlicerOpenIGTLink/blob/master/SuperBuild/External_OpenIGTLink.cmake to be identical (or almost).

lassoan:
I've tested the current version on Windows. Build completed successfully both incrementally and starting a clean build from scratch.

leochan2009:
@lassoan , would you check the build of SlicerOpenIGTLink on windows platform?  if you build using "visual studio vs14 2015", the extension should be built successful, then you should have video streaming available. if you use other compiler, the build should still be successful, though without video streaming feature. 
@jcfr , thanks for your suggestion for openigtlink superbuild. Yes, i actually ran into some issue, your suggestion might solve my issue . The issue i have is that, as Openigtlink is already super built in SlicerOpenIGTLink extension, any other extension (for example, Sequence module) that is depending on SlicerOpenIGTLink should have access to OpenIGTLink library as well.  I need to dig into the cmake files, probably will come to you for further suggestions.

lassoan:
SlicerIGT and some other extensions will depend on SlicerOpenIGTLink and will have access to OpenIGTLink through that. 

Sequences may depend on SlicerOpenIGTLink in the short term, but at some point the dependency has to be removed by moving the video compression API into the Slicer core. Extensions can then register video compression codecs and other extensions can use them, but there will not be compile-time dependency.

jcfr:
We will completely remove the support for OpenIGTLink from  Slicer when the extension will built on all platform and will be renamed to SlicerOpenIGTLink

xref https://github.com/openigtlink/SlicerOpenIGTLink/issues/3
xref https://github.com/openigtlink/SlicerOpenIGTLink/issues/4

jcfr:
Closing. Integrated in [r26972](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26972)

Extension is now available:

screenshot from 2018-02-28 21-51-52

@leochan2009  Would be nice to have the tests associated with OpenIGTLinkIF executed

leochan2009:
sure, will add the test soon

PR #874, opened 2018-01-19
BUG: Fix menu/groupboxes display misalignenents in Qt5

vovythevov:
Since Qt5 now relies on the Fusion style instead of cleanlooks, we can
remove the cleanlooks-specific code (introduced in c6cb27dbc9) that caused
the menus and groupboxes text to be misaligned.

### DIFF
--- a/Base/QTGUI/qSlicerStyle.cxx
+++ b/Base/QTGUI/qSlicerStyle.cxx
@@ -146,6 +146,7 @@ QRect qSlicerStyle::subControlRect(ComplexControl control, const QStyleOptionCom
   /// the following code aims at overriding that value by setting it to 4.
   switch(control)
     {
+#if (QT_VERSION < QT_VERSION_CHECK(5, 0, 0))
     case CC_GroupBox:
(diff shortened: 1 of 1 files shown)


Punzo:
@vovythevov thanks, I tested on ubuntu 17.10 and works! 

Is also the smaller icon issue part of the style? or are the comboboxs (e.g., in the slicecontroller : https://user-images.githubusercontent.com/7985338/34866317-8887fc54-f77c-11e7-86d9-22764052f151.png) bigger (and then the icons seem smaller)?

jcfr:
> Is also the smaller icon issue part of the style?

I wonder if it is related to the `qSlicerIconEnginePlugin`

vovythevov:
@Punzo: I think the issue is unrelated to the style as changing style didn't seem to affect it. I'll investigate it separatly from this. 
@jcfr: Thanks ! That seems like a good place to start indeed !

vovythevov:
Commited with [r26846](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26846).

PR #875, opened 2018-01-22
ENH: Generate license-<projectname>.txt file for each external projects

jcfr:
This will help compose the list of open source packages used in a Slicer
based application based on the enables options.

### DIFF
--- a/CMake/ExternalProjectGenerateProjectDescription.cmake
+++ b/CMake/ExternalProjectGenerateProjectDescription.cmake
@@ -33,14 +33,29 @@
 #! ExternalProject_GenerateProjectDescription_Step(<projectname>
 #!   [NAME <name>]
 #!   [VERSION <version>]
+#!   [LICENSE_FILES <file> [...]]
 #!   [SOURCE_DIR <source_dir>]
 #!   )
 #!
-#! Generate a project description file named 'version-<projectname>.txt'
-#! containing one line of the form '<projectname> <version>'.
+#! Generate the project description files describing its version
+#! and the license.
+#!
+#! The generate files are:
+#!
+#! * 'version-<projectname>.txt' containing one line of the form '<projectname> <version>'.
+#!
+#! * 'license-<projectname>.txt' containing the project licenses.
 #!
 #! VERSION If no <version> is specified, the version is extracted from the
 #!         <source_dir> running the command '${GIT_EXECUTABLE} describe --always'.
+#!
+#! LICENSE_FILES If no license files is specified, the first file found in this list
+#!               "NOTICE COPYRIGHT Copyright COPYING LICENSE License" is appended to 'license-<projectname>.txt'.
+#!               Files with the following extension are also considered: .md .rst .txt
+#!
+#!               If one or multiple URLs and/or paths relative to the project source
+#!               directory and/or absolute paths are specified, their content is appended
+#!               to 'license-<projectname>.txt'.
 #!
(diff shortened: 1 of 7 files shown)


CMake/ExternalProjectGenerateProjectDescription.cmake
```diff
@@ -59,33 +72,97 @@ function(ExternalProject_GenerateProjectDescription_Step projectname)
...
 
-  file(WRITE "${script}"
-"execute_process(
-  COMMAND \"${GIT_EXECUTABLE}\" describe --always
-  OUTPUT_VARIABLE output
-  )
-file(WRITE \"${CMAKE_BINARY_DIR}/${description_file}\" \"${name} \${output}\")
-")
+  if(NOT _epgpd_LICENSE_FILES)
+    foreach(filename IN ITEMS NOTICE COPYRIGHT Copyright COPYING LICENSE License)
```
fedorov: I think instead of this hack, which does not guarantee presence of the actual license in the content, we introduce a requirement that 

1) every extension has a LICENSE file that should contain the license text (note that [current extension contribution instructions](https://www.slicer.org/wiki/Documentation/Nightly/Developers/Tutorials/BuildTestPackageDistributeExtensions) don't even mention "license"!)

2) presence of the LICENSE in the source code repository is a mandatory, and extension should not be merged until that file is present and content is valid

jcfr: This PR doesn't address the case of Slicer extension but focus on external project (e.g OpenSSL, ...). The  build will fail if no license information was found.

That said, here are some answer to your questions:

> every extension has a LICENSE file that should contain the license text (note that current 
> extension contribution instructions don't even mention "license"!)

While this is not explicitly listed in the step, the reviewer integrating extension to the index should check for this.

To find out about the requirement, clicking on the link at the top of the contribution page _To learn what is an extension, see [here](https://www.slicer.org/wiki/Documentation/Nightly/Developers/FAQ#What_is_an_extension_.3F)_ will outline the different level of compliance for extension. 

Adding a checklist for reviewer would be a great idea. Would  you like to take the lead on this ? 


> presence of the LICENSE in the source code repository is a mandatory, and extension should 
> not be merged until that file is present and content is valid

Agreed

fedorov: Makes sense, I missed that part. Sure, you can assign the issue to me. Working on a grant deadline today.

lassoan: Nice work!

I'm not sure the heuristics is needed. We could just skip license information if no LICENSE_FILES is specified.

Is there an option to specify custom string or local repository file (not in the external project's repository), just in case none of the files in the external repository are usable? (license is provided in a portal word file, license is not in separate file, etc.)

jcfr:
@fedorov:

>  extension contribution instructions [...] Sure, you can assign the issue to me

See https://issues.slicer.org/view.php?id=4500



@lassoan:

> not sure the heuristics is needed.

 I could definitively remove the heuristic and explicitly specify the relative path.  That said, I made license information a requirement and I am currently specifying it only where needed. 

>  or local repository file (not in the external project's repository)

* [ ]  I could add support for absolute path

> to specify custom string

* [ ] That would be nice too.

jcfr:
Closing. Integrated in [r26853](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26853)

PR #876, opened 2018-01-22
ENH: Try to download sample data again after loading fails

cpinter:
Related to https://discourse.slicer.org/t/loading-of-mrhead-sample-data-set-failed/1716

### DIFF
--- a/Modules/Scripted/SampleData/SampleData.py
+++ b/Modules/Scripted/SampleData/SampleData.py
@@ -180,12 +180,17 @@ class SampleDataWidget(ScriptedLoadableModuleWidget):
     self.log = qt.QTextEdit()
     self.log.readOnly = True
     self.layout.addWidget(self.log)
-    self.logMessage('<p>Status: <i>Idle</i>\n')
+    self.logMessage('<p>Status: <i>Idle</i>')
 
     # Add spacer to layout
     self.layout.addStretch(1)
 
-  def logMessage(self,message):
+  def logMessage(self, message, logLevel=logging.INFO):
+    # Set text color based on log level
+    if logLevel >= logging.ERROR:
+      message = '<font color="red">' + message + '</font>'
+    elif logLevel >= logging.WARNING:
+      message = '<font color="orange">' + message + '</font>'
     # Show message in status bar
     doc = qt.QTextDocument()
     doc.setHtml(message)
@@ -195,6 +200,7 @@ class SampleDataWidget(ScriptedLoadableModuleWidget):
     self.log.insertPlainText('\n')
     self.log.ensureCursorVisible()
     self.log.repaint()
+    logging.log(logLevel, message)
     slicer.app.processEvents(qt.QEventLoop.ExcludeUserInputEvents)
 
 #
@@ -292,6 +298,9 @@ class SampleDataLogic:
     """Given a uri and and a filename, download the data into
     a file of the given name in the scene's cache"""
     destFolderPath = slicer.mrmlScene.GetCacheManager().GetRemoteCacheDirectory()
+    if not os.access(destFolderPath, os.W_OK):
+      errorMessage = '<b>Cache folder %s is not writable!</b>' % destFolderPath
+      self.logMessage(errorMessage, logging.ERROR)
     return self.downloadFile(uri, destFolderPath, name)
 
   def downloadSourceIntoCache(self, source):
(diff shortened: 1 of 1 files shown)


Modules/Scripted/SampleData/SampleData.py
```diff
@@ -185,7 +185,7 @@ def setup(self):
     # Add spacer to layout
     self.layout.addStretch(1)
 
-  def logMessage(self,message):
+  def logMessage(self, message, isError=False):
```
jcfr: instead of `isError`, you could have a `logLevel` parameter where value could be `logging.INFO` or `logging.ERROR`. See https://docs.python.org/2/library/logging.html#logging-levels

Modules/Scripted/SampleData/SampleData.py
```diff
@@ -195,6 +195,10 @@ def logMessage(self,message):
     self.log.insertPlainText('\n')
     self.log.ensureCursorVisible()
     self.log.repaint()
+    if isError:
```
jcfr: replacing this with `logging.log(message, lvl=logLevel)` should do it then

Modules/Scripted/SampleData/SampleData.py
jcfr: I also suggest to improve `self.log.insertHtml(message)` so that the message is displayed in orange if `logging.WARNING`, and red if above that level.

Modules/Scripted/SampleData/SampleData.py
```diff
@@ -309,7 +316,15 @@ def downloadFromSource(self,source):
...
       filePath = self.downloadFileIntoCache(uri, fileName)
       if nodeName:
-        nodes.append(self.loadNode(filePath, nodeName, source.loadFileType, source.loadFileProperties))
+        loadedNode = self.loadNode(filePath, nodeName, source.loadFileType, source.loadFileProperties)
+        if loadedNode is None:
+          self.logMessage('<b>Load failed! Trying to download again...</b>\n', True)
+          if os.access(filePath, os.W_OK):
+            os.remove(filePath)
+          else:
+            self.logMessage('<b>Load failed! Unable to delete and try again loading %s!</b>\n' % filePath, True)
```
jcfr: The message is reported only if the file doesn't exist not if it couldn't be deleted.
Some exception would have to be caught.  
See https://docs.python.org/2/library/os.html#os.remove

Using the Qt implementation  may have a more robust  behavior across platform.

cpinter:
Thanks, @jcfr! I updated the commit.

Modules/Scripted/SampleData/SampleData.py
```diff
@@ -180,12 +180,17 @@ def setup(self):
...
-    self.logMessage('<p>Status: <i>Idle</i>\n')
+    self.logMessage('<p>Status: <i>Idle</i>')
 
     # Add spacer to layout
     self.layout.addStretch(1)
 
-  def logMessage(self,message):
+  def logMessage(self, message, logLevel=logging.INFO):
+    # Set text color based on log level
+    if logLevel is logging.ERROR:
```
jcfr:
```
if logLevel >= logging.ERROR:
```

Modules/Scripted/SampleData/SampleData.py
```diff
@@ -180,12 +180,17 @@ def setup(self):
...
+      message = '<font color="red">' + message + '</font>'
+    elif logLevel is logging.WARNING:
```
jcfr:
```
elif logLevel >= logging.WARNING:
```

Modules/Scripted/SampleData/SampleData.py
```diff
@@ -309,7 +318,14 @@ def downloadFromSource(self,source):
     for uri,fileName,nodeName in zip(source.uris,source.fileNames,source.nodeNames):
       filePath = self.downloadFileIntoCache(uri, fileName)
       if nodeName:
-        nodes.append(self.loadNode(filePath, nodeName, source.loadFileType, source.loadFileProperties))
+        loadedNode = self.loadNode(filePath, nodeName, source.loadFileType, source.loadFileProperties)
+        if loadedNode is None:
+          self.logMessage('<b>Load failed! Trying to download again...</b>', logging.ERROR)
+          file = qt.QFile(filePath)
+          if not file.remove():
+            self.logMessage('<b>Load failed! Unable to delete and try again loading %s!</b>' % filePath, logging.ERROR)
```
jcfr: If it failed to remove the file, I suggest to completely fail and not try again.

cpinter:
@jcfr Another round done.

cpinter:
Good to go then?

jcfr:
Nitpick: Since the message seems to be systematically bold, you could have `<b>` set in `logMessage` function.

Assuming you tested locally, merging as it is would be perfectly fine too .

cpinter:
I did, it works well. Good point about the bold font, I'll do that, it takes only a second. Thanks

cpinter:
It's not consistently bold, see
https://github.com/Slicer/Slicer/blob/master/Modules/Scripted/SampleData/SampleData.py#L183
or
https://github.com/Slicer/Slicer/blob/master/Modules/Scripted/SampleData/SampleData.py#L390
I think I'll just go ahead and integrate

jcfr:
>  not consistently bold,

Thanks for looking into it

> I'll just go ahead and integrate

👍

cpinter:
Integrated in https://github.com/Slicer/Slicer/commit/09d9918a828788ba8643f2cbdcc3b5b86ab871e1

PR #877, opened 2018-01-22
ENH: Hide background segment after Grow from seeds effect

cpinter:
Determine whether a segment is background by checking corner voxel values (if 5 or more are non-zero, then background)

### DIFF
--- a/Modules/Loadable/Segmentations/EditorEffects/Python/AbstractScriptedSegmentEditorAutoCompleteEffect.py
+++ b/Modules/Loadable/Segmentations/EditorEffects/Python/AbstractScriptedSegmentEditorAutoCompleteEffect.py
@@ -49,6 +49,22 @@ class AbstractScriptedSegmentEditorAutoCompleteEffect(AbstractScriptedSegmentEdi
     self.delayedAutoUpdateTimer.stop()
     self.observeSegmentation(False)
 
+  @staticmethod
+  def isBackgroundLabelmap(labelmapOrientedImageData):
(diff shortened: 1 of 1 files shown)


cpinter:
@lassoan We talked about this change. What do you think?

Modules/Loadable/Segmentations/EditorEffects/Python/AbstractScriptedSegmentEditorAutoCompleteEffect.py
```diff
@@ -49,6 +49,22 @@ def __del__(self, scriptedEffect):
...
+  def isBackgroundLabelmap(labelmapOrientedImageData):
+    if labelmapOrientedImageData is None:
+      return False
+    # If five or more corner voxels of the image contain non-zero, then it is background
+    extent = labelmapOrientedImageData.GetExtent()
+    numberOfFilledCorners = 0
+    for i in [0,1]:
+      for j in [2,3]:
+        for k in [4,5]:
+          if labelmapOrientedImageData.GetScalarComponentAsFloat(extent[i],extent[j],extent[k],0) > 0:
```
lassoan: Add a check that the extent is not empty (extent[i]<=extent[i+1] along all three axes).

jcfr:
👍 After addressing @lassoan  comments

cpinter:
Integrated in https://github.com/Slicer/Slicer/commit/08890700b0692ef079eec2c09ff84b3fc7cb8c49

PR #880, opened 2018-01-25
ENH: Added Median to LabelMapStatistics

moselhy:


jcfr:
Thanks for the patch 👍 

Closing. Integrated in [r26869](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26869)

PR #881, opened 2018-01-25
ENH: Allow excluding views from main layout management WIP

cpinter:
Re https://github.com/KitwareMedical/SlicerOpenVR/issues/18

### DIFF
--- a/Libs/MRML/Core/vtkMRMLAbstractLayoutNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLAbstractLayoutNode.cxx
@@ -1,0 +1,66 @@
+/*==============================================================================
+
+  Copyright (c) Laboratory for Percutaneous Surgery (PerkLab)
+  Queen's University, Kingston, ON, Canada. All Rights Reserved.
+
+  See COPYRIGHT.txt
+  or http://www.slicer.org/copyright/copyright.txt for details.
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+  This file was originally developed by Csaba Pinter, PerkLab, Queen's University
+  and was supported through CANARIE’s Research Software Program, and Cancer
+  Care Ontario.
+
+==============================================================================*/
+
+// MRML includes
+#include "vtkMRMLAbstractLayoutNode.h"
+
+//----------------------------------------------------------------------------
+vtkMRMLAbstractLayoutNode::vtkMRMLAbstractLayoutNode()
+{
+  this->HideFromEditors = 1;
+}
+
+//----------------------------------------------------------------------------
+vtkMRMLAbstractLayoutNode::~vtkMRMLAbstractLayoutNode()
+{
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLAbstractLayoutNode::WriteXML(ostream& of, int nIndent)
+{
+  Superclass::WriteXML(of, nIndent);
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLAbstractLayoutNode::ReadXMLAttributes(const char** atts)
+{
+  //int disabledModify = this->StartModify();
+
+  Superclass::ReadXMLAttributes(atts);
+
+  //this->EndModify(disabledModify);
+}
+
+//----------------------------------------------------------------------------
+// Copy the node's attributes to this object.
+// Does NOT copy: ID, FilePrefix, LabelText, ID
+void vtkMRMLAbstractLayoutNode::Copy(vtkMRMLNode *anode)
+{
+  //int disabledModify = this->StartModify();
+
+  Superclass::Copy(anode);
+
+  //this->EndModify(disabledModify);
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLAbstractLayoutNode::PrintSelf(ostream& os, vtkIndent indent)
+{
+}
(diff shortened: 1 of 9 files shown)


Libs/MRML/Core/vtkMRMLAbstractLayoutNode.cxx
```diff
@@ -0,0 +1,66 @@
...
+//----------------------------------------------------------------------------
+void vtkMRMLAbstractLayoutNode::WriteXML(ostream& of, int nIndent)
+{
+  Superclass::WriteXML(of, nIndent);
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLAbstractLayoutNode::ReadXMLAttributes(const char** atts)
+{
+  int disabledModify = this->StartModify();
```
lassoan: We don't need Start/EndModify now. We can add them when we add reading of some values.

Libs/MRML/Core/vtkMRMLAbstractLayoutNode.cxx
```diff
@@ -0,0 +1,66 @@
...
+
+  this->EndModify(disabledModify);
+}
+
+//----------------------------------------------------------------------------
+// Copy the node's attributes to this object.
+// Does NOT copy: ID, FilePrefix, LabelText, ID
+void vtkMRMLAbstractLayoutNode::Copy(vtkMRMLNode *anode)
+{
+  int disabledModify = this->StartModify();
```
lassoan: Add later, when needed.

Libs/MRML/Core/vtkMRMLAbstractLayoutNode.h
```diff
@@ -0,0 +1,61 @@
...
+  virtual void Copy(vtkMRMLNode *node) VTK_OVERRIDE;
+
+protected:
+  vtkMRMLAbstractLayoutNode();
+  ~vtkMRMLAbstractLayoutNode();
+
+  vtkMRMLAbstractLayoutNode(const vtkMRMLAbstractLayoutNode&);
+  void operator=(const vtkMRMLAbstractLayoutNode&);
+
+  //TODO: Add features to handle heads-up display. It is unlikely that the VR layouts can use
```
lassoan: I would not say it is a TODO. You can just describe (preferably in the class documentation, so that it shows up in doxygen) what are the intended roles and responsibilities of this class.

Libs/MRML/Core/vtkMRMLAbstractViewNode.cxx
```diff
@@ -599,3 +602,21 @@ void vtkMRMLAbstractViewNode::SetAxisLabel(int labelIndex, const char* label)
...
 }
+
+//----------------------------------------------------------------------------
+vtkMRMLNode* vtkMRMLAbstractViewNode::GetParentLayoutNode()
+{
+  return this->GetNodeReference(this->ParentLayoutNodeReferenceRole);
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLAbstractViewNode::SetAndObserveParentLayoutNode(vtkMRMLNode* node)
```
lassoan: Pointer-based reference setting is a convenience function, but we always need to have ID-based setting (vtkMRMLAbstractViewNode::SetAndObserveParentLayoutNodeID) as well, because pointer-based setting does not work in all cases (for example, when we need to set node reference before adding the node to the scene).

Libs/MRML/Core/vtkMRMLAbstractViewNode.h
```diff
@@ -110,9 +110,19 @@ class VTK_MRML_EXPORT vtkMRMLAbstractViewNode
   /// \sa GetVisibility()
   /// \sa IsViewVisibleInLayout()
   /// \sa vtkMRMLLayoutNode::SetViewArrangement()
+//TODO: Remove
```
lassoan: Are you sure we can remove this? Why?

Libs/MRML/Core/vtkMRMLAbstractViewNode.h
```diff
@@ -110,9 +110,19 @@ class VTK_MRML_EXPORT vtkMRMLAbstractViewNode
...
   virtual int IsMappedInLayout();
   virtual void SetMappedInLayout(int value);
 
+  /// Get parent layout node.
+  /// Default is no reference, meaning that the view is mapped in the main layout.
```
lassoan: I think this should be "...meaning that the view is managed by the main layout".

Even if a layout manager manages a view, in a particular layout a view may or may not be mapped.

Libs/MRML/Core/vtkMRMLAbstractLayoutNode.cxx
cpinter: I think they make it easy to write these functions later. What aobut commenting them out if you don't want them to execute?

Libs/MRML/Core/vtkMRMLAbstractLayoutNode.h
cpinter: Right, this belongs in vtkMRMLVRLayoutNode anyway.

Libs/MRML/Core/vtkMRMLAbstractLayoutNode.cxx
lassoan: Yes, commenting out would be better. It's just better to have less code in general, but of course it means extra work to put it back when you need that code, so we need a good balance (we could event take out the methods completely).

Libs/MRML/Core/vtkMRMLAbstractViewNode.h
cpinter: We can indicate the same status with the parent layout reference, can't we?

cpinter:
If there are no further comments then I'll integrate this in the afternoon.

cpinter:
Integrated in https://github.com/Slicer/Slicer/commit/2e316564d343be446c7f8251ec86430661566e17

PR #882, opened 2018-01-25
BUG: Plotting fixes

lassoan:
1. Prevent changing chart node in all viewers when it is changed in one

When plot chart node selection was changed in one plot widget, it changed plot chart node selection in all other plot widgets.
This prevented showing different charts in different widgets.

Active node in Selection node should not be used for immediately updating all views, but instead it should only be done when vtkMRMLApplicationLogic::PropagatePlotChartSelection is called.

2. Made PlotView input setting more robust: prevent crash when scene is set after plot view node, allow using the view without setting color logic (log warning only if not set).

3. Removed modification of MRML nodes in qMRMLPlotViewControllerWidget::updateWidgetFromMRML

Problem: Common PlotData type, x column name, and marker style was stored in vtkMRMLPlotChartNode. However, this information may get out of sync with actual common properties in associated PlotData nodes. Consolidation of redundant information happened in qMRMLPlotViewControllerWidget::updateWidgetFromMRML, by modifying PlotData nodes. In general, MRML nodes must not be modified in updateWidgetFromMRML calls and data redundancy should be avoided.

Solution: Do not store redundant information in PlotChart node. Common PlotData type, x column name, and marker style is determined when needed, by looking into associated PlotData nodes. When user changes a common property on the GUI, PlotData nodes are updated immediately.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLPlotChartNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLPlotChartNode.cxx
@@ -333,3 +333,120 @@ int vtkMRMLPlotChartNode::GetPlotIDs(std::vector<std::string> &plotDataNodeIDs)
 
   return static_cast<int>(plotDataNodeIDs.size());
 }
+
+// --------------------------------------------------------------------------
+void vtkMRMLPlotChartNode::SetPropertyToAllPlots(PlotDataNodeProperty plotProperty, const char* value)
+{
+  if (!this->GetScene())
+    {
+    vtkErrorMacro("vtkMRMLPlotChartNode::SetPropertyToAllPlots failed: invalid scene");
+    return;
+    }
+
+  int numPlotDataNodes = this->GetNumberOfNodeReferences(this->GetPlotDataNodeReferenceRole());
+
+  std::vector<int> plotDataNodesWasModifying(numPlotDataNodes, 0);
+
+  // Update all plot nodes and invoke modified events at the end
+
+  for (int plotIndex = 0; plotIndex < numPlotDataNodes; plotIndex++)
+    {
+    vtkMRMLPlotDataNode *plotDataNode = vtkMRMLPlotDataNode::SafeDownCast(this->GetNthNodeReference(this->GetPlotDataNodeReferenceRole(), plotIndex));
+    if (!plotDataNode)
+      {
+      continue;
+      }
+    plotDataNodesWasModifying[plotIndex] = plotDataNode->StartModify();
+
+    if (plotProperty == PlotType)
+      {
+      plotDataNode->SetType(value);
+      }
+    else if (plotProperty == PlotXColumnName)
+      {
+      plotDataNode->SetXColumnName(value);
+      }
+    else if (plotProperty == PlotYColumnName)
+      {
+      plotDataNode->SetYColumnName(value);
+      }
+    else if (plotProperty == PlotMarkerStyle)
+      {
+      plotDataNode->SetMarkerStyle(plotDataNode->GetMarkersStyleFromString(value));
+      }
+    }
+
+  for (int plotIndex = 0; plotIndex < numPlotDataNodes; plotIndex++)
+    {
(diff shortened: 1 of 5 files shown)


Punzo:
Hi @lassoan , the modifications make sense. I'll review in detail tomorrow morning as first thing.

Libs/MRML/Core/vtkMRMLPlotChartNode.cxx
```diff
@@ -333,3 +333,120 @@ int vtkMRMLPlotChartNode::GetPlotIDs(std::vector<std::string> &plotDataNodeIDs)
 
   return static_cast<int>(plotDataNodeIDs.size());
 }
+
+// --------------------------------------------------------------------------
+void vtkMRMLPlotChartNode::SetPropertyToAllPlots(PlotDataNodeProperty plotProperty, const char* value)
```
jcfr: Should that be `const PlotDataNodeProperty& plotProperty`

Libs/MRML/Core/vtkMRMLPlotChartNode.cxx
```diff
@@ -333,3 +333,120 @@ int vtkMRMLPlotChartNode::GetPlotIDs(std::vector<std::string> &plotDataNodeIDs)
...
+    if (!plotDataNode)
+      {
+      continue;
+      }
+    plotDataNode->EndModify(plotDataNodesWasModifying[plotIndex]);
+  }
+}
+
+// --------------------------------------------------------------------------
+std::string vtkMRMLPlotChartNode::GetPropertyFromAllPlots(PlotDataNodeProperty plotProperty)
```
jcfr: const PlotDataNodeProperty& plotProperty

Libs/MRML/Core/vtkMRMLPlotChartNode.cxx
jcfr: I see this is an enum. It means you could discard my comment

Punzo:
Hi @lassoan , I have a couple of feedback:

1) you are right, but then we need call vtkMRMLApplicationLogic::PropagatePlotChartSelection when the plot button in table module is pushed. The users expect that when you push the button the plot will appear immediately. 

Specifically PropagatePlotChartSelection should go at the end of qMRMLTableView::plotSelection
https://github.com/lassoan/Slicer/blob/b74cdbe5f846de400639cf429ef579817761d0e7/Libs/MRML/Widgets/qMRMLTableView.cxx#L450

2) thanks for fixing!

3) I understand the issue (i.e., if the Type of plotChart is Line, the Type variable of all the referenced PlotDataNodes can not be modified. On the other hand the developer just need to remember to set back the plotChartNode Type to Custom).

I am fine in changing, however or I am missing something or maybe you didn't push a commit. If you remove these attributes https://github.com/Slicer/Slicer/pull/882/files#diff-98ac64d8460a34f09ea360befd5b4aacL203

we need to further update the infrastructure at least (but I think there also other places):

https://github.com/lassoan/Slicer/blob/b74cdbe5f846de400639cf429ef579817761d0e7/Libs/MRML/Core/vtkMRMLPlotChartNode.cxx#L76-L79

https://github.com/lassoan/Slicer/blob/b74cdbe5f846de400639cf429ef579817761d0e7/Libs/MRML/Widgets/qMRMLPlotViewControllerWidget.cxx#L277-L315

https://github.com/lassoan/Slicer/blob/b74cdbe5f846de400639cf429ef579817761d0e7/Libs/MRML/Widgets/qMRMLPlotViewControllerWidget.cxx#L332-L364

https://github.com/lassoan/Slicer/blob/b74cdbe5f846de400639cf429ef579817761d0e7/Libs/MRML/Widgets/qMRMLPlotViewControllerWidget.cxx#L784

https://github.com/lassoan/Slicer/blob/b74cdbe5f846de400639cf429ef579817761d0e7/Libs/MRML/Widgets/qMRMLTableView.cxx#L595

Punzo:
@lassoan  can we divide this pull request in two steps? a first one with point 1 and 2 (for me good to go) and a second pull request for point 3. For point 3 I think we need to work on it.

Punzo:
I mean you are still using the PlotChartNode Attributes
screenshot from 2018-01-26 10-52-22

lassoan:
I've missed to push some changes. I'll integrate 1 and 2 and update the pull request for 3.

PR #883, opened 2018-01-26
RFC: make retrieveExtensionMetadata public, accept QVariantMap?

ihnorton:
Follow-up to https://github.com/Slicer/Slicer/pull/863#issuecomment-358000228 before I forget:

I noticed that qRestAPI::Parameters is just a QMap<string,string> typedef, so #883 could be reworked for more general use. Then the parameter is automatically wrapped for auto-conversion from a Python dict, and we could use it from Python to query arbitrary metadata combinations instead of only extension-by-name (for example to simplify the Midas query implementation in ExtensionStats).

- [ ] **One question though**: should we add an API to return an array if there are multiple results? Right now there are internal checks whether more than one result was returned from the query, in which case the function just returns an empty dict.

### DIFF
--- a/Base/QTCore/qSlicerExtensionsManagerModel.cxx
+++ b/Base/QTCore/qSlicerExtensionsManagerModel.cxx
@@ -220,9 +220,6 @@ public:
 
   void saveExtensionDescription(const QString& extensionDescriptionFile, const ExtensionMetadataType &allExtensionMetadata);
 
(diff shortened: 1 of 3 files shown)


jcfr:
Since the name of the midas API parameter is currently an implementation details, for now I would prefer not expose such a method.

After we transition to the girder based implementation of the extension server (currently [work in progress](https://github.com/girder/slicer_extension_manager)) where we would document and version the API, we could look into having consistent use of property names and expose them.

PR #884, opened 2018-01-30
ENH: Added median to Segmentation statistics

moselhy:
Added median to Segmentation statistics


jcfr:
Thanks 👍 

This is now integrated in [r26904](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26904)

PR #886, opened 2018-02-01
BUG: Plotting fixes

lassoan:
**just for review, not to merge it yet**

1. Added macros for reading, writing, copying, and printing MRML node properties

2. vtkPlot was stored in vtkPlotDataNode and was used in multiple charts. This caused plots to malfunction when multiple plot views were visible (for example, reset axes did not work).

A vtkPlot cannot be used in multiple charts, as vtkPlot holds pointers to axes and charts own axes objects. Fixed it by storing only properties in vtkPlotDataNode (not a vtkPlot object), vtkPlot objects are vreated in qMRMLPlotView (based on properties stored in vtkPlotDataNode).

3. Store vtkMRMLPlotChartNode properties in MRML node members access them using direct get/set methods, instead of using generic node attributes.

This node broke away from the MRML node convention by storing properties in custom attributes. Attributes are less easy to discover, more complicated to change (need to make sure the correct attribute name is used and value is properly converted to/from string).

4. Instead of using "Indexes" special column value, empty string means that X axis should be value index.

Now "Indexes" is no longer a reserved column name; it can be used freely in tables.

5. Replaced "Custom" option in plot controller widgets by invalid selection

6. Removed modification of MRML nodes in qMRMLPlotViewControllerWidget::updateWidgetFromMRML

Problem: Common PlotData type, x column name, and marker style was stored in vtkMRMLPlotChartNode. However, this information may get out of sync with actual common properties in associated PlotData nodes. Consolidation of redundant information happened in qMRMLPlotViewControllerWidget::updateWidgetFromMRML, by modifying PlotData nodes. In general, MRML nodes must not be modified in updateWidgetFromMRML calls and data redundancy should be avoided.

Solution: Do not store redundant information in PlotChart node. Common PlotData type, x column name, and marker style is determined when needed, by looking into associated PlotData nodes. When user changes a common property on the GUI, PlotData nodes are updated immediately.

**Known limitation**:
* Default plot colors are not set yet (they should be set when a  new plot is created)


Punzo:
Hi @lassoan thanks for the rework of the plotting and the bug fixes.

It is very good, I have just few general remarks:

1) Plot types:  
  a) The Line and Scatter have been merged in only one (Scatter). The new Scatter type is a vtkPlotLine.
      This means that there is no option anymore for plotting vtkPlotPoints. I really would prefer to 
      maintain both a Line and Scatter options (respectively vtkPlotLine and vtkPlotPoints). Sometime is 
      really useful to plot scatter without connecting the points with a line  (e.g., table with no ordered "X" 
      axis). 
  b) I could not understand how to use the box and pie type, maybe will be good to add a couple of 
      examples in the documentation: https://www.slicer.org/wiki/Documentation/Nightly/Developers/Plots

2) BUG: If in a table I select only one column and I push the plot button, the string XColumnName for the 
    vthMRMLPlotData is automatically set to "Indexes". The XColumnName should be " " (as it should be 
    in the new infrastructure) instead it is set to "indexes" .

lassoan:
1a. Yes, of course we need the option of not showing the line in a scatter plot. I've added this to the list of known limitations that should be fixed.

I've tried to follow Excel's naming convention (as probably this is the most widely used software in the world for simple plotting).
- Scatter:  XY plots, regardless of a line is shown or not - same as now in Slicer (we just have to expose the option of showing/hiding line)
- Pie: same as in Slicer (just implementation is incomplete in Slicer)
- Box: same as in Slicer (just implementation is incomplete in Slicer)
- Column: it's called "Bar" in Slicer (Excel uses Bar for horizontal bars), we should probably rename to Column
- Line: it is for XY plots that don't have X values. I think it could be a good idea to use this in Slicer, too. Then we would not need to handle "Indexes" special case, but the plot type would determine if X column is needed (and if it is specified it should be used as X coordinate or as label for equally spaced values).

1b. Implementation is incomplete. I've added to the list of known limitations.

2. Good catch, added to list of known issues.

3. It is nice if we generate colors by default. Colors should be unique an application-wide service (not just in a specific chart). It could be a static helper function in the plot data node that would look at all plot data nodes in the scene and choose a color that has not been used yet. Nevertheless, a color picker could be useful. I'm thinking about adding a dedicated Plot module to view/edit charts and plots, which would be more convenient and easier-to-find than the advanced section of View Controllers module.

---
Known issues:
- Default plot colors are not set yet (they should be set when a new plot is created)
- Change chart type for box and pie plots (or remove these plot types)
- Add option to show/hide line in scatter plots
- Rename plot types (scatter, line, column, pie, box)
- Fix plot X column index specification when choosing only one column in table view
- Plot/Chart property editing is difficult to find, create dedicated module

lassoan:
Pushed some few features and fixes. Remaining issues:
- Default plot colors are not set yet (they should be set when a new plot is created)
- Plot/Chart property editing is difficult to find, create dedicated module

Punzo:
Hi @lassoan,

A) default color: I agree with you.

B) I received the same feedback from users regarding the GUI for the Plot/Chart property editing. It is too tricky to find it and it is not documented anywhere. I agree in creating a separate module will help a lot.

Other feedback on commit 88124df:

1) Line (plotting one Column and the X-axes is indexed) and Scatter (plotting two Columns) definitions now can be very confusing for the user: i.e. the user will interact mainly with the viewcontroller widget and when clicking Line or Scatter will expect that the aesthetics of the plot will change (Scatter or Line). Instead now the plot will have the same aesthetics, but the difference is on the X-axes (indexed or from a Column).
Two simple solutions can be: i) changing automatically the Markers and Line styles when switching from/to Scatter/Line types; ii) changing the name of the type of the plots (something like: Scatter and Indexed Scatter or Line and Indexed Line).

2) When the "general" type (in the plotview controller) is Line, then the menu for choosing the "X-axes" 
should be disabled.

3) When the "general" type (in the plotview controller) is Bar, then the menu for choosing the Markers style should be disabled.

4) We should add in the plotView controller a menu for choosing the general Line style as well. 
I think there is room to reduce the width dimension of the menu boxes of grid, legend, x-axes, plot type and markers style.

5) The tooltips should be updated as well. 

6) I didn't manage to understand how to use the label Column, is this feature not complete? In this case, is it not better to incorporate it later on? 

7) for bar plot should be possible to plot both providing two Columns and with an indexed X-Axes (or even better a label Column). The first case is needed for example in the case of Histograms.

lassoan:
> the user will interact mainly with the viewcontroller widget

We either have to improve the controller widget (to expose more options and allow editing individual plot data properties and instead of only allowing overwriting properties of all plot data nodes) or remove most controls (keep only some high-level chart properties) and let the user edit all plot and chart properties in the Plots module. Do you have a preference? 

> changing the name of the type of the plots (something like: Scatter and Indexed Scatter or Line and Indexed Line)

How do you like these? Scatter (or scatter line), scatter bar, line, bar, pie, box (the last two will be added later) 

> I didn't manage to understand how to use the label Column, is this feature incomplete?

Labels are defined for each point, so they are shown in the tooltips (when you hover over a data point). For showing label on the X axis, we could add a chart property that references a plot data node and that node's labels will be used as axis labels. 

I find PlotData name hard to use. "Data" word is not adding any information (all nodes store data). It should really be PlotSeries instead. This name is used in Excel, Matlab, etc. so it should be clear and specific enough for all users.  Can I rename the class?

Punzo:
> We either have to improve the controller widget (to expose more options and allow editing individual 
> plot data properties and instead of only allowing overwriting properties of all plot data nodes) or 
> remove most controls (keep only some high-level chart properties) and let the user edit all plot and 
> chart properties in the Plots module. Do you have a preference?

I think the plotview controller GUI is already too crowded. If we do a plots module, I would go for a lighter and simpler plotview controller with only some high level chart properties. 

> How do you like these? Scatter (or scatter line), scatter bar, line, bar, pie, box (the last two will be 
> added later)

I don't have better names to express them: scatter line, scatter bar (in which scatter means it is a two columns plot), line, bar, pie, box  will work for me (maybe we can put a short explanation in a tooltip). Maybe somebody else can give a feedback on the names? @jcfr @pieper ?

> Labels are defined for each point, so they are shown in the tooltips (when you hover over a data 
> point). For showing label on the X axis, we could add a chart property that references a plot data node 
> and that node's labels will be used as axis labels.

ah perfect, I didn't try to put the cursor over the data. This is nice!

> I find PlotData name hard to use. "Data" word is not adding any information (all nodes store data). It 
> should really be PlotSeries instead. This name is used in Excel, Matlab, etc. so it should be clear and 
> specific enough for all users. Can I rename the class?

yes of course, at that time I had no big preference for the two names either and thanks for improving the plotting infrastructure.

jcfr:
> I don't have better names to express them: scatter line, scatter bar (in which scatter means it is a two columns plot), line, bar, pie, box will work for me 

I don't have a lot of experience with the Slicer plotting module. That said, reusing concept from other software is sensible. Regarding API and naming, it may be worth looking at https://doc.qt.io/qt-5/qtcharts-overview.html 

> (maybe we can put a short explanation in a tooltip).

👍

lassoan:
I've added a "Plots" module that has detailed GUI for changing most chart and series properties; and simplified the plot view controller widget.

I've also created a reusable widget for chart and series property editing, added tests, and made many other small fixes.

Libs/MRML/Core/vtkMRMLNodePropertyMacros.h
```diff
@@ -0,0 +1,249 @@
+/*=auto=========================================================================
```
lassoan: Macros that simplify MRML node property read/write XML, copy, and printself.
@jcfr @pieper @cpinter

Libs/MRML/Core/vtkMRMLPlotChartNode.cxx
```diff
@@ -142,6 +117,21 @@ void vtkMRMLPlotChartNode::WriteXML(ostream& of, int nIndent)
 {
   // Start by having the superclass write its information
   Superclass::WriteXML(of, nIndent);
```
lassoan: Example of using macros for simplifying MRML node property read/write XML, copy, and printself.
@jcfr @pieper @cpinter

pieper: I like the macros a lot 👍

Libs/MRML/Core/vtkMRMLPlotSeriesNode.cxx
```diff
@@ -0,0 +1,422 @@
...
+
+  This file was originally developed by Davide Punzo, Kapteyn Astronomical Institute,
+  and was supported through the European Research Council grant nr. 291531.
+
+==============================================================================*/
+
+#include "vtkMRMLPlotSeriesNode.h"
+
+// MRML includes
+#include "vtkMRMLColorTableNode.h "
```
Punzo: please remove the space after the .h in linux gives error

Punzo:
please remove the .h in:
https://github.com/lassoan/Slicer/blob/e1363614f93066f8fbb38d247dd08711e77be48d/Libs/MRML/Widgets/qMRMLPlotViewControllerWidget.cxx#L28

also gives error in linux qt5

Punzo:
this 
https://github.com/lassoan/Slicer/blob/e1363614f93066f8fbb38d247dd08711e77be48d/Libs/MRML/Widgets/qMRMLPlotView.cxx#L90-L107

is now unused it can be removed

Punzo:
these are unused as well:
https://github.com/lassoan/Slicer/blob/e1363614f93066f8fbb38d247dd08711e77be48d/Libs/MRML/Widgets/qMRMLPlotView.cxx#L342

and 

https://github.com/lassoan/Slicer/blob/e1363614f93066f8fbb38d247dd08711e77be48d/Libs/MRML/Widgets/qMRMLPlotView.cxx#L752

Modules/Loadable/Plots/qSlicerPlotsModule.cxx
```diff
@@ -0,0 +1,143 @@
...
+QStringList qSlicerPlotsModule::dependencies() const
+{
+  return QStringList();
+}
+
+//-----------------------------------------------------------------------------
+void qSlicerPlotsModule::setup()
+{
+  this->Superclass::setup();
+  vtkSlicerPlotsLogic* PlotsLogic =
```
Punzo: unused logic, it can be removed

Punzo:
Hi @lassoan. a part few small compilation warnings and errors, for me it is good to go the PR. Will you also update the documentation? 
https://www.slicer.org/wiki/Documentation/Nightly/Developers/Plots
I moved you from contributor to author, please add also needed logos or acknowledgements

Punzo:
Very nice the menu for the interaction mode!!!

Punzo:
the single colors of each plotdataseries are still not init, are they?

lassoan:
Thank you for reviewing!

Series have unique colors now: a new unique color is generated when the user adds a series through the GUI. If you add a series programmatically then you have to call plotSeriesNode->SetUniqueColor().

Libs/MRML/Core/vtkMRMLNodePropertyMacros.h
```diff
@@ -0,0 +1,249 @@
...
+
+// Macros for writing node properties to XML
+//----------------------------------------------------------------------------
+
+#define vtkMRMLWriteXMLBeginMacro(of) \
+  { \
+  ostream& xmlWriteOutputStream = of;
+
+#define vtkMRMLWriteXMLEndMacro() \
+  }
```
jcfr: I wonder if these two macros are needed: vtkMRMLWriteXMLBeginMacro / vtkMRMLWriteXMLEndMacro

Libs/MRML/Widgets/qMRMLPlotView_p.h
```diff
@@ -81,21 +89,23 @@ public slots:
   void onPlotChartNodeChanged();
 
   void RecalculateBounds();
-  void switchSelectionMode();
+  void switchInteractionMode();
   void switchLeftAndMiddleClick();
```
Punzo: I guess this method can be removed now that the pan is done with left click.

Punzo:
> Series have unique colors now: a new unique color is generated when the user adds a series through 
> the GUI.

ok just tested now. But If I create a plot in the table module by clicking the plot button, teh color is always black. Is it possible to init the color also there by calling plotSeriesNode->SetUniqueColor().?
 
> If you add a series programmatically then you have to call plotSeriesNode->SetUniqueColor().
ok perfect, then this should go in the example in the documentation

Libs/MRML/Core/vtkMRMLNodePropertyMacros.h
```diff
@@ -0,0 +1,249 @@
...
+
+#define vtkMRMLWriteXMLBooleanMacro(xmlAttributeName, propertyName) \
```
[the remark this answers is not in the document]
lassoan: There are several things that we couldn't do with functions. If we use macros, internally we can always move as much as we can to functions (without changing anything in MRML nodes that use the macros), but if we use a function then the possibilities are quite limited.

Libs/MRML/Core/vtkMRMLNodePropertyMacros.h
lassoan: Two reasons:
1. Hide xmlWriteOutputStream local variable (a name clash is unlikely, but just to be on the safe side).
2. Future-proofing: this allows us to improve the macros with any additional initializations/cleanups without requiring any change in the MRML nodes that use them.

Libs/MRML/Widgets/qMRMLPlotView_p.h
lassoan: Yes. The mechanism was also a bit fragile (releasing shift when mouse was clicked outside the view left the view in inconsistent state).

lassoan:
> But If I create a plot in the table module by clicking the plot button, teh color is always black

The first color in the table is black but all the other series should get different colors. I've just tested and it works well for me. Could you check again?

Libs/MRML/Widgets/qMRMLPlotView_p.h
Punzo: ok, perfect, then I please remove it

Punzo:
> The first color in the table is black but all the other series should get different colors. I've just tested and it 
> works well for me. Could you check again?

ah yes, just didn't notice that the first color is black. It is good as it is then.

Punzo:
P.S.: I noticed only now that the ClickAndDragX/Y bool variables have been moved in the PlotViewNode (now EnablePointMoveAlongX/Y). I think that it is fine, but we should expose them in the GUI to enable/disable them in the advanced controls in the GUI of ViewControllers.

Libs/MRML/Core/vtkMRMLNodePropertyMacros.h
jcfr: In that case, 👍  to use macros. May be it could be as an implementation details, function could be used.

@finetjul What do you think ?

lassoan:
> Will you also update the documentation?
https://www.slicer.org/wiki/Documentation/Nightly/Developers/Plots

Done.

lassoan:
> I noticed only now that the ClickAndDragX/Y bool variables have been moved in the PlotViewNode (now EnablePointMoveAlongX/Y). I think that it is fine, but we should expose them in the GUI to enable/disable them in the advanced controls in the GUI of ViewControllers.

If you have a strong need for adjusting it from the GUI then we can add it, but there are reasons for not doing it:
- Moving of points is a very advanced feature. Restricting point motion to certain axes is even more advanced. Adding two checkboxes would clutter the user interface but users would never or very rarely use them.
- If someone needs point motion for curve fitting or other special purposes then the module that manages the workflow can set the flags. If the constraints are not exposed on the GUI then users cannot easily override them, this preventing accidental data corruption.
- Little bit of extra work to add these options back to the GUI.

Punzo:
ok, I agree it is fine. Specialized modules and extensions can handle it automatically or give the user the capability. On the other hand, I think the problem is that the two attributes should be attributes of the PlotChartNode class. Is there any specific reason why you moved the two attributes from the plotChartNode to the plotViewNode? 

The point is: the user may change the chartNode that he/she is visualizing in a PlotView. The new Chart and the previous can be two chartNodes which require different settings for the ClickAndDrag. If a module needs to change the click and drag settings (now in the plotViewNode) automatically will require an observer and modifying the parameters in the plotViewNode.

lassoan:
> Is there any specific reason why you moved the two attributes from the plotChartNode to the plotViewNode?

Interaction mode is defined in the view node and these constraints further specify the interaction mode properties. It was very odd when interaction mode was set to "move" but I still could not move points because some chart settings prevented that.

Having the constraints in the view node also gives more flexibility. You can have a read-only view and an editable view of the same chart.

For more refined control, you can add a plot view in your module GUI and then you have full control over which plot is shown, what is allowed to be edited, etc.

Anyway, if this is a strong need for you, we can add point move locking at chart level (in addition to view level lock).

Punzo:
> Is there any specific reason why you moved the two attributes from the plotChartNode to the 
>plotViewNode?

> Interaction mode is defined in the view node and these constraints further specify the interaction 
> mode properties. It was very odd when interaction mode was set to "move" but I still could not move 
> points because some chart settings prevented that.

>Having the constraints in the view node also gives more flexibility. You can have a read-only view and 
>an editable view of the same chart.

ah yeah now understand why. 

> For more refined control, you can add a plot view in your module GUI and then you have full control 
> over which plot is shown, what is allowed to be edited, etc.

> Anyway, if this is a strong need for you, we can add point move locking at chart level (in addition to 
> view level lock).

I'd like to go for the second option even if it means a duplication of parameters (I use various chart with different drag and click settings in multiple modules and the user may be willing to see some plot independently by which is the current active module, etc...). Not sure, however, how will be implemented, maybe two solutions can be:
1) in default the chart has the dragAndClick attribute empty. Only in the case they will be modified/set, they will overwrite the plotView settings
2) adding an attribute in plotView to allow the overwrite of the plotView settings by the chart settings. 

I went through the modification a second time and for me this PR is really nice and good to go (if you can add the chart dragAndClick settings will be super nice).

lassoan:
How about this solution: Add EnablePointMoveAlongX/Y properties to vtkMRMLPlotChartNode. In the plot view, points can be moved along an axis if it is enabled in both vtkMRMLPlotChartNode **and** vtkMRMLPlotViewNode. By default EnablePointMoveAlongX/Y is enabled in both chart and view.

Libs/MRML/Core/vtkMRMLNodePropertyMacros.h
lassoan: I've created a separate, small pull request for node property setting helper macros: https://github.com/Slicer/Slicer/pull/890

Punzo:
> How about this solution: Add EnablePointMoveAlongX/Y properties to vtkMRMLPlotChartNode. In the 
> plot view, points can be moved along an axis if it is enabled in both vtkMRMLPlotChartNode and 
> vtkMRMLPlotViewNode. By default EnablePointMoveAlongX/Y is enabled in both chart and view.

for me this will be perfect.

lassoan:
Added Add EnablePointMoveAlongX/Y properties to vtkMRMLPlotChartNode. If there are no more comments, I'll squash the commits and integrate tomorrow.

Punzo:
for me it is good to go

Punzo:
@lassoan in linux Qt5 after the last commits I get crashes when selecting any Plot layout. I am investigating and look where the crash occurs.

Punzo:
ok found: you need a check on this->MRMLPlotChartNode here:

https://github.com/lassoan/Slicer/blob/32f8328e1489caca5fff71c8f57c758a5c0476ed/Libs/MRML/Widgets/qMRMLPlotView.cxx#L517

lassoan:
Squashed, rebased to latest master, now running tests. If tests pass then I integrate.

jcfr:
Note: It indicates it can't merge because of  the warning fixes for vtkMRMLNodePropertyMacros I just integrated ... it should not be a problem when you will `git svn dcommit`

lassoan:
I've resolved the conflicts locally. I'm running tests now and integrate in the morning if everything works as expected.

lassoan:
Thank you for all the feedbacks. These plotting improvements have been merged into r26906.

Punzo:
Hi @lassoan a small compilation error :
https://github.com/Slicer/Slicer/blob/082edc40ce0102532578884553a326250742569a/Libs/MRML/Core/vtkMRMLPlotSeriesNode.cxx#L149 gives error, I guess should be simply:   vtkMRMLCopyBeginMacro(anode);

lassoan:
Thanks, I'll fix it. Interestingly, Visual Studio did not complain.

PR #887, opened 2018-02-02
BUG: Fixing the empty dialog on startup

hina-shah:
The progress dialog of SlicerExtensionRestoreWidget was enabled at the initialization causing it to show up. Right now closing it during the initialization so that the pop up does not show. Ideally maybe the instance of the ProgressDialog should be created and deleted only when it's needed!


jcfr:
Thanks @hina-shah 

Locally tested. Now integrated as [r26879](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26879)

PR #888, opened 2018-02-05
ENH: Support passing markups to CLI specifying IJK coordinate

jcfr:
This commit allows associating any markupsNode with a reference image. The
association is then used to automatically convert to/from IJK coordinate
using SetMarkupPointIJK()/GetMarkupPointIJK() used in vtkMarkupsNode::WriteCLI().

Co-authored-by: Deepak Chittajallu <deepak.chittajallu@kitware.com>

### DIFF
--- a/Modules/CLI/ExecutionModelTour/ExecutionModelTour.xml
+++ b/Modules/CLI/ExecutionModelTour/ExecutionModelTour.xml
@@ -190,8 +190,22 @@
     <description><![CDATA[Parameters that describe point sets.]]></description>
     <point multiple="true" coordinateSystem="ras">
       <name>seed</name>
-      <label>Seeds</label>
-      <longflag>--seed</longflag>
+      <label>Seeds RAS</label>
+      <longflag>--seed_ras</longflag>
+      <description><![CDATA[Lists of points in the CLI correspond to slicer fiducial lists]]></description>
+      <default>0,0,0</default>
+    </point>
+    <point multiple="true" coordinateSystem="lps">
+      <name>seed_lps</name>
+      <label>Seeds LPS</label>
+      <longflag>--seed_lps</longflag>
+      <description><![CDATA[Lists of points in the CLI correspond to slicer fiducial lists]]></description>
+      <default>0,0,0</default>
+    </point>
+    <point multiple="true" coordinateSystem="ijk">
+      <name>seed_ijk</name>
+      <label>Seeds IJK</label>
+      <longflag>--seed_ijk</longflag>
       <description><![CDATA[Lists of points in the CLI correspond to slicer fiducial lists]]></description>
       <default>0,0,0</default>
     </point>
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsFiducialStorageNode.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsFiducialStorageNode.cxx
@@ -330,11 +330,10 @@ int vtkMRMLMarkupsFiducialStorageNode::ReadDataInternal(vtkMRMLNode *refNode)
               }
             else
               {
-              // IJK not implemented yet, assume RAS
-              markupsNode->SetMarkupPoint(thisMarkupNumber,0,x,y,z);
+              markupsNode->SetMarkupPointIJK(thisMarkupNumber,0,x,y,z);
               }
 
-            // orientatation
+            // orientation
             getline(ss, component, ',');
             ow = atof(component.c_str());
             getline(ss, component, ',');
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
@@ -21,6 +21,7 @@
 #include "vtkMRMLMarkupsNode.h"
 #include "vtkMRMLMarkupsStorageNode.h"
 #include "vtkMRMLTransformNode.h"
+#include "vtkMRMLVolumeNode.h"
 
 // Slicer MRML includes
 #include "vtkMRMLScene.h"
@@ -40,6 +41,9 @@
 #include <sstream>
 #include <algorithm>
 
+const char* vtkMRMLMarkupsNode::ReferenceImageNodeReferenceRole = "referenceImage";
+const char* vtkMRMLMarkupsNode::ReferenceImageNodeReferenceMRMLAttributeName = "referenceImageNodeRef";
+
 //----------------------------------------------------------------------------
 vtkMRMLNodeNewMacro(vtkMRMLMarkupsNode);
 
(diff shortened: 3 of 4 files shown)


Modules/CLI/ExecutionModelTour/ExecutionModelTour.xml
```diff
@@ -190,8 +190,22 @@
     <description><![CDATA[Parameters that describe point sets.]]></description>
     <point multiple="true" coordinateSystem="ras">
       <name>seed</name>
-      <label>Seeds</label>
-      <longflag>--seed</longflag>
```
pieper: Should we leave in --seed as a synonym for --seed_ras for backward compatibility?

jcfr:
After getting initial feedback, we will add corresponding tests.

The current implementation does not:
* automatically associated the markup with the reference image:
  * After a markup is manually placed on an image, we could automatically associate it with the underlying image ? If there are multiple layers, it could be foreground (if set) or background (if set)
* provide a UI to associate a markups with a reference image:
  * we could add a combo box in the markups module



Cc: @cdeepakroy @lassoan @pieper @fedorov

fedorov:
Few comments:
* in the past, we used to have a similar functionality associating fiducial placed on the image with the SOPInstanceUID of the annotated image; I would suggest you consider capturing this information when the underlying image is DICOM. If you like, I can follow up and dig the specifics.
* if you pass seed in IJK, it is meaningless without explicitly specifying what is the image defining the IJK coordinate space. Is there a plan to address this? Perhaps you could add a reference to the input image node in the point description? I know it is potentially stretching the limits of CLI, but I am not sure how it can be meaningful otherwise...

jcfr:
> in the past, we used to have a similar functionality associating fiducial placed on the image with the 
> SOPInstanceUID of the annotated image; I would suggest you consider capturing this information 
> when the underlying image is DICOM. If you like, I can follow up and dig the specifics.

Good idea. While relevant in the context of DICOM images, I think this is outside the scope of this particular PR. 

That said, it may be worth documenting the specifics in an issue.

> if you pass seed in IJK, it is meaningless without explicitly specifying what is the image defining the 
> IJK coordinate space. Is there a plan to address this? Perhaps you could add a reference to the input 
> image node in the point description? I know it is potentially stretching the limits of CLI, but I am not 
> sure how it can be meaningful otherwise...

Agreed.

1. We could improve the CLI user interface so that it displays the reference image if one is associated with a selected markup.

2. Then, we could recommend that fiducial list are named following a convention like `<refImageName>_<listName>`. In the context of `ExtractSkeleton`, the name of the parameter could be `arg0_seeds_ijk`

Initially, we looked into adding support for `reference` to the `point` element but didn't go anywhere with the idea .... 
 
```
<point multiple="true" coordinateSystem="ijk" reference="arg0">
```

lassoan:
I don't think it is a good idea to pass around fiducial positions in IJK. In general, all computations are to be performed in physical space.

If someone needs IJK then just pass the image (or some kind of reference to the image) or the IJK To LPS transform?

lassoan:
> In the context of ExtractSkeleton

I've fixed ExtractSkeleton module 1-2 weeks ago. It now properly generates centerline point coordinates and have the option of returning them as fiducial points (of course in physical coordinates) so that you can feed it to MarkupsToModel or CurveMaker extension to create a model.

jcfr:
> I've fixed ExtractSkeleton module 1-2 weeks ago

👍 

I meant `ExecutionModelTour` and not `ExtractSkeleton`

@dzenanz @cdeepakroy  Could you comment on what would work best for your use cases ?

lassoan:
Having a method such as `vtkMRMLMarkupsNode::SetMarkupPointIJK` sounds like a very bad idea to me. Behavior would be complex and error-prone if the volume is modified or transformed, potentially with non-linear transforms.

I'm working on updating markups widgets (to add not just point list but other widgets, such as angle and curve, improve performance to support thousands of points, etc.). Markups are already very complex, so I would not even consider complicating it with supporting manipulations in image-specific coordinate systems.

jcfr:
Agreed.

Having method with IJK in the their name that silently pass coordinate in RAS was also very confusing.

What about updating the SEM XML parser to throw a warning when "ijk" is specified as a coordinate system and suggest an alternative approach.

As you suggested, the CLI could have an explicit image argument that could be used for the conversion. 

I would like to provide some convenience function in the SlicerExecutionModel to avoid duplicating code.

cdeepakroy:
Users may want to pass in seeds in IJK as they are the most human friendly compared to LPS or RAS. When i am providing a small set of seeds, it is not easy to calculate their LPS or RAS in the mind. But it is easy to record the pixel/voxel locations (IJK coords) and pass it to the CLI.

The CLI developer standpoint,  as @jcfr @fedorov suggested making it mandatory to specify a `referenceImage` attribute for `point` and `pointfile` parameters in IJK space as follows will resolve the ambiguity

```
<point multiple="true" coordinateSystem="ijk" referenceImage="arg0">
```

Then the auto-generated UI within Slicer  for ijk point and pointfile parameters should also ask the user to select an associated reference image

dzenanz:
@cdeepakroy While that is convenient from CLI writer's point of view, it does not look like it is easy to implement in SEM and GUI (as @jcfr hinted [above](https://github.com/Slicer/Slicer/pull/888#issuecomment-363233603)).

pieper:
I don't mind that this is closed, but we should keep in mind that the CLI is designed to be an easy on-ramp for algorithm developers who have limited experience with concepts such as nonlinear transforms and frame of reference UIDs.

Not to generalize, but a lot of MICCAI papers are written by people who only used matlab with one or two sample datasets so while their math may be good they stumble hard when trying to make a usable tool.  I believe the original intent of this feature is to have Slicer handle the complexities while letting the CLI operate in a simpler (but still correct) pixel space.  This has the side benefit of centralizing the implementation of these transforms so that any improvements or fixes only need to happen inside Slicer.

lassoan:
I know that people struggle even with LPS/RAS transform and understanding why or how to multiply a point by a matrix is beyond their capabilities. However, we should be able to handle this with documentation, examples, and help via the Slicer forum.

> letting the CLI operate in a simpler (but still correct) pixel space

It is so rare that you can do operation in pixel space (ignoring pixel spacing). Most people do it by mistake.

Anyway, we could offer seed definition by voxel indices if Slicer's CLI infrastructure could take care of all transforms (apply them before sending to CLIs), but it would be a significant development task.

pieper:
I'm not excusing "algorithm developers" who don't understand pixel spacing.  I certainly expect developers to know the difference between algorithms that can ignore pixel spacing and those that cannot.  But the code to map through transforms and identify pixel coordinates should be handled in a uniform and correct manner.  Expecting each CLI to implement it from scratch is an obviously error prone approach.

Another solution could be to make sure that mrml logic classes can be easily re-used in CLIs to handle transform and spaces.  Or perhaps better, that we have clear examples about how to achieve the same using the corresponding ITK based methods.

PR #889, opened 2018-02-05
ENH: RFC bundle fixup script in python

ihnorton:
Add a new BundleFixup.py script to replace the CMakeScript
"fixup_bundle" routine. Advantages are:

  - speed: `make package` runs in ~3 min as compared to ~1 hr
  - debuggability: can be debugged in-place with standard pdb
    (note: For debugging, it is very helpful to create a git
           image of the package directory after the install step.
           Then the script can be run/re-run stand-alone, and
           reset without deleting or re-running other CMake steps)

There are two steps to the process:

  - find all dependencies. For this, we start from the `libs`
    passed by SlicerCPackBundleFixup (generated based on configured
    libraries). We look at all dependencies, and based on various
    criteria, may copy the dependency into the bundle. Similar to
    existing script, and uses modified versions of the existing regex
    filters to map source_lib -> target_dir in the bundle.

  - fixup step: this is a change from the existing script. We look
    at *all* Mach-O files in the potential bundle, and correct dylib
    references based on *which files are actually in the bundle*.
    System library or allow-listed references are ignored (currently
    only pqsql and mysql re allow-listed -- they are needed by the
    optional qt database drivers).

      - implicit verification -- bundling will fail if the dependency
        is not found or allow-listed.

      - we remove all absolute RPATH references. This should allow to
        run tests from the bundle itself on the factory, as a
        separate verification step (we can't check dlopen calls, but
        assuming Slicer only dlopens checked dylibs from inside the
        bundle...).


ihnorton:
Note, not ready to merge. To do:

- [ ] Extension support
- [ ] Windows support
- [ ] Linux support

PR #890, opened 2018-02-07
ENH: Added MRML node property macros

lassoan:
Added macros for reading, writing, copying, and printing MRML node properties.

It reduces amount of copy-pasted code and also allows higher quality implementation. For example, all special characters in strings are now properly encoded and error handling of reading enums or numbers is improved.

Updated vtkMRMLAbstractViewNode and vtkMRMLCameraNode to use these macros. In the future, more nodes should be updated in a similar way.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLAbstractViewNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLAbstractViewNode.cxx
@@ -36,25 +36,25 @@ static const char* DEFAULT_AXIS_LABELS[vtkMRMLAbstractViewNode::AxisLabelsCount]
 
 //----------------------------------------------------------------------------
 vtkMRMLAbstractViewNode::vtkMRMLAbstractViewNode()
+: LayoutLabel(NULL)
+, ViewGroup(0)
+, Active(0)
+, Visibility(1)
+, OrientationMarkerEnabled(false)
+, OrientationMarkerType(OrientationMarkerTypeNone)
+, OrientationMarkerSize(OrientationMarkerSizeMedium)
+, RulerEnabled(false)
+, RulerType(RulerTypeNone)
 {
-  this->LayoutLabel = NULL;
-  this->ViewGroup = 0;
-  this->Active = 0;
-  this->Visibility = 1;
-
-  double black[3] = {0.,0.,0.};
-  memcpy(this->BackgroundColor, black, 3 * sizeof(double));
-  memcpy(this->BackgroundColor2, black, 3 * sizeof(double));
+  this->BackgroundColor[0] = 0.0;
+  this->BackgroundColor[1] = 0.0;
+  this->BackgroundColor[2] = 0.0;
+  this->BackgroundColor2[0] = 0.0;
+  this->BackgroundColor2[1] = 0.0;
+  this->BackgroundColor2[2] = 0.0;
 
   this->SetLayoutLabel("1");
   this->SetHideFromEditors(0);
-
-  this->OrientationMarkerEnabled = false;
-  this->OrientationMarkerType = OrientationMarkerTypeNone;
-  this->OrientationMarkerSize = OrientationMarkerSizeMedium;
-
-  this->RulerEnabled = false;
-  this->RulerType = RulerTypeNone;
 
   this->AxisLabels = vtkSmartPointer<vtkStringArray>::New();
   for (int i=0; i<vtkMRMLAbstractViewNode::AxisLabelsCount; i++)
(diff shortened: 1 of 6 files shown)


Libs/MRML/Core/vtkMRMLNodePropertyMacros.h
```diff
@@ -0,0 +1,253 @@
...
+
+  See COPYRIGHT.txt
+  or http://www.slicer.org/copyright/copyright.txt for details.
+
+=========================================================================auto=*/
+
+#ifndef __vtkMRMLNodePropertyMacros_h
+#define __vtkMRMLNodePropertyMacros_h
+
+
```
jcfr: would be nice to have the macros documented in doxygen

jcfr: And mention that they have been designed to be used in method `vtkMRMLNode::Copy(), ... `

jcfr:
Thanks for working on this 👍 

Nitpick: Would be nice to have the macro documented in doxygen

Libs/MRML/Core/vtkMRMLAbstractViewNode.cxx
```diff
@@ -36,26 +36,26 @@ static const char* DEFAULT_AXIS_LABELS[vtkMRMLAbstractViewNode::AxisLabelsCount]
...
-
-  double black[3] = {0.,0.,0.};
-  memcpy(this->BackgroundColor, black, 3 * sizeof(double));
-  memcpy(this->BackgroundColor2, black, 3 * sizeof(double));
+  this->BackgroundColor[0] = 0.0;
+  this->BackgroundColor[1] = 0.0;
+  this->BackgroundColor[2] = 0.0;
+  this->BackgroundColor2[0] = 0.0;
+  this->BackgroundColor2[1] = 0.0;
+  this->BackgroundColor2[2] = 0.0;
```
ihnorton: These could also go in the initializer list as `BackgroundColor{}, BackgroundColor2{}`

lassoan: Good point. Earlier VS versions did not allow this, but this works now in VS2015:

<pre>
vtkMRMLAbstractViewNode::vtkMRMLAbstractViewNode()
  : LayoutLabel(NULL) 
  ...
  , Color{0.0, 0.0, 0.0}
{
...
</pre>

Do you know if this works on MacOSX and Linux compilers that Slicer-4.9 uses?

ihnorton:
Nice.

> In the future, more nodes should be updated in a similar way.

Now that we have C++11 enabled, at some point it might be worth considering [cereal](https://uscilab.github.io/cereal/index.html). The initial investment would be higher, but we would get JSON "for free" (in quotes because I'm sure that's ambitious and we would need to do some customization). Might be useful for the MRML-> nosql/json database project @pieper (and others?) were working on in the past.

Libs/MRML/Core/vtkMRMLAbstractViewNode.cxx
ihnorton: I checked clang on mac sdk 10.12 already -- works fine. I'm not sure if the linux compiler version is finalized yet, but this *should* be ok in any C++11 compliant compiler.

lassoan: OK, I'll update the initialization then.

lassoan:
Cereal looks OK but I'm not sure if it fits well our requirements. Anyway, if we serialize all data by macros then it could be possible to change the underlying serialization library with no or minimal modifications of MRML nodes.

Libs/MRML/Core/vtkMRMLAbstractViewNode.cxx
jcfr: Waiting Linux is also migrated. Could you hold on c++11 only feature ?

Libs/MRML/Core/vtkMRMLNodePropertyMacros.h
finetjul: I agree with @jcfr, I don't see what macros here do better than functions. 

```
ostream& vtkMRMLWriteXMLBoolean(ostream & xmlWriteOutputStream , const char* attributeName, bool propertyName)
{
  xmlWriteOutputStream << " "<< attributeName <<  "=\"" << propertyName ? "true" : "false" << "\"";
}
...

vtkMRML***Node::WriteXML(ostream& of, int nIndent)
{
...
   vtkMRMLWriteXMLBoolean(of, "layoutLabel", this->GetLayoutLabel());
...
}
```

Templated functions could also be used up to some extend.

lassoan: There are several reasons why macros more appropriate choice for this:

1. Simpler and more consistent syntax. Have a look at the current syntax and try to reproduce it without using macros.

2. Greater flexibility. We can create block with local variables, use the same argument as string and a method name, etc. allowing better performance, robustness, and keeping implementation simpler.

3. Future-proofing. Without changing existing MRML nodes, we can switch to use custom templated functions, serialization library, or (in a few years) standard C++ reflection.

jcfr: Hi @lassoan , 

Sounds good to me, I  just wanted to make sure we discuss it.

Could you add the three points written above in the commit message ?

jcfr:
Great. 👍  

After integrating few other fixes for tests, I will manually trigger the build and this change will be built and tested on all platform

ihnorton:
One side-effect here is that any change to `vtkMRMLNodePropertyMacros.h` affects vtkMRMLNode and causes recompilation of most of Slicer (twice today for me). I guess it will stabilize over time 😄

lassoan:
Yes, it's essentially in the vtkMRMLNode header, so if there is a change then a lot of things are re-built. We could move internal implementation into functions but probably would not worth the effort, as changes should be quite rare.

PR #891, opened 2018-02-07
BUG: Fix crash on exit when chart view has been opened

vovythevov:
The QWebEngineView would crash on exit. This seems related to https://bugreports.qt.io/browse/QTBUG-50160?focusedCommentId=305211&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-305211
The work-around is to delete manually the QWebEngineView (in this case
the qMRMLChartView) before the application has exited.

See issue #4479 - https://issues.slicer.org/view.php?id=4479

### DIFF
--- a/Libs/MRML/Widgets/qMRMLChartWidget.cxx
+++ b/Libs/MRML/Widgets/qMRMLChartWidget.cxx
@@ -16,6 +16,7 @@
 ==============================================================================*/
 
 // Qt includes
+#include <QCoreApplication>
 #include <QDebug>
 #include <QFileInfo>
 #include <QGridLayout>
@@ -89,6 +90,14 @@ void qMRMLChartWidgetPrivate::init()
                    this->ChartView, SLOT(setMRMLScene(vtkMRMLScene*)));
   QObject::connect(q, SIGNAL(mrmlSceneChanged(vtkMRMLScene*)),
                    this->ChartController, SLOT(setMRMLScene(vtkMRMLScene*)));
+
+#ifndef MRML_WIDGETS_HAVE_WEBKIT_SUPPORT
+  // The QWebEngineView crashes when the application quits.
+  // Work-around for now seems to be to delete the view before the application
+  // quits.
+  QObject::connect(QCoreApplication::instance(), SIGNAL(aboutToQuit()),
+                   q, SLOT(onAppAboutToQuit()));
+#endif
 }
 
 // --------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


Libs/MRML/Widgets/qMRMLChartWidget.cxx
```diff
@@ -161,3 +172,14 @@ vtkMRMLColorLogic* qMRMLChartWidget::colorLogic()const
   Q_D(const qMRMLChartWidget);
   return d->ChartView->colorLogic();
 }
+
+#ifndef MRML_WIDGETS_HAVE_WEBKIT_SUPPORT
+//---------------------------------------------------------------------------
+void qMRMLChartWidget::onAppAboutToQuit()
```
jcfr: Would be best to move this slot into the pimpl

Libs/MRML/Widgets/qMRMLChartWidget.cxx
```diff
@@ -161,3 +172,14 @@ vtkMRMLColorLogic* qMRMLChartWidget::colorLogic()const
...
+{
+  Q_D(qMRMLChartWidget);
```
jcfr: May be sensible to also add:

```
d->ChartView->setVisible(false);
d->ChartView->setParent(0);
```

jcfr:
Well done. Nice investigative work.

Could you update the chart view to use qSlicerWebWidget as a base class, and move the workaround into `qSlicerWebWidget` ?

Libs/MRML/Widgets/qMRMLChartWidget.cxx
vovythevov: Does it matter ? The whole application is shutting down, so that shouldn't change anything ?

Libs/MRML/Widgets/qMRMLChartWidget.cxx
vovythevov: You cannot have a slot in a pimpl when the class declaration is in the cpp. I would have to move the class declaration in a private header so Qt can run moc on it.
I thought against since it would involve moving a lot of code, but it might be worth it ? What do you think ?

vovythevov:
Make sense. Will do.

vovythevov:
Arf, qSlicerBaseQTGUI depends on qMRMLWidgets, so we cannot inherit from qSlicerWebWidget.
I'll add the hack to qSlicerWebWidget however to make sure doesn't crash the application

vovythevov:
@jcfr: I updated the commit to reflects your comments if you want to take a look.

jcfr:
Thanks @vovythevov 

After tweaking few things, the fix have been integrated in [r26902](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26902)

PR #893, opened 2018-02-20
Fix comment for deprecated method

fepegar:


jcfr:
Thanks for the patch.

Integrated in [r26935](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26935)

PR #896, opened 2018-02-21
BUG: Propagation of the ${additional_project}_SOURCE_DIR variable to the Slicer inner build

laurapascal:
The propagation of this variable is needed during the compilation of a Slicer Project Template in order to create the slicer's version: https://github.com/Slicer/Slicer/blob/94a26e017a3aba4a97c17182ffb913d872f897bc/CMake/SlicerVersion.cmake#L29


jcfr:
Thanks for the patch. Equivalent fix has been integrated in [r26954](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26954)

PR #898, opened 2018-02-23
Plenty chinese i18n ts files and code modification

u8621011:
Below is what this PR includes
1. traditional/simplified Chinese translation files
2. i18n related cmake script for loadable module
3. i18n related code fixes.


pieper:
I haven't reviewed in detail, but in principle this looks great!

I'd be mostly concerned about unintended side effects, so maybe we could find a way to generate to packages from this branch that people can test thoroughly.

jcfr:
I rebased and re-organize the changes.

I will add comment directly on each commit

jcfr:
A lot of the changes have been reviewed, fixed and integrated. Including support for Qt5.

Remaining changes to review are the one in this PR

Base/QTGUI/qSlicerDataDialog.cxx
```diff
@@ -94,6 +94,11 @@ qSlicerDataDialogPrivate::qSlicerDataDialogPrivate(QWidget* _parent)
   QPushButton* resetButton = this->ButtonBox->button(QDialogButtonBox::Reset);
   connect(resetButton, SIGNAL(clicked()), this, SLOT(reset()));
 
+  // do the internationalization setup for ButtonBox buttons
+  this->ButtonBox->button(QDialogButtonBox::Ok)->setText(tr("OK"));
+  this->ButtonBox->button(QDialogButtonBox::Cancel)->setText(tr("Cancel"));
+  this->ButtonBox->button(QDialogButtonBox::Reset)->setText(tr("Reset"));
```
jcfr: @u8621011 This will work for the two occurrences  found  of dialog but I think we should look for a more general approaches.

* Is there `.ts` files already available for Qt base classes ? 

* also wonder if using `QT_TRANSLATE_NOOP` as hinted [here](https://forum.qt.io/topic/47564/translate-qdialogbuttonbox/2) and [here](http://www.qtcentre.org/threads/16714-QDialogButtonBox-translation-of-standard-button-text-(-quot-OK-quot-quot-Cancel-quot-etc-)) could be more general

Base/QTCLI/qSlicerCLIModuleUIHelper.cxx
```diff
@@ -667,6 +668,9 @@ QWidget* qSlicerCLIModuleUIHelperPrivate::createImageTagWidget(const ModuleParam
   QString imageLabel = QString::fromStdString(moduleParameter.GetLabel());
   QString imageName = QString::fromStdString(moduleParameter.GetName());
 
+  imageLabel = QCoreApplication::translate(
+          this->CLIModuleWidget->moduleName().toLatin1(), imageLabel.toLatin1());
```
jcfr: @u8621011  It would be nice to add translation for all the other tags.

jcfr:
Few more issues:
* failed to initialize the dataprobe module: `No Data Probe frame - cannot create DataProbe`
* And few file that failed to be loaded in the translator:

```
The File  "qSlicerBaseQTCore_zh_cn.qm"  hasn't been loaded in the translator
The File  "qSlicerBaseQTApp_zh_cn.qm"  hasn't been loaded in the translator
The File  "SubjectHierarchy_zh_cn.qm"  hasn't been loaded in the translator
The File  "VolumeRendering_zh_tw.qm"  hasn't been loaded in the translator
```

Few more things we need to address:
* CLI module translation:
  * I envision that we would distribute a file named like `<modulename>_zh_cn.qm.ts`. We should then update https://github.com/Slicer/SlicerExecutionModel to ensure the logic generating and installing the translation files is used by the macro [SEMMacroBuildCLI](https://github.com/Slicer/SlicerExecutionModel/blob/master/CMake/SEMMacroBuildCLI.cmake)
* distribution of `*.qm` files within Slicer extension


Otherwise, I was able to start Slicer built with Qt5:

slicer-i18n

Base/QTCLI/qSlicerCLIModuleUIHelper.cxx
u8621011: OK, I will add translation for all the other tags later.

lassoan: It would be also better not to use the CLI module name as translation context but indicate that it is a CLI module translation, by using a prefix, such as "CLI.". Something like this:

    CLINodeTranslationContextPrefix = "CLI."
    ...
    imageLabel = QCoreApplication::translate((CLINodeTranslationContextPrefix+this->CLIModuleWidget->moduleName()).toLatin1(), imageLabel.toLatin1());

u8621011: @lassoan why better not to use the CLI module name as translation context?

lassoan: Just the module name is not specific enough. We need to know that a particular context belongs to a CLI module.

u8621011: But loadable modules either don't have prefix to identifiy?

jcfr: Loadable module are associated with the full classname, which ensure the context is unique enough. 

What Andras suggested is way to achieve a similar things but for CLIs.

u8621011: I got it,  I have added for other tags with older naming pattern and will change to what Andras suggested before commit it.

jcfr: Thanks @u8621011 

Could you share an example of translation file for a CLI ? (at list only few field)

Does it have to be created by hand ?  

If yes, it could probably be generated automatically based on on the XML file. (using a python script)

u8621011: I do the translation by hand.

When I translating the CLI module, i found qt can direct parse xml for us please refer the link below

https://stackoverflow.com/questions/28166081/can-i-use-tr-and-lupdate-on-xml-files

But I don't know how to script cmake to do that job automatically so I do it handily.

any idea?

u8621011: @jcfr the CLI Module we translate likes steps below.

1. modify our version xml file, partial BRAINSFit.xml for example
```
  <parameters advanced="false">
    <label>Output Settings (At least one output must be specified)</label>
    <transform fileExtensions=".h5,.hdf5,.mat,.txt" type="linear" reference="movingVolume">
      <name>linearTransform</name>
      <longflag>linearTransform</longflag>
      <label>QT_TRANSLATE_NOOP("BRAINSFit","Slicer Linear Transform")</label>
      <description>(optional) Output estimated transform - in case the computed transform is not BSpline. NOTE: You must set at least one output object (transform and/or output volume).</description>
      <channel>output</channel>
    </transform>
    <transform fileExtensions=".h5,.hdf5,.mat,.txt" type="bspline" reference="movingVolume">
      <name>bsplineTransform</name>
      <longflag>bsplineTransform</longflag>
      <label>QT_TRANSLATE_NOOP("BRAINSFit","Slicer BSpline Transform")</label>
      <description>(optional) Output estimated transform - in case the computed transform is BSpline. NOTE: You must set at least one output object (transform and/or output volume).</description>
      <channel>output</channel>
    </transform>
    <image>
      <name>outputVolume</name>
      <longflag>outputVolume</longflag>
      <label>QT_TRANSLATE_NOOP("BRAINSFit","Output Image Volume")</label>
      <description>(optional) Output image: the moving image warped to the fixed image space. NOTE: You must set at least one output object (transform and/or output volume).</description>
      <channel>output</channel>
    </image>
  </parameters>
```

2. run qt lupdate over our xml
```
C:\D\Support\qt-4.8.7-64-vs2013-deb\bin\lupdate BRAINSFit.xml -ts  BRAINSFit_zh_tw.ts
```

3. translated the generated ts file.
4. copy lupdate generated ts context to some cmake generated ts file by hand.

the step 2 & 4 should possibly be done in cmake but we don't know how to script it.
Any ideas?

jcfr: Thanks, now you provided the step it will be easy to automate the generation.

For step 2, we could ensure that an intermediate file named `BRAINSFit.xml.tr` is generated based on the original xml file. This file could then be given as input to `lupdate`.

Here are two possible approaches:
* (1) all strings in the same context (e.g `BRAINSFit`)
* (2) label/description associated with their corresponding parameter group context (e.g `BRAINSFit.Input Images ...`

```
import os
import xml.etree.ElementTree as ET

def tr(context, value):
    return 'QT_TRANSLATE_NOOP("%s", "%s")' % (context, value)

def generate_lupdate_input_v1(xmlFile, candidates=["label"]):
    """Process CLI XML description file and generate an new XML file
    including `QT_TRANSLATE_NOOP()` for each `candidates` xml tags.
    """
    moduleName = os.path.splitext(xmlFile)[0]

    tree = ET.parse(xmlFile)
    root = tree.getroot()

    for candidate in candidates:
        for label in root.iter(candidate):
            label.text = tr(moduleName, label.text)

    tree.write(xmlFile + ".v1.tr")
    
def generate_lupdate_input_v2(xmlFile, candidates=["label"]):
    """Process CLI XML description file and generate an new XML file
    including `QT_TRANSLATE_NOOP()` for each `candidates` xml tags.
    """
    moduleName = os.path.splitext(xmlFile)[0]

    tree = ET.parse(xmlFile)
    root = tree.getroot()
    
    for parameters in root.findall("./parameters"):
        context = moduleName + "." + parameters.find("label").text
        labels = parameters.findall(".//label")
        for label in labels:
            label.text = tr(context, label.text)

    tree.write(xmlFile + ".v2.tr")

generate_lupdate_input_v1("BRAINSFit.xml")
generate_lupdate_input_v2("BRAINSFit.xml")
```

... and corresponding screenshot for the `linguist` application:

### Without parameter group context

linguist_v1

### With parameter group context

linguist_v2

jcfr: I think  approach  (2) with context will make translation work easier. What do you think ? 

After we agree on an approach, it will be easy to integrate it with the CLI build system.

u8621011: I feel if the CLI cmake script can parse and compile ts into qm for every cli module directory, approach should be ok.(eg, each CLI module has its own ts/qm file)

Approach 2 looks too long the name but if we have to put every cli module sentences into same/single ts file, approach 2 is necessary.

lassoan: Need to be careful with enumerated strings. While the values should be translated in the GUI, the original text should be sent on the command-line.

For example:

https://github.com/Slicer/Slicer/blob/aa07f6b274778f4cbab626b35228bea8768d38ce/Modules/CLI/ThresholdScalarVolume/ThresholdScalarVolume.xml#L31-L40

User will see "outside", "below", "above" translated on GUI, but when the selection is made, the CLI module must receive the original "outside", "below", "above" values.

This goes the other direction, too. Returning strings is rare, but eventually, any strings returned by the CLI should be possible to translate.

PR #900, opened 2018-02-23
ENH: Separated volume rendering preset combobox from module widget

cpinter:
The Volume Rendering module widget handled everything, which made the class quite big, and also prevented reusing the preset combobox in other modules or slicelets.

- Added reusable qSlicerVolumeRenderingPresetComboBox class that contains the combobox and the shift slider
- Added EffectiveRange member to vtkMRMLVolumePropertyNode. It sets a custom extent to the transfer functions so that only the meaningful part of it is shown (and not the long tails). Thus the shift slider is much less sensitive and is more user friendly
- Added effectiveRange attribute to presets.xml. For now the effective range is between the second and last second points for CTs, first and second last for MRs
- Removed calculation of transfer function extent based on volume scalar range. The effective range is always valid, and contains a meaningful range (if not then encompasses the transfer function bounds), so this is not necessary anymore (and the combobox class is also cleaner)
- Instead of only moving the middle points of the transfer function, all the points are moved. This fixes the bug of squashing the transfer function if it was shifted beyond its extreme points (bounds)


cpinter:
Known issue: when the shift slider is dragged so that the extent (effective range) of the transfer function is outside the bounds, the extent shrinks. This does not cause any noticeable issues, but the user may want to resize the extent afterwards. The reason I could not preserve the size of the extent is that if the bounds changes (which does on moveAllPoints, when the transfer function's control points are translated), then the extent is enforec to be valid at the vtkAxis level (SetMinimumLimit), and by the time I would have the chance in qSlicerVolumeRenderingModuleWidget, it is already changed (or if done too early then neither are changed). I tried to calculate the new bounds based on the offset, but it wasn't stable. I decided to leave this in, as this is already much better than truncating the transfer function when it is shifted beyond the bounds, as it was before.

pieper:
> Instead of only moving the middle points of the transfer function, all the points are moved. This fixes the bug of squashing the transfer function if it was shifted beyond its extreme points (bounds)

Great to see this bug finally fixed!

jcfr:
Well done. 

CTK updated in [r26953](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26953)

jcfr:
Also, could you put all style/indent/... changes in one commit ?

cpinter:
Thanks @pieper and @jcfr! I've integrated this in https://github.com/Slicer/Slicer/commit/f4b50114908e98b1b6226f4b2e75bd95eb01e75b

jcfr:
Thanks for integrating. 

Look there is a small issue, when grabbing the range and moving it, the VR is not updated.

vr-not-updated

lassoan:
That's just for adjusting what portion of the transfer function is displayed. It does not modify the function.

cpinter:
Yes those sliders only affect the extent of the transfer functions that the shift slider on the top considers, so even before this change they didn't change the volume rendering transfer functions.

jcfr:
Good point. Makes sens.

PR #901, opened 2018-02-25
ENH: Added accessor functions to layout manager and factory

cpinter:
- Layout manager factory can be accessed from python
- View widgets can be retrieved by their layout label
- 3D views can be retrieved by name

### DIFF
--- a/Libs/MRML/Widgets/qMRMLLayoutManager.cxx
+++ b/Libs/MRML/Widgets/qMRMLLayoutManager.cxx
@@ -961,6 +961,13 @@ qMRMLThreeDWidget* qMRMLLayoutManager::threeDWidget(int id)const
 }
 
 //------------------------------------------------------------------------------
+qMRMLThreeDWidget* qMRMLLayoutManager::threeDWidget(const QString& name)const
+{
+  return qobject_cast<qMRMLThreeDWidget*>(
+    this->mrmlViewFactory("vtkMRMLViewNode")->viewWidget(name));
+}
+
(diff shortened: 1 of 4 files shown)


lassoan:
LGTM!

Libs/MRML/Widgets/qMRMLLayoutManager.cxx
```diff
@@ -960,6 +960,13 @@ qMRMLThreeDWidget* qMRMLLayoutManager::threeDWidget(int id)const
     this->mrmlViewFactory("vtkMRMLViewNode")->viewWidget(id));
 }
 
+//------------------------------------------------------------------------------
+qMRMLThreeDWidget* qMRMLLayoutManager::threeDWidget(const QString& name)const
+{
+  return qobject_cast<qMRMLThreeDWidget*>(
+    this->mrmlViewFactory("vtkMRMLViewNode")->viewWidget(name));
```
jcfr: Check for null pointer

jcfr:
:+1:

Libs/MRML/Widgets/qMRMLLayoutManager.cxx
cpinter: OK I'll add that to the other similar functions too then.

cpinter:
Thanks for the reviews! Integrated in https://github.com/Slicer/Slicer/commit/7ffed00965a12b3dd5d97f6f1be303c4961cc8b4 with the required changes.

PR #902, opened 2018-02-25
STYLE: Simplify Main.cxx introducing (pre|post)InitializeApplication function

jcfr:
This allows to remove duplicated code from Slicer-based application and
streamline the associated maintenance.

### DIFF
--- a/Applications/SlicerApp/Main.cxx
+++ b/Applications/SlicerApp/Main.cxx
@@ -32,10 +32,8 @@ namespace
 //----------------------------------------------------------------------------
 int SlicerAppMain(int argc, char* argv[])
 {
-  typedef qSlicerAppMainWindow SlicerMainWindowType;
-  typedef qSlicerStyle SlicerAppStyle;
-
-  qSlicerApplicationHelper::preInitializeApplication<SlicerAppStyle>(argv[0]);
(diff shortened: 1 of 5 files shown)


jcfr:
Since both style and main window are instantiated instantiated in the
methods at different time , this is why. There are also independent
objects.

I will look at the python issue.

lassoan:
It is just highly unusual to sneak in a constructor by using a templated method to create an object inside. Normally you just create the object outside and pass it to the method as an argument.

jcfr:
Agreed and thanks for the review. I am updating the approach.

jcfr:
> Why can't we create the object outside and pass it to the method as an argument?

The rational was to create the object only when needed. 

This means that when Slicer is executed with `--no-main-window`, or any other command that do not required the main windows (eg `--help`, ...), it is not instantiated.

Since the Style is lightweight, I moved it as parameter.

I will investigate if there is an overhead of always creating the main window even if not needed

lassoan:
Thank you, it looks good.

jcfr:
Closing. Integrated in [r26964](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26964) and [r26965](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26965)

PR #904, opened 2018-02-27
Minor bug fixes

dzenanz:


pieper:
LGTM

lassoan:
It looks good, but I think it is not enough to fix it there. For example, these lines should be updated, too:

https://github.com/Slicer/Slicer/blob/ac8c769012bf6dc3253ea1ead778cea6bc2c77d8/Base/Logic/vtkSlicerApplicationLogicRequests.h#L37

https://github.com/Slicer/Slicer/blob/ac8c769012bf6dc3253ea1ead778cea6bc2c77d8/Base/Logic/vtkSlicerApplicationLogic.cxx#L492

dzenanz:
Thanks for pointing that out Andras!

dzenanz:
circleCI compile error `/usr/src/Slicer/Libs/MRML/Widgets/qMRMLPlotView.cxx:547:21: error: 'nullptr' was not declared in this scope` has nothing to do with this PR. Merge or rebase once the error is fixed?

lassoan:
I guess the error is due to Linux factory still not has been updated to support cxx11. I think you can ignore the error, as it is unrelated to your changes.

jcfr:
CircleCi will be updated in the next few days.

In the mean time, we have been working on improving the associated build images based on centos. (now moved into this project: https://github.com/dockcross/centos-build)

jcfr:
Thanks @dzenanz 

Integrated in [r26968](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26968) and [r26969](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26969)

PR #906, opened 2018-03-02
Add tooltip for Reduction in Decimation section of Surface Toolbox module

fepegar:
It might not be clear if 0.25 means reduce the size to 75% or to 25%. I've adapted the tooltip from VTK docs: https://www.vtk.org/doc/nightly/html/classvtkDecimatePro.html#a2889d06be4ebee23a5c02e69e8b2a946


cpinter:
Based on my memories (without looking at any doc or code) 0.25 means that the 75% of the vertices will be preserved.

fepegar:
Yes, that how it works. But not everyone knows and it was not clear to me, that's why I created this PR.

cpinter:
Oh I see, I thought it was a question :) Makes sense!

lassoan:
I'll merge this now.

fepegar:
Thanks Andras.

lassoan:
Committed r26987

PR #907, opened 2018-03-04
BUG: Fixed hang when volume rendering in one-up 3D view

lassoan:
If annotation ROI widget was placed or handle picker was used without initializing the render window then it caused application GUI to stop updating.
Fixed by adding checks to prevent widget use until render window is at least rendered once.

Fixes https://issues.slicer.org/view.php?id=4516.

### DIFF
--- a/Modules/Loadable/Annotations/MRMLDM/vtkMRMLAnnotationROIDisplayableManager.cxx
+++ b/Modules/Loadable/Annotations/MRMLDM/vtkMRMLAnnotationROIDisplayableManager.cxx
@@ -38,6 +38,7 @@
 #include <vtkProperty.h>
 #include <vtkRenderer.h>
 // for picking
+#include <vtkRenderWindow.h>
 #include <vtkRenderWindowInteractor.h>
 #include <vtkTransform.h>
 #include <vtkTransformPolyDataFilter.h>
@@ -419,6 +420,11 @@ void vtkMRMLAnnotationROIDisplayableManager::PropagateMRMLToWidget(vtkMRMLAnnota
 /// Propagate properties of MRML node to widget.
 void vtkMRMLAnnotationROIDisplayableManager::PropagateMRMLToWidget2D(vtkMRMLAnnotationNode* node, vtkAbstractWidget * widget)
 {
+  if (!this->GetRenderer() || !this->GetRenderer()->GetRenderWindow() || this->GetRenderer()->GetRenderWindow()->GetNeverRendered())
+    {
+    return;
+    }
+
   vtkMRMLSliceNode *sliceNode = this->GetSliceNode();
   if (!sliceNode)
     {
--- a/Modules/Loadable/Annotations/VTKWidgets/vtkAnnotationROIRepresentation2D.cxx
+++ b/Modules/Loadable/Annotations/VTKWidgets/vtkAnnotationROIRepresentation2D.cxx
@@ -37,6 +37,7 @@
 #include <vtkProperty2D.h>
 #include <vtkPropPicker.h>
 #include <vtkRenderer.h>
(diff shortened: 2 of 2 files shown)


Modules/Loadable/Annotations/MRMLDM/vtkMRMLAnnotationROIDisplayableManager.cxx
```diff
@@ -419,6 +420,11 @@ void vtkMRMLAnnotationROIDisplayableManager::PropagateMRMLToWidget(vtkMRMLAnnota
 /// Propagate properties of MRML node to widget.
 void vtkMRMLAnnotationROIDisplayableManager::PropagateMRMLToWidget2D(vtkMRMLAnnotationNode* node, vtkAbstractWidget * widget)
 {
+  if (!this->GetRenderer() || !this->GetRenderer()->GetRenderWindow() || this->GetRenderer()->GetRenderWindow()->GetNeverRendered())
+    {
+    return;
```
jcfr: Out of curiosity, is this triggered by a ResetCamera ? 

See https://github.com/Kitware/VTK/blob/81004dd80253ba8f19caeea06f4ab51ac782d2c5/Rendering/Label/vtkLabelPlacer.cxx#L378-L386

lassoan: All the problems occurred when 3D-only view was enabled (and slice views were never instantiated).

1. Window updates stopped working when entering Volume Rendering module due to Pick operation invoked using a hardware-based picker, on an unitinitalized RenderWindow. It **probably would be good to check in vtkPropPicker::Pick if the render window has been ever rendered and return with empty picking result if not (instead of messing up the OpenGL render context)**.

After adding a check that prevented the above problem, another scenario came up:

2. Window updates stopped working when ROI visibility was set to true. I could not pinpoint exactly what caused the problem here, but adding a check that prevented all widget updates until the render window was rendered once.

lassoan:
Committed r27000

Modules/Loadable/Annotations/MRMLDM/vtkMRMLAnnotationROIDisplayableManager.cxx
jcfr: Thanks for the explanation.

Do you think you could submit a MR to upstream VTK ? The idea would be to update:
* `vtkPropPicker::PickProp`
* `vtkOpenVRHardwarePicker::PickProp`

lassoan: Yes, it should be definitely fixed. I'm not sure at what level this should be captured or handled. It seems that OpenGL2 backend is somehow more sensitive to this (especially when combined with volume rendering), so maybe it is an underlying bug somewhere at lower level, which should be fixed there, instead of masked by a high-level check.

jcfr: Good point.

@sankhesh @martinken Do you any insight on this ?

msmolens: Note that this fixes the `py_VolumeRenderingThreeDOnlyLayout` test for the OpenGL2 backend. I introduced that test while fixing a similar bug in the old backend when starting the app in 3D-only layout and entering Volume Rendering (see http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25239).

In the old backend it was sufficient to call `Render()` on the render window to ensure that an OpenGL context is created even if the widget is not visible. This isn't sufficient for the OpenGL2 backend.

PR #908, opened 2018-03-04
ENH: Extend volume if non-linear transform is hardened

lassoan:
Voxel grid of the volume was not changed when non-linear transform was hardened on it, which could cause clipping of the volume.

Now volume is automatically extended to contain the transformed volume.

Fixes https://issues.slicer.org/view.php?id=4433


lassoan:
Integrated in r27006

PR #910, opened 2018-03-08
ENH: Speedup by 50% handling of argument expected to display a msg and exit

jcfr:
On my workstation, executing "Slicer --help" or "Slicer --version" went
down from ~1.3s to ~0.6s.

This commit introduces the function quickExit allowing to quickly
terminate the executable.

After supporting exclusively c++11, the corresponding function from
the standard library will be used.
See http://en.cppreference.com/w/cpp/utility/program/quick_exit

### DIFF
--- a/Base/QTCore/qSlicerCoreApplication.cxx
+++ b/Base/QTCore/qSlicerCoreApplication.cxx
@@ -393,6 +393,17 @@ void qSlicerCoreApplicationPrivate::init()
 }
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


jcfr:
Integrated in [r27015](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27015)

lassoan:
Does this disable VTK's memory leak reporting? If yes, then maybe quick exit should only be enabled if leak reporting is disabled.

PR #911, opened 2018-03-08
ENH: Added performance related options to GPU volume rendering

cpinter:
- Added option to use the mapper's feature LockSampleDistanceToInputSpacing to calculate it internally instead of setting sampling distance manually. Off by default. This applies only for GPU rendering. If it's on, then AutoAdjustSampleDistances is disabled, because it prevents using this feature. Because of this, if the flag is on, then the MaximumQuality and AdaptiveQuality setting does not work
- Added UseJittering option to GPU display node. If turned on, then uses random noise to remove the wood-grain artifact. Off by default. Not exposed on the UI yet
- Widened range of FPS slider for adaptive mode to accommodate for new virtual reality use cases (maximum used to be 20, now it is 100)
- Removed empty "Misc" tab from Volume rendering module UI
- Fixed ctkDoubleSlider related warning message
- Modernized vtkMRMLVolumeRenderingDisplayNode to use node references and new property macros. This fixed several bugs as well (e.g. not reading/writing attributes)

### DIFF
--- a/Modules/Loadable/VolumeRendering/MRML/vtkMRMLGPURayCastVolumeRenderingDisplayNode.cxx
+++ b/Modules/Loadable/VolumeRendering/MRML/vtkMRMLGPURayCastVolumeRenderingDisplayNode.cxx
@@ -34,6 +34,8 @@ vtkMRMLNodeNewMacro(vtkMRMLGPURayCastVolumeRenderingDisplayNode);
 vtkMRMLGPURayCastVolumeRenderingDisplayNode::vtkMRMLGPURayCastVolumeRenderingDisplayNode()
 {
   this->RaycastTechnique = vtkMRMLGPURayCastVolumeRenderingDisplayNode::Composite;
+  this->UseJittering = 0;
+  this->LockSampleDistanceToInputSpacing = 0;
 }
 
 //----------------------------------------------------------------------------
@@ -46,20 +48,11 @@ void vtkMRMLGPURayCastVolumeRenderingDisplayNode::ReadXMLAttributes(const char**
 {
   this->Superclass::ReadXMLAttributes(atts);
 
-  const char* attName;
-  const char* attValue;
-  while (*atts != NULL)
-    {
-    attName = *(atts++);
-    attValue = *(atts++);
-    if (!strcmp(attName,"raycastTechnique"))
-      {
-      std::stringstream ss;
-      ss << attValue;
-      ss >> this->RaycastTechnique;
-      continue;
-      }
-    }
+  vtkMRMLReadXMLBeginMacro(atts);
+  vtkMRMLReadXMLIntMacro(raycastTechnique, RaycastTechnique);
+  vtkMRMLReadXMLIntMacro(useJittering, UseJittering);
+  vtkMRMLReadXMLIntMacro(lockSampleDistanceToInputSpacing, LockSampleDistanceToInputSpacing);
+  vtkMRMLReadXMLEndMacro();
 }
 
 //----------------------------------------------------------------------------
@@ -67,7 +60,11 @@ void vtkMRMLGPURayCastVolumeRenderingDisplayNode::WriteXML(ostream& of, int nInd
 {
   this->Superclass::WriteXML(of, nIndent);
 
-  of << " raycastTechnique=\"" << this->RaycastTechnique << "\"";
+  vtkMRMLWriteXMLBeginMacro(of);
+  vtkMRMLWriteXMLIntMacro(raycastTechnique, RaycastTechnique);
+  vtkMRMLWriteXMLIntMacro(useJittering, UseJittering);
+  vtkMRMLWriteXMLIntMacro(lockSampleDistanceToInputSpacing, LockSampleDistanceToInputSpacing);
+  vtkMRMLWriteXMLEndMacro();
 }
 
 //----------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


cpinter:
This is an experimental change about possible performance options to use, so I'd appreciate if those ho are interested tried it out and gave feedback about how they worked for them. How to try the new options:
`vrdn=getNode('vtkMRMLGPURayCastVolumeRenderingDisplayNode1')` 
`vrdn.SetLockSampleDistanceToInputSpacing(1)`
`vrdn.SetUseJittering(1)`

Performance optimization is needed because in the new VR use cases the interactive mode where the sampling distance is adjusted to increase frame rate is not available, and volume rendering can be very slow especially in the case of large volumes. Low frame rate may cause health problems too, so this is something to address.

Best result seems to be with both options turned on. LockSampleDistanceToInputSpacing increases performance while UseJittering increases quality while performance stays the same. Please read commit message for details.

This commit also contains general code improvements, which should go to the core no matter what the decision will be about the new options.

Thanks for your reviews!

cpinter:
Good point @jcfr! I'll run the tests and fix baselines as needed.

I'd appreciate getting feedback about the changes you experience with these options on all platforms.

As the volume rendering method changed a lot in VTK, the current defaults and UI options are no longer relevant (just see difference between 4.8.1 and master: with 4.8.1 you get a consistent frame rate with all volume sizes, and the master behaves very differently). It would be great if we could eventually decide what defaults and UI options we want for the new volume rendering method.

lassoan:
These new options make volume rendering much much nicer (faster AND higher quality).

I would add a checkbox for jittering (find a user-friendlier name, maybe "Surface smoothing") and enable LockSampleDistanceToInputSpacing in "Adaptive mode".

hherhold:
This is tons better. Rotating and scaling are now much faster and very useable.

I've enabled both:

`vrdn.SetLockSampleDistanceToInputSpacing(1)`
`vrdn.SetUseJittering(1)`

I will do some more testing with larger volumes (both in dimensions and 8 vs 16 bit data) and update. This is on MacOS 10.12.6 with a Radeon R9 M370X 2048MB.

Thanks!!

cpinter:
It seems that so far there is an agreement that using the two new options is better than the default.

It would make sense to change the default to use both options, but first I'd like to ask for more opinions: @pieper @jcfr @ihnorton ?
Good to keep in mind that the LockSampleDistanceToInputSpacing makes the MaximumQuality option unusable, so if we decide to turn on the new settings by default, this will need to be dealt with.

lassoan:
Maximum quality still provides higher-quality rendering (even if on most systems, the difference is small) it is already implemented, and users already know this option. It is also useful when we temporarily want to enable highest-quality rendering (e.g., for acquiring screen captures).

hherhold:
I think I saw this elsewhere and it probably doesn't have anything to do with this commit, but changing rendering quality has no effect.

lassoan:
> I think I saw this elsewhere and it probably doesn't have anything to do with this commit, but changing rendering quality has no effect.

Enabling maximum quality does improve volume rendering quality. https://issues.slicer.org/view.php?id=1996 The reason is that by enabling maximum quality, you aggressively oversample the volume, which may result in visible improvement at surfaces. With jittering enabled, the difference is even smaller, but is still there, as jittering does not increase the sample count, just slightly randomizes the sampling, making the woodgrain artifact less visible (in some cases completely invisible).

lassoan:
If we add the jittering option then I think https://issues.slicer.org/view.php?id=1996 can be set to resolved. The difference is very small that it should not be a problem in general (and for special cases, such as taking screenshots, maximum option can be enabled).

pieper:
These all sound like great improvements to me - 👍

hherhold:
I've tried setting rendering quality to Adaptive and Maximum, and I see no difference in quality or frame rate. (This is contrasted with 4.8.1 where there is a big difference in quality and frame rate.) I can post a screen capture movie if that would help.

cpinter:
This is known, please read above

On March 10, 2018 13:52:51 Hollister Herhold <notifications@github.com> wrote:

I've tried setting rendering quality to Adaptive and Maximum, and I see no difference in quality or frame rate. (This is contrasted with 4.8.1 where there is a big difference in quality and frame rate.) I can post a screen capture movie if that would help.

hherhold:
Ugh, I completely missed the "Because of this, if the flag is on, then the MaximumQuality and AdaptiveQuality setting does not work". My mistake - sorry about that.

cpinter:
Yes, these issues are the reason I asked for as many opinions as possible, because results seem somewhat ambivalent, especially compared to 4.8.1 :)

hherhold:
4.8.1 vs this version - well, the volume renderer obviously operates quite differently. Opacity mapping values from 4.8.1 do not work the same as with the newer VTK, but that is possibly a comment for another discussion...? 

As far as performance goes, the frame rate with these modifications is pretty much identical to 4.8.1 at maximum quality, which seems to be consistent with comments from above.

My overall uneducated impression is that the new VTK volume renderer "looks better" - I feel like I can see more detail, but it's kind of hard to quantify "quality".

cpinter:
Integrated in https://github.com/Slicer/Slicer/commit/4e01304f0efb5e87c43edac465d2b064f4b8230a with the following changes compared to this PR:
* LockSampleDistanceToInputSpacing is on by default in Adaptive mode when using GPU
* Maximum quality mode is the same as before
* Added jittering on UI in advanced rendering properties, with the name Surface smoothing (jittering does not say much to a user), with a tooltip explaining what it is
* Fixed vtkMRMLVolumePropertyNodeTest1 that failed because it expected warnings due to the obsolete reference handling method of the display node. As the display node is now updated, this warning does not appear

PR #912, opened 2018-03-08
ENH: added functionality to set/get operator name on application level

che85:
- name will be stored under QSettings "OperatorName"

for the reference: https://github.com/QIICR/QuantitativeReporting/issues/201#issuecomment-371258260

### DIFF
--- a/Base/QTGUI/qSlicerApplication.h
+++ b/Base/QTGUI/qSlicerApplication.h
@@ -101,6 +101,10 @@ public:
   Q_INVOKABLE qSlicerLayoutManager* layoutManager()const;
   Q_INVOKABLE void setLayoutManager(qSlicerLayoutManager* layoutManager);
 
+  /// Set/Get operator name
+  Q_INVOKABLE QString operatorName()const;
+  Q_INVOKABLE void setOperatorName(const QString &name);
+
   /// Return a pointer on the main window of the application if any.
   QMainWindow* mainWindow()const;
 
--- a/Base/QTGUI/qSlicerApplication.cxx
+++ b/Base/QTGUI/qSlicerApplication.cxx
@@ -483,6 +483,23 @@ qSlicerLayoutManager* qSlicerApplication::layoutManager()const
 }
 
 //-----------------------------------------------------------------------------
+void qSlicerApplication::setOperatorName(const QString &name)
+{
+  QSettings* appSettings = this->userSettings();
+  Q_ASSERT(appSettings);
+  appSettings->setValue("OperatorName", name);
+}
+
+//-----------------------------------------------------------------------------
+QString qSlicerApplication::operatorName()const
+{
+  Q_D(const qSlicerApplication);
+  QSettings* appSettings = this->userSettings();
+  Q_ASSERT(appSettings);
+  return appSettings->value("OperatorName", "").toString();
+}
+
+//-----------------------------------------------------------------------------
 QMainWindow* qSlicerApplication::mainWindow()const
 {
   foreach(QWidget * widget, this->topLevelWidgets())


Base/QTGUI/qSlicerApplication.h
```diff
@@ -101,6 +101,10 @@ class Q_SLICER_BASE_QTGUI_EXPORT qSlicerApplication : public qSlicerCoreApplicat
   Q_INVOKABLE qSlicerLayoutManager* layoutManager()const;
   Q_INVOKABLE void setLayoutManager(qSlicerLayoutManager* layoutManager);
 
+  /// Set/Get operator name
+  Q_INVOKABLE QString operatorName()const;
+  Q_INVOKABLE void setOperatorName(const QString &name);
```
jcfr: Thanks for your contribution.

Could you:

* provide a little more context by describing what is an `OperatorName` ?

* move the method into the `qSlicerCoreApplication`

* Remove the use of `Q_INVOKABLE` and instead use `Q_PROPERTY`

fedorov:
Just an idea - should we also consider "OperatorLoginName", "OperatorUID"?

che85:
@fedorov "OperatorLoginName" sounds reasonable to me. "OperatorUID" is problematic because you cannot really make sure that it's a unique name.

jcfr:
Thinking more about this ...

We can definitively add the method to the application, that said I think this is too specific. Also, at that stage the same effect can easily be achieved using `slicer.app.userSettings().setValue("OperatorName", "John Doe")`

Also, instead of storing this at the top-level of the settings,  I suggest to add them into a settings [group](https://doc.qt.io/qt-5/qsettings.html#beginGroup) that could be named (e.g `User`, or `Operator`, or `Profile` ...)

While the concepts and naming convention are ironed out, what about having a `UserProfile` or `Provenance` or `Audit`extension ?  

There, a settings panel could be developed, mrml node to store info in scene, logic to audit operation, ..

lassoan:
What is required in the DICOM standard? Real person name and/or some user ID?

che85:
image

http://dicom.nema.org/medical/dicom/current/output/chtml/part06/chapter_6.html#para_1d93f33c-3935-49db-97c0-2dcad981f4aa

image

http://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_6.2.html#para_31eaa241-35b0-4fd1-b2b9-53185f5d50e5

lassoan:
We should probably add support for these. They could be stored in application settings, under one group, something like "OperatorIdentification", ""Operator", "UserInformation". Within that, we could have Name, LoginName, Organizationname, RoleInOrganization, RoleInProcedure, IdentifierWithinRole, and encode them in string as usual (PNAME as text, with separators between name parts as defined in DICOM, ^; CODE as other coded values - coding scheme, value, meaning; and the rest as simple text).

pieper:
+1

lassoan:
We may use these values in different places, wherever patient identification is needed, for example we would want to be able to populate Person Identification Macro Attributes (http://dicom.nema.org/medical/dicom/2014c/output/chtml/part03/chapter_10.html#table_10-1.) from the same fields.

I would not use this very specific field names, as it would suggest a that it cannot be used in other contexts: an operator does not necessarily observe but can perform or verify a procedure etc. Also, the names ("Person observer name", etc.) are somewhat cryptic - it is difficult to guess what it means and why the name is exactly this.

fedorov:
@lassoan I agree with you. 

How about the hierarchy:

```
UserInformation
> Name
> Login
> Email
> Organization
# refer to http://dicom.nema.org/medical/dicom/current/output/chtml/part16/sect_CID_7452.html for reference
> OrganizationRole 
# refer to http://dicom.nema.org/medical/dicom/current/output/chtml/part16/sect_CID_7453.html for reference
> ProcedureRole
```

Anything else that anyone feels should be included?

We can keep all of those empty by default. I am not sure we need a popup on startup, since this functionality will not be needed by most modules, and it creates a potential for identifiable information leakage. For that same reason I am not convinced it should be stored in the MRML scene, but only maintained in the local settings file.

We can also add a "User information" panel to the applications settings, but we can't spend too much time on this, since this is just a component for several other applications. We really don't have time to develop an extension and then iron things out. 

What is the level of detail that would be acceptable by @jcfr and @lassoan and at the same time manageable by @che85? Let's try to reach some middle ground?

lassoan:
I completely agree with all points. Adding a settings tab is quite easy, but not necessary right now - it can be added later. The important part is to have API and settings group/key names and storage format that will not need to be changed later.

fedorov:
Ok, sounds great - so how about the following:

1) we add `Q_PROPERTY` to `qSlicerCoreApplication` for the items I mentioned above, and include URLs I referenced into the comments.
2) these properties are stored under the `UserInformation` in the settings
3) we add settings panel to the application settings keeping all of those blank by default

Once we have green light from @jcfr, @che85 can move ahead updating the PR.

jcfr:
Since the amount of "user" info to manage will grow, instead of having a specific getter/setter/property for each one of them and bloat the interface of the qSlicerApplication, I suggest we consider an  API to set/get any user info.

### (1) Dedicated enum to identify keys

```
void qSlicerCoreApplication::setUserInfo(int key, const QVariant & value)
QVariant qSlicerCoreApplication::userInfo(int key)
```

From python, getting the user email would then be:

```
print(slicer.app.userInfo(slicer.app.Email))
```

The enum would be:

```
[...]
   Q_ENUMS(UserInformation)
public:

  enum UserInformation {
    Name = 0
    Login,
    Email,
    ...
    };

```

### (2) Free "form" key string

```
void qSlicerCoreApplication::setUserInfo(const QString& key, const QVariant & value)
QVariant qSlicerCoreApplication::userInfo(const QString& key)
```

While more flexible, issue and typo in key name will be harder to track. 

Will have to document with the key are case sensitive or not



### summary

Any of this approach will allow to change how the information is stored and managed in the future without breaking existing code.


Also related https://www.slicer.org/wiki/Documentation/Nightly/SlicerApplication/ApplicationSettings#Information_for_Advanced_Users

pieper:
Instead of enums, how about introducing a utility class for managing user info.  That could include default constructors, validators, de/serialization, helper method, etc.  That would also keep the app interface less cluttered.  Not sure if this should be a vtk or qt class - I suspect we may need to use it from logic code, so maybe a vtk class is the most flexible.

lassoan:
I would vote for enums. If needed, we can still add API for storing custom fields later.

jcfr:
> utility class for managing user info

Since we have an API added to the application, this could be added later I think.

Also agree, we need to think about that in the context of tracking provenance

> constructors, validators, de/serialization, helper method

This seems to lead back to a MRML node ... 


> I would vote for enums.

Sounds good. Having specific identifier is more robust and less error-prone than free form strings.


To summarize, we agree that:
* dedicated API is relevant
* enum makes sense
* should not bloat the main application API

and potentially that:
* we may want to support operation like 
* anticipate use from logic code (currently written in VTK)

Considering these points, we could have an object in the QTCore library named `qSlicerUserInformation` that would have the following API:
* `info/setInfo(int key, const QVariant& value)`
* `settingsFilePath/setSettingFilePath` : This the file where properties are stored

For convenience, Q_PROPERTY and setting/getter for common info could be added too.

From python, it would like this:

```
slicer.app.userInfo().info(slicer.qSlicerUserInformation.Email)
```

or if there is a convenience property:

```
slicer.app.userInfo().email
```

jcfr:
And regarding validation, in the future using https://doc.qt.io/qt-5/qvalidator.html could help too

lassoan:
Actually, it would be great to have a MRML node so that we can store user info on n the scene as well. The application could set/get user info by using this MRML node type as argument. Even if not MRML node, it should be a VTK node so that it can be accessed from module logic classes.

fedorov:
Ok, sounds great - so how about the following:

* we add enum for keeping the keys and generic get/set user info API call in `slicer.app.userInfo()`
* these properties are stored under the UserInformation in the settings
* we add settings panel to the application settings keeping all of those blank by default

I am not sure if adding a VTK class or MRML class is a good idea - the information is really specific to the application. I do not see the benefit of adding another class. Perhaps that can be done later if such a need arises.

@jcfr @pieper @lassoan are we all in agreement on this plan to move forward?

pieper:
I don't mind the enums for now and we can see how it works out.

My guess is that there will ultimately be need for some validation code for things like email addresses, utilities for converting to/from dicom Person Name conventions (with the ^s) etc.  Centralizing that in a vtk class (perhaps in vtkAddons) would make that re-usable for other parts of slicer.

fedorov:
@pieper yes, I agree it makes sense. I am just trying to make concise and manageable definition of the task we all agree on, so Christian can go ahead. 

I have to say that I do like the simple interface for setters/getters, and I am thinking your idea about validator could be handled transparently inside the `setUserInfo()` call? We could also support this by adding an optional flag to the call, and change the return type, similar to how this is done in DCMTK:

```
int qSlicerCoreApplication::setUserInfo(int key, const QVariant & value, bool validate=false)
```

lassoan:
I think it's a good idea to create this separate VTK helper class for storing use info. For now, it could be really simple, just a couple of get/set methods. It would be better to have this as a VTK class so that it can be used from module logics. A separate class is also useful because it can be utilized whenever we want to store any person information in the future. vtkAddon folder would be a good location. That is accessible for all software components, both in C++ and Python.

fedorov:
Andras, how will that class be linked with the application and application settings?

lassoan:
Similarly to vtkCodedEntry, the vtkPersonInformation (or something like that) class could have string serialization methods:

    std::string GetAsString();
    bool SetFromString(const std::string& content);

The application's SetUserInformation/GetUseInformation methods could either take a string or a vtkPersonInformation class as input.

fedorov:
@lassoan right, but unlike vtkCodedEntry, for which the set of attributes is fixed, vtkPersonInformation will be more open-ended, as we discussed earlier in this thread, so I am having troubles balancing this with the suggested idea for validation.

After this long discussion, and difficulties reaching specific plan that satisfies everyone, I now feel that indeed we should implement a custom solution in our extensions, and later on, if and when it makes sense, migrate to an application-wide solution. We will have a better idea after the extension-based experiment, and in any case this was the idea floated earlier by @jcfr, so perhaps this will be a reasonable middle ground.

lassoan:
> After this long discussion, and difficulties reaching specific plan that satisfies everyone, I now feel that indeed we should implement a custom solution in our extensions

It was a productive design discussion, which resulted in a simple-to-implement and extensible design. For me, the next step is to implement it. It's not clear why you concluded to abandon this design. If you don't have time to implement it, then just say so. I'll send a pull request with an implementation today.

fedorov:
Andras, I am not the one that would be implementing it, and I had troubles formulating the work package for @che85. If you can have a concise formulation, and everyone else agrees on it (including Christian), I am fine with moving forward. Discussions are great, but I find it difficult to formulate the specific steps that need to be implemented, and we really need to move forward.

fedorov:
@che85 it would be helpful if you could chime in with your understanding and any issues you have in mind.

che85:
If we don't find an agreement here, why don't we have a video call? It might be much more comfortable and faster when doing that.

fedorov:
> It might be much more comfortable and faster when doing that.

Well, I've just spent my day in back-to-back meetings from 9am until 3:40pm, with breaks for lunch and moving from place to place. Speaking for myself, I am *not* looking to add more meetings to my agenda, there's just no time to do work, but I am fine if you guys move forward and discuss without me, and just add the agreed upon conclusion to this thread. Whatever moves this issue forward, since it's been 19 days since the initial PR, and this topic is a blocker for @chribaue.

lassoan:
Let's continue the discussion at this pull request:
https://github.com/Slicer/Slicer/pull/926

It implements most things that we've discussed, but should be reviewed before we merge it in the core.

fedorov:
Thank you Andras, this is great. I do think however that it would have been good to reach agreement and formulate the task for Christian. If our only way to reach agreement is to have you implement the idea, we clearly have a communication problem.

jcfr:
> implement the idea ... only way to reach agreement is to have ... implement 

@lassoan Thanks for prototyping, this will definitively help the discussion.

I think there is still opportunity to iterate on the change proposed by Andras or even evaluate an other prototype like the one that was originally submitted by @che85 . 

My understanding is that both prototypes (the one from Christian and the one from Andras) are contributing toward the final solution, and it is clearly the outcome of an open discussion. A lot of interesting point were discussed in this pr.

Independently of the final version, I think the authorship should be given to Christian with everyone else as `Co-authored-by:` or `Suggested-by` ... . 

:love_letter:

lassoan:
Considering the amount of time we spent with discussion, @fedorov's conclusion,  and the tiny development task, it seemed more efficient to jump to implementation rather than further discussing it.

PR #913, opened 2018-03-09
BUG: Fixed old bug with volume rendering preset selection when changi…

cpinter:
…ng volume

When the rendered volume was changed, then the preset combobox selection stayed at the previous selection, but that selection was not applied. Because of this, many times the user had to change selection to something else than back. In addition it was confusing.
Now the selected preset is remembered, and set when the volume changes.

### DIFF
--- a/Modules/Loadable/VolumeRendering/MRML/vtkMRMLVolumePropertyNode.cxx
+++ b/Modules/Loadable/VolumeRendering/MRML/vtkMRMLVolumePropertyNode.cxx
@@ -152,11 +152,15 @@ void vtkMRMLVolumePropertyNode::CopyParameterSet(vtkMRMLNode *anode)
     return;
     }
 
+  vtkMRMLCopyBeginMacro(anode);
+  vtkMRMLCopyVectorMacro(EffectiveRange, double, 2);
+  vtkMRMLCopyEndMacro();
+
+  // VolumeProperty
   this->VolumeProperty->SetIndependentComponents(node->VolumeProperty->GetIndependentComponents());
   this->VolumeProperty->SetInterpolationType(node->VolumeProperty->GetInterpolationType());
 
-  //VolumeProperty
-  for (int i=0;i<VTK_MAX_VRCOMP;i++)
+  for (int i=0; i<VTK_MAX_VRCOMP; i++)
     {
     this->VolumeProperty->SetComponentWeight(i,node->GetVolumeProperty()->GetComponentWeight(i));
     //TODO: No set method for GrayTransferFunction, ColorChannels, and DefaultGradientOpacity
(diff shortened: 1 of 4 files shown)


Modules/Loadable/VolumeRendering/MRML/vtkMRMLVolumePropertyNode.h
```diff
@@ -206,6 +211,11 @@ class VTK_SLICER_VOLUMERENDERING_MODULE_MRML_EXPORT vtkMRMLVolumePropertyNode
   /// independent value-wise, and they do not have third and fourth axes.
   double EffectiveRange[2];
 
+  /// MRML node ID of the selected preset volume property node
+  /// Important! The node with this ID belongs to the presets private scene in volume rendering
+  ///            logic, and not the slicer MRML scene this node is in!
+  char* PresetVolumePropertyNodeID;
```
lassoan: In this case, I think it would be better to store a "preset name" string instead, which would be matched with the node name of the volume property preset.

cpinter: Maybe less confusing, but I see it as an unnecessary conversion back and forth, because it's the ID that is used.

lassoan: It is more than confusing, thinking a bit more about it, it may actually lead to incorrect behavior if we use node IDs.

Node IDs in the private scene are assigned in undeterministic way, because extensions can register their own presets and random order. But even for built-in presets, node IDs may change if we add/remove/reorder them.

Also, if you store something like 'vtkMRMLPropertyNode1' in the scene, it is very uninformative, compared to seeing "CTCardiac1" instead. This case is different from other node references, as we cannot look up what 'vtkMRMLPropertyNode1' means, as it is not in the scene file.

cpinter:
I did @lassoan's suggestions. Thus the code has been simplified as well.

I needed to add a check in qSlicerVolumeRenderingPresetComboBox::applyPreset to prevent the preset node from overwriting the volume property node when the rendered volume is changed. So the selection and icon are updated, but the transfer functions are not reset.

cpinter:
Integrated in http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27039

(for some reason the GitHub mirror is not updating, I hope it's just temporary)

lassoan:
@jcfr GitHub-SVN sync seems to be down. Could you please check?

cpinter:
Just an update that the GitHub mirror still hasn't synced.

I did my usual workflow (cherry pick + DCommit), so I think it's unrelated.

jcfr:
That was it .. the command `crontab -e` is for editing the list, and `crontab -r` is to remove all the jobs. And the `e` and the `r` are close to each other on the keyboard.

jcfr:
Since I always do a `crontab -l` before editing, I add a copy in the buffer of `tmux`.

Also, grepping through the history confirmed my hypothesis.

In the next few seconds, github should be up-to-date

PR #915, opened 2018-03-10
BUG: Fix transform hierarchy errors after scene load

lassoan:
The problem was that qMRMLSceneTransformModel tries to determine the parent transform node as soon as the node is added to the scene, but at the time FirstTransformNode was added to the scene, the parent SecondTransformNode had not yet added to the scene, so it got NULL as parent transform.

In general, node IDs and references are not valid until the import is completed, therefore they cannot be added one by one.

Solution: do scene import always in batch mode.

Fixes https://issues.slicer.org/view.php?id=4080

### DIFF
--- a/Libs/MRML/Widgets/qMRMLSceneModel.cxx
+++ b/Libs/MRML/Widgets/qMRMLSceneModel.cxx
@@ -1219,8 +1219,10 @@ void qMRMLSceneModel::onMRMLSceneNodeAdded(vtkMRMLScene* scene, vtkMRMLNode* nod
   Q_ASSERT(scene == d->MRMLScene);
   Q_ASSERT(vtkMRMLNode::SafeDownCast(node));
 
-  if (d->LazyUpdate && d->MRMLScene->IsBatchProcessing())
+  if (d->MRMLScene->IsImporting() || (d->LazyUpdate && d->MRMLScene->IsBatchProcessing()))
     {
+    // Node IDs and references are not valid until the import is completed, therefore do not attempt
+    // to add a node during importing (see https://issues.slicer.org/view.php?id=4080).
     return;
     }
   this->insertNode(node);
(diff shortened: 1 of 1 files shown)


lassoan:
@cpinter @jcfr You've worked a lot with the scene model. Do you see any risk/drawback in updating the whole scene at the end of scene import?

lassoan:
@cpinter

cpinter:
Sorry for the delay. I think it should be OK. There should be no side effects other than the extra time it takes. Subject hierarchy also updates the whole hierarchy after import and batch processing

lassoan:
Thanks, I'll merge it then.

lassoan:
Committed r27077

jcfr:
>  than the extra time it takes

The idea is too avoid the extra time.

Here is a topic that I think is related and should help address the problem: https://github.com/Slicer/Slicer/compare/master...jcfr:4081-ensure-reference-are-up-to-date-on-import

It adds the following method:

```
+  /// \brief Update ChangeIDs map to ensure proper update of node with
+  /// ID conflicts on scene imports or explicit node addition.
+  ///
+  /// In case the scene needs to change the ID of some nodes to add, the new
+  /// ID should not be one already existing in the scene nor one of the
+  /// imported scene. To ensure the map of `oldID` to `newID` is up-to-date so
+  /// that reference are updated before invoking the NodeAddedEvent, this
+  /// function will call  GenerateUniqueID(vtkMRMLNode* node) for node that will
+  /// conflict and will keep track of the association using
+  /// AddChangedID(const char* oldID, const char* newID).
+  ///
+  /// Additionally, the following nodes will be marked as reserved so that
+  /// the ID generator doesn't choose them:
+  /// * all the node IDs of the scene that conclict with the one in \a nodes.
+  /// * all generated IDs.
+  ///
+  /// \sa Import(), AddNode(vtkMRMLNode* nodeToAdd), CopyNode(vtkMRMLNode* nodeToCopy)
+  /// \sa AddChangedID(const char* oldID, const char* newID)
+  /// \sa IsReservedID(const std::string& nodeID), AddReservedID(const char* nodeID)
+  /// \sa GenerateUniqueID(vtkMRMLNode* node)
+  void PrepareImport(vtkCollection* nodes);
+
```

cpinter:
> The idea is too avoid the extra time

If you load a scene you need to wait long seconds. Compared to that, the extra fraction of a second is not noticeable.

lassoan:
Full update takes 1-2 seconds for the default scene of 200 nodes (mostly color nodes). I don't think the extra delay is noticeable. Also, if you import a large scene then loading should be faster now. 

It would be nice to enable lazy update for batch processing, too

lassoan:
I would like to improve the scene model and get your topic integrated but I still need to work on widget improvements.

PR #917, opened 2018-03-13
BitStreamVolumeNode and  CompressionDeviceNode for compressed image data 

leochan2009:
1. vtkMRMLBitStreamVolumeNode is implemented for storing the encoded bit stream and key frame bit stream. It derives from VectorVolumeNode and  provides interface for attaching different "vtkMRMLCompressionDeviceNode".  
2. vtkMRMLCompressionDeviceNode is a generic node for codecs.  Any nodes that encapsulates codec should derive from the vtkMRMLCompressionDeviceNode. See this derived node for more detail: [DerivedCompressionNode](https://github.com/openigtlink/SlicerOpenIGTLink/blob/BitStreamNodeRemoval/OpenIGTLinkIF/MRML/vtkMRMLIGTLIOCompressionDeviceNode.h)

@lassoan , would be great if we could have a discussion for this initial implementation.

Best,
Longquan


Libs/MRML/Core/vtkMRMLBitStreamVolumeNode.cxx
```diff
@@ -0,0 +1,235 @@
+#include "vtkMRMLBitStreamVolumeNode.h"
+
+// OpenIGTLink includes
+//#include "igtlioVideoDevice.h"
```
pieper: Please remove dead code

pieper:
These look nice and useful, but I wonder if they have any use cases outside of OpenIGTLink.  I'm thinking they should be part of the extension and then could get moved to the core if there is a future need to do so.

leochan2009:
Hi Steve,

Thanks for the review. The two nodes added here provide generic API for registering any compression codec. 
Yes, Sequence module needs to access the BitStreamVolumeNode for recording the compressed video stream and replay the video stream data. Sequence is an extension that should not depend on "SlicerOpenIGTLink". @lassoan, maybe you have some thoughts for sequence? 

Best,
longquan

lassoan:
Yes, we need video codecs, not just for network transfer of images but for file I/O. The generic interface allows extensions to download and register codecs that Slicer core or other extensions can use.

Several codecs cannot be built into the core due to licensing complications.

Thank you very much Longquan, I'll review the code by the end of the week.

Libs/MRML/Core/vtkMRMLBitStreamVolumeNode.cxx
```diff
@@ -0,0 +1,235 @@
...
+vtkMRMLNodeNewMacro(vtkMRMLBitStreamVolumeNode);
+
+//----------------------------------------------------------------------------
+// vtkMRMLBitStreamVolumeNode methods
+
+//-----------------------------------------------------------------------------
+vtkMRMLBitStreamVolumeNode::vtkMRMLBitStreamVolumeNode()
+{
+  this->compressionDevice = NULL;
+  IsCopied = false;
```
jcfr: For sake of consistency:
```
this->IsCopied = false;
this->KeyFrameReceived
...
```

Libs/MRML/Core/vtkMRMLBitStreamVolumeNode.cxx
```diff
@@ -0,0 +1,235 @@
...
+//----------------------------------------------------------------------------
+vtkMRMLNodeNewMacro(vtkMRMLBitStreamVolumeNode);
+
+//----------------------------------------------------------------------------
+// vtkMRMLBitStreamVolumeNode methods
+
+//-----------------------------------------------------------------------------
+vtkMRMLBitStreamVolumeNode::vtkMRMLBitStreamVolumeNode()
+{
+  this->compressionDevice = NULL;
```
jcfr: Per VTK convention, private member should start with upper case letter. ->`CompressionDevice`

Libs/MRML/Core/vtkMRMLBitStreamVolumeNode.cxx
```diff
@@ -0,0 +1,235 @@
...
+  KeyFrameUpdated = false;
+  MessageBufferValid = false;
+  KeyFrameBuffer = "";
+  MessageBuffer = "";
+}
+
+//-----------------------------------------------------------------------------
+vtkMRMLBitStreamVolumeNode::~vtkMRMLBitStreamVolumeNode()
+{
+  if (compressionDevice != NULL)
```
jcfr: Add missing brackets

Libs/MRML/Core/vtkMRMLBitStreamVolumeNode.cxx
```diff
@@ -0,0 +1,235 @@
...
+     compressionDevice->Delete();
+}
+
+//---------------------------------------------------------------------------
+int vtkMRMLBitStreamVolumeNode::ObserveOutsideCompressionDevice(vtkMRMLCompressionDeviceNode* device)
+{
+  if (device)
+  {
```
jcfr: Fix indent

Libs/MRML/Core/vtkMRMLBitStreamVolumeNode.cxx
```diff
@@ -0,0 +1,235 @@
...
+    device->AddObserver(device->GetDeviceContentModifiedEvent(), this, &vtkMRMLBitStreamVolumeNode::ProcessDeviceModifiedEvents);
+    //------
+    std::string bitStream = std::string(device->GetContent()->frameMessage);
```
jcfr: What if `device->GetContent()->frameMessage` is null ? I anticipate it will seg fault ..

Libs/MRML/Core/vtkMRMLBitStreamVolumeNode.cxx
```diff
@@ -0,0 +1,235 @@
...
+void vtkMRMLBitStreamVolumeNode::SetMessageStream(std::string buffer)
+{
+  this->MessageBuffer.resize(buffer.length());
+  char * array = new char[buffer.length()];
+  memcpy(array, buffer.c_str(), buffer.length());
+  this->MessageBuffer.assign(array, buffer.length());
+  this->MessageBufferValid = true;
+};
+
+std::string vtkMRMLBitStreamVolumeNode::GetMessageStream()
```
jcfr: Method should probably not include the word `Stream` as the return value is a string

Libs/MRML/Core/vtkMRMLBitStreamVolumeNode.cxx
```diff
@@ -0,0 +1,235 @@
...
+{
+  std::string returnMSG(this->MessageBuffer);
+  return returnMSG;
+};
+
+void vtkMRMLBitStreamVolumeNode::SetKeyFrameStream(std::string buffer)
```
jcfr: Consider passing using `const std::string& buffer` to avoid unneeded copy

Libs/MRML/Core/vtkMRMLBitStreamVolumeNode.cxx
```diff
@@ -0,0 +1,235 @@
...
+{
+  this->KeyFrameBuffer.resize(buffer.length());
+  char * array = new char[buffer.length()];
+  memcpy(array, buffer.c_str(), buffer.length());
+  this->KeyFrameBuffer.assign(array, buffer.length());
```
jcfr: The code both seems to be duplicated in few location, I suggest to introduce an helper function

Libs/MRML/Core/vtkMRMLBitStreamVolumeNode.cxx
```diff
@@ -0,0 +1,235 @@
...
+};
+
+std::string vtkMRMLBitStreamVolumeNode::GetKeyFrameStream()
+{
+  std::string returnMSG(this->KeyFrameBuffer);
+  return returnMSG;
+};
+
+//---------------------------------------------------------------------------
+void vtkMRMLBitStreamVolumeNode::DecodeMessageStream(vtkMRMLCompressionDeviceNode::ContentData* deviceContent)
```
jcfr: method could probably return a boolean to indicate success

jcfr: Or an integer as it is done with `EncodeImageData`

Libs/MRML/Core/vtkMRMLBitStreamVolumeNode.h
```diff
@@ -0,0 +1,110 @@
...
+#define __vtkMRMLBitStreamVolumeNode_h
+
+// MRML includes
+#include <vtkMRML.h>
+#include <vtkMRMLNode.h>
+#include <vtkMRMLStorageNode.h>
+#include "vtkMRMLVectorVolumeDisplayNode.h"
+#include "vtkMRMLVectorVolumeNode.h"
+#include "vtkMRMLVolumeArchetypeStorageNode.h"
+#include "vtkMRMLCompressionDeviceNode.h"
```
jcfr: Order include statements alphabetically

Libs/MRML/Core/vtkMRMLBitStreamVolumeNode.h
```diff
@@ -0,0 +1,110 @@
...
+
+// VTK includes
+#include <vtkStdString.h>
+#include <vtkImageData.h>
+#include <vtkObject.h>
+
+class  VTK_MRML_EXPORT vtkMRMLBitStreamVolumeNode : public vtkMRMLVectorVolumeNode
```
jcfr: Could you add a high-level description of the  class, its usage and relation ship to other class.

Libs/MRML/Core/vtkMRMLBitStreamVolumeNode.h
```diff
@@ -0,0 +1,110 @@
...
+  ///
+  /// Copy the node's attributes to this object
+  virtual void Copy(vtkMRMLNode *node) VTK_OVERRIDE;
+  
+  ///
+  /// Get node XML tag name (like Volume, Model)
+  virtual const char* GetNodeTagName() VTK_OVERRIDE
+  {return "BitStreamVolume";};
+
+  void SetKeyFrameReceived(bool value){this->KeyFrameReceived = value;};
```
jcfr: Remove extra `;` from the end of the line

Libs/MRML/Core/vtkMRMLCompressionDeviceNode.cxx
```diff
@@ -0,0 +1,145 @@
...
+#include <vtkXMLUtilities.h>
+
+//---------------------------------------------------------------------------
+vtkMRMLNodeNewMacro(vtkMRMLCompressionDeviceNode);
+//---------------------------------------------------------------------------
+vtkMRMLCompressionDeviceNode::vtkMRMLCompressionDeviceNode()
+{
+  Content = new ContentData();
+  Content->image = NULL;
+  Content->frameType = -1; // -1 is an invalid value
```
jcfr: Should this be a constant somewhere ? instead of an hardcoded `-1`

Libs/MRML/Core/vtkMRMLCompressionDeviceNode.h
```diff
@@ -0,0 +1,100 @@
...
+
+class VTK_MRML_EXPORT vtkMRMLCompressionDeviceNode : public vtkMRMLNode
+{
+public:
+  
+  enum {
+    DeviceModifiedEvent         = 118961, //Todo, should have a different id, as it is conflict with openigtlinkIO VideoDeviceModified event
+  };
+  
+  struct ContentData
```
jcfr: does it make sense to expose this struct as part of the public API ?

jcfr:
And thanks @leochan2009  for contributing the changes back into the core :+1:

Libs/MRML/Core/vtkMRMLBitStreamVolumeNode.cxx
leochan2009: thanks for pointing out ,  device->GetContent() might return NULL. 
if device->GetContent() is not NULL. 
device->GetContent()->frameMessage won't be NULL, as frameMessage is of type string.

Libs/MRML/Core/vtkMRMLBitStreamVolumeNode.cxx
```diff
@@ -0,0 +1,247 @@
...
+
+void vtkMRMLBitStreamVolumeNode::CopySrcStringToDestString(const std::string& srcString, std::string& destString)
+{
+  destString.resize(srcString.length());
+  char * array = new char[srcString.length()];
+  memcpy(array, srcString.c_str(), srcString.length());
+  destString.assign(array, srcString.length());
+}
+
+void vtkMRMLBitStreamVolumeNode::SetFrameMessage(std::string& buffer)
```
lassoan: It would be nice if we would not be forced to make a deepcopy. For example, if we use vtkUnsignedCharArray (or other VTK object) to store the buffer instead of std::string then we can pass the ownership without actual memory copy.

"Message" should not be included in the name, as we may decode from file. "Set" indicates that previous content is overwritten, while here it is updated/combined/appended. So a name like UpdateFromDataStream or something similar would be better.

Libs/MRML/Core/vtkMRMLBitStreamVolumeNode.cxx
```diff
@@ -0,0 +1,247 @@
...
+{
+  this->CopySrcStringToDestString(buffer, this->FrameMessage);
+  this->FrameMessageValid = true;
```
lassoan: It would be better to not have a separate flag for validity (it may be inconsistent with FrameMessage content), instead, we could just set FrameMessage to NULL or empty when it is not valid.

Libs/MRML/Core/CMakeLists.txt
```diff
@@ -223,6 +223,8 @@ if(MRML_USE_vtkTeem)
     vtkMRMLNRRDStorageNode.cxx
     vtkMRMLTensorVolumeNode.cxx
     vtkMRMLVectorVolumeNode.cxx
+    vtkMRMLBitStreamVolumeNode.cxx
+    vtkMRMLCompressionDeviceNode.cxx
```
lassoan: Codec should not be a MRML node, as nodes are only intended to be used for data storage and not for processing. If we put the codec in a MRML node then it cannot have access to application or logic classes. Instead, we should define a simple vtkObject-based class as codec interface class, concrete codec classes (in the Slicer core we could have only one codec, a trivial "Uncompressed" codec that just copies data), and a codec factory where you can register custom codecs. Similarly how it is done in segmentation converter rules: http://apidocs.slicer.org/master/classvtkSegmentationConverterFactory.html

Volumes logic class would contain the codec factory. When a streaming volume node is added to the scene, volumes logic would read the codec name from the node, instantiate the corresponding codec class, and set it in the streaming volume node.

Libs/MRML/Core/vtkMRMLBitStreamVolumeNode.cxx
```diff
@@ -0,0 +1,247 @@
...
+
+//----------------------------------------------------------------------------
+// vtkMRMLBitStreamVolumeNode methods
+
+//-----------------------------------------------------------------------------
+vtkMRMLBitStreamVolumeNode::vtkMRMLBitStreamVolumeNode()
+{
+  this->CompressionDevice = NULL;
+  this->FrameUpdated = false;
+  this->KeyFrameReceived = false;
```
lassoan: There are 32 potential potential states of  these 5 flags, which is quite difficult to keep track of and probably not all combinations are correct. Would it be possible to use enum(s) to enumerate valid states?

Libs/MRML/Core/vtkMRMLBitStreamVolumeNode.cxx
```diff
@@ -0,0 +1,247 @@
...
+      {
+      this->SetKeyFrameReceived(true);
+      this->SetKeyFrameMessage(compressedStream);
+      }
+      return 1;
+    }
+  return 0;
+}
+
+std::string vtkMRMLBitStreamVolumeNode::GetCodecType()
```
lassoan: Probably the interface should have a CodecName property (simple string), that would be saved in MRML scene; and a Codec property (VTK class based vtkStreamingVideoCodec interface class) that would be set by the Volumes logic.

Libs/MRML/Core/vtkMRMLBitStreamVolumeNode.cxx
```diff
@@ -0,0 +1,247 @@
...
+{
+  Superclass::WriteXML(of, nIndent);
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLBitStreamVolumeNode::ReadXMLAttributes(const char** atts)
+{
+  int disabledModify = this->StartModify();
+  
+  Superclass::ReadXMLAttributes(atts);
```
lassoan: Codec name must be read/written/printed/copied. Use the recently added helper macros https://github.com/Slicer/Slicer/blob/master/Libs/MRML/Core/vtkMRMLCameraNode.cxx#L110

Libs/MRML/Core/CMakeLists.txt
```diff
@@ -223,6 +223,8 @@ if(MRML_USE_vtkTeem)
     vtkMRMLNRRDStorageNode.cxx
     vtkMRMLTensorVolumeNode.cxx
     vtkMRMLVectorVolumeNode.cxx
+    vtkMRMLBitStreamVolumeNode.cxx
```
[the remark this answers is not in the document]
leochan2009: Agree

Libs/MRML/Core/CMakeLists.txt
leochan2009: Great Suggestion!
Should i create a folder "vtkStreamingVideoCore" in the Slicer/Libs, same level as the vtkSegmentationCore folder?
And place the vtkStreamingVideoCodec interface class, a trival "Uncompressed" codec class and the vtkStreamingVideoCodecFactory in the folder?

leochan2009: Probably the interface class should be call vtkStreamingVolumeCodec. and the folder "vtkStreamingVolumeCore", as the data might be normal volume instead of Video

lassoan: Code in Slicer/Libs cannot use any MRML or Slicer module logic, so only the codec abstract class and uncompressed codec class could be there. Adding a new library to Slicer/Libs is quite a lot of small changes, so just for 2 classes I would not do it.

These codecs are only for image encoding, so I would put them in Libs/MRML/Core and/or Volumes module MRML and Logic folders.

lassoan: > vtkStreamingVolumeCodec

OK. Even if the most commonly used codecs are for 2D+t video data, the infrastructure would support streaming of volumes.

Libs/MRML/Core/vtkMRMLBitStreamVolumeNode.cxx
leochan2009: I remove one unnecessary flag now we have 4 flags:
FrameUpdated, KeyFrameReceived, KeyFrameDecoded, KeyFrameUpdated.
FrameUpdated and the other three key frame flags are independent. 
And most combination of the key frame flags are valid.  i will keep the implementation as it's now.

leochan2009:
@lassoan 
Just modified the code according to your suggestions.
Still need to work on the SlicerVolumeLogic to add the StreamingVolumeNode.

Libs/MRML/Core/vtkMRMLBitStreamVolumeNode.cxx
leochan2009: HI Andras,
i made the frame and the keyframe of type vtkUnsignedCharArray in StreamingVolumeNode.
Also Set/Get frame functions are implemented. So there wouldn't be deep copy among StreamingVolumeNodes in Slicer. 
However, do you think it is better to make the vtkStreamingVolumeCodec use vtkUnsignedCharArray instead of std::string?  Currently i still make copy of the std::string from vtkStreamingVolumeCodec into  StreamingVolumeNode.  As vtkStreamingVolumeCodec is the wrapping class of external codec libraries, the encoded/decode bit stream from external codec libraries will be saved in the std::string structure in vtkStreamingVolumeCodec, deep copy of std::string into StreamingVolumeNode would be safer to avoid the memory segmentation fault.  What do you think? 
Thanks!

lassoan: Unlike vtkObject-based objects, std::string does not give a chance to share or transfer ownership, therefore it gives us more flexibility. You may still DeepCopy a vtkUnsignedCharArray, if needed.

Libs/MRML/Core/CMakeLists.txt
leochan2009: Addressed

Libs/MRML/Core/vtkMRMLBitStreamVolumeNode.cxx
leochan2009: Addressed, using CodecName and CodecDeviceType

Libs/MRML/Core/vtkMRMLBitStreamVolumeNode.cxx
leochan2009: Addressed

Libs/MRML/Core/vtkMRMLBitStreamVolumeNode.cxx
leochan2009: vtkUnsignedCharArray is used for all now

leochan2009:
@lassoan 
Hi, Andras,

I have some question regarding the design pattern.
The "codec device type" is added in the StreamingVolumeNode attributes and is saved in the mrmlScene file. 
When we close slicer and reload the mrmlscene, we should have the StreamingVolumeNode with the codec device specified by the  "codec device type". As the codec device is not saved, we need to create the codec device using the vtkStreamingVolumeCodecFactory (which i already added in the "Volumes" module). 
Do you have any suggestion how(when and where) we can add this codec device into the StreamingVolumeNode?

Thanks,
Longquan

lassoan:
> When we close slicer and reload the mrmlscene, we should have the StreamingVolumeNode with the codec device specified by the "codec device type".

"device" is such on overused term in IGT (and not very well defined) that we should avoid using it. We can just call it "codec type".

> As the codec device is not saved, we need to create the codec device using the vtkStreamingVolumeCodecFactory (which i already added in the "Volumes" module).
Do you have any suggestion how(when and where) we can add this codec device into the StreamingVolumeNode?

It is quite common that when you load a node, a module logic class, which is primarily responsible for those nodes creates additional classes. In this case, Volumes module would observe the scene, whenever a StreamingVolumeNode is added to the scene, the volumes logic reads the codec name, instantiates a corresponding codec class, and sets it in the node.

leochan2009:
@lassoan  
Modified according to your suggestions, see the latest commit. 306398e
Thanks,
Best,
Longquan

lassoan:
Great, thank you. I'll review in detail tomorrow.

leochan2009:
@lassoan 
Hi Andras, 
Here is some example how the StreamingVolumeNode and StreamingVolumeCodec can be used in other projects:
1. StreamingVolumeCodec as an interface for external compression library:
 [Interfacing with OpenIGTLinkIO VideoDevice](https://github.com/openigtlink/SlicerOpenIGTLink/blob/BitStreamNodeRemoval/OpenIGTLinkIF/MRML/vtkIGTLStreamingVolumeCodec.cxx)
2. StreamingVolumeNode in Sequence recording and replaying:
[Recoding and Replay in Sequence](https://github.com/leochan2009/Sequences/blob/BitStreamForVideo/Sequences/MRML/vtkMRMLNodeSequencer.cxx#L207)
I hope the above information will be helpful for the code review process.

Best,
Longquan

PR #918, opened 2018-03-15
BUG: fix #4485 - use of known null

ihnorton:
Ran in to this earlier when I returned null from a subclass `::GetOutputMeshConnection` (temporary testing change in order to short-circuit pipeline).

Also clean up the flow.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLModelDisplayNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLModelDisplayNode.cxx
@@ -217,10 +217,20 @@ vtkPointSet* vtkMRMLModelDisplayNode::GetOutputMesh()
     }
 
   vtkAlgorithmOutput* outputConnection = this->GetOutputMeshConnection();
-  vtkAlgorithm* producer =  outputConnection? outputConnection->GetProducer() : 0;
+  if (!outputConnection)
+    {
+    return NULL;
+    }
+
+  vtkAlgorithm* producer = outputConnection->GetProducer();
+  if (!producer)
+    {
+    return NULL;
+    }
+
   producer->Update();
   return vtkPointSet::SafeDownCast(
-    producer ? producer->GetOutputDataObject(outputConnection->GetIndex()) : 0);
+           producer->GetOutputDataObject(outputConnection->GetIndex()));
 }
 
 //---------------------------------------------------------------------------


pieper:
Seems reasonable to me.

cpinter:
Just curious: is there any advantage of using auto, or it's just more convenient?

jcfr:
As discussed during the hangout, I suggest we start using `c++11` only feature after cutting the next release, that will make it clear and help dependent project to transition. 

`v5.0` will be the last tag known to compile with `c++98`, and beyond is `c++11`. Thanks for understanding.

lassoan:
I've been so looking forward to using auto. But now, that I see a piece of code that I have to review, it's quite bad. To review the code, I either have to build the code and attach a debugger (lot of extra work) or try and look up the types in the source code manually. Couple this with non-meaningful variable names `oc`, `pr`) and now the code is barely readable. To optimize readability, I'm tempted to only encourage use of auto for trivial cases only.

Maybe my point of view will change over time, but it is hard to let go the certainty that explicit types give. Especially because when we are dealing with pointers, a mistake may cause a hard crash. In Python, if I mess up something, the result is very likely not be catastrophic, so I'm not that worried.

Another small remark is that I strongly prefer early returns (return immediately if there is an error) instead of deep nesting. With deep nesting, I need to review the code as many times as number of if/else (I have to read the code until I find a a return). It also wastes valuable horizontal space. With early returns, I just need to read the code once and there are no nested if/else branches. Deep nesting provides single point of return, which is nice if you want to clean up stuff on return, but that is handled better by guard classes anyway (impossible to forget, exception-safe).

ihnorton:
Updated.

---

@jcfr the requirement of C++11 going forward was announced when [Slicer 4.8.1 was released](https://discourse.slicer.org/t/slicer-4-8-1-release-and-transition-to-qt5-and-vtk9/1551), and anyone who wants to still use Qt4 needs to do their own integration testing anyway because there are no nightlies for that anymore. So at this point, telling people to backport patches as-needed too would seem very reasonable.

@lassoan 
- For what it's worth even Visual Studio Code gives a bubble with return type of all of these calls (no static type inference for the var itself like the compiler does, but I believe "real" Visual Studio Intellisense has that). Anyway, I hope we can find a good balance going forward. It's worth a quick browse through a few files in e.g. LLVM or Swift some time to see `auto` in good use (when readability is in favor).

 - I like early return too. But there are multiple, dependent conditions here and nesting all-true conditions, with assign-and-test seemed cleaner. Fair enough though, I changed it to several multiple negation/return tests.

lassoan:
> For what it's worth even Visual Studio Code gives a bubble with return type of all of these calls 

I mostly review code in GitHub, just by reading it. It would be useful to build and test it using a debugger, but it would take more time, so I normally don't do it.

> I hope we can find a good balance going forward

Me too. There are several advantages of auto, we just have to find ways how and when to use it. Probably it also helps if the code base is modern and consistent, for example you don't have many types of strings and pointers. For example as long as we use both const char* and std::string for referring to strings, it is quite important to explicitly state what it is. If we get rid of char* then the risk of not knowing the exact type of a string would be much lower.

ihnorton:
Will run tests and then merge later today if no objection.

ihnorton:
in r27100

jcfr:
Thanks @ihnorton 

Nitpick: In future commit message, consider adding some context.

Instead of `BUG: fix #4485 - use of known null `, prefer `BUG: vtkMRMLModelDisplayNode::GetOutputMesh: Use of known null. Fix #4485`

lassoan:
For me either adding bug id at the front or back is OK, as long as we are consistently use the same format. Is the commit comment format described somewhere?

jcfr:
> is the commit comment format described somewhere?

See https://www.slicer.org/wiki/Documentation/Nightly/Developers/Style_Guide#Commits

ihnorton:
There is a tight line to walk between descriptive comments, long class names, and the pre-commit hook 😄

PR #919, opened 2018-03-20
ENH: Added direct file export from segmentation

lassoan:
Users still find it difficult to create STL files from segmentations (export from segmentation to model nodes, open Save dialog, select model nodes, select output folder, change file type to STL for each model node).

It allows export segments (all or just visible) directly to STL or OBJ files, without the need to export to nodes first.

Segments may be exported to separate STL files or one merged file.
OBJ files contain color and opacity information.

The feature is available both from Segment editor (Segmentations button menu)
and in Segmentations module (new Export to files section).

### DIFF
--- a/Modules/Loadable/Segmentations/Logic/vtkSlicerSegmentationsModuleLogic.cxx
+++ b/Modules/Loadable/Segmentations/Logic/vtkSlicerSegmentationsModuleLogic.cxx
@@ -35,6 +35,8 @@
 #include "vtkSlicerTerminologyEntry.h"
 
 // VTK includes
+#include <vtkActor.h>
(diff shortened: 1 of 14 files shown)


lassoan:
The new export function was added to the "Segmentations" button menu in Segment editor, as it was easier to implement than adding a new button (and modules that use the segment editor don't need to be updated to potentially hide the new button).

image

It opens this popup window:

image

Also made the export available in the Segmentations module, as a new section, to make it easily discoverable:

image

pieper:
Looks sweet!  I'm traveling this week so I didn't build but the code looks nice 👍

lassoan:
Thanks for the reviews. Committed in r27089.

PR #920, opened 2018-03-21
ENH: remove unused vtkXNATPermissionPrompter class

ihnorton:
ref: https://discourse.slicer.org/t/slicer-5-0-deprecation-discussion-wiki/2377

I checked all registered (git-based) extensions by cloning all [1] and then grepping for `vtkXNAT` The only remaining svn extensions are below, and they probably don't use XNAT.

ABC.s4ext
IntensitySegmenter.s4ext
MarginCalculator.s4ext
Scoliosis.s4ext

[1]
```
for f in `grep -rl "scm git" ExtensionsIndex/`; do
  url=`cat $f | grep scmurl | cut -d" " -f2`
  git -C sources clone $url --depth 1
done
```


jcfr:
> extensions by cloning all

Back when not all extensions were using git, I created the following CLI to easily checkout source code of all extensions:

jcfr:
I grepped through all extensions and `vtkXNATPermissionPrompter` is not referenced in any of them.

jcfr:
Integrated in [r27093](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27093)

PR #921, opened 2018-03-22
ENH: rename vtkNRRDReader,vtkNRRDWriter -> vtkTeemNRRD{R,W}

ihnorton:
Discussed on hangout this week. `vtkNRRDReader` is ambiguous with `vtkNrrdReader`, both to a casual reader and to the build system on case-insensitive filesystems.

There are five extensions using these classes:

- Sequences: https://github.com/SlicerRt/Sequences/pull/47
- MultiVolumeExplorer: https://github.com/fedorov/MultiVolumeExplorer/pull/40
- SlicerDMRI: https://github.com/SlicerDMRI/SlicerDMRI/pull/103
- UKFTractography: https://github.com/pnlbwh/ukftractography/pull/100
- SlicerCIP: https://github.com/acil-bwh/SlicerCIP/pull/23

~~I will make PRs to those projects and link them here.~~ Need to coordinate merges.


jcfr:
_This comment is related to this PR but not outside the scope of this PR. In other word, it shouldn't prevent the integration of these changes._

Did you consider updating the reader available in upstream VTK or ITK ? Like we just did for @lassoan contribution, it can quickly be reviewed and integrated in VTK.

Also, I wonder what are the limitation of the ITK implementation ? We could also look into improving it.

@pieper  @thewtex  I discussed this in an other thread/topic/issue but can't find it.

Independently of VTK or ITK, we would need to vendorize teem into upstream VTK or ITK. The good news is that teem is [considering the transition to gitHub](https://discourse.itk.org/t/add-a-parallel-compression-method-to-nrrd-and-or-metaimage/696/17) (this would just make the process easier)

ihnorton:
> Did you consider updating the reader available in upstream VTK or ITK ? 

- ITK already has sufficient functionality (we use it directly in MRMLIDImageIO).
- Because no VTK-only users/developers needed this functionality in so many years (apparently), to me it seems better for maintenance to keep it closer to where it is actively used and more developers are familiar with the format.

ihnorton:
Will merge later today if no objection.

lassoan:
We could wrap ITK's nrrd reader the same way as we wrap other ITK classes in vtkitk (it would be vtkitkNRRDReader). But it can be done later, I'm OK with integrating this as is.

ihnorton:
in r27102

PR #922, opened 2018-03-22
BUG: SegmentStatistics: Changing parameter set now update selected plugin list

chribaue:
Patch for https://issues.slicer.org/view.php?id=4520 Segment Statistics selected plugin list not updated when parameter set is changed


lassoan:
Thank you, it looks good. If there are no other comments then I'll integrate early afternoon.

cpinter:
LGTM, thanks!

jcfr:
:+1: 

Please consider changing the commit message to something like this:

```
BUG: SegmentStatistics: Changing parameter set now update selected plugin list

See https://issues.slicer.org/view.php?id=4520
```

chribaue:
> Please consider changing the commit message
I think there's no easy way to do this now on github after issuing the PR. Do you want me to issue another PR after changing the commit message?

cpinter:
You can amend the commit locally (only editing the message), then force push the topic branch.

lassoan:
Committed r27099

PR #925, opened 2018-03-24
ENH: Update VTK

jcfr:
In addition to the list of changes listed below, this commit allows to
disable the effect of these Slicer commits when building against VTK9:
* r23656 (BUG: 3808: fix failing test - add some pointer checks)
* r23648 (BUG: fixes for 3808 fiducial picking issue)
* r25672 (BUG: Fixed picking of markups in 3D view)

Indeed, thanks to commit kitware/VTK@8e4f2f7 from "Johan Andruejol <johan.andruejol@kitware.com>"
and "Ken Martin <ken.martin@kitware.com>", the workaround specific to Slicer/VTK fork
originally implemented by "Steve Pieper <pieper@bwh.harvard.edu>" is not needed anymore.

This properly address issue http://na-mic.org/Bug/view.php?id=3808

<s>A side effect is that fiducial are not disabled anymore in lightbox mode.</s> 

Highlighted list of changes
---------------------------

* kitware/VTK@3b814e9 (from Andras Lasso <lasso@queensu.ca>)
  ENH: Improve scene exporters and STL, OBJ, and PLY file IO

    - Allow reading/writing custom comments in STL, PLY, and OBJ files. This allows
    storing metadata, such as coordinate system, unit, or color in the files.
    - Allow reading/writing binary comments (that can contain 0 character) for STL
    files. This allows reading/writing color information in Mimics-style (Mimics
    software writes RGBA color as 4 bytes in STL file binary header).
    - Allow specifying renderer for exporting: Some exporters rejected to export
    a scene when more renderers were associated with a render window. This was an
    issue because it prevented scene export when additional renderers were used for
    displaying various annotations on the render window. Instead of hardcoding using
    the first renderer, added a ActiveRenderer member, which defines which renderer
    content should be exported. If not set then the first renderer is used, so the
    behavior is backward-compatible.
    - Fixed writing of of .mtl file path in .obj file (full path of .mtl file was
    written into the .obj file).

  Slicer: Used in Segmentation and SegmentEditor modules

* kitware/VTK@8e4f2f7 (from Johan Andruejol <johan.andruejol@kitware.com>)
  Fix picking through a disabled widget with picking manager on

    With picking manager on, it was impossible to pick a widget if a disabled
    widget was in front of it.
    To fix this, we have:
      - In the old style widgets (derived only from vtkInteractorObserver), we
    went through all the implementation and added the unregistration/registration
    of the pickers in the SetEnabled() method.
     - In new style widgets (derived from vtkAbstractWidget and
    vtkWidgetRepresentation) the method RegisterPickers and UnRegisterPickers
    have been moved into the representation public API to be able to be called
    by the widget's SetEnabled() method. This allows the widget to register and
    unregisters the pickers as necessary when enabled and disabled.

    In both cases we also implemented the SetPickingManaged method as previously
    the PickingManaged property was never used anywhere in the code base.
    Similarly, the PickersModified() method was removed as its purpose seemed
    redundant to SetPickingManaged().

    The test TestPickingManagerSeedWidget2 was added to demonstrate/test the
    picking behind a disabled widget.

    For more background information, see https://issues.slicer.org/view.php?id=3808

    Co-Authored-by: Ken Martin <ken.martin@kitware.com>
    Co-Authored-by: Jean-Christophe Fillion-Robin <jchris.fillionr@kitware.com>
    Thanks: Steve Pieper <pieper@bwh.harvard.edu>

  Slicer: Allow to remove workaround and Slicer/VTK specific patch associated with issue #3808

* kitware/VTK@d36b129 (from Andrew Bauer <andy.bauer@kitware.com>)
  Improve vtkAppendFilter efficiency (used by vtkAppendPolyData itself used in

    If there's a single unstructured grid in the composite dataset we
    now just do a shallow copy of that to the output.

  Slicer: used in vtkSlicerTransformLogic, vtkSlicerSegmentationsModuleLogic,
  vtkOrientedImageDataResample, vtkMRMLVolumeNode, MergeModels CLI,
  and ScriptedSegmentEditor templates

* kitware/VTK@a7988f5 (from Francois Budin <francois.budin@kitware.com>)
  OpenGL2: Check that context exists before trying to pop context.

    On some linux drivers (such as nvidia version 384.111 and 387.34) setting the
    context to zero causes a segfault so check before setting in cases where the
    destruction may be trying to pop to a zero context.

* kitware/VTK@a7988f5 (from Ken Martin <ken.martin@kitware.com>)
  Add support for writing larger unstructured grids

    Patch from gitlab Constantine @Butakoff

  Slicer: used in vtkMRMLModelStorageNode and ProbeVolumeWithModel

Complete list of changes
------------------------

<removed from the pull request description - complete list in commit message>


jcfr:
@pieper  Thanks to your initial fix pushed on the Slicer/VTK fork and detailed report, @vovythevov with input from @martinken  was able to address the issue so that we do not need a custom Slicer/VTK patch.  Note that following this patch the explicit disabling of annotation picking in lightbox mode is not active anymore. It was originally implemented in [r24726] (http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=24726)

@lassoan  your patche for IO is also available following this update.

jcfr:
Since there is no apparent regression (I confirmed that fiducial are effectively disabled in lightbox). I will integrate, we will follow up with VTK patch if needs to be.

jcfr:
Integrated in [r27104](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27104)

pieper:
Nice! 💯

PR #926, opened 2018-03-27
ENH: User information can be saved to application settings - WIP - do not merge yet

lassoan:
This PR supersedes https://github.com/Slicer/Slicer/pull/912

User information can be accessed through application logic:

    userInfo = slicer.app.applicationLogic().GetUserInformation()

If the userInfo object is changed, the updated values are automatically written into Slicer.ini.

    userInfo.SetName("John Doe")

Currently, validation is only performed for email address, just to illustrate how validation works.

   userInfo.SetEmail("invalid email")
   userInfo.SetEmail("valid.email@somewhere.com")

Limitations:
- There is no API yet for deleting keys.
- There is no GUI for users to change their user information.
- No tests have been added yet.

### DIFF
--- a/Libs/vtkAddon/vtkPersonInformation.cxx
+++ b/Libs/vtkAddon/vtkPersonInformation.cxx
@@ -116,11 +116,14 @@ std::string vtkPersonInformation::WriteToString()
   for (std::map<std::string, std::string>::iterator it = this->Data.begin();
     it != this->Data.end(); ++it)
     {
-    if (!output.empty())
+    if (!it->second.empty())
       {
-      output += ";";
+       if (!output.empty())
+        {
+        output += ";";
+        }
+      output += this->EncodeString(it->first) + ":" + this->EncodeString(it->second);
       }
-    output += this->EncodeString(it->first) + ":" + this->EncodeString(it->second);
     }
   return output;
 }
--- a/Libs/vtkAddon/vtkPersonInformation.cxx
+++ b/Libs/vtkAddon/vtkPersonInformation.cxx
@@ -1,0 +1,175 @@
+/*=auto=========================================================================
+
+  Portions (c) Copyright 2005 Brigham and Women's Hospital (BWH) All Rights Reserved.
+
+  See COPYRIGHT.txt
+  or http://www.slicer.org/copyright/copyright.txt for details.
+
+=========================================================================auto=*/
+
+#include "vtkPersonInformation.h"
+
+#include "vtkInformation.h"
+#include "vtkInformationStringKey.h"
+#include "vtkObjectFactory.h"
+
+#include <vtksys/SystemTools.hxx>
+
+// STD includes
+#include <sstream>
+
+vtkStandardNewMacro(vtkPersonInformation);
+
+#define vtkGetSetStringInformationCxxMacro(key) \
+  bool vtkPersonInformation::Set##key(const std::string& value) \
+  { \
+    this->SetCustomString(#key, value); \
+    return true; \
+  } \
+  std::string vtkPersonInformation::Get##key() \
+  { \
+    return this->GetCustomString(#key); \
+  }
+
+#define vtkGetSetValidatedStringInformationCxxMacro(key) \
+  bool vtkPersonInformation::Set##key(const std::string& value) \
+  { \
+    if (!this->Is##key##Valid(value)) \
+    { \
+      return false; \
+    } \
+    this->SetCustomString(#key, value); \
+    return true; \
+  } \
+  std::string vtkPersonInformation::Get##key() \
+  { \
+    return this->GetCustomString(#key); \
+  }
+
+vtkGetSetStringInformationCxxMacro(Name);
+vtkGetSetValidatedStringInformationCxxMacro(Email);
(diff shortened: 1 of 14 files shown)


lassoan:
This pull request should not be merged yet, as the API has to be discussed with other developers.

Libs/vtkAddon/vtkPersonInformation.cxx
```diff
@@ -0,0 +1,175 @@
...
+    int colonIndex = keyValuePair.find(':');
+    std::string key = this->DecodeString(keyValuePair.substr(0, colonIndex));
+    std::string value = this->DecodeString(keyValuePair.substr(colonIndex + 1));
+    this->SetCustomString(key, value);
+    }
+  return true;
+}
+
+//----------------------------------------------------------------------------
+std::string vtkPersonInformation::EncodeString(const std::string& value)
```
pieper: We have a generic url encode/decode function available vtkMRMLNode.  I wonder if that should move to vtkAddon and then be used from here and from mrml.

https://github.com/Slicer/Slicer/blob/8198bffc09899bc55140f6f375452ace4714c598/Libs/MRML/Core/vtkMRMLNode.cxx#L931

lassoan: There are already two encoders (URL and XML) in vtkMRMLNode, yet none of them are suitable for encoding these kind of property lists (where special characters are ```;``` and ```:```). The XML encoder would be usable if we stored the person information in an XML tree, which would be very flexible, but seems that it would be somewhat an overkill.

Regardless, we could move XML, URL, and generic property list encoders/decoders to vtkAddon.

pieper: Ah, so this is specific to property lists.  I guess that could be more explicit in the method name but otherwise okay to keep as a method here.

Libs/vtkAddon/vtkPersonInformation.cxx
```diff
@@ -0,0 +1,175 @@
...
+  std::string output = value;
+  // decode percent sign and semicolon (semicolon is a special character because it separates attributes)
+  vtksys::SystemTools::ReplaceString(output, "%3B", ";");
+  vtksys::SystemTools::ReplaceString(output, "%3A", ":");
+  vtksys::SystemTools::ReplaceString(output, "%25", "%");
+  return output;
+}
+
+//----------------------------------------------------------------------------
+bool vtkPersonInformation::IsEmailValid(const std::string& value)
```
jcfr: And since `kwsys` provide a [RegularExpression](https://gitlab.kitware.com/utils/kwsys/blob/master/RegularExpression.hxx.in) class, if needed, this type of check could be more sophisticated ... 

and by optionally associating regex if each field, UI could leverage them ...

jcfr:
:+1:

fedorov:
Looks good to me

fedorov:
Do we need further discussions of the API with the developers, or can this be merged?

fedorov:
@che85 @chribaue can one of you take on those tasks?

che85:
Sure I can take a look at it

che85:
@lassoan 
> There is no GUI for users to change their user information.

Is this supposed to be a settings panel then?

lassoan:
Yes, it would be nice to expose at least some of the user information in application settings GUI.

che85:
Ok.

> There is no API yet for deleting keys.

Did you mean deleting keys from the QSettings?

lassoan:
All properties (key/value pairs) are stored in a single QSettings. What I meant by deleting keys is deleting properties completely (not just setting the value to empty but remove the key). We could delete the key if the value is set to empty string, or add a "RemoveCustomString(const std::string& key)" method. I don't have any preference. If you don't think you need to distringuish between empty and missing value then we can go with the simpler option of deleting a key if its value is set to empty string.

lassoan:
Committed r27142.
Thanks @che85 for adding application settings panel and automatic test!

fedorov:
Thank you @lassoan and @che85!

fedorov:
@lassoan I downloaded the nightly today, and the app fails to start for me (or takes forever to start, not sure) - I have splash screen and the message below in the console, which looks related to this PR:

```
% /Applications/Slicer.app/Contents/MacOS/Slicer
QObject::connect: No such slot qSlicerSettingsUserInformationPanel::updateFromUserInformation(vtkObject*)
QObject::connect:  (receiver name: 'qSlicerSettingsUserInformationPanel')
Failed to obtain reference to 'qSlicerAppMainWindow'
```

lassoan:
The error message related to connect is most probably unrelated (I’ve fixed the issue yesterday - see https://github.com/Slicer/Slicer/commit/74d9a5cb75bdfb05331255854926506dd99ec1e7).

The startup hang happened to Ron, too. He "fixed" it by disconnecting from the network during startup. Probably it is caused by the extension update checker.

fedorov:
Thanks for the hint! For me it was yet another issue - I added a note to the existing thread on discourse here: https://discourse.slicer.org/t/splash-screen-does-not-disappear-after-startup-and-covers-another-popup/2155/5.

PR #927, opened 2018-03-31
BUG: Ensure markup is visible after clicking in 3d in persistent mode

jcfr:
After toggling persistent mode, placing a fiducial in 3D with Slice in the
background caused the fiducial to visible right way but clicking anywhere
else (without any slice in the background) was not causing the fiducial to
appear.

This commit ensures the rendering pipeline is triggered after adding
a fiducial in persistent mode.

Fixes https://issues.slicer.org/view.php?id=4512


Punzo:
Tested and it works. For me it is good to go.

jcfr:
Thanks for reviewing.

jcfr:
Closing. Fix integrated in [r27122](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27122)

PR #931, opened 2018-04-06
ENH: Added option to use TBB as SMP backend in VTK

lassoan:
It significantly improves vtk speed (typically by 20-30%, in some situations by a factor of 10x or more).

The change has been tested on Windows (building, running, packaging).

Linux build may need a patch (see TODO:) and Mac build may need some work with rpaths.

To test: when configuring Slicer build, set ```Slicer_VTK_SMP_IMPLEMENTATION_TYPE``` CMake variable to ```TBB```.

This is a follow-up of https://github.com/Slicer/Slicer/pull/930 (probably this will make that pull request obsolete).


lassoan:
I am tempted to integrate this and enable TBB by default on Windows, so that we can start collecting feedback from users.

TBB would be disabled on Linux and Mac until somebody can test&fix the build on these operating systems.

jcfr:
Makes sense. I will try to review and fix it before leaving on Monday.

SuperBuild/External_VTKv9.cmake
```diff
@@ -67,6 +67,20 @@ if((NOT DEFINED VTK_DIR OR NOT DEFINED VTK_SOURCE_DIR) AND NOT ${CMAKE_PROJECT_N
       -DModule_vtkGUISupportQtOpenGL:BOOL=ON
     )
   endif()
+  if(Slicer_USE_TBB)
+    list(APPEND ${proj}_DEPENDENCIES tbb)
```
jcfr: This should be added before the call to ` ExternalProject_Include_Dependencies`

lassoan: Feel free to make any changes. You should be able to push them directly to this branch.

jcfr:
Will do.

SuperBuild/External_VTKv9.cmake
lassoan: Order is fixed.

lassoan:
Committed r27132.
Builds and runs on Windows computers. Needs work for Linux and Mac.

pieper:
@lassoan join this if you want to discuss https://bit.ly/slicer-googlemeet-hosted-by-kitware

I was planning to try TBB on mac.

PR #932, opened 2018-04-07
ENH: Make model clipping GUI more user friendly

lassoan:
Clipping controls (enable clipping for selected model; clipping planes) were shown quite far from each other on Models module GUI. The user had to click enable clipping, disable backface culling, scroll down to clipping planes section, open it, and select clipping planes. Very often users could not figure out that they needed all these steps.

Implemented solution:

Clipping planes are applied globally, to all models, therefore separation from display properties of the selected model is good. But made it easier to get from model-specific clipping enable/disable checkbox to clipping plane definition by adding a "Configure" button next to the clipping enable checkbox, which opens the clipping planes section and scrolls the module widget to ensure the section is visible.

image

Added a checkbox to clipping planes section, which allows the user to enable/disable clipping of the selected model at the same place where clipping planes are defined. If user enables clipping then backface culling is automatically disabled and red slice clipping is automatically enabled.

image

### DIFF
--- a/Modules/Loadable/Models/Resources/UI/qSlicerModelsModuleWidget.ui
+++ b/Modules/Loadable/Models/Resources/UI/qSlicerModelsModuleWidget.ui
@@ -216,6 +216,9 @@
           <verstretch>0</verstretch>
          </sizepolicy>
         </property>
+        <property name="clippingConfigurationButtonVisible">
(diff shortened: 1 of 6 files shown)


pieper:
LGTM - nice work!

lassoan:
Committed r27141

PR #934, opened 2018-04-11
Multi volume rendering

cpinter:
Volume rendering displayable manager has been completely reimplemented to support displaying multiple volumes.


cpinter:
There are two outstanding issues:
* This is more a question than a problem: Clipping plane normals needed to be inverted (in vtkMRMLVolumeRenderingDisplayableManager::vtkInternal::UpdatePipelineROIs) for the cropping to work. The reason for the normals having pointed in the opposite direction is unknown, as the code that calculates the normals did not change. I implemented a workaround (in the function mentioned above) that flips the normals, but it would be nice to know why they were inverted in the first place.
* This is an actual bug, but the problem occurs in the OpenGL rendering pipeline, and I don't have any ideas how to track this down. The bug is that if surface smoothing is turned on for a shown volume, the other (also shown) one disappears. If surface smoothing is turned on and off for the disappearing volume then it shows up again, and then it starts working properly (i.e. changing the setting does not hide volumes anymore). The message is the following: "Generic Warning: In C:\e\ScD\VTKv9\Rendering\VolumeOpenGL2\vtkOpenGLGPUVolumeRayCastMapper.cxx, line 1322	Error after glDrawElements in RenderVolumeGeometry! 1 OpenGL errors detected:   0 : (1282) Invalid operation"

cpinter:
Update: It seems that the problem about overlapping is that instead of multiple mappers, a common multi-volume mapper needs to be used, see in [this example](https://github.com/Kitware/VTK/blob/master/Rendering/Volume/Testing/Cxx/TestGPURayCastMultiVolumeOverlapping.cxx). If this assumption is correct, then the first and third issue described above can be easily solved. What I'll do in another commit within this PR:
* Change the displayable manager to have one vtkMultiVolume and one mapper, and the pipelines are propagated to these common instances
* Blending mode, jittering, GPU memory and other mapper-specific properties will be moved from the volume rendering display node to a new singleton node that stores these now-global properties

pieper:
That is so funny @cpinter I had just built and tested the multivolume mode and was going to comment on the overlapping depth issue.  I hope the unified mapper strategy works!  I'll test it when it's ready.  Otherwise the code looks good.  

Probably the visibility 'eye' should be a checkable combobox to be more obvious that multiple selections are possible.  (like in the per-view visibility selection).

lassoan:
> Probably the visibility 'eye' should be a checkable combobox to be more obvious that multiple selections are possible. (like in the per-view visibility selection).

We could even have a subject hierarchy tree (filtered to show only volume nodes) instead of a simple listbox. It would make it much easier to show/hide multiple volumes. No need to change this in this pull request, just an idea for the future.

cpinter:
Thanks, Steve, for building the branch, and sorry for omitting the depth issue in my initial description! That issue is critical, but so far we assumed the issue is in VTK, that's why I looked at it differently.

I agree, there could be an easier way of showing/hiding volumes. The reason I'd keep Volume Rendering module as is, and implement something better in subject hierarchy instead (for example separate "2D" and "3D" visibility controls), is that all modules dealing with nodes allow selecting one node, and set its properties. The only exception currently is the Models module with the tree. So I think with this very simple combobox + eye icon setting, the module conforms to the Slicer conventions. If we need better controls in the Volume Rendering module - which makes sense because to show a volume in 3D the user needs to set at least the preset but usually other things too - then a subject hierarchy combobox with only the 3D visibility controls visible would be what I'd go for.

pieper:
I like the idea of a SH combo box.  I think the main reason to change it would be as a visual hint so that users know that the option is even available.  Otherwise they might think there is still the old on-at-a-time limitation.

cpinter:
I have been working on using vtkMultiVolume with a common mapper. Although it seems to handle overlapping well, there are some blocking issues:
* vtkMultiVolume seems not to support shading
* It's very slow, although there are no CPU mappers instantiated
* Issues with handling UserTransform
  * The first volume's UserTransform is ignored, and it appears as it were isotropic with 1,1,1 spacing
  * When changing the UserTransform of one volume, the other volume is also transformed in volume rendering when the bounds start to approach each other

Not sure yet which of these are bugs in VTK (I suspect that at least the shading and user transform problems are in VTK).

I'll keep investigating these issues, but I doubt there will be a perfect solution in the near future. There seems to be an agreement that I integrate this PR as is (see [here](https://github.com/Slicer/Slicer/pull/936)). As @lassoan is updating VTK today, I'll integrate this tomorrow, so that there is a nightly release between the two, to be able to track down bugs in the future.

cpinter:
Integrated in https://github.com/Slicer/Slicer/commit/0f8926d1391f5abc2be845890dae369453465317

PR #935, opened 2018-04-11
BUG: Fixed conventional layout size

lassoan:
Fixed view sizes in layouts where splitter was used. Splitter size is now explicitly set to get the desired size ratio between views.

Fixes https://issues.slicer.org/view.php?id=4530

Requires this CTK improvement: https://github.com/commontk/CTK/pull/795


lassoan:
Integrated in http://svn.slicer.org/Slicer4/trunk@27147

PR #936, opened 2018-04-16
ENH: Added new volume rendering quality option: Normal

lassoan:
SetLockSampleDistanceToInputSpacing usually provides good quality rendering at acceptable speed, but on slow computers or in virtual reality views (where very high frame rate is needed), it is still necessary to allow choose between volume rendering speed and quality.

Added a new option to clearly distinguish between quality settings:
- adaptive: sampling computed from desired fps
- normal: sampling computed from input volume
- maximum quality: sampling computed from input volume, with 10x oversampling

### DIFF
--- a/Modules/Loadable/VolumeRendering/MRML/vtkMRMLGPURayCastVolumeRenderingDisplayNode.h
+++ b/Modules/Loadable/VolumeRendering/MRML/vtkMRMLGPURayCastVolumeRenderingDisplayNode.h
@@ -59,14 +59,10 @@ public:
   vtkSetMacro(RaycastTechnique, int);
 
   // Description:
-  // Jittering
-  vtkGetMacro(UseJittering, int);
-  vtkSetMacro(UseJittering, int);
-
-  // Description:
-  // Lock sample distance to volume spacing
-  vtkGetMacro(LockSampleDistanceToInputSpacing, int);
-  vtkSetMacro(LockSampleDistanceToInputSpacing, int);
+  // Reduce wood grain artifact to make surfaces appear smoother.
+  // For example, by applying jittering on casted rays.
+  vtkGetMacro(SurfaceSmoothing, bool);
+  vtkSetMacro(SurfaceSmoothing, bool);
 
 protected:
   vtkMRMLGPURayCastVolumeRenderingDisplayNode();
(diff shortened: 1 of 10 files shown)


pieper:
LGTM!

Modules/Loadable/VolumeRendering/MRML/vtkMRMLGPURayCastVolumeRenderingDisplayNode.h
```diff
@@ -81,14 +77,8 @@ class VTK_SLICER_VOLUMERENDERING_MODULE_MRML_EXPORT vtkMRMLGPURayCastVolumeRende
...
 
-  /// Use jittering to reduce the wood-grain effect if on. Off by default
-  int UseJittering;
-
-  /// Compute sample distance from volume spacing if on. Off by default
-  /// If turned on, then AutoAdjustSampleDistances and explicit SampleDistance
-  /// will be ignored, and only the internal auto-calculation will determine
-  /// the sampling distance.
-  int LockSampleDistanceToInputSpacing;
+  /// Make surface appearance smoother. Off by default
```
cpinter: I think the part of the comment about the wood-grain atrifact is meaningful. I don't think making the documentation more vague is desirable.

cpinter:
Looks good to me.

I propose that we integrate the [multi-volume rendering PR ](https://github.com/Slicer/Slicer/pull/934) soon, so that it does not get much harder to merge it after changes like this. Although it is not perfect yet, it is an incremental improvement without it breaking anything, so even until the overlapping issue is solved, it could be integrated.

Modules/Loadable/VolumeRendering/MRML/vtkMRMLGPURayCastVolumeRenderingDisplayNode.h
lassoan: I moved these details to the get/set method description so that is included in the API documentation.

We can include details here as well, but I thought it is easier to maintain documentation if there is less duplication.

cpinter: OK no problem. I didn't see the set/get part, just saw that this info was not there anymore.

lassoan:
Committed r27153

PR #939, opened 2018-04-22
ENH: Add ability to load custom volume rendering presets

cpinter:
Added the option to use custom presets scene instead of the default one by calling the LoadCustomPresetsScene function of the volume rendering logic. This might be needed for specific applications when only a subset of the presets need to be shown, or if new presets need to be added.

Added showIcons property to the presets combobox classes. This is useful if new presets in the custom presets scene do not have icons, or simply when showing icons are not desired. Default is showing icons.

### DIFF
--- a/Modules/Loadable/VolumeRendering/Logic/vtkSlicerVolumeRenderingLogic.cxx
+++ b/Modules/Loadable/VolumeRendering/Logic/vtkSlicerVolumeRenderingLogic.cxx
@@ -1173,3 +1173,19 @@ void vtkSlicerVolumeRenderingLogic::RemovePreset(vtkMRMLVolumePropertyNode* pres
     }
   presetScene->RemoveNode(preset);
 }
+
+//---------------------------------------------------------------------------
+int vtkSlicerVolumeRenderingLogic::LoadCustomPresetsScene(const char* sceneFilePath)
+{
+  if (!this->PresetsScene)
+    {
+    this->PresetsScene = vtkMRMLScene::New();
+    }
+  else
+    {
+    this->PresetsScene->Clear(1);
+    }
+
+  this->PresetsScene->SetURL(sceneFilePath);
+  return this->PresetsScene->Import();
+}
--- a/Modules/Loadable/VolumeRendering/Logic/vtkSlicerVolumeRenderingLogic.h
+++ b/Modules/Loadable/VolumeRendering/Logic/vtkSlicerVolumeRenderingLogic.h
@@ -44,8 +44,6 @@ class vtkVolumeProperty;
 #include <map>
 #include <vector>
 
-class vtkStringArray;
-
 /// \ingroup Slicer_QtModules_VolumeRendering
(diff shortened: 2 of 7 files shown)


Modules/Loadable/VolumeRendering/Logic/vtkSlicerVolumeRenderingLogic.cxx
```diff
@@ -1173,3 +1173,19 @@ void vtkSlicerVolumeRenderingLogic::RemovePreset(vtkMRMLVolumePropertyNode* pres
     }
   presetScene->RemoveNode(preset);
 }
+
+//---------------------------------------------------------------------------
+int vtkSlicerVolumeRenderingLogic::LoadCustomPresetsScene(const char* sceneFilePath)
```
lassoan: This helper function is not really needed, as you can call clear the scene by simply calling

    logic->GetPresetsScene()->Clear();

or import a scene as described here: https://www.slicer.org/wiki/Documentation/Nightly/Modules/VolumeRendering#How_to_register_custom_volume_rendering_presets

but I guess this should work, too:

    logic->GetPresetsScene()->SetURL(...);
    logic->GetPresetsScene()->Import();

If you think helper functions are needed then it would be better to keep the name and behavior of the called methods. For example, instead of LoadCustomPresersScene() it would be better to name it ImportCustomPresetsScene().

Modules/Loadable/VolumeRendering/Logic/vtkSlicerVolumeRenderingLogic.cxx
```diff
@@ -1173,3 +1173,19 @@ void vtkSlicerVolumeRenderingLogic::RemovePreset(vtkMRMLVolumePropertyNode* pres
...
+{
+  if (!this->PresetsScene)
+    {
+    this->PresetsScene = vtkMRMLScene::New();
```
lassoan: This would break the GUI: all the widgets that are already created would not be updated anymore.

Modules/Loadable/VolumeRendering/Logic/vtkSlicerVolumeRenderingLogic.cxx
cpinter: I don't see why having a utility function for something is undersired. We have tons of functions that save the developers writing a few lines of code.

"Import" implies that whatever was in the presets scene before stays. This convenience function however imports the file to an empty scene. So the name "load" was intentional.

Modules/Loadable/VolumeRendering/Logic/vtkSlicerVolumeRenderingLogic.cxx
cpinter: I don't understand. Creating a new scene only happens if the scene was NULL before.

lassoan: OK, good then.

PR #940, opened 2018-04-25
BUG: Added API to add border around fullscreen app window

lassoan:
On Windows, when turning on OpenGL and using the full screen mode (mainWindow=slicer.util.mainWindow().showFullScreen()),
menus and tooltips were no longer visible.

By enabling setHasBorderInFullScreen, a one-pixel border is added around the window, which fixes the problem.

See http://doc.qt.io/qt-5/windows-issues.html#fullscreen-opengl-based-windows


lassoan:
Committed r27162

PR #941, opened 2018-04-25
ENH: Add color support to 3D view controllers

cpinter:
The layout view property 'viewcolor' was only available for slice views. By sinking the LayoutColor member variable from vtkMRMLSliceNode to vtkMRMLAbstractViewNode, and adding accessor functions to qMRMLThreeDWidget and controller similar to those in the slice widgets, this property can now be set to 3D views as well.


Libs/MRML/Widgets/qMRMLSliceWidget.h
```diff
@@ -85,27 +85,27 @@ class QMRML_WIDGETS_EXPORT qMRMLSliceWidget : public qMRMLWidget
 
   /// \sa qMRMLSliceControllerWidget::sliceViewName()
   /// \sa setSliceViewName()
-  QString sliceViewName()const;
+  Q_INVOKABLE QString sliceViewName()const;
 
   /// \sa qMRMLSliceControllerWidget::sliceViewName()
   /// \sa sliceViewName()
-  void setSliceViewName(const QString& newSliceViewName);
+  Q_INVOKABLE void setSliceViewName(const QString& newSliceViewName);
```
lassoan: Why not make these properties instead?
I think we usually define properties for variable types whenever it is possible, and for QString it should be possible.

cpinter:
Thanks Andras! I've just rebased to latest, build and test one more time, then integrate.

lassoan:
Ok. Have you seen my comment about property vs for nvokable?

cpinter:
Yes, that's why I tested it again. Integrated in https://github.com/Slicer/Slicer/commit/2504cf9c9092b18d933c2322bb91f2e7f47275e1

Thanks!

PR #943, opened 2018-04-28
COMP: remove hard QtPositioning dependency

ihnorton:
It is only a soft (Qt) build-time dependency so we don't need to enforce it here:

  https://bugreports.qt.io/browse/QTBUG-57418

### DIFF
--- a/CMake/SlicerBlockInstallQt.cmake
+++ b/CMake/SlicerBlockInstallQt.cmake
@@ -37,13 +37,11 @@ else()
   # WebEngine Dependencies
   if("Qt5::WebEngine" IN_LIST QT_LIBRARIES)
     find_package(Qt5 REQUIRED COMPONENTS
-      Positioning
       Qml
       Quick
       QuickWidgets
       )
     list(APPEND QT_LIBRARIES
-      "Qt5::Positioning"
       "Qt5::Qml"
       "Qt5::Quick"
       "Qt5::QuickWidgets"


pieper:
I read the linked bug report to be saying that if WebEngine was built with Positioning then it is a runtime requirement.  Have you tested that this change won't break packaging?  Positioning is bundled with the binary downloads of Qt so I would think the WebEngine is built with it.

ihnorton:
Packaging looks directly at the rpath dependencies, so it should be fine. But I'll double-check next week.

jcfr:
This is currently only true on macOS. For other platforms, it needs to be
explicit.

ihnorton:
How about using `OPTIONAL_COMPONENTS Positioning` instead? Then I think it should still show up in the library list if-found.

jcfr:
Assuming the Qt5 cmake config module support it, that would work.

jcfr:
I confirm that on a regular Qt install on linux, QtPositioning is hard dependency:

```
$ cd /home/jcfr/Software/Qt5.11.1/5.11.1/gcc_64
$ for lib in $(find . | ack -i \.so$); do deps=$(ldd $lib | ack Positio); [[ $deps != "" ]] && (echo $lib; echo "  $deps"); done
./lib/libQt5WebEngineCore.so
  	libQt5Positioning.so.5 => /home/jcfr/Software/Qt5.11.1/5.11.1/gcc_64/./lib/libQt5Positioning.so.5 (0x00007fa433ac6000)
./lib/libQt5WebEngine.so
  	libQt5Positioning.so.5 => /home/jcfr/Software/Qt5.11.1/5.11.1/gcc_64/./lib/libQt5Positioning.so.5 (0x00007f92d8358000)
[...]
```

jcfr:
Closing. This has been integrated in [r28330](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28330) leveraging `OPTIONAL_COMPONENTS`

PR #944, opened 2018-05-03
BUG: fix QtWebEngineProcess for extension manager on linux

ihnorton:
QWebEngineProcess could not find resource paths due to an upstream bug.

The fix is not available until Qt 5.11, so work-around is to include qt.conf
file next to the QtWebEngineProcess executable, setting the relative path
to 'resources/'

  https://issues.slicer.org/view.php?id=4544

upstream
  bug: https://bugreports.qt.io/browse/QTBUG-66346
  fix: https://codereview.qt-project.org/#/c/220734/


pieper:
Looks good to me - I'd say go ahead and commit it as it's hard for me to imagine a downside.

ihnorton:
In r27179

PR #946, opened 2018-05-15
BUG: Fix install path for python-real

cquammen:
python-real is inside the same directory as SlicerPython, so the
installation subdirectory should be . instead of Slicer_BIN_DIR.

Fixes https://issues.slicer.org/view.php?id=4553.


cquammen:
cc @jcfr

jcfr:
Thanks @cquammen 

Changed tested locally and integrated in [r27189](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27189)

cquammen:
Thanks!

PR #947, opened 2018-05-16
BUG: Fix PYTHONPATH in installation so that vtk can be imported

cquammen:
Add an additional path (different on each OS) so that the vtk
Python package can be found.

Fixes https://issues.slicer.org/view.php?id=4556.

### DIFF
--- a/SuperBuild/python_configure_python_launcher.cmake
+++ b/SuperBuild/python_configure_python_launcher.cmake
@@ -126,6 +126,21 @@ set(PYTHONLAUNCHER_PYTHONPATH_INSTALLED
   "<APPLAUNCHER_DIR>/../${Slicer_BIN_DIR}/Python"
   )
 
+# Add paths necessary to import VTK
+if (APPLE)
+  list(APPEND PYTHONLAUNCHER_PYTHONPATH_INSTALLED
+    "${PYTHONHOME}/lib/Slicer-4.9"
+    )
+elseif(MSVC)
+  list(APPEND PYTHONLAUNCHER_PYTHONPATH_INSTALLED
+    "${PYTHONHOME}/${Slicer_BIN_DIR}/Lib/site-packages"
+  )
+else()
+  list(APPEND PYTHONLAUNCHER_PYTHONPATH_INSTALLED
+    "${PYTHONHOME}/lib/Slicer-4.9/python2.7/site-packages"
+  )
+endif()
+
 #
 # Notes:
 #


cquammen:
cc @jcfr

SuperBuild/python_configure_python_launcher.cmake
```diff
@@ -126,6 +126,21 @@ set(PYTHONLAUNCHER_PYTHONPATH_INSTALLED
   "<APPLAUNCHER_DIR>/../${Slicer_BIN_DIR}/Python"
   )
 
+# Add paths necessary to import VTK
+if (APPLE)
+  list(APPEND PYTHONLAUNCHER_PYTHONPATH_INSTALLED
+    "${PYTHONHOME}/lib/Slicer-4.9"
```
cquammen: @jcfr Do you know the CMake-ish to put in in place of `Slicer-4.9`?

SuperBuild/python_configure_python_launcher.cmake
```diff
@@ -126,6 +126,21 @@ set(PYTHONLAUNCHER_PYTHONPATH_INSTALLED
...
+    )
+elseif(MSVC)
+  list(APPEND PYTHONLAUNCHER_PYTHONPATH_INSTALLED
+    "${PYTHONHOME}/${Slicer_BIN_DIR}/Lib/site-packages"
```
lassoan: I think this should be:
"<APPLAUNCHER_DIR>/../${Slicer_BIN_DIR}/Lib/${PYTHON_SITE_PACKAGES_SUBDIR}"

SuperBuild/python_configure_python_launcher.cmake
lassoan: Slicer_LIB_DIR is set to "lib/Slicer-4.9", so you might use it here.

SuperBuild/python_configure_python_launcher.cmake
```diff
@@ -126,6 +126,21 @@ set(PYTHONLAUNCHER_PYTHONPATH_INSTALLED
   "<APPLAUNCHER_DIR>/../${Slicer_BIN_DIR}/Python"
   )
 
+# Add paths necessary to import VTK
+if (APPLE)
+  list(APPEND PYTHONLAUNCHER_PYTHONPATH_INSTALLED
+    "${PYTHONHOME}/lib/Slicer-4.9"
+    )
+elseif(MSVC)
```
lassoan: We very rarely use MSVC variable in Slicer CMake files. Instead, we mostly use UNIX, APPLE, and WIN32. So, it could be if (APPLE) ... elseif (UNIX) ... else() ... endif()

SuperBuild/python_configure_python_launcher.cmake
```diff
@@ -126,6 +126,21 @@ set(PYTHONLAUNCHER_PYTHONPATH_INSTALLED
...
+  list(APPEND PYTHONLAUNCHER_PYTHONPATH_INSTALLED
+    "${PYTHONHOME}/${Slicer_BIN_DIR}/Lib/site-packages"
+  )
+else()
+  list(APPEND PYTHONLAUNCHER_PYTHONPATH_INSTALLED
+    "${PYTHONHOME}/lib/Slicer-4.9/python2.7/site-packages"
```
lassoan: probably you could use ${PYTHON_SITE_PACKAGES_SUBDIR} here

SuperBuild/python_configure_python_launcher.cmake
cquammen: Great, thanks.

SuperBuild/python_configure_python_launcher.cmake
cquammen: 50/50, I chose wrong :-). Will fix.

jcfr:
Thanks for the patch :+1: 

In the comment below are discussed the `approach 1` proposed by this PR as well as some other approaches (2 and 3), including one that I would recommend (Approach 4). 


## Background

Originally, `SlicerPython` was created to be the equivalent of `python`. This was needed to ensure the corresponding python library and openssl libraries as well as standard package could be found during the build step of Slicer that are using python.

Later one, we thought it would be sensible to distribute it.

## Possible way forward

### Approach 1: Ensure SlicerPython is configured with launcher settings for VTK/ITK/Slicer/... 

Considering the `SlicerPython` launcher is configured at the same time of python external project but before VTK/ITK/Slicer python packages are configured, the approach proposed by this PR is to to duplicate settings like `PYTHONPATH` to support the usage of the launcher to import ITK/VTK/Slicer/... python packages.

**Problem**: While duplicating the settings would work in the short term, it would make the maintenance more difficult. Indeed, PYTHONPATH settings specific to an external project (e.g VTK) would have to be added in both `External_VTK` file and also in `External_python`. This would break the "modularity" allowing us to build custom application and addition other external project.

**Mitigation:** leverage the two passes nature of the build system

To mitigate this, we could leverage the fact [ExternalProjectDependency](https://github.com/commontk/Artichoke) (integrated [here](https://github.com/Slicer/Slicer/blob/master/CMake/ExternalProjectDependency.cmake)) used to load all `External_*` files is doing two passes:
* pass 1: include all `External_*` to create the complete list of dependencies
* pass 2: effectively include enabled projects

Launcher settings could then be set as a global property during the first pass ... and then be available in the second pass in `External_python` file.

**Problem:** this would defeat the purpose of having two passes. In some case, settings can only be set after the first pass is complete.


### Approach 2: Use of Slicer launcher

An other approach could be to use Slicer launcher directly:

```
./Slicer --launch python-real /path/to/script
```

**Problem:** The issue is that this will work only on Linux and Windows ... because on macOS, `Slicer` is not a launcher.

### Approach 3: Leverage `--launcher-additional-settings`

Considering all additional settings can already be found in `SlicerLauncherSettings.ini` available and distributed on all platform (including macOS)

I wonder if the following would be a reasonable solution:

```
./SlicerPython --launcher-additional-settings ./path/to/SlicerLauncherSettings.ini  /path/to/script
```

Also if the python interpreter needs to be directly used, ones could easily source the environment doing something like this:

```
eval $(./SlicerPython --launcher-additional-settings ./path/to/SlicerLauncherSettings.ini --launcher-show-set-environment-commands)
```

**Problem:** To ensure that the installed settings are relocatable, the special string `<APPLAUNCHER_DIR>` is used. This means that the settings can **NOT** be used with an other launcher located in a different directory. To mitigate this, we would instead support `<APPLAUNCHER_CURRENT_SETTINGS_DIR>`.


### Approach 4: Improve launcher to support `<APPLAUNCHER_CURRENT_SETTINGS_DIR>` and support hard-coded location of additional launcher settings

I think this approach is the most reasonable.

Doing so, we can easily leverage the settings already configured for Slicer


## Some more pointers

Currently VTK PYTHONPATH are configured in the associated external project. See [here](https://github.com/Slicer/Slicer/blob/992cdf895387b77f3fdc4aeb80b319eb55f5dfde/SuperBuild/External_VTKv9.cmake#L190-L253)

>  Do you know the CMake-ish to put in in place of Slicer-4.9?

See [CMake/SlicerBlockCTKAppLauncherSettings.cmake](https://github.com/Slicer/Slicer/blob/992cdf895387b77f3fdc4aeb80b319eb55f5dfde/CMake)/SlicerBlockCTKAppLauncherSettings.cmake#L277-L298]

SuperBuild/python_configure_python_launcher.cmake
cquammen: Is `PYTHON_SITE_PACKAGES_SUBDIR` set to `python2.7/site-packages`? I could check, but I don't have a Slicer build set up locally (and I'm trying to avoid having one if possible.

lassoan:
@jcfr Are we talking about SlicerPython in the build tree or in the install tree?
I guess installation tree should be quite trivial.

My main use case for the custom Python interpreter would be to set it for the Python IDE during Python module development. So, if it works in the install tree then it would be enough for me for now.

For Python-based CLI development (and automatic testing during build time) it would be certainly useful to make the same libraries available in SlicerPython in the build tree as in the install tree.

ihnorton:
> because on macOS, Slicer is not a launcher.

Can you remind me what is the reason for this? (I know it is not necessary, but it's always confusing that `Slicer-build/Slicer` is a launcher in a local mac build, but in a package, `Contents/MacOS/Slicer` is not)

lassoan:
The Python interpreter must be a simple executable. You cannot set in a Python IDE to use `./Slicer --launch python-real` as Python interpreter.

It seems that even the executable name of the Python interpreter is restricted: it must start with python. Exceptions are added for well-known interpreters, such as ipy or jython, but we cannot expect that for SlicerPython. So, if we want documentation and auto-complete in PyCharm then SlicerPython will have to be renamed to PythonSlicer or pythonslicer or something similar.

lassoan:
Do I understand correctly, that Approach 4 would add an `additionalLauncherSettings` variable to `SlicerPythonLauncherSettings.ini` that would point to `...\Slicer-build\SlicerLauncherSettings.ini`? It seems quite clean solution and also simple to implement.

jcfr:
@lassoan 

> Do I understand correctly, that Approach 4 would add an additionalLauncherSettings variable to SlicerPythonLauncherSettings.ini that would point to ...\Slicer-build\SlicerLauncherSettings.ini? It seems quite clean solution and also simple to implement.

Exactly.

To support this, we would have to also support `<APPLAUNCHER_CURRENT_SETTINGS_DIR>`


@ihnorton 

>  because on macOS, Slicer is not a launcher. 
>> Can you remind me what is the reason for this?

On macOS, the never needed a launcher because dependencies to libraries are all resolved using `@rpath`.  But for consistency, environment variables are also loaded using the launcher settings. See https://github.com/Slicer/Slicer/blob/992cdf895387b77f3fdc4aeb80b319eb55f5dfde/Base/QTCore/qSlicerCoreApplication.cxx#L222-L268

ihnorton:
@lassoan SlicerPython doesn't currently work as an interactive interpreter; does PyCharm communicate through stdin/stdout or some other method? (CTKApplauncher would probably need to dup the handles, which I didn't see a way to do with QProcess; there's a related [Qt3-era blog post](http://blog.qt.io/blog/2006/03/16/starting-interactive-processes-with-qprocess/)...)

lassoan:
PyCharm works perfectly with SlicerPython.exe as interpreter if I rename it to PythonSlicer.exe.

cquammen:
I am totally happy with whatever approach you all choose.

ihnorton:
@lassoan that's great. PyCharm works because it runs a script to start a server and communicate over sockets. What I meant is that SlicerPython can not provide an *interactive prompt*. I would like that to be possible, but it will take some unrelated changes and the discussion can happen elsewhere.

lassoan:
Yes, it does not work by simply starting SlcerPython. But it does work through PyCharm. I don't know how it works internally, I just specify PythonSlicer.exe as a Python interpreter in PyCharm and open a Python command prompt in PyCharm and it works. I can enter import slicer, import vtk, enter commands that use vtk, auto-complete works in the console, etc.

You can give it a try, too. I have professional edition (free for everybody with a university email address), but it should also work in the free community edition.

PR #949, opened 2018-05-16
BUG: Fix pydicom check when using system python

phcerdan:


phcerdan:
@jcfr

jcfr:
Thanks for the patch

Instead of installing `pydicom`, you could instead install `dicom`


@fedorov @pieper @lassoan 

It turns out that the version 0.9.9 we use in Slicer is 4 year old and was indeed associated with a `dicom` module. See https://github.com/pydicom/pydicom/tree/v0.9.9/source

It may be sensible to update the version of pydicom to the latest release. See https://pydicom.github.io/pydicom/stable/transition_to_pydicom1.html  We could also keep the update for Slicer 5.0

phcerdan:
Nice, dicom works too. I got confused reading recent documentation, where they `import pydicom`.

phcerdan:
However I got this warning using dicom at configure time:
```bash
This code is using an older version of pydicom, which is no longer
maintained as of Jan 2017.  You can access the new pydicom features and API
by installing `pydicom` from PyPI.
See 'Transitioning to pydicom 1.x' section at pydicom.readthedocs.org
for more information.

```

phcerdan:
Yes, `pip install dicom` is exactly that version already. Same message with `pip install -U dicom==0.9.9.post1`. But all good, python compiles with system libraries:

`pip install dicom chardet appdirs packaging pyparsing smmap gitdb GitPython PyGithub nose numpy`

jcfr:
In that case, the following should install the expected version:

```
pip install -U git+https://github.com/pydicom/pydicom.git@v0.9.9#subdirectory=source
```

jcfr:
Closing this PR for now. We will revisit when we effectively update the version of pydicom used in Slicer.

PR #951, opened 2018-05-17
BUG: Closed slicer.util displays references deleted

jamesobutler:
Fixes https://issues.slicer.org/view.php?id=4536

Slicer.util displays (infoDisplay, warningDisplay, errorDisplay, etc) that use `messageBox` were still visible when hovering over Slicer's Windows 10 taskbar preview even though they were closed. This was causing confusion about whether the displays had actually been closed.

References are now scheduled for deletion when the display is closed.

### DIFF
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -1091,6 +1091,7 @@ def messageBox(text, parent=None, **kwargs):
   for key, value in kwargs.iteritems():
     if hasattr(mbox, key):
       setattr(mbox, key, value)
+  mbox.deleteLater()
   return mbox.exec_()
 
 def createProgressDialog(parent=None, value=0, maximum=100, labelText="", windowTitle="Processing...", **kwargs):


lassoan:
Can you still reproduce this bug? I've tried it on 4 different Windows 10 computers and none of them shows popup windows at all, in any software (not just Slicer) in the taskbar thumbnails. Do you have some special settings on that computer (non-default taskbar thumbnail settings, Windows updates disabled, ...)?

When a Python variable goes out of scope then the associated Qt object should be deleted automatically. It should not require an additional deleteLater call.

jamesobutler:
Sorry if I wasn't explaining it clearly/correctly.

Yes I am able to reproduce and on multiple Windows 10 computers that are all using default settings. The closed popups aren't visible in the taskbar previews themselves, but when you hover the mouse over the taskbar preview to peek into the window you can still see the closed popup.  Then when you click the preview to go to the window the closed popup will no longer be in view. Minimizing/Maximizing the Slicer window causes this issue to go away, or calling deleteLater.
slicer1
slicer2

lassoan:
I still cannot reproduce it. Popups never show up in peek views and the full window is not shown when I hover over the taskbar. This video shows what I'm seeing: https://1drv.ms/v/s!Arm_AFxB9yqHtJR5upBXJqmlaQ2qzA

Is that Windows fully up-to-date? I have Microsoft Windows 10 Pro 10.0.16299 Build 16299.
I think it was an old behavior that full windows showed up when you hovered over the taskbar and not the case anymore.

Your patch looks harmless and the whole issue is not worth a lot of time to be spent on, it would be just nice to understand what's happening.

jamesobutler:
From your video, once you closed the dialog you didn't hover over the taskbar preview.

I'm running the latest Windows 10 Home version.  I'm actually more up-to-date then you. I'm running Build 17134, but I know the issue was also happening for build 16299.  Maybe this is an issue with ctk since errorDisplay is actually calling ctkMessageBox and not a QMessageBox directly?

Here's a video of what I'm doing:
https://1drv.ms/v/s!Apu1i4aiHcpijg2NPd-MsHUucefa

jamesobutler:
Here's another video demonstrating that it happens with a recent slicer nightly build.  
https://1drv.ms/v/s!Apu1i4aiHcpijg7Eq-PJEcokwKOr
Opening the dicom browser, multiple util displays, and closed progress bar dialogs can result in a confusing view when peeking into the window because all the closed popups are visible.

This PR was only for the message box displays because that's what I use most often and care about, but my fix wouldn't solve these similar problems.

lassoan:
I was able to reproduce this with Slicer. The behavior is the same with Paraview (for example: Edit / Find data dialog), MeshMixer (File/Preferences). The problem is that peek shows all hidden but not destroyed windows. Probably it would need to be fixed in Qt or Windows, but the current workaround is reasonable.

Committed r27195.

jamesobutler:
I was able to find an old bug report in Qt's tracker system that is exactly what I was describing. One of the developers punted it saying it was a Windows 10 aero peek issue and suggested the workaround of disabling aero peek for the window (see [Qt Windows Extras](http://doc.qt.io/qt-5/qtwinextras-overview.html) originally introduced in Qt5.2).
- [QTBUG-48855](https://bugreports.qt.io/browse/QTBUG-48855) "Windows 10: Closed dialogs become visible in live preview when hovering mouse over the application on taskbar"

The only other workaround that I observed that doesn't involve destroying the window is to change the parent of the window because Windows 10 aero peek is evidently showing all child windows of an application even if they are in the hidden state.

PR #952, opened 2018-05-22
ENH: C++11 modernize NULL and 0 to nullptr.

phcerdan:
ITK5.0 uses c++11 nullptr instead of NULL and 0.
I was getting errors explained here: https://marcofoco.com/use-nullptr-if-you-can-0-if-you-need-but-please-dont-use-null/
```
error: call to 'a' is ambiguous
```

STEPs to use clang-tidy:
Generate compile_commands.json with CMake option: `-DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=ON`.
This generates a `compile_commands.json` in your build directory at configure time.

install clang and clang tools, that includes clang-tidy and run-clang-tidy.py.
(In Archlinux: sudo pacman -S clang, in other systems might be clang-extra-tools or similar)
Docs: http://clang.llvm.org/extra/clang-tidy/

And in the build folder of slicer, execute:
```
run-clang-tidy.py -header-filter='.*' -checks='-*,modernize-use-nullptr' -fix
```
This will iteratively check files for updates. It takes a while. The sources are only modified at the end of the process, so you could cancel anytime before with no consequences.

Issue in Mantis: https://issues.slicer.org/view.php?id=4560


jcfr:
Closing. This is superseded by https://github.com/Slicer/Slicer/pull/1102

PR #953, opened 2018-05-23
Fix scissors effect free-form glyph path

YiiGuxing:


lassoan:
The current behavior of the scissors effect looks good to me. Did you add curve points by simulating mouse events?

YiiGuxing:
No, I just move quickly when mouse  released. 
The last point is not connected to the previous point in code logic.

lassoan:
The contour is immediately deleted when I release the mouse button. While I'm moving the mouse, I always see a closed contour, no matter how fast I move and how slow the rendering is.

image

I would like to reproduce the error so that I can test that the solution is correct. I think in a line cell you can only have two points, so adding 3 points in the final case does not seem to be correct (if two line segments have to be added then there should be two InsertNextCell calls).

YiiGuxing:
Did you mean that a line cell can only be a minimum cell? It can contain only two points? Why? (I'm learning VTK)

I can't reproduce on Slicer, because it handles events too fast, I think that if the event is handled slowly enough, it can reproduce.

PR #954, opened 2018-05-23
BUG: Fix launcher settings generation with system qt.

phcerdan:
First, it sets Slicer_USE_SYSTEM_QT=ON in the case of Qt5.

And also fix appending QT_BINARY_DIR in the case of
the system qt5.
(this was generating problems with system python and virtualenvs)

Co-authored-by: Jean-Christophe Fillion-Robin <JChris.FillionR@kitware.com>

### DIFF
--- a/CMake/SlicerBlockCTKAppLauncherSettings.cmake
+++ b/CMake/SlicerBlockCTKAppLauncherSettings.cmake
@@ -103,8 +103,12 @@ endforeach()
 #-----------------------------------------------------------------------------
 set(SLICER_PATHS_BUILD
   <APPLAUNCHER_SETTINGS_DIR>/bin/<CMAKE_CFG_INTDIR>
-  ${QT_BINARY_DIR}
   )
+if(NOT Slicer_USE_SYSTEM_QT)
+  list(APPEND SLICER_PATHS_BUILD
+    ${QT_BINARY_DIR}
+    )
+endif()
 
 if(Slicer_BUILD_CLI_SUPPORT AND Slicer_BUILD_CLI)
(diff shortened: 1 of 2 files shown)


jcfr:
Thanks for the PR

Do you know if `Qt5.cmake` is always found in a sub-directory of `/usr/lib64` ? Should we add `/usr/lib` ? And would also be nice to find out where the file is installed when using homebrew of macport on macos.

phcerdan:
From [Archlinux](https://wiki.archlinux.org/index.php/Frequently_asked_questions) wiki (search lib64 there), `/usr/lib64` is a symlink to `/usr/lib`, and it follows convention in distros with systemd. I think we should add `/usr/lib` and maybe `/usr/lib32`?

From the [homebrew formula of qt](https://github.com/Homebrew/homebrew-core/blob/master/Formula/qt@5.5.rb) it seems like it is installing in `[prefix]/lib` where `[prefix]` I guess is the homebew prefix.

jcfr:
I squashed and tweaked the logic. Could you review ? 

> removed lib64 if it is symlink to lib

The fact it is a symlink doesn't really matter here. If the path includes "lib64", it should be added to the list.

phcerdan:
You are right, added back the `lib64`

jcfr:
Thanks for checking. I added the path with `lib64`. 

After you confirm it works, we will integrate.

jcfr:
After I checked it was working with Qt5, and after @phcerdan confirmed directly to me it was working for Qt5, I integrated the changes.

Closing. Fix integrated in [r27202](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27202)

PR #956, opened 2018-05-25
Added setWorldPosition utility method to  qSlicerReformatModuleWidget

mschumak:
Added a utility method to qSlicerReformatModuleWidget to make it easier to invoke setWorldPosition from Python.


jcfr:
Thanks for the contribution

To ensure the property is available from python in read/write, I suggest to do the following instead:

```
Q_PROPERTY(QVector3D worldPosition READ worldPosition WRITE setWorldPosition)

[...]

QVector3D worldPosition() const;
void setWorldPosition(const QVector3D& position);
```

mschumak:
Thanks, I'll try that.

mschumak:
I see that I have another way to accomplish what I'm trying to do, so for now, I'm going to close this request. I acknowledge that adding Python read/write capability and consistent naming would be good improvements, though.

PR #957, opened 2018-05-29
BUG: Add missing typename keywords.

phcerdan:
Also remove unused type.


lassoan:
Sorry about the missing typenames. Visual C++ does not need them and does not even log a warning if they are not there, so it is always easy to miss them. I've pushed a fix yesterday but it seems it arrived just after the nightly build was cut.

jcfr:
No problem at all.

This fix will also address the continuous build failure reported by CircleCI. See https://circleci.com/gh/Slicer/Slicer/2733?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

PR #963, opened 2018-05-29
BUG: Update VTK to fix volume clip bug and improve threading management 

jcfr:
This PR should **NOT** be integrated integrated. First, remaining issues discussed in [kitware/vtk#4338](https://gitlab.kitware.com/vtk/vtk/merge_requests/4338?commit_id=5c6bde06d788ea82e91b90bc1b95205a934f8eb7#note_416096) should be addressed.

Cc: @lassoan 

Highlighted list of changes
---------------------------

* https://gitlab.kitware.com/vtk/vtk/merge_requests/4338/commits (from Allison Vacanti <allison.vacanti@kitware.com>)
  Fix volume clip bug and add regression test.

    When clipping a volume using in-shader clipping planes, it was possible for the starting point of the ray cast to lie beyond the data volume. The raycast code is written such that the first sample is always taken before testing termination criteria, and in these cases we would always take a single sample outside of the volume, leading to artifacts.

    Fixed this behavior by checking that the starting position calculated by AdjustSampleRangeForClipping is indeed
    inside of the volume bounds and aborting the raycast if it is not.

    The existing TestGPURayCastClipping test would have caught this, except that the vase.vti volume used for testing
    has all 0's at the boundaries, so the rendering was correct even with the edge-clamp repetition outside of the volume
    (the faulty samples always computed RGBA=vec4(0)). I replaced the vase.vti of this test with a wavelet with finite
    boundary values that will catch this problem if there's a regression.

* https://gitlab.kitware.com/vtk/vtk/merge_requests/4175/commits (from David Gobbi)
  Resolve "Multithreader creates many unnecessary threads"

    Creating threads is expensive, so creating threads that do nothing should be avoided.

    - Limit number of threads in vtkImageHistogram
    - Remove redundant UpdateExtent calculation.
    - Use vector to allocate arrays instead of new
    - 17279: Limit num threads to num pieces.

Complete list of changes
------------------------

$ git shortlog 10e8cdc..9095631 --no-merges

Alessandro Genova (6):
      Add a test for the axis->setPrecision(0) issue on Windows
      Increase the font size of the axis label
      Fix the Precision = 0 bug on MSVC compilers
      Address compiler warning
      Cleanup comments and unused includes
      Resolve last compiler warning

Allison Vacanti (3):
      Retain renderer to recompute right rendered rectangle.
      Populate DataArraySelection objects for vtm files.
      Fix volume clip bug and add regression test.

Andreas Buykx (2):
      Fix issue #17211: crash on pick of composite data
      Backport fix for issue #17211 to release.

Andrew Bauer (7):
      Fixing issues with the HyperTreeGrid and improving documentation.
      Hypertreegrid cleanup.
      Update the HyperTreeGridSource to work in parallel.
      Fixing vtkImageData::FindPoint() bug
      Fixing typo in error message.
      Fixing bug with unstructured grid ghost cell generator.
      Fixing issue with writing structured XML field data

Aron Helser (5):
      Remove MomentInvariants filter, prep to make remote
      Remote Modules, add MomentInvariants, and ParallelMI
      Remote: change ParallelMoments to use branch
      Remote: parallel code included in single module
      MomentInvariants final remote repo location and hash

Berk Geveci (2):
      Introduced new composite datasets for partitioned data.
      Fixed bug in the XML structured data writer.

Bill Lorensen (1):
      Add NumberToString conversion class.

Brad King (10):
      ParallelMomentInvariants: Exclude from all pending fixes
      cmake: Set policy CMP0075 explicitly to avoid warnings
      CTestCustom: Drop unused Utilities/ matches
      pre-commit: Add support for ExternalData SHA512 objects
      git-gitlab-push: Add support for ExternalData SHA512 objects
      ExternalData: Switch from MD5 to SHA512 for new content links
      ExternalData: Add URLs to fetch from Girder on data.kitware.com
      ExternalData: Drop URLs using midas3.kitware.com
      ExternalData: Update URLs for vtk.org to use https
      ArchiveTestingDataOnGirder: Switch default to VTK collection

Carson Brownlee (2):
      fixing unnecessary volume updates in ospray on updates to tf.
      Fix material params with textures and add principled materials.

Cory Quammen (9):
      Use slightly clearer name for variables
      Stop drawing line segments outside image buffer bounds
      Use vtkMath::Min/Max instead of std::min/max
      Fix logic when tolerance is 0
      Clamp tolerance to range [0.0, VTK_DOUBLE_MAX]
      Set the field type for block selections
      Change algorithm for computing selections
      Rename vtkSelectionOperator to vtkSelector
      Change type of InsidednessArrayName member variable

Dan Lipsa (22):
      Fix SegY3D reader.
      Use SegYReader for both 2D and 3D data.
      Fix RequestInformation
      Use X,Y coordinates instead of inline,crossline.
      Use X, Y for SegY 3D.
      Test changed because of switch to XY coordinate system.
      Simplify computation.
      Use crossline, inline, sampleCountPerTrace order to match OpendTect.
      Clean up 2D reader and apply options to both readers.
      Add an option to create a StructuredGrid for a 3D SegY dataset.
      Update Progress.
      Flip data if increment is negative.
      Fix ParaView parameters and remove printouts.
      Make StructuredGrid the default for 3D data.
      Fix SegY warnings.
      Fix uninitialized variables.
      Fix warnings.
      Add SEGY to StandAlong group and warning fixes.
      Override and vector init warnings.
      Use a standard 4 byte unsigned int instead of long.
      Add cstdint include.
      Fix leaks on FileName.

David E. DeMarle (12):
      work around a bug where multiple threads delete image together
      silently ignore objects that lack requested valuepass array
      Increment version to VTK 8.1.1
      fix a memory leak and vtkstyle this for member
      Plug a vtk2osp leak and avoid a copy while exposing more formats.
      A change to the texture has to cause a redraw.
      Take advantage of zero copy textures in ospray.
      Fix a huge memory leak.
      dont leak the light either
      don't leak ospray buffer either
      use osp release not delete with osp volumes to prevent crashes
      suppress and fix comp warnings

David Gobbi (11):
      17265: Fix OpenGL2 vtkImageResliceMapper OOB issue
      17279: Limit num threads to num pieces.
      Use vector to allocate arrays instead of new
      Remove redundant UpdateExtent calculation.
      Limit number of threads in vtkImageHistogram
      17265: Fix OpenGL2 vtkImageResliceMapper OOB issue
      Fix thin border around vtkImageResliceMapper
      Backport 'reslicemapper-border' to release branch.
      Fix misuse of purportedly boolean Border.
      Fix documentation, use K for slice index.
      Fix typo in comments, xform to sform

Francois Budin (1):
      Check that context exists before trying to pop context.

Joachim Pouderoux (1):
      Fix destructor override warning

John Tourtellott (1):
      Added GDAL reprojection filter from vtkMap

Julien Schueller (1):
      Cast in vtkGetLibraryPathForSymbol macro

KWSys Upstream (2):
      KWSys 2018-04-23 (aa0f06ec)
      KWSys 2018-04-24 (85c211d6)

Ken Martin (26):
      clean up the state API and fix Qt widget
      do not use glDGetDouble on OpenGL ES
      add an option for a floor to the vtkSkybox class
      fix an access violation happening in StructureGridGeom filter
      fix uninitialized index in DataArray.cxx
      fix case where depth buffer was underflowing
      fix crash when scene has scalar values
      remove failihg part of test
      improve the picking on the pointgaussian mapper
      significantly improve performance for many blocks
      fix an access violation happening in StructureGridGeom filter
      add better support for using mesa when needed on windows
      add support for writing larger unstructured grids
      make sure we configure CTestCustom.cmake.in
      fix handling of large textures
      dataset mapper was not forwarding relative coincident parameters
      remove old 32bit simulator for ios
      add an executable to do opengl detection
      fix issues with the array calc and string arrays
      fixed case where the array name differed from
      Remove vtkProbeOpebGLSupport executable from Mobile
      comparison was being done for the wrong case
      fix ios build issues
      remove remaining opengl32 status methods
      fix buffer overrun
      cleanup and improve sphere and stick mappers in opengl

Kitware Robot (1):
      ExternalData: Convert content links from MD5 to SHA512

Lucas Gandel (5):
      Add WindowStereoTypeChangedEvent fired by vtkRenderWindow
      Fix memory exception with checkerboard stereo
      Invoke WindowResizeEvent on vtkWindow::SetSize() call
      Fix viewport size before FXAA pass on OSX
      Improve QVTKOpenGLWidget and add QVTKOpenGLWindow class based on QOpenGLWindow

Maik Froechtenicht (4):
      Fix for vtkTriangle::TrianglesIntersect() Issue #17092
      Undo copy vars for handover, format
      Fix for vtkTriangle::TrianglesIntersect() Issue #17092
      Fix for vtkTriangle::TrianglesIntersect() Issue #17092

Mathieu Westphal (2):
      Set vtkInteractionStylePython-TestStyleRubberBandZoomPerspective to run serial
      Adding new StepTime related features to LPT

Matt McCormick (1):
      ArchiveTestingDataOnGirder: Add script to upload to data.kitware.com

Michael Migliore (3):
      Add option in OpenFOAM reader to copy data to cell zones
      Add capacity to add a point outside of the current polyline
      Add tone mapping render pass

Nicolas Vuaille (3):
      Introduce Molecule Append filter
      Molecule to polydata conversion
      Improve pointSetToMolecule : optionally convert lines into bonds

Norman Juchler (2):
      Fix for QVTKRenderWindowInteractor for better support of retina displays
      Style improvement for my fix in QVTKRenderWindowInteractor

Robert Maynard (1):
      Update Accelerators/Vtkm to not produce any cpp check warnings

Sankhesh Jhaveri (12):
      Module Geovis GDAL
      Fix compile warnings for module GeovisGDAL
      Volume 2D transfer function issue causing undefined variable shader errors
      Fixed raster reprojection filter for celldata and GDAL 2
      Include GDAL header directory for module vtkGeovisGDAL
      Fix unused parameter and unused variable warnings in vtkGeovisGDAL
      Add API for GDAL reader to provide projection string in PROJ.4 format
      Ability to directly provide the PROJ.4 string to vtkGeoProjection
      Fix crash due to unitialized pointer
      Free up GDAL created memory in a GDAL safe way
      Fix unused variable warnings in vtkGeovisGDAL module
      Fix compiler warning in GDAL vector reader

Scott Wittenburg (1):
      Keep track of threads/workers and kill them explicitly

Seacas Upstream (1):
      exodusII 2018-05-07 (e4638682)

Sean McBride (1):
      Added a #warning on code that I’m pretty sure is dead

Sebastien Jourdain (4):
      Add ThreadedImageWriter
      Improve ThreadedWriter test
      Try to improve TestThreadedWriter
      Remove ThreadedWriter timing failure

Shawn Waldon (23):
      Add vtkFrustumSelector a selection operator for frustum selection
      vtkExtractSelection2: compiles but only partly implemented
      vtkExtractSelection2: selection with frustums compiles
      Move selection operator to Filters/Extraction
      Add test for combining extraction nodes via expression
      Refactor vtkSelectionOperator to use vtkDataObject API
      Change selection operator to return a boolean success value
      Handle selecting points with containing cells
      Add missing include
      Give the selection array the expected name
      Fix compile error
      Update test to set association for cell selection
      Fix typos
      Ensure output array has correct name
      Handle INVERSE property on selection nodes
      Handle legacy 1-component 2 tuple threshold selection arrays
      Change TestExtraction to test both old and new filters
      vtkExtractSelection2: add support for extracting rows from a vtkTable
      Add test of extracting rows
      Fix dashboard errors and warnings
      Deprecate and rename old vtkExtractSelection filter
      Add API documentation
      Unmark New() as legacy

Steven Hahn (2):
      Use public instead of protected member functions
      Use public instead of protected member functions

Sujin Philip (3):
      Update VTK-m
      Update to latest VTK-m master
      Fix some warnings in Accelerators/Vtkm

Tim Rowley (1):
      Add cell valued unstructured volume support to OSPRay mapper

Todd Martin (5):
      Add custom wrap compiler directive
      Fix definition name
      Add macro size hints
      Eliminate Windows name mangling workaround for GCCXML wrap
      Remove redundant code

Utkarsh Ayachit (39):
      Handle CFG files with other statements.
      put code blocks to simplify debugging
      Fix a shadowed var warning.
      extend grammar to include '-' in file names.
      updates to fix position file motions.
      Ensures output has as many rows as timesteps.
      Use `vtkCharArray` for mask array
      Add new doubleconversion 3rd party module.
      Update vtkNumberToString API to overcome build issues
      use correct package name when locating sustem doubleconversion.
      update thirdparty repo to point to the official fork.
      avoid segfault during app exit.
      Extend vtkSelection to support expressions.
      add missing return
      Add vtkValueSelector
      use vtkValueSelector in vtkExtractSelection2 (and its test)
      Add vtkLocationSelector to extract selected locations.
      Update TestExtractionExpression to test vtkLocationSelector.
      Update TestExtractionExpression to test threshold filter.
      fixes after API changes
      Misc vtkExtractSelection2 improvements
      simplify `vtkExtractSelection2::GetAttributeTypeOfSelection`
      exclude 3rd party warnings.
      vtkNumberToString doesn't need to delete assignment operator.
      add missing include.
      Don't use VTK_LEGACY with out set/get macros.
      vtkExtractSelection: remove unreachable code.
      fix missing `std::to_string` error
      fix warnings: remove testing of legacy filter
      fix missing `override` warnings
      warnings: fix `assignment operator could not be generated`
      docs: make `Return value: New reference.` explicit.
      fix warnings.
      exclude ospray warnings
      finally fix the implicit constructor/operator warnings.
      add MSVC 2013 exclusion.
      disable TestNumberToString on MSVC 2013.
      add tests to read/write multiblock with partial arrays
      Handle mismatch is chosen and available arrays.

Will Schroeder (6):
      Fast 2D clipper for image data
      Avoid reading into undefined heap; optimizing comparison on unsigned char
      Input is now generalized to vtkPointSet
      Removed unnecessary variables
      Faster computation of point inside/outside
      New regression test images

Yuri (1):
      Correction to when stat64 is used: FreeBSD doesn't use stat64.

double-conversion Upstream (2):
      doubleconversion 2018-03-29 (0e0b255d)
      doubleconversion 2018-05-11 (f3ac1585)


jcfr:
Closing. Latest version of VTK was integrated in [r27347](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27347)

See https://github.com/Slicer/VTK/tree/slicer-v9.0.0-2018-08-10-584314a73

PR #964, opened 2018-05-30
ENH: Improve slicer.util.exit documentation

lassoan:
Fixes https://issues.slicer.org/view.php?id=4470

### DIFF
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -12,7 +12,25 @@ def quit():
   exit(EXIT_SUCCESS)
 
 def exit(status=EXIT_SUCCESS):
+  """Exits the application with the specified exit code.
+  The method does not stops the process immediately but lets
+  pending events to be processed.
+  If exit() is called again while processing pending events,
+  the error code will be overwritten.
+
+  To make the application exit immediately, this code can be used.
+  Note that forcing the application to exit may result in
+  improperly released files and other resources.
+
+  .. code-block:: python
+
+    import sys
+    sys.exit(status)
+
+  """
+
   from slicer import app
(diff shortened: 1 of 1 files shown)


Base/Python/slicer/util.py
```diff
@@ -12,7 +12,25 @@ def quit():
...
+
+  .. code-block:: python
+
+    import sys
+    sys.exit(status)
+
+  """
+
   from slicer import app
+  # Prevent automatic exit overwrite this manually set exit code
```
jcfr: this manually set exit code -> this manually to set exit code

jcfr: That said, I am not sure of the meaning of "overwrite this to manually set exit code"

lassoan:
I mean overwriting the exit code that is set in this call

Base/Python/slicer/util.py
jcfr:
```
# Prevent automatic exit and ensure the "status" exit code passed to the following
# "slicer.exit()" call is considered
```

lassoan: how about this?

    Prevent automatic exit (for example, triggered by "--testing") overwrite the exit code that we set now

jcfr: `..  overwrite` -> `.. and overwrite ...`

jcfr:
Closing. Fix integrated in [r27220](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27220) and [r27224](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27224)

lassoan:
Thank you!

PR #967, opened 2018-06-07
ENH: VectorToScalar optionally extract single components.

phcerdan:
This commit improves the user interface adding support for choosing
the conversion mode.

Available conversion modes are now:
 * vector (RGB, RGBA) images to luminance (default mode keeping backward
        compatibility with original module implementation)
 * vector images to single component (user select which component to extract)

To support reusing of components by other scripted modules, a logic
and widget have also been created:
 * VectorToScalarVolumeConversionModeWidget
 * VectorToScalarVolumeLogic

### DIFF
--- a/Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
+++ b/Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
@@ -1,7 +1,49 @@
-import vtk, qt, ctk, slicer
+import vtk, qt, ctk
+import slicer
 import logging
 
 from slicer.ScriptedLoadableModule import *
+from slicer.util import NodeModify, VTKObservationMixin
+
+from contextlib import contextmanager
+@contextmanager
+def MyScopedQtPropertySetter(qobject, properties):
+  """ Context manager to set/reset properties"""
+  # TODO: Move it to slicer.utils and delete it here.
+  previousValues = {}
+  for propertyName, propertyValue in properties.items():
+    previousValues[propertyName] = getattr(qobject, propertyName)
+    setattr(qobject, propertyName, propertyValue)
+  yield
+  for propertyName in properties.keys():
+    setattr(qobject, propertyName, previousValues[propertyName])
+
+@contextmanager
+def MyObjectsBlockSignals(*qobjects):
+  """
+  Context manager to block/reset signals of any number of input qobjects.
+  Usage:
+  with MyObjectsBlockSignals(self.aComboBox, self.otherComboBox):
+  """
+  # TODO: Move it to slicer.utils and delete it here.
+  previousValues = list()
+  for qobject in qobjects:
+    # blockedSignal returns the previous value of signalsBlocked()
+    previousValues.append(qobject.blockSignals(True))
+  yield
+  for (qobject, previousValue) in zip(qobjects, previousValues):
+    qobject.blockSignals(previousValue)
+
+def getNode(nodeID):
+  if nodeID is None:
+    return None
+  return slicer.mrmlScene.GetNodeByID(nodeID)
+
+def getNodeID(node):
+  if node is None:
+    return ""
+  else:
+    return node.GetID()
 
 #
 # VectorToScalarVolume
@@ -15,15 +57,16 @@ class VectorToScalarVolume(ScriptedLoadableModule):
     self.parent.dependencies = []
     self.parent.contributors = ["Steve Pieper (Isomics)",
                                 "Pablo Hernandez-Cerdan (Kitware)",
-                                "Jean-Christophe Fillion-Robin (Kitware)",]
+                                "Jean-Christophe Fillion-Robin (Kitware)", ]
     self.parent.helpText = """
     <p>Make a scalar (1 component) volume from a vector volume.</p>
 
     <p>It provides multiple conversion modes:</p>
 
     <ul>
-    <li>convert RGB images to scalar using luminance as implemented in vtkImageLuminance (scalar = 0.30*R + 0.59*G + 0.11*B).</li>
     <li>extract single components from any vector image.</li>
+    <li>convert RGB images to scalar using luminance as implemented in vtkImageLuminance (scalar = 0.30*R + 0.59*G + 0.11*B).</li>
+    <li>computes the mean of all the components.</li>
     </ul>
     """
     self.parent.acknowledgementText = """
@@ -36,10 +79,23 @@ NIH Roadmap for Medical Research, Grant U54 EB005149."""
 # VectorToScalarVolumeWidget
 #
 
-class VectorToScalarVolumeWidget(ScriptedLoadableModuleWidget):
+class VectorToScalarVolumeWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
+  """
+  The user selected parameters are stored in a parameterNode.
+  """
+
+  def __init__(self, parent=None):
+    ScriptedLoadableModuleWidget.__init__(self, parent)
+    VTKObservationMixin.__init__(self)
+    self.logic = None
+    self._parameterNode = None
 
   def setup(self):
     ScriptedLoadableModuleWidget.setup(self)
+
+    self.logic = VectorToScalarVolumeLogic()
+    # This will use createParameterNode with the provided default options
+    self.setParameterNode(self.logic.getParameterNode())
 
     # Collapsible button
     self.selectionCollapsibleButton = ctk.ctkCollapsibleButton()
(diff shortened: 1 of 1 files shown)


Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
```diff
@@ -76,6 +103,26 @@ def setup(self):
     # Add vertical spacer
     self.layout.addStretch(1)
 
+  def onRGBCheckBox(self, state):
+    if (state == 2): # Checked
+      # Disable other box.
+      self.singleComponentBox.setChecked(False)
```
cpinter: I don't know exactly how the UI looks like, but if you have to do this, then probably radio button would be a better widget to use.

phcerdan: Definitely right, thanks @cpinter

pieper:
LGTM - thanks! 👍

Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
```diff
@@ -76,6 +103,26 @@ def setup(self):
...
+      self.componentsComboBox.setEnabled(False)
+
+  def onSingleComponentCheckBox(self, state):
+    if (state == 2): # Checked
```
jcfr: Parenthesis are now needed in python.

jcfr: For future reference, Qt enum value can be accessed using `qt.Qt.Checked`

phcerdan: `qt.Qt.Checked` good to know.

Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
```diff
@@ -76,6 +103,26 @@ def setup(self):
     # Add vertical spacer
     self.layout.addStretch(1)
 
+  def onRGBCheckBox(self, state):
```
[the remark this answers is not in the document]
phcerdan: I removed one slot, and just work with a RadioButton toggle, checking the state of the buttons themselves, not the boolean signal.

phcerdan:
@cpinter @jcfr I have added QRadialButtons, see my replies to @jcfr for extra details. Thanks for the reviews!

cpinter:
Excellent, thanks for the contribution!

Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
phcerdan: I couldn't get the ratio buttons to be exclusive just with the parent (maybe because there are frames?). Anyway, I created an exclusive QButtonGroup, and the buttons there. It is a few lines more verbose but explicit. Let me know what you think.

phcerdan:
The GUI looks like:

peekslicervectortoscalar

Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
cpinter: Common parent and setAutoExclusive(true) usually does the trick. Maybe you forgot the latter? Jc says it's not needed, but maybe it depends on the order you do things. What happens if you set it to true explicitly? That said I wouldn't mind the button group if that's how you can make it work.

lassoan:
If you only select between integers then use a spinbox instead of a combobox. But I think it would be even better to make Luminance option the first option in the combobox. Then we could get rid of all the radiobuttons and have a simple label (Component) and a combobox (Luminance, 0, 1, 2,...).

Note that luminance is not the same as mean(r, g, b). I think we use the latter. If you really need luminance then it could be an additional entry in the combobox, which is only displayed for 3 or maybe 4 component images (assuming rgb and rgba). 

Number of components should be updated whenever a new volume is selected. We could also add an observer to the image change event of the node so that we can update the widget when some module changes the volume in the background or the user changes things in the Python interactor.

phcerdan:
@lassoan thanks a lot for the input.

 > Number of components should be updated whenever a new volume is selected.
That is/was already implemented.

> We could also add an observer to the image change event of the node so that we can update the widget when some module changes the volume in the background or the user changes things in the Python interactor.
Sounds good, maybe in another PR?

> Note that luminance is not the same as mean(r, g, b). I think we use the latter.
This module was using the vtkLuminanceImage, so it was computing luminance, not mean. But adding mean or any other function that reduces dimensionality to scalar sounds good.

> If you only select between integers then use a spinbox instead of a combobox.
Nice, I'll do.

> But I think it would be even better to make Luminance option the first option in the combobox. Then we could get rid of all the radiobuttons and have a simple label (Component) and a combobox (Luminance, 0, 1, 2,...).
Yeah, I liked the idea, and @jcfr also mentioned it.
But my reasoning was to keep the extract single components separated from any other operator. Mainly, because it doesn't require any extra checking or logic of the input vector volume, it always work for any dimension of the vector. And it might be a common for some applications. On the contrary, Luminance requires at least 3 components, and other functions might impose other conditions.
And also from a coding point of view, it was convenient to link the current index of the combo/spin directly to the component. Adding entries will require tracking and extra mapping.
I am not against this change, but separating the extraction of a single component from other operations working on all the dimensions seemed sensible. Let me know what you think.

phcerdan:
Current state:

peekslicervectortoscalarupdatejc

lassoan:
The GUI looks nice now, so you can leave it like this.

If you consider using a single combobox: Using combox is easy, as you can set custom data for each item, so you don't need to maintain an external mapping.

https://github.com/Slicer/Slicer/blob/f051e24c803ae0a067e04ea2b2a9a178040d7415/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorSmoothingEffect.py#L39-L45

https://github.com/Slicer/Slicer/blob/f051e24c803ae0a067e04ea2b2a9a178040d7415/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorSmoothingEffect.py#L102-L103

Observing image data should be quite simple, too (about 6-8 lines of code) and all core modules have this kind of always-up-to-date GUI (you don't need to click around the GUI to get it updated). If you don't have the time to add this small amount of logic then use a spinbox with very high max value instead of enumerating specific values.

Unfortunately, this scripted module has other fundamental issues (it does not use parameter node, so user selections are not saved into the scene; there is no logic class - everything is just dumped into the widget class), which should be addressed at some point in the future.

Thank you for working on this and sorry for being picky, but we try to keep Slicer core modules quality as high as possible, as they may be used in many other modules and extensions; and they also often serve as example for custom modules.

jcfr:
> we try to keep Slicer core modules quality as high as possible, as they may be used in many other modules and extensions; and they also often serve as example for custom modules.

@lassoan Thanks for the careful review :+1: 

> (1) Observing image data should be quite simple
> (2) use parameter node

@phcerdan I suggest you add support for these two features. Let me know if you need help.

>  there is no logic class - everything is just dumped into the widget class), which should be addressed at some point in the future.

Good news is that this PR now add a logic class

phcerdan:
@jcfr @lassoan thanks for the inputs and review. I am working on (1) and (2) and learning slicer guts at the same time. I'll update this soon.

phcerdan:
New commit because there was quite a few changes, and didn't want to lose the original changes. I have added:
(1) Observing image data should be quite simple
(2) use parameter node

And the new method to compute the average of all components.

phcerdan:
Current state

peekslicervectortoscalarupdateaverage

phcerdan:
I have an issue with the automatic visualization of the generated output.
The new average conversion method returns a ScalarVolume with ScalarType Double, independently of the ScalarType of the components of the input, so, when I re-use that ScalarVolume for any of the other computation (that have the same ScalarType than the input, in this case `unsigned char`), the visualization range is not updated for the new type, and the following happens:

peekslicervectortoscalarissueview


Is there any way to force re-compute the map between pixel-value range <-> visualization at the application level, or any method of the VolumeNode? Thanks!

lassoan:
You can add a vtkImageCast filter to convert double pixel type to the same pixel type as the input vector volume.

If automatic window width/leve (brightness/contrast) settings are not good then you can copy window width/level values from the display node of the input vector volume.

phcerdan:
Thanks @lassoan, I casted to the same Scalar Type than input for consistency with the other methods.

phcerdan:
@cpinter @lassoan let me know if I can do anything else to push this forward. Thanks!

Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
```diff
@@ -26,82 +63,420 @@ def __init__(self, parent):
...
+                                        self.setConversionMethod(self.methodSelectorComboBox.itemData(currentIndex))
+                                        )
+
+    self.componentsComboBox.connect('currentIndexChanged(int)', self.setComponentToExtract)
+
+  def inputVolumeNode(self):
+    return self._inputVolumeNode
+
+  def setInputVolumeNode(self, node):
+    if isinstance(node, basestring):
```
lassoan: This kind of ambiguity - accepting a node or some string (node id? node name? what if there are multiple nodes with the same name?) - is acceptable in testing and development tools for convenience, but not in production code.

For example, slicer.util.getNode has this behavior exactly because it is often used in the Python console to interactively get a node. However, in modules you would use slicer.mrmlScene.GetNode()or slicer.mrmlScene.GetFirstNodeByName().

Since the widget is not supposed to be used from any other modules (you only interact through MRML nodes and sometimes by using another module's logic), I would remove this method completely.

Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
```diff
@@ -26,82 +63,420 @@ def __init__(self, parent):
...
+      node = getNode(node)
+    self._inputVolumeNode = node
+
+  def componentToExtract(self):
```
lassoan: no need for get/set methods - widget is not supposed to be used from anywhere else

Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
```diff
@@ -26,82 +63,420 @@ def __init__(self, parent):
...
+    self.conversionMethodWidget.componentsComboBox.connect('currentIndexChanged(int)', self.updateParameterNodeFromGui)
+
+    self.applyButton.connect('clicked(bool)', self.onApply)
+
+    self.setDefaultParameters()
+
+  def cleanup(self):
+    self.removeObservers()
+
+  def setDefaultParameters(self):
```
lassoan: It should be possible to create a parameter node (with default values) by only using the logic class, so default value initialization should be done in the logic class. See createParameterNode method.

Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
```diff
@@ -26,82 +63,420 @@ def __init__(self, parent):
...
+
+  def updateGuiFromMRML(self, caller=None, event=None):
+    """
+    Query all the parameters in the parameterNode,
+    and update the GUI state accordingly if something has changed.
+    """
+    self.applyButton.enabled = all([self.inputVolumeNode(), self.outputVolumeNode(), self.parameterNode()])
+    if self.parameterNode is None:
+      return
+    self.setInputVolumeNode(self._parameterNode.GetParameter("InputVectorVolume"))
```
lassoan: need to call blocksignals for each widget when modifying from parameter node to make sure updateMRMLFromGUI method is not called in the middle of updateGUIFromMRML.

Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
```diff
@@ -26,82 +63,420 @@ def __init__(self, parent):
...
+  CONVERSION_METHODS = (LUMINANCE, AVERAGE, SINGLE_COMPONENT)
+
+
+  def __init__(self, parent = None):
+    ScriptedLoadableModuleLogic.__init__(self, parent)
+
+  @staticmethod
+  def isValidInputOutputData(inputVolumeNode, outputVolumeNode, conversionMethod, componentToExtract):
+    """
+    Validate parameters using the parameterNode.
```
lassoan: document what is the return value, especially when it is not trivial (not just a simple Boolean)

Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
```diff
@@ -26,82 +63,420 @@ def __init__(self, parent):
...
+      if nrbOfComponents < 3:
+        msg = 'input has only %d components but requires ' \
+              'at least 3 components for luminance conversion.' % nrbOfComponents
+        logging.debug("isValidInputOutputData failed: %s" % msg)
+        return False, msg
+
+    return True, None
+
+
+  def run(self, inputVolumeNode, outputVolumeNode, conversionMethod, componentToExtract):
```
lassoan: If we use a parameter node then the primary method for running the processing should take a parameter node as input.

In addition to that method, you may add a convenience method, which takes, inputVolume, outputVolume, conversionMethod, etc., creates a parameter node and calls the primary method with that.

Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
```diff
@@ -26,82 +63,420 @@ def __init__(self, parent):
...
-    selectionNode.SetReferenceActiveVolumeID(outputVolume.GetID())
-    slicer.app.applicationLogic().PropagateVolumeSelection(0)
+    inputVolumeNode.GetIJKToRASMatrix(ijkToRAS)
+    outputVolumeNode.SetIJKToRASMatrix(ijkToRAS)
+
+    nrbOfComponents = inputVolumeNode.GetImageData().GetNumberOfScalarComponents()
+    weightedSum = vtk.vtkImageWeightedSum()
+    weights = vtk.vtkDoubleArray()
+    weights.SetNumberOfValues(nrbOfComponents)
+    evenWeight = 1.0/nrbOfComponents
```
lassoan: numberOfComponents

Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
jcfr: The `getNode` function only lookup node ID (defined at the top of the file). There are no risk of finding a node by name.

The issue is that the name of the function is the same as `slicer.util.getNode`

lassoan: OK then. Still, I find all these set/get functions unnecessary, as the updateMRML… and updateGUI… methods take care of keeping the widget and MRML node in sync.

Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
```diff
@@ -26,82 +63,420 @@ def __init__(self, parent):
...
+      self._parameterNode.SetParameter("ComponentToExtract", str(self.conversionMethodWidget.componentToExtract()))
+
   def onApply(self):
-    inputVolume = self.inputSelector.currentNode()
-    outputVolume = self.outputSelector.currentNode()
-    # check for input data
-    if not (inputVolume and outputVolume):
-      slicer.util.errorDisplay('Input and output volumes are required for conversion', windowTitle='Luminance')
+
+    with ScopedQtPropertySetter(self.applyButton, {"enabled": False, "text": "Working..."}):
```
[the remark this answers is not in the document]
jcfr: > move it to slicer.util so that it can be used elsewhere.

That was indeed the plan. I suggest we do not use this call in this PR and introduced it only later.

Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
jcfr: Since the widget is setting the value of the parameter node to what it is already, no node modified event should be invoked.

As a precautionary measure, I suggest we introduce a context managed named like `BlockedObjectSignal`

Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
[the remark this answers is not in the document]
phcerdan: I just renamed it (it is too handy now to not use it!). If you want to move it to slicer.util afterwards with the good name, I can remove the local function here afterwards.

Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
```diff
@@ -26,82 +63,420 @@ def __init__(self, parent):
...
+class VectorToScalarVolumeConversionMethodWidget(qt.QWidget):
+  """
+  Widget to interact with conversion parameters.
+  """
+
+  EXTRACT_COMPONENT_NONE = -1
+
+  def __init__(self, parent=None):
+    qt.QWidget.__init__(self, parent)
+    self._inputVolumeNode = None
```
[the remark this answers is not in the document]
phcerdan: Done

Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
phcerdan: Removed variables in widget, the values are only stored in the ComboBoxes. And the 'upper' widget makes the connections to keep updated the parameterNode.

Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
phcerdan: The time invested here was motivated to integrate this functionality in a external module. So I do need to reuse the GUI. Kept the get method for convenience, but removed the set methods.

Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
phcerdan: Done, moved to logic overriding the createParameterNode.

Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
phcerdan: I haven't block anything, but created a context manager to block signals in a list of qobjects.

Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
phcerdan: Done

Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
phcerdan: Done

Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
phcerdan: Done

phcerdan:
Thanks a lot @lassoan for the careful review, really appreciated, let me know what you think about my latest change. I think I addressed all your suggestions, except the blocking signals. I have tested it, and there is no recursive calling in either case, changing from MRML or from GUI.

phcerdan:
As a node, I really do need to reuse the GUI widget to justify the time spent on this. I have removed the internal variables, but I am sure it is still usable from external modules with a little bit of plumbing and connections.

lassoan:
It all looks good! Thank you very much for taking the time to implement this so nicely. 

One last thing: I did not understand what XXX: meant. If it is a todo then rename to TODO: (if something else then rename to something that is more clear or more standard).

jcfr:
> XXX: meant

I guess I influenced this. This is an other convention. See https://en.wikipedia.org/wiki/Comment_(computer_programming)#Tags

phcerdan:
Changed XXX to TODO.

lassoan:
There are hundreds of TODO tags in VTK and ITK, and maybe 4-5 XXX. I think it is better to use TODO if that code must be changed, or use plain comments if you just want to tell something to future developers.

lassoan:
Thank you, I have no more comments!

phcerdan:
Thanks @lassoan and @jcfr for your time. I learnt quite a bit!

@cpinter the changes you requested were solved (I implemented only one combo-box), but github doesn't know that yet. Let me know if you have any other suggestion.

cpinter:
Thanks for all the work, LGTM!

phcerdan:
I have successfully integrated the conversion method widget in an external module! It wasn't that much plumbing, and looks good!

peekslicervectortoscalarexternal

phcerdan:
@lassoan @cpinter may I help in something for this to be finally merged?
Thanks!

jcfr:
@phcerdan  In the gif above, it looks like the input combobox is updated ? Is this expected ?

phcerdan:
In the extension, yes.

jcfr:
Gotcha. Makes sense.

Will integrate later today

lassoan:
Thank you! Integrated in 27278.

PR #968, opened 2018-06-08
BUG: Load python dependent widget during module instantiation

vovythevov:
This let python module access widget/logic or any C++ classes they may
have as soon as they are instantiated. This let users do the following:
"var = slicer.qMyClassWidget.MyEnumValue" for example.

### DIFF
--- a/Base/QTGUI/qSlicerScriptedLoadableModule.cxx
+++ b/Base/QTGUI/qSlicerScriptedLoadableModule.cxx
@@ -192,28 +192,6 @@ bool qSlicerScriptedLoadableModule::setPythonSource(const QString& newPythonSour
 void qSlicerScriptedLoadableModule::setup()
 {
   Q_D(qSlicerScriptedLoadableModule);
-
-  qSlicerCoreApplication * app = qSlicerCoreApplication::application();
-  if (app)
-    {
-    // Set to /path/to/lib/Slicer-X.Y/qt-scripted-modules
-    QString modulePath = QFileInfo(this->path()).absolutePath();
-    // Set to /path/to/lib/Slicer-X.Y
-    modulePath = QFileInfo(modulePath).absolutePath();
-    // Set to /path/to/lib/Slicer-X.Y/qt-loadable-modules
-    modulePath = modulePath + "/" Slicer_QTLOADABLEMODULES_SUBDIR;
-
-    bool isEmbedded = app->isEmbeddedModule(this->path());
-    if (!isEmbedded)
-      {
-      if (!qSlicerLoadableModule::importModulePythonExtensions(
-            app->corePythonManager(), app->intDir(), modulePath, isEmbedded))
-        {
-        qWarning() << "qSlicerLoadableModule::setup - Failed to import module" << this->name() << "python extensions";
-        }
-      }
-    }
-
   this->registerFileDialog();
   this->registerIO();
   d->PythonCppAPI.callMethod(Pimpl::SetupMethod);
--- a/Base/QTGUI/qSlicerScriptedLoadableModuleFactory.cxx
+++ b/Base/QTGUI/qSlicerScriptedLoadableModuleFactory.cxx
@@ -77,6 +77,18 @@ qSlicerAbstractCoreModule* ctkFactoryScriptedItem::instanciator()
   module->setInstalled(qSlicerUtils::isPluginInstalled(this->path(), app->slicerHome()));
   module->setBuiltIn(qSlicerUtils::isPluginBuiltIn(this->path(), app->slicerHome()));
 
+  // Set to /path/to/lib/Slicer-X.Y/qt-scripted-modules
+  QString modulePath = QFileInfo(this->path()).absolutePath();
+  // Set to /path/to/lib/Slicer-X.Y
+  modulePath = QFileInfo(modulePath).absolutePath();
+  // Set to /path/to/lib/Slicer-X.Y/qt-loadable-modules/Release|(...)|Debug
+  modulePath = modulePath + "/" Slicer_QTLOADABLEMODULES_SUBDIR + "/" + app->intDir();
+  if (!qSlicerLoadableModule::importModulePythonExtensions(
+        app->corePythonManager(), app->intDir(), modulePath, app->isEmbeddedModule(this->path())))
+    {
+    return 0;
+    }
+
   bool ret = module->setPythonSource(this->path());
   if (!ret)
     {


lassoan:
We have been accessing constants and methods in scripted module by simply importing them. For example:

    import SampleData
    sampleDataLogic = SampleData.SampleDataLogic()
    print("Getting MR Head Volume")
    mrHeadVolume = sampleDataLogic.downloadMRHead()

I'm afraid this change may break many things, because during module instantiation, other modules that our module depends on are not guaranteed to be instantiated yet. By the time setup() is called, it is guaranteed that all required modules are already set up (because in setup() we may need to use those other modules).

lassoan:
I don't fully understand the use case (what classes you need to use, from where, and when), so maybe I misunderstood what your did. Some more information on this would be useful.

vovythevov:
@lassoan:

In my case, I am accessing a C++ widget that I have compiled for my scripted module. I am accessing this widget from outside of the python class, like so:

```c++
class qMyItemDelegate: public QStyledItemDelegate
{
  Q_OBJECT
public:
  qMyItemDelegate(QObject *parent = 0);
  virtual ~qMyItemDelegate();
  enum ItemDataRole
    {
    /// List of strings.
    StringListRole = Qt::UserRole + 1,
    DecimalValueRole,
    IntegerValueRole,
    LastItemDataRole
    };
  Q_ENUMS(ItemDataRole)
(...)
```

```python
import slicer

MyItemDataRole = slicer.qMyItemDelegate.LastItemDataRole + 1

class MyModule():
  (...)
```

For most cases, you access the method/widget from within your python class, once everything has been setup.
During the module instantiation, it's ignored as python doesn't run that code.
However, with my above test case, the widget is *not* available during the instantiation which causes the module to not be instantiated (and not loaded). 
Note that after the modules have been loaded however, the *setup()* method of is run and then you can access your widget.

Does that make sense ?

jcfr:
Thanks @vovythevov :+1: 

And thanks for the review. @lassoan @ihnorton 

This change is the same as [r26315](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26315), the module are only imported earlier in the process.

I anticipate it will  not have any impact on startup.

To measure the impact, the following can be done before and after the change:

```
python /path/toApplications/SlicerApp/Testing/Python/MeasureStartupTimes.py --normal -n 5  /path/to/Slicer-build/Slicer
```

lassoan:
Thanks for the additional information. Unfortunately, I still don't understand why would you want to instantiate a class in the global scope in a scripted module file. So far the rule has been to never do any operations in the global scope (outside the module, widget, or logic classes), because that would happen before all the required modules for your module are loaded. That's why currently all initializations that must be done at application startup, are performed in a callback triggered by startupCompleted signal.

The only limitation of this solution is that startupCompleted signal is sent to modules in random order. Therefore, if Python modules depend on each other then it is possible that a Python module will get startupCompleted signal before all its required prerequisite modules get startupCompleted.

Ideally, scripted module logic objects would be instantiated by Slicer (the same way as Slicer instantiates module logic classes for C++ loadable modules) - respecting module dependencies. However, as I understand, this would make module reloading more complicated.

I think the correct solution for allowing scripted modules to perform initialization steps during application startup would be to let Slicer core to create the scripted module's logic class. You could perform all initializations in that logic class' constructor.

jcfr:
> I still don't understand why would you want to instantiate a class in the global scope

To clarify, no class is being instantiated in the global scope.

We are only talking about a reference to static member living within a C++ wrapped class.

See example below where the reference `slicer.qFooItemDelegate.LastRole` is expected to be valid at the time the source interpretation.

And since in that case `qFooItemDelegate` is provided by a C++ wrapped class, we just need to make sure the logic/mrml/pythonqt c++ wrapped class are loaded.

```python
import qt
import slicer
import vtk
from slicer.util import NodeModify
from slicer.util import VTKObservationMixin

[...]

BarIDRole = slicer.qFooItemDelegate.LastRole + 1

class BarTableWidget(qt.QTableWidget, VTKObservationMixin):
    # ---------------------------------------------------------------------------
    def __init__(self, *args, **kwargs):
        """
        Specialized QTableWidget displaying ... 
        [...]
        """
[...]
```

Agreed, the entire module (or widget) could be ported to C++ .. but this is a priority currently.


> Ideally, scripted module logic objects would be instantiated by Slicer

Definitively agree, but in that particular case some helper widgets are built within the scripted module using `SlicerMacroBuildModuleWidgets`.

An other approach would be to create a loadable module with a hidden UI that would take care of loading these.

lassoan:
> To clarify, no class is being instantiated in the global scope.

Sorry, I was not accurate. I meant that we've been always telling people not to do any real work in the file scope, because at that point Slicer startup has not been completed yet and none of the modules are guaranteed to be available.

As I see, the change would not cause any issue directly, but it would allow/encourage people to do some "work" in the file scope when the file is loaded. If modules started to do this, it would make Slicer core development really difficult, because any change that we make in how Slicer starts up could break some modules. Even seemingly simple things, such as instantiating a GUI widget may not work fully correctly before application startup has been completed.

vovythevov:
@lassoan:
Thanks for your input ! I see your point and it makes sense to me. I will close this PR and I will only access my custom Qt widget from within the class itself.
I updated the Python FAQ to document the behavior here: https://www.slicer.org/wiki/Documentation/Nightly/Developers/FAQ/Python_Scripting#Why_can.27t_I_access_my_C.2B.2B_Qt_class_from_python.
Feel free to edit if you want !

Thanks everyone for their feedback, that was really helpful !

jcfr:
Out of curiosity, does the following work ?

```python
import qt
import slicer
from slicer.util import VTKObservationMixin

[...]

class BarTableWidget(qt.QTableWidget, VTKObservationMixin):

    BarIDRole = slicer.qFooItemDelegate.LastRole + 1

    def __init__(self, *args, **kwargs):
        """
        Specialized QTableWidget displaying ... 
        [...]
        """
[...]
```

vovythevov:
@jcfr: No that doesn't work.
Only
```python
import qt
import slicer
from slicer.util import VTKObservationMixin

[...]

class BarTableWidget(qt.QTableWidget, VTKObservationMixin):
    # ---------------------------------------------------------------------------
    def __init__(self, *args, **kwargs):
        """
        Specialized QTableWidget displaying ... 
        [...]
        """
        self.BarIDRole = slicer.qFooItemDelegate.LastRole + 1
[...]
```

jcfr:
Thanks for clarifying.

I tweaked the wiki further mentioning `scope of instantiated python class` instead of `scope of python class`.

See https://www.slicer.org/w/index.php?title=Documentation%2FNightly%2FDevelopers%2FFAQ%2FPython_Scripting&type=revision&diff=59103&oldid=59101

lassoan:
Thanks to you both, we've learned things and updated docs, so the discussion was useful.

PR #973, opened 2018-06-18
ENH: Add Set/Get and XML macros for iterable STL container properties

Sunderlandkyl:
Adds macros to Read/Write/Print/Copy iterable properties.

Example usage:

- Numeric container:
`vtkMRMLWriteXMLStdVectorMacro(property, Property, std::deque<int>);`

- String container:
`vtkMRMLWriteXMLStdStringVectorMacro(property, Property, std::vector);`

See also this related merge request in VTK: https://gitlab.kitware.com/vtk/vtk/merge_requests/4427
It's not required for this change, but it adds some convenience macros for setting/getting STL containers.

Also added generic Set/Get macros that work with STL containers.
Currently, STL containers cannot use the generic vtkSetMacro/vtkGetMacro types since those macros require a "<<" operator within vtkDebugMacro

https://github.com/openigtlink/SlicerOpenIGTLink/pull/36

cc: @lassoan

### DIFF
--- a/Libs/MRML/Core/Testing/vtkMRMLNodeTest1.cxx
+++ b/Libs/MRML/Core/Testing/vtkMRMLNodeTest1.cxx
@@ -23,6 +23,14 @@
 #include <vtkNew.h>
 #include <vtkObjectFactory.h>
 #include <vtkStringArray.h>
+#include <vtkXMLDataElement.h>
+#include <vtkXMLUtilities.h>
+
+// VTKSYS includes
+#include <vtksys/SystemTools.hxx>
+
+// vtkAddon includes
+#include "vtkAddonSetGet.h"
 
 // STD includes
 #include <sstream>
@@ -70,12 +78,24 @@ public:
   void SetOtherNodeID(const char* id);
   vtkGetStringMacro(OtherNodeID);
 
+  vtkSetStdVectorMacro(TestingStringVector, std::vector<std::string>);
+  vtkGetStdVectorMacro(TestingStringVector, std::vector<std::string>);
+
+  vtkSetStdVectorMacro(TestingIntVector, std::vector<int>);
+  vtkGetStdVectorMacro(TestingIntVector, std::vector<int>);
+
+  vtkSetStdVectorMacro(TestingFloatVector, std::vector<float>);
+  vtkGetStdVectorMacro(TestingFloatVector, std::vector<float>);
+
   virtual void SetSceneReferences() VTK_OVERRIDE;
   virtual void UpdateReferenceID(const char *oldID, const char *newID) VTK_OVERRIDE;
   virtual void WriteXML(ostream& of, int nIndent) VTK_OVERRIDE;
   virtual void ReadXMLAttributes(const char** atts) VTK_OVERRIDE;
 
-  char *OtherNodeID;
+  char*                     OtherNodeID;
+  std::vector<std::string>  TestingStringVector;
+  std::vector<int>          TestingIntVector;
+  std::vector<float>        TestingFloatVector;
 
   vtkObject* LastMRMLEventCaller;
   unsigned long LastMRMLEventId;
@@ -194,6 +214,7 @@ bool TestNodeReferenceSerialization();
 bool TestClearScene();
 bool TestImportSceneReferenceValidDuringImport();
 int TestSaveLoadSpecialCharacters();
+int TestReadWriteXMLProperties();
 
 //---------------------------------------------------------------------------
 int vtkMRMLNodeTest1(int , char * [] )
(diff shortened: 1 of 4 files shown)


lassoan:
Thanks a lot Kyle, it looks good! I would just ask for one change.

We may need different implementation in the future for managing float/double and integer vectors. Therefore, it would be more future-proof if we created separate macros for float and integer (e.g., set large number of decimal digits for floats for loss-less writing to file; handle special values, such as nan, inf, ...). vtkMRMLWriteXMLStdFloatVectorMacro and vtkMRMLWriteXMLStdIntVectorMacro name should work (and of course not just the writer macros but others would need to be renamed as well). If you don't need integer vector macros then for you can just rename vtkMRML...StdVectorMacro to vtkMRML...StdFloatVectorMacro.

Sunderlandkyl:
Makes sense. I'll add the int macro as well. I believe that there are some integer parameters.

Sunderlandkyl:
Any other changes to make here?

jcfr:
Unless you plane to use the macro  in the core in a follow up commit, could you test the new macros by either adding a new test or extending [this one](https://github.com/Slicer/Slicer/blob/master/Libs/MRML/Core/Testing/vtkMRMLNodeTest1.cxx) ?

Sunderlandkyl:
Sure, I'll do that.

Sunderlandkyl:
Macros have been added to vtkMRMLNodeTest1. Vector XML methods tested in TestReadWriteXMLProperties().

Libs/MRML/Core/Testing/vtkMRMLNodeTest1.cxx
```diff
@@ -2854,6 +2863,69 @@ bool TestClearScene()
...
+  testingFloatVector.push_back(3);
+  testingFloatVector.push_back(4);
+  testingFloatVector.push_back(5);
+  node1->SetTestingIntVector(testingIntVector);
+
+  std::stringstream ss;
+  ss << "<MRMLNode ";
+  node1->WriteXML(ss, NULL);
+  ss << " />";
+  std::string testString = ss.str();
```
jcfr: `testString` is unused

Libs/MRML/Core/Testing/vtkMRMLNodeTest1.cxx
```diff
@@ -2854,6 +2863,69 @@ bool TestClearScene()
...
+  vtkSmartPointer<vtkXMLDataElement> element = vtkSmartPointer<vtkXMLDataElement>::Take(vtkXMLUtilities::ReadElementFromStream(ss));
+
+  const char *atts[] = {
+    "TestingStringVector", element->GetAttribute("TestingStringVector"),
+    "TestingFloatVector", element->GetAttribute("TestingFloatVector"),
+    "TestingIntVector", element->GetAttribute("TestingIntVector"),
+    NULL };
+
+  vtkNew<vtkMRMLNodeTestHelper1> node2;
+  node2->ReadXMLAttributes(atts);
```
jcfr: It seems `atts` is not used anywhere

Libs/MRML/Core/Testing/vtkMRMLNodeTest1.cxx
```diff
@@ -2854,6 +2863,69 @@ bool TestClearScene()
...
+
+  if (node1->GetTestingStringVector() != node2->GetTestingStringVector())
+  {
```
jcfr: Fix indent.

jcfr (changes requested):
Thanks for adding the test :+1: 

After addressing few nitpicks, we should be all set.

Libs/MRML/Core/Testing/vtkMRMLNodeTest1.cxx
Sunderlandkyl: atts is used in:
```node2->ReadXMLAttributes(atts);```
To populate the parameters of node2.

Libs/MRML/Core/Testing/vtkMRMLNodeTest1.cxx
Sunderlandkyl: Done

Libs/MRML/Core/Testing/vtkMRMLNodeTest1.cxx
Sunderlandkyl: Done

jcfr (approved):
:+1: Will integrate after local build completes

jcfr:
Closing. Integrated in [r27266](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27266)

PR #977, opened 2018-06-22
BUG: Generate unique name in AddNewNodeByClass

phcerdan:
Even though the documentation mentioned it, the implementation
wasn't generating unique names. Fix it.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLScene.cxx
+++ b/Libs/MRML/Core/vtkMRMLScene.cxx
@@ -1279,7 +1279,7 @@ vtkMRMLNode* vtkMRMLScene::AddNewNodeByClass(
     }
   if (!nodeBaseName.empty())
     {
-    nodeToAdd->SetName(nodeBaseName.c_str());
+    nodeToAdd->SetName(this->GenerateUniqueName(nodeBaseName).c_str());
     }
   return this->AddNode(nodeToAdd);
 }


lassoan:
We intentionally not override names that developers explicitly request. Automatic node renames only happen at the GUI or application level (for user's convenience) and never at MRML library level. We assume developers know what they want to do. It would be extremely frustrating for a developer if he would need to double-check that the node actually got the name that he asked for and if not then he would need to call SetName again. If a developer needs to add a new node with unique name then he should simply call:

`myNewNode = slicer.mrmlScene.AddNewNodeByClass(className, slicer.mrmlScene.GetUniqueNameByString(name))
`

I would recommend to keep the current behavior.

phcerdan:
That makes sense @lassoan.

I have no much experience with the library, so forgive my probably non-sense questions, but if this code is generating two nodes with the same code:
```
myNewNode = slicer.mrmlScene.AddNewNodeByClass(className, name)
myNewNodeWithSameName = slicer.mrmlScene.AddNewNodeByClass(className, name)
```

Wouldn't be nice to have something like:

```
myNode = slicer.mrmlScene.AddNewNodeOrOverrideByClass(className, name)
myNode = slicer.mrmlScene.AddNewNodeOrOverrideByClass(className, name)
```

What is the good/canonical way of discarding existing node and create a new one, but keeping the name? Thanks

lassoan:
Nodes are identified by their ID. IDs are always unique. Names are just labels, which may be the same.

> What is the good/canonical way of discarding existing node and create a new one, but keeping the name?

You want to avoid discarding a node, because that would remove all node references and node observations that various modules have added. If you remove a node and add a new node to the scene with the same name then the new node will not replace the old one in any way; the new node will have no relationship with the old one at all.

phcerdan:
> You want to avoid discarding a node

I see, thanks!

PR #980, opened 2018-06-28
ENH: added 3d views linking capability

Punzo:
Added Link button in 3DViewController;
Added MRMLViewLogic: set the interactions;
Added MRMLViewLinkLogic: propagate the events.

next step for Slicer 4.10 (no backward compatibility): merge vtkMRMLCameraNode into vtkMRMLViewNode

see also https://github.com/NA-MIC/ProjectWeek/blob/master/PW28_2018_GranCanaria/Projects/3DViewsLinking/README.md#progress-and-next-steps point (a)

### DIFF
--- a/Libs/MRML/Logic/vtkMRMLViewLinkLogic.cxx
+++ b/Libs/MRML/Logic/vtkMRMLViewLinkLogic.cxx
@@ -57,14 +57,13 @@ vtkMRMLViewLinkLogic::~vtkMRMLViewLinkLogic()
 }
 
 //----------------------------------------------------------------------------
-void vtkMRMLViewLinkLogic::SetMRMLSceneInternal(vtkMRMLScene * newScene)
+void vtkMRMLViewLinkLogic::SetMRMLSceneInternal(vtkMRMLScene* newScene)
 {
   // List of events the slice logics should listen
   vtkNew<vtkIntArray> events;
   vtkNew<vtkFloatArray> priorities;
 
   float normalPriority = 0.0;
-  // float highPriority = 0.5;
 
   // Events that use the default priority.  Don't care the order they
   // are triggered
@@ -171,7 +170,7 @@ void vtkMRMLViewLinkLogic::PrintSelf(ostream& os, vtkIndent indent)
 }
 
 //----------------------------------------------------------------------------
-void vtkMRMLViewLinkLogic::BroadcastCameraNodeEvent(vtkMRMLCameraNode *cameraNode)
+void vtkMRMLViewLinkLogic::BroadcastCameraNodeEvent(vtkMRMLCameraNode* cameraNode)
 {
   // only broadcast a camera node event if we are not already actively
   // broadcasting events and we are actively interacting with the node
@@ -189,13 +188,13 @@ void vtkMRMLViewLinkLogic::BroadcastCameraNodeEvent(vtkMRMLCameraNode *cameraNod
     }
 
   int requiredViewGroup = viewNode->GetViewGroup();
-  vtkMRMLCameraNode* sNode;
+  vtkMRMLCameraNode* sNode = NULL;
   vtkCollectionSimpleIterator it;
   vtkSmartPointer<vtkCollection> nodes;
   nodes.TakeReference(this->GetMRMLScene()->GetNodesByClass("vtkMRMLCameraNode"));
   for (nodes->InitTraversal(it);
       (sNode = vtkMRMLCameraNode::SafeDownCast(nodes->GetNextItemAsObject(it)));)
-    {
+    {  
     if (!sNode || sNode == cameraNode)
       {
       continue;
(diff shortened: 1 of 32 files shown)


Punzo:
Hi @jcfr, @lassoan, @pieper, @cpinter 

I have added support for liner transformation to the camera when the 3D view linking is active in 
https://github.com/Slicer/Slicer/pull/980/commits/03e7baaf24a2cc594cb52d617fa9d41eafa05a11
(plus some BUG fixes)

Here a video how it works: 
https://www.dropbox.com/s/dgjg8raozt851m6/3dviewlink?dl=0

I don't think that specialized interface is required in the Cameras module (it will just a duplication of the GUI already in the transformation module).

I'd like to merge this pull request within next week, please let me know if you can review/test.

Davide.

P.S.: probably I'll not have time to implement point c (right click to open menu edit proprieties in the 3D views) of the project (https://na-mic.github.io/ProjectWeek/PW28_2018_GranCanaria/Projects/3DViewsLinking/)

pieper:
Hi @Punzo !  I don't see any harm in this, so it's probably okay as is.  I'm still curious though about what we discussed in Las Palmas - I noticed that when you had the second camera inside the transform you only dragged left-right so things worked well.  But isn't it the case that if you drag up-down the second view will 'roll' in a kind of unexpected way?

Punzo:
Hi @pieper 

here is another video:
https://www.dropbox.com/s/mbnigb37dnhhaxl/3dviewlink.mkv?dl=0

I think this is exactly what we were talking about, isn't it? Another thing maybe to discuss is the user-friendliness of the GUI. In fact, in the last part of this video (on purpose), I started to put random angles which made no sense. I think we have two options: 1) we leave as it is (using the transform GUI) and let the user to have the full options on the transformation and after few attempts he/she will know how to use it (i.e., which angles he/she has to modify to get the desired "mirror" camera); 2) or we add specialized interface in the Camera module in which the user can give only a couple of Euler angles or/and some presets of angle combinations (however, I guess this will depends on user cases and applications).  I am more inclined to option (1), but if you have time to test it and find several nice transforms which give useful mirror cameras, I can implement them like push button in the Camera module GUI.

pieper:
Thanks for the extra video - I agree this looks really useful.  +1 👍 

Yes, a more streamlined GUI would be nice but it can be added later as needed.

Punzo:
ok, perfect!

Punzo:
@lassoan @jcfr I rebased against the last master branch and fixed a couple of tests failing. Please let me know if it is ok merging or you will review it. Thanks!

Libs/MRML/Logic/vtkMRMLViewLinkLogic.cxx
```diff
@@ -0,0 +1,462 @@
...
+          {
+          vtkNew<vtkTransform> CameraTransform;
+          vtkNew<vtkMatrix4x4> CameraTransformMatrix;
+          // Assumption: mrmlCamera nodes are only linearly transformed
+          vtkMRMLTransformNode::GetMatrixTransformBetweenNodes
+            (NULL, sNode->GetParentTransformNode(), CameraTransformMatrix.GetPointer());
+          CameraTransformMatrix->Invert(CameraTransformMatrix, CameraTransformMatrix);
+          CameraTransform->Concatenate(CameraTransformMatrix.GetPointer());
+
+          double Position[3];
```
cpinter: Initialize variable please

Libs/MRML/Logic/vtkMRMLViewLinkLogic.cxx
```diff
@@ -0,0 +1,462 @@
...
+  vtkMRMLViewNode* viewNode = vtkMRMLViewNode::SafeDownCast
+    (this->GetMRMLScene()->GetNodeByID(cameraNode->GetActiveTag()));
+
+  if (!viewNode)
+    {
+    return;
+    }
+
+  int requiredViewGroup = viewNode->GetViewGroup();
+  vtkMRMLCameraNode* sNode;
```
cpinter: Please always initialize variables everywhere. It will save a lot of trouble (e.g. these may cause a crash in release mode but not in debug).

Libs/MRML/Widgets/qMRMLThreeDView.cxx
```diff
@@ -440,7 +505,7 @@ void qMRMLThreeDView::resetFocalPoint()
   if (this->renderer())
     {
     this->renderer()->ResetCameraClippingRange();
-    }
+  }
```
cpinter: Please fix indentation

jcfr:
> Instead of 0 or NULL, now we should use nullptr

Let's use `nullptr` only after the coming release. Clang based tools will allow us to update the code in one pass.

Punzo:
Hi Csaba,

thanks for the review.

> Why do we need to have those ViewLogics containers? As far as I see they are not used anywhere. Only to ensure they are not destructed?

No you are right they are not used. However I thought to leave the same infrastructure as the slice logics. In fact, the slice controller widget has all of them in a vtkcollection because there few methods which require all the logics (e.g.: https://github.com/Slicer/Slicer/blob/master/Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx#L2398-L2424

 and  
https://github.com/Slicer/Slicer/blob/master/Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx#L2073-L2095). Therefore, I think it is good to leave them in the case somebody want to add a method in the threeDView controller which require all the logics. Do you think that the vtkcollection of 3dview logic should be removed?  

> Instead of 0 or NULL, now we should use nullptr. There are lots of these, please do all.
>> Let's use nullptr only after the coming release. Clang based tools will allow us to update the code in one pass.

OK

> I see some inconsistencies with formatting, for example whether you put the space after the * or before, etc. Not a big deal, but as you go through the code please try to keep the Slicer style.

you are right, I fixed all the formatting error that I could find. Please let me know if you want that I squash  the commit in a single one and if it is ok to merge.

Thanks!

cpinter:
Thanks a lot, Davide! As far as I'm concerned it's good to go.

Punzo:
Hi @jcfr, @lassoan or @pieper, may you check and merge this PR? (of course, if you think it is ready). It will be very nice if I can have this merged within this week! Thanks!

lassoan:
I'll test this on Windows and if everything goes well then integrate it today or tomorrow.

Punzo:
@lassoan did you have time to test on windows? Thanks!

Punzo:
merged in https://github.com/Slicer/Slicer/commit/1c07409013bb3a8578b0ddfea7e2914452fe2228 thanks @lassoan

PR #981, opened 2018-07-02
COMP: add GetPointer to vtkNew'd matrices

pieper:
On ubuntu 16.04 (gcc 5.4) the vtkMatrix4x4 is not coerced to a
pointer automatically.  Adding GetPointer is consistent with
other usages in this file and solves the compile issue.


lassoan:
Thank you. I guess this is a Qt4/VTK7 build. vtkNew automatic conversion to pointer was added in VTK8 or so.

pieper:
Yes, it's Qt4.8.7 and VTK 7.1.

lassoan:
Thanks. We agreed to make Slicer-4.10 compatible with Qt4/VTK7 (and require VTK9 support after that), so this fix is needed.

lassoan:
Committed r27269

PR #982, opened 2018-07-03
ENH: update MultiVolumeImporter hash

fedorov:
Confidence for the loadables was revisited to prefer loadables that have temporal resolution as needed for quantitative analysis applications. Issue identified using a GE Signa HDxt DSC MRI dataset provided by a user.


lassoan:
merged

PR #983, opened 2018-07-05
BUG: Log full formatted python string

jamesobutler:
This solves an issue where python debug log messages were not having their strings formatted correctly when it was added to the Slicer application log.

This is based on a [discussion](https://discourse.slicer.org/t/python-logging-with-string-formatting-issue/3366) from the discourse forum with @lassoan.


lassoan:
Thank you! Committed 27276.

PR #984, opened 2018-07-06
ENH: update MultiVolumeExplorer hash

fedorov:
New feature added to allow extraction of individual multivolume frames without overwriting. Thanks to @threarth for the contribution!


lassoan:
merged

PR #741, opened 2017-06-30
ENH: Add diverging colormap

fepegar:
Slicer lacked a perceptually uniform diverging colormap. According to [matplotlib](https://matplotlib.org/):
> Diverging: change in lightness and possibly saturation of two different colors that meet in the middle at an unsaturated color; should be used when the information being plotted has a critical middle value, such as topography or when the data deviates around zero.

More info about why perceptually uniform colormaps should be used over older colormaps like Rainbow or Jet:
- [matplotlib colormaps](https://matplotlib.org/users/colormaps.html)
- [Diverging colormaps](http://www.kennethmoreland.com/color-maps/)
- [MATLAB's posts about new perceptually uniform colormaps](http://blogs.mathworks.com/steve/2014/10/13/a-new-colormap-for-matlab-part-1-introduction/)

Some perceptually uniform colormaps have already been added to Slicer before, but they were not diverging: #582 

Some examples:
- Correlation between localization of the electrode contacts in a normalized space and some clinical score (left: Rainbow; middle: ColdToHot; right: new diverging BlueRed):
corr
- Visualization of compression / expansion of a deformation field (Jacobian) after a non-linear registration. When using the new diverging colormap, blue means expansion and red means compression (up: Rainbow; middle: ColdToHot; down: new diverging BlueRed):
jacob

When Rainbow or ColdToHot (similar to Jet) are used:
- It's difficult to mentally map a color with a value
- It's not easy to know where values are close to 0
- There is a lot of green areas, because green take a lot of room in the colormap
- Since yellow takes just a bit of room in the colormap, there seems to be a quick jump between green and red and a yellow line between them

Code used to generate the file:
```
import numpy as np
import matplotlib.pyplot as plt
colors = plt.get_cmap('RdBu')(range(255))
colors = colors[::-1] * 255
colors = colors.astype(np.uint8)
link = 'https://matplotlib.org/users/colormaps.html'
outputPath = '/tmp/BlueRed.txt'
opacity = 255
with open(outputPath, 'w') as f:
    f.write('# Inspired form matplotlib colormaps: {}\n'.format(link))
    for i, color in enumerate(colors):
        s = '%d %d %d %d %d 255\n' % (i, i, color[0], color[1], color[2])
        f.write(s)
```

PR #989, opened 2018-07-11
Change default colormap for model scalars and displacement volumes

fepegar:
See https://github.com/Slicer/Slicer/pull/741

Also, in my vessels network with `Radius` set as Active scalar:

Using Viridis, I can easily figure out that dark lines represent thin vessels and yellow ones are thicker:
viridis

Using Rainbow, the current default, the result is more fun but more confusing:
rainbow

### DIFF
--- a/Modules/Loadable/Transforms/Logic/vtkSlicerTransformLogic.cxx
+++ b/Modules/Loadable/Transforms/Logic/vtkSlicerTransformLogic.cxx
@@ -542,7 +542,7 @@ vtkMRMLVolumeNode* vtkSlicerTransformLogic::CreateDisplacementVolumeFromTransfor
       vtkNew<vtkMRMLVectorVolumeDisplayNode> displayNode;
     }
     scene->AddNode(displayNode.GetPointer());
-    displayNode->SetAndObserveColorNodeID("vtkMRMLColorTableNodeRainbow");
+    displayNode->SetAndObserveColorNodeID("vtkMRMLColorTableNodeFileViridis.txt");
     outputVolumeNode->SetAndObserveNthDisplayNodeID(0, displayNode->GetID());
   }
 


fepegar:
Feature request: is it possible to visualize the colorbar of the scalars in the 3D view, as the one shown by Data Probe on the 2D views?

lassoan:
D You can show the color bar in 3D views using Colors module (there is a section to configure it, near the bottom).

fepegar:
Yes, but it would be great to see the values changing to reflect the "Display Range" widget limits.

pieper:
The pull request makes sense to me.  +1 for merging.

lassoan:
There are 4 distinct colors (red, green, blue, purple) in rainbow colormap, which might be helpful, as it provides iso-lines/surfaces. However, the iso-values are usually just randomly assigned (by scaling the colormap to the scalar range), highlighting lines/surfaces that are not particularly interesting.

Viridis provides continuous transition, so there are no specific iso-values that are displayed more prominently.

Due to this, viridis seems to be a more generic colormap, which I agree that seems more suitable as a default colormap than rainbow.

I would slightly prefer plasma (another smoothly transitioning colormap), as its colors are associated with cold-to-hot temperature; while viridis transitions between two neutral colors (purple-to-yellow).

'R' programming language went with viridis (https://www.r-bloggers.com/using-the-new-viridis-colormap-in-r-thanks-to-simon-garnier/), so it might be one more reason to choose the same (or not, to be more distinctive?).

Rainbow:
image

Viridis:
image

Plasma:
image

fepegar:
Matplotlib also uses Viridis as default. Originally they had only the other three, but apparently [people prefer Viridis](https://youtu.be/xAoljeRJ3lU?t=16m38s) for some reason. They have all been designed on the same principles anyway.

I agree that Plasma might be a bit closer to the hot-cold idea but not because of the extremes, which are very similar, but because it goes through red/orange instead of green.

Do you think it'd be a good idea to add this default to the PET and DTI presets?

pieper:
@fedorov do you have preferences for PET?  @ljod and @ihnorton how about dMRI?

lassoan:
Merged https://github.com/Slicer/Slicer/commit/589e6b8d29cb13b434c97bd504f386d6e7503038

PR #987, opened 2018-07-08
COMP: Add missing headers.

phcerdan:
Solve incomplete type errors when using latest VTKv9:
982ec1c561868f3e2442d11df0c44eaea1e56834

```
Software/Slicer/Slicer-src/Base/QTApp/qSlicerApplicationHelper.cxx: In static member function ‘static void qSlicerApplicationHelper::preInitializeApplication(const char*, ctkProxyStyle*)’:
Software/Slicer/Slicer-src/Base/QTApp/qSlicerApplicationHelper.cxx:98:18: error: variable ‘QSurfaceFormat format’ has initializer but incomplete type
   QSurfaceFormat format = QVTKOpenGLWidget::defaultFormat();
                  ^~~~~~
Software/Slicer/Slicer-src/Base/QTApp/qSlicerApplicationHelper.cxx:98:59: error: invalid use of incomplete type ‘class QSurfaceFormat’
   QSurfaceFormat format = QVTKOpenGLWidget::defaultFormat();
                                                           ^
In file included from Software/Slicer/Slicer-src/Base/QTApp/qSlicerApplicationHelper.cxx:59:
Software/Slicer/build-relwithdebinfo/VTKv9/GUISupport/Qt/QVTKOpenGLWidget.h:23:7: note: forward declaration of ‘class QSurfaceFormat’
 class QSurfaceFormat;
       ^~~~~~~~~~~~~~
Software/Slicer/Slicer-src/Base/QTApp/qSlicerApplicationHelper.cxx:100:19: error: incomplete type ‘QSurfaceFormat’ used in nested name specifier
   QSurfaceFormat::setDefaultFormat(format);
```

PR #990, opened 2018-07-16
COMP: Add missing header when using QVTKOpenGLWidget

phcerdan:
Same fix than #987 but for all tests.

Missing header when `Slicer_VTK_USE_QVTKOPENGLWIDGET` is defined.

Tested using VTKv9 from this branch:
https://github.com/phcerdan/VTK/tree/slicer-v9.0.0-2018-07-06-982ec1c561

Solve error:
```bash
In file included from
/Software/Shape/SALT/build/slicersources-src/Libs/MRML/Widgets/Testing/qMRMLDisplayNodeWidgetTest1.cxx:37:
/Software/Shape/SALT/build/VTKv9/GUISupport/Qt/QVTKOpenGLWidget.h:23:7:
note: forward declaration of ‘class QSurfaceFormat’
 class QSurfaceFormat;
        ^~~~~~~~~~~~~~
        /Software/Shape/SALT/build/slicersources-src/Libs/MRML/Widgets/Testing/qMRMLDisplayNodeWidgetTest1.cxx:48:19:
        error: incomplete type ‘QSurfaceFormat’ used in nested name
        specifier
           QSurfaceFormat::setDefaultFormat(format);
```

### DIFF
--- a/Applications/SlicerApp/Testing/Cpp/qSlicerAppMainWindowTest1.cxx
+++ b/Applications/SlicerApp/Testing/Cpp/qSlicerAppMainWindowTest1.cxx
@@ -31,6 +31,7 @@
 
 // VTK includes
 #ifdef Slicer_VTK_USE_QVTKOPENGLWIDGET
+#include <QSurfaceFormat>
 #include <QVTKOpenGLWidget.h>
 #endif
 
--- a/Base/QTGUI/Testing/Cxx/qSlicerDirectoryListViewTest1.cxx
+++ b/Base/QTGUI/Testing/Cxx/qSlicerDirectoryListViewTest1.cxx
@@ -31,6 +31,7 @@
 
 // VTK includes
 #ifdef Slicer_VTK_USE_QVTKOPENGLWIDGET
+#include <QSurfaceFormat>
 #include <QVTKOpenGLWidget.h>
 #endif
 
--- a/Base/QTGUI/Testing/Cxx/qSlicerLayoutManagerTest1.cxx
+++ b/Base/QTGUI/Testing/Cxx/qSlicerLayoutManagerTest1.cxx
@@ -31,6 +31,7 @@
 
 // VTK includes
 #ifdef Slicer_VTK_USE_QVTKOPENGLWIDGET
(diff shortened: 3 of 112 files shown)


pieper:
Technically this is a Qt include that should be up with the other Qt includes and not mixed in with the VTK includes.  But since it's in the ifdef I can see the point of keeping them together.  Soon this ifdef will go away since it won't be possible to build against the old classes - I'm a little afraid we'll forget to move this include when that time comes.

lassoan:
Would it be possible to add `#include <QSurfaceFormat>` to somewhere in QVTKOpenGLWidget.h?

jcfr:
> Would it be possible to add #include <QSurfaceFormat> to somewhere in QVTKOpenGLWidget.h?

Is is currently forward declared.
See https://github.com/Kitware/VTK/blob/200888e78daff01ec9896ac7075bcdc4974a10ec/GUISupport/Qt/QVTKOpenGLWidget.h#L23

We could patch the Slicer/VTK fork ... but I think including the header is sensible

lassoan:
> including the header is sensible

Good point. Actually, the problem is that default surface format setting is an internal detail that QVTKOpenGLWidget-using applications should not care about (except special cases). Application should just call a method like `QVTKOpenGLWidget::setDefaultSurfaceFormat()`, which should perform this:

    // Set default surface format for QVTKOpenGLWidget
    QSurfaceFormat format = QVTKOpenGLWidget::defaultFormat();
    format.setSamples(0);
    QSurfaceFormat::setDefaultFormat(format);


This would not require adding extra includes. It would also allow changing default surface format setup at a single place.

phcerdan:
Should we merge this to fix the compiling errors, and improve the solution when the proposed modification to `QVTKOpenGLWidget`, `QVTKOpenGLWidget::setDefaultSurfaceFormat()` is merged in VTK, and the VTK updated in Slicer?

jcfr:
Closing. Integrated in [r27346](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27346)
This was required to move forward with the VTK update

PR #991, opened 2018-07-17
BUG: Fix scene views module

lassoan:
Use native Qt widget instead of web widget.
Also fixed writing of scene view description to scene file (HTML formatting can be saved into the scene file).

Fixes
- https://issues.slicer.org/view.php?id=4542
- https://issues.slicer.org/view.php?id=4508

image


lassoan:
Thanks for testing this. Merged in r27283.

I could only get errors logged and not crash, but scene views certainly still unstable. We should try to fix the crash and maybe implement proper fix after Slicer-4.10 is released.

PR #992, opened 2018-07-18
STYLE: Fix comment explaining the smoothing factor

fepegar:
I fixed the numbers explaining the mapping between the smoothing factor and the pass band of the smoothing filter used in `vtkSegmentationCore`.


lassoan:
@fepegar it would be nice if you could check if it made sense to take pixel spacing into account for computation of the passband value.

lassoan:
Yes. When voxel size is larger then there are less details to preserve and larger steps to smooth, so limit of maximum allowed spatial frequencies in the output mesh should be lower.

PR #993, opened 2018-07-19
ENH: Changed Slicer's Python executable to PythonSlicer

lassoan:
PythonSlicer is recognized as Python executable by developer tools such as PyCharm, while SlicerPython is not recognized.

SlicerPython is kept for now for backward compatibility but will be removed in the future.

### DIFF
--- a/CMake/SlicerBlockInstallPython.cmake
+++ b/CMake/SlicerBlockInstallPython.cmake
@@ -75,24 +75,32 @@ if(Slicer_USE_PYTHONQT)
   endif()
 
   # Install Slicer python launcher settings
-  install(
-    FILES ${python_bin_dir}/SlicerPythonLauncherSettingsToInstall.ini
-    DESTINATION ${Slicer_INSTALL_BIN_DIR}
-    RENAME SlicerPythonLauncherSettings.ini
-    COMPONENT Runtime
-    )
 
-  # Install Slicer python launcher
-  set(_launcher CTKAppLauncher)
-  if(Slicer_BUILD_WIN32_CONSOLE)
-    set(_launcher CTKAppLauncherW)
-  endif()
-  install(
-    PROGRAMS ${CTKAppLauncher_DIR}/bin/${_launcher}${CMAKE_EXECUTABLE_SUFFIX}
-    DESTINATION ${Slicer_INSTALL_BIN_DIR}
-    RENAME SlicerPython${CMAKE_EXECUTABLE_SUFFIX}
-    COMPONENT Runtime
-    )
+  macro(_install_python_launcher executablename)
+    # Install Slicer python launcher settings
+    install(
+      FILES ${python_bin_dir}/${executablename}LauncherSettingsToInstall.ini
+      DESTINATION ${Slicer_INSTALL_BIN_DIR}
+      RENAME ${executablename}LauncherSettings.ini
+      COMPONENT Runtime
+      )
+    # Install Slicer python launcher
+    set(_launcher CTKAppLauncher)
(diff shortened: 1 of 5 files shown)


lassoan:
Thanks @ihnorton, I've updated all the remaining references.

lassoan:
Integrated into http://svn.slicer.org/Slicer4/trunk@27289

Packaging was successful on Windows, so hopefully the CircleCI test error was just a false alarm. I'll check the packaging on the dashboard tomorrow.

jcfr:
>  the CircleCI test error was just a false alarm

Most likely. Since the commit touches external projects, we have to wait for a new base image to be pushed on dockerhub.

ihnorton:
@lassoan @jcfr @pieper I wonder if it makes sense to override `sys.executable` to point to `PythonSlicer`? The [documentation (link)](https://docs.python.org/2/library/sys.html#sys.executable) says that it is:

> A string giving the absolute path of the executable binary for the Python interpreter, on systems 
> where this makes sense. If Python is unable to retrieve the real path to its executable, 
> sys.executable will be an empty string or None.

Currently we set `sys.executable = slicer.app.applicationPath`, but that is not a usable interpreter. This can break software which expects to use `sys.executable` to do things like spawn pre-compilation/caching processes.

jcfr:
Now that `PythonSlicer` use the same environment as Slicer (thanks to [27198](viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27198) and [27247](viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27247)), I think it makes sense.

* For build tree: yes
* For install tree: we first need to make sure we package `PythonSlicer` on the three platform

lassoan:
PythonSlicer.exe is included in the install package on Windows.

ihnorton:
There is a note in `Utilities/Scripts/CMakeLists.txt` which says:

```
  # Determine name of Python interpreter
  #
  # NOTE: Do NOT use ${PYTHON_EXECUTABLE}; when using our own Python, it is set
  #       to a ctkAppLauncher wrapper, and we don't want that!
```

Would it be better to use `python-real`? The downside is it will not be usable outside of Slicer, on all platforms, so if some software is relying on `sys.executable` to write a "permanent" path for future reference, it won't work. The upside is: it is not an applauncher, so we don't have to risk conflicting arguments or other issues.

jcfr:
This comment applies only to the slicerExtensionWizard configured for the build tree and in that case the python interpreter is not renamed.

Are you suggesting we rename the python executable in the build tree ? 

If yes, the logic currently done in the script configured for the install tree

https://github.com/Slicer/Slicer/blob/fd5fb920990cc57fae235640c00f4d2c35fa02c5/Utilities/Scripts/slicerExtensionWizard-install.sh.in#L6-L8

should be applied in to the script configured for the build tree [slicerExtensionWizard.sh.in](https://github.com/Slicer/Slicer/blob/master/Utilities/Scripts/slicerExtensionWizard.sh.in).

ihnorton:
> Are you suggesting we rename the python executable in the build tree ?

No, just wondering which is better to use:
```
sys.executable = '[...]/PythonSlicer'
#or
sys.executable = '[...]/python-real'
```

(by the way, the specific use-case right now is SlicerJuptyer, because jedi does some caching/precompilation step where it wants to shell out to a usable interpreter)

jcfr:
I see.

Then, that would be set here:

https://github.com/Slicer/Slicer/blob/fd5fb920990cc57fae235640c00f4d2c35fa02c5/Base/QTCore/qSlicerCoreApplication.cxx#L900-L909

PR #995, opened 2018-07-27
BUG: Ensure Slicer process has a name in macOS Activity monitor

jcfr:
See https://github.com/Slicer/SlicerJupyter/issues/9

Reported-by: Steve Pieper <pieper@bwh.harvard.edu>


jcfr:
Question: Instead of "Slicer" we could have the "Slicer <full version>". Would that be better ?

jcfr:
Option 1:

image


Option 2:
image

jcfr:
Cc: @ihnorton

ihnorton:
I prefer the full date information because I usually have 3+ slicer instances running.

(semi-related: I've been meaning to make a PR to add the PID to the application title in developer mode)

PR #996, opened 2018-07-30
COMP: Fail early building script with PythonQt disabled

phcerdan:
```bash
CMake Error at Software/Slicer/SlicerCustomAppTemplate/build-SlicerSALT/slicersources-src/CMake/SlicerMacroBuildScriptedModule.cmake:94 (ctkMacroCompilePythonScript):
  Unknown CMake command "ctkMacroCompilePythonScript".
Call Stack (most recent call first):
  Software/Slicer/SlicerCustomAppTemplate/build-SlicerSALT/Sequences/SequenceSampleData/CMakeLists.txt:17 (slicerMacroBuildScriptedModule)
```

This is caused because the cmake macro `ctkMacroCompilePythonScript` is only included when PythonQt is on.

### DIFF
--- a/CMake/SlicerMacroBuildScriptedModule.cmake
+++ b/CMake/SlicerMacroBuildScriptedModule.cmake
@@ -37,9 +37,9 @@ macro(slicerMacroBuildScriptedModule)
     "${multiValueArgs}"
     ${ARGN}
     )
-    
+
   message(STATUS "Configuring Scripted module: ${MY_SLICER_NAME}")
-  
+
   # --------------------------------------------------------------------------
   # Print information helpful for debugging checks
   # --------------------------------------------------------------------------
@@ -82,6 +82,12 @@ macro(slicerMacroBuildScriptedModule)
       endif()
     endforeach()
   endforeach()
+
+  if(NOT Slicer_USE_PYTHONQT)
+      message(FATAL_ERROR
+        "Attempting to build the Python scripted module '${MY_SLICER_NAME}'"
+        " when Slicer_USE_PYTHONQT is OFF")
+  endif()
 
   set(_no_install_subdir_option NO_INSTALL_SUBDIR)
   set(_destination_subdir "")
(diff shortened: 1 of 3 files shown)


jcfr:
This probably means that the module using the CMake function `slicerMacroBuildScriptedModule` should be excluded when `Slicer_USE_PYTHONQT` is `OFF`

lassoan:
It could also help to fail early, with a meaningful error message Slicer_USE_PYTHONQT is FALSE and a Python scripted module is attempted to be added. You can use MESSAGE(FATAL_ERROR ...) for this.

phcerdan:
@jcfr it makes sense. @lassoan where do you suggest adding the error message? In `CMake/SlicerMacroBuildScriptedModule.cmake`?
Something like:
 ```cmake
if(NOT Slicer_USE_PYTHONQT )
  MESSAGE(FATAL_ERROR "Attempting to build a Python scripted module when Slicer_USE_PYTHONQT is OFF")
endif()
```

**Comment by jcfr** [comment]
What about:

```
if(NOT Slicer_USE_PYTHONQT)
  message(FATAL_ERROR "Attempting to build the Python scripted module '${MY_SLICER_NAME}' when Slicer_USE_PYTHONQT is OFF")
endif()
```

> where do you suggest adding the error message?

here: https://github.com/Slicer/Slicer/blob/2c43c10bf9308277a3769b8d2c70f51857e29600/CMake/SlicerMacroBuildScriptedModule.cmake#L85

phcerdan:
Thanks @jcfr, added, please have a look.

jcfr:
Thanks, so that the message is really useful could you move this call from:

https://github.com/Slicer/Slicer/blob/87e8a89274eca8c0a249df8785ded16a060006ab/Modules/Scripted/CMakeLists.txt#L2

to here:

https://github.com/Slicer/Slicer/blob/87e8a89274eca8c0a249df8785ded16a060006ab/CMakeLists.txt#L387-L396

phcerdan:
Moved to the top level @jcfr

jcfr:
Thanks @phcerdan . Closing, integrated in [r27345](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27345)

PR #999, opened 2018-07-31
ENH: Add Segmentations closed surface to fractional labelmap conversion test

Sunderlandkyl:
Fixed and moved from SlicerRT to a more relevant location.

Tests the conversion of a sphere closed surface to a fractional labelmap using vtkClosedSurfaceToFractionalLabelmapConversionRule.

### DIFF
--- a/Libs/vtkSegmentationCore/Testing/vtkClosedSurfaceToFractionalLabelMapConversionTest1.cxx
+++ b/Libs/vtkSegmentationCore/Testing/vtkClosedSurfaceToFractionalLabelMapConversionTest1.cxx
@@ -1,0 +1,137 @@
+/*==============================================================================
+
+  Copyright (c) Laboratory for Percutaneous Surgery (PerkLab)
+  Queen's University, Kingston, ON, Canada. All Rights Reserved.
+
+  See COPYRIGHT.txt
+  or http://www.slicer.org/copyright/copyright.txt for details.
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+  This file was originally developed by Kyle Sunderland, PerkLab, Queen's University
+  and was supported through the Applied Cancer Research Unit program of Cancer Care
+  Ontario with funds provided by the Ontario Ministry of Health and Long-Term Care
+
+==============================================================================*/
+
+// VTK includes
+#include <vtkImageAccumulate.h>
+#include <vtkNew.h>
+#include <vtkPolyData.h>
+#include <vtkSphereSource.h>
+
+// vtkSegmentationCore includes
+#include "vtkClosedSurfaceToFractionalLabelmapConversionRule.h"
+#include <vtkOrientedImageData.h>
+#include <vtkPolyDataToFractionalLabelmapFilter.h>
+#include <vtkSegment.h>
+#include <vtkSegmentation.h>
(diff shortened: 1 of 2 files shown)


Sunderlandkyl:
@cpinter The issue with the test was a result in a change to the default conversion geometry.

Libs/vtkSegmentationCore/Testing/vtkClosedSurfaceToFractionalLabelMapConversionTest1.cxx
```diff
@@ -0,0 +1,137 @@
...
+    sphereSegment->GetRepresentation(vtkSegmentationConverter::GetSegmentationFractionalLabelmapRepresentationName()) );
+
+  vtkNew<vtkImageAccumulate> imageAccumulate;
+  imageAccumulate->SetInputData(fractionalLabelmap);
+  imageAccumulate->Update();
+
+  int expectedVoxelCount = 234484;
+  int voxelCount = imageAccumulate->GetVoxelCount();
+  if (voxelCount != expectedVoxelCount)
+  {
```
jcfr: nitpick: Consider using the `CHECK_*` macro. For example see https://github.com/Slicer/Slicer/blob/56e72e1b147834a386f67ade2f71987f0a48a7b3/Modules/Loadable/Volumes/Testing/Cxx/vtkSlicerVolumesLogicTest1.cxx#L81

See http://apidocs.slicer.org/master/vtkAddonTestingMacros_8h.html#define-members

jcfr (changes requested):

jcfr:
Thanks for the contribution. Beside of few nitpick and the review of @cpinter , we should be good.

Libs/vtkSegmentationCore/Testing/vtkClosedSurfaceToFractionalLabelMapConversionTest1.cxx
lassoan: I think vtkSegmentationCore does not depend on vtkAddon. It'll be advertised as a separate library (and a copy will be made available in a standalone git repository, independently from Slicer) and if there is enough interest then it may be actually spun out and maintained as a separate library. So, making it depend on vtkAddon might not be preferable.

jcfr: Makes sense..

Whenever you think it makes sens, we could extract the history and create a dedicated repo.

Then we could either add an `UpdateFromUpstream.sh` script or an other external project ...

jcfr (approved):
Based on @lassoan comment, using the `CHECK_` macros is not relevant here.

Pending @cpinter review, :+1:

Libs/vtkSegmentationCore/Testing/vtkClosedSurfaceToFractionalLabelMapConversionTest1.cxx
cpinter: I agree, we shouldn't add dependencies other than regular VTK.

The dedicated repo is in place: https://github.com/PerkLab/PolySeg
I created this for a publication, but it's identical to vtkSegmentationCore (except this one commit). We can make it an external project whenever.

cpinter:
Thanks, Kyle! If it compiles on Linux/Mac, and the test passes, then I think it's good to go.

Sunderlandkyl:
Successfully compiled and tested on an experimental dashboard build here: http://slicer.cdash.org/buildSummary.php?buildid=1338089

cpinter:
Integrated in https://github.com/Slicer/Slicer/commit/6a4e983445d032248d7a7b3a63a95580899ee28e

PR #1000, opened 2018-08-01
ENH: Move Editor module to Legacy category

lassoan:
Now Segment Editor should be used instead of Editor module.
To reduce user confusion, moved Editor module to Legacy category and removed Editor module from default Favorite module list.


pieper:
I'd thought we'd wait until after the 4.10 release but this will reduce confusion.

lassoan:
Yes, it's for reducing confusion. The module is still available for those who look for it.

cpinter:
I'm OK with this if everybody is.

It will definitely steer new users to use Segment Editor, so we'll need to answer "use Segment Editor" a bit less on discourse.

PR #1004, opened 2018-08-10
BUG: Fix slice view corner annotations in DataProbe

fepegar:
Annotations in 2D views were not hidden when the box was unchecked and:
 1. There was an image in the label layer or
 2. There were images in background and foreground


jcfr:
Thanks for the contribution. I tweaked the commit message and integrated in [r27344](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27344)

fepegar:
Thanks @jcfr

PR #1006, opened 2018-08-23
Fix AddPointToNthMarkup method

vovythevov:
The method never returned anything but 0 and would not trigger a modified
when the point was added.


jcfr:
:+1:  

After changing the commit message to `BUG: Ensure vtkMRMLMarkupsNode::AddPointToNthMarkup return value on failure`, commit after the CI returns green..

jcfr:
Thanks @vovythevov

PR #1010, opened 2018-08-31
PREVIEW: Planes and splines

vovythevov:
### **DO NOT MERGE**
This is a preview PR intended to discuss the Splines and Planes module.

CC: @lassoan, @jcfr, @cpinter


jcfr:
@lassoan @Sunderlandkyl  Is there anything useful from this PR that would make sense to integrate ?

lassoan:
I've reviewed this PR and integrated everything that I could in https://github.com/Slicer/Slicer/pull/4938 (the only major difference is that I don't use an intermediate representation to convert between MRML and JSON).

lassoan:
We looked at and used some of this when we reworked markups and I don't think there is anything more that need to be ported from here to the current Slicer core.

PR #1015, opened 2018-09-05
ENH: add Ext. Wizard paths by default

ihnorton:
This sets the "Add selected modules to search path" button to checked by default (this is probably what people want most of the time):

image

Suggested by @ljod


jcfr:
That sounds reasonable.

Should these lines be removed too:

https://github.com/Slicer/Slicer/pull/1015/files#diff-34f5b47e8fee89d93fbacf67f4e21f45R70

https://github.com/Slicer/Slicer/pull/1015/files#diff-34f5b47e8fee89d93fbacf67f4e21f45R76

ihnorton:
Those seem fine, they just disable the checkbox if nothing is selected.

jcfr:
Thanks for checking.

ihnorton:
Thanks. r27395

PR #1016, opened 2018-09-06
ENH: Add pauseRender() and resumeRender() functions to qMRMLLayoutManager

Sunderlandkyl:
When modifying or setting up nodes that may trigger a render event, pauseRender can be called to temporarily stop the views from rendering intermediate steps.
Once all modifications have been completed, calling resumeRender will restore the previously stored pause states.

Depends on https://github.com/commontk/CTK/pull/827

-----

Edited by @jcfr: 

To understand the motivation see https://github.com/commontk/CTK/pull/827#issue-213722434 and exchange between @finetjul  and @lassoan  copied from https://github.com/commontk/CTK/pull/827#issuecomment-419428173 

> > From @finetjul :
> > Please explain how different it is from renderEnabled.

> From @lassoan 
> Good point. There is a very important difference (rendering requests arriving while renderEnabled(false) makes rendering requests to be forgotten, while they are just delayed while rendering is paused) but it may not be obvious at first.
> 
> Another important advantage of pauseRender is that it can be called from high-level application code, where it is not clear if a certain change will eventually require re-rendering or not. When pauseRender is turned off, rendering will only be performed if it was requested while rendering was paused (in contrast, after renderEnabled(false) you must always re-render).
> 
> I think renderEnabled is mostly made obsolete by the new mechanism.
> 

Cc: @ihnorton

### DIFF
--- a/Libs/MRML/Widgets/qMRMLLayoutManager.cxx
+++ b/Libs/MRML/Widgets/qMRMLLayoutManager.cxx
@@ -24,6 +24,7 @@
 
 // MRMLWidgets includes
 #include <qMRMLLayoutManager_p.h>
+#include <qMRMLSliceView.h>
 #include <qMRMLSliceWidget.h>
 #include <qMRMLSliceControllerWidget.h>
 #include <qMRMLChartView.h>
@@ -1329,4 +1330,37 @@ QWidget* qMRMLLayoutManager::viewWidget(vtkMRMLNode* viewNode) const
   Q_D(const qMRMLLayoutManager);
 
   return d->viewWidget(viewNode);
+}
+
+//------------------------------------------------------------------------------
+void qMRMLLayoutManager::setRenderPaused(bool pause)
+{
+  // Note: views that are instantiated between pauseRender() calls will not be affected
+  // by the specified pause state
+  Q_D(qMRMLLayoutManager);
+  qMRMLLayoutViewFactory* sliceViewFactory = this->mrmlViewFactory("vtkMRMLSliceNode");
+  foreach(const QString& viewName, sliceViewFactory->viewNodeNames())
+    {
+    ctkVTKAbstractView* view = this->sliceWidget(viewName)->sliceView();
+    view->setRenderPaused(pause);
+    }
+
+  qMRMLLayoutViewFactory* threeDViewFactory = this->mrmlViewFactory("vtkMRMLViewNode");
+  foreach(const QString& viewName, threeDViewFactory->viewNodeNames())
+    {
+    ctkVTKAbstractView* view = this->threeDWidget(viewName)->threeDView();
+    view->setRenderPaused(pause);
+    }
+}
+
+//------------------------------------------------------------------------------
+void qMRMLLayoutManager::pauseRender()
+{
+  this->setRenderPaused(true);
+}
+
+//------------------------------------------------------------------------------
+void qMRMLLayoutManager::resumeRender()
+{
+  this->setRenderPaused(true);
 }
--- a/Base/QTGUI/qSlicerApplication.cxx
+++ b/Base/QTGUI/qSlicerApplication.cxx
@@ -993,3 +993,28 @@ void qSlicerApplication::logApplicationInformation() const
          additionalModulePaths.isEmpty() ? "(none)" : qPrintable(additionalModulePaths.join(", ")));
 
 }
+
+//-----------------------------------------------------------------------------
+void qSlicerApplication::setRenderPaused(bool pause)
+{
+  Q_D(qSlicerApplication);
+
+  if (d->LayoutManager)
+    {
+    d->LayoutManager->setRenderPaused(pause);
+    }
+
+  emit renderPaused(pause);
+}
+
+//------------------------------------------------------------------------------
+void qSlicerApplication::pauseRender()
+{
+  this->setRenderPaused(true);
+}
+
+//------------------------------------------------------------------------------
+void qSlicerApplication::resumeRender()
+{
+  this->setRenderPaused(false);
+}
--- a/Base/QTGUI/qSlicerApplication.h
+++ b/Base/QTGUI/qSlicerApplication.h
@@ -171,6 +171,21 @@ public slots:
   /// See http://doc.qt.io/qt-5/windows-issues.html#fullscreen-opengl-based-windows
   void setHasBorderInFullScreen(bool);
 
+  /// Calls setRenderPaused(pause) on the current layout manager.
+  /// Emits pauseRenderRequested() if pause is true and resumeRenderRequested() if pause is false.
+  /// The caller is responsible for making sure that each setRenderPaused(true) is paired with
+  /// setRenderPaused(false).
+  /// \sa qMRMLLayoutManager::setRenderPaused()
+  void setRenderPaused(bool pause);
+
+  /// Equivalent to setRenderPaused(true)
+  /// \sa setRenderPaused
+  void pauseRender();
+
+  /// Equivalent to setRenderPaused(false)
+  /// \sa setRenderPaused
+  void resumeRender();
+
 signals:
 
   /// Emitted when the startup phase has been completed.
@@ -183,6 +198,10 @@ signals:
   ///
   /// \sa qSlicerAppMainWindow::initialWindowShown()
   void startupCompleted();
+
+  /// Emitted when setRenderPaused() is called
+  /// \sa setRenderPaused
+  void renderPaused(bool);
 
 protected:
   /// Reimplemented from qSlicerCoreApplication
--- a/Libs/MRML/Widgets/qMRMLLayoutManager.h
+++ b/Libs/MRML/Widgets/qMRMLLayoutManager.h
@@ -243,6 +243,20 @@ public slots:
   /// \sa qMRMLSliceControllerWidget::fitSliceToBackground(), vtkMRMLSliceLogic::FitSliceToAll()
   void resetSliceViews();
 
+  /// Calls setPauseRender(pause) on all slice and 3D views
+  /// Tracks the previous pause state which is restored using resumeRender()
+  /// Each pauseRender(true) should always be accompanied by a corresponding pauseRender(false) call
+  /// \sa pauseRender
+  void setRenderPaused(bool pause);
+
+  /// Equivalent to setRenderPaused(true)
+  /// \sa setRenderPaused
+  void pauseRender();
+
+  /// Equivalent to setRenderPaused(false)
+  /// \sa setRenderPaused
+  void resumeRender();
+
 signals:
   void activeMRMLThreeDViewNodeChanged(vtkMRMLViewNode* newActiveMRMLThreeDViewNode);
   void activeMRMLChartViewNodeChanged(vtkMRMLChartViewNode* newActiveMRMLChartViewNode);


pieper:
LGTM 👍

Libs/MRML/Widgets/qMRMLLayoutManager.cxx
```diff
@@ -1330,3 +1330,81 @@ QWidget* qMRMLLayoutManager::viewWidget(vtkMRMLNode* viewNode) const
 
   return d->viewWidget(viewNode);
 }
+
+//------------------------------------------------------------------------------
+void qMRMLLayoutManager::pauseRender(bool pause)
```
lassoan: Since we have two methods (pause and resume render), I don't think we need the bool input argument.

Libs/MRML/Widgets/qMRMLLayoutManager.cxx
```diff
@@ -1330,3 +1330,81 @@ QWidget* qMRMLLayoutManager::viewWidget(vtkMRMLNode* viewNode) const
...
+
+  qMRMLLayoutManagerPrivate::PausedViewVector currentPausedViews;
+
+  QString sliceNodeType = "vtkMRMLSliceNode";
+  qMRMLLayoutViewFactory* sliceViewFactory = this->mrmlViewFactory(sliceNodeType);
+  foreach(const QString& viewName, sliceViewFactory->viewNodeNames())
+    {
+    ctkVTKAbstractView* view = (ctkVTKAbstractView*)this->sliceWidget(viewName)->sliceView();
+    qMRMLLayoutManagerPrivate::PausedView pausedView;
+    pausedView.Name = viewName.toStdString();
```
lassoan: try to minimize conversions between QString and std::string. In Qt classes, Qt types are preferred, but if you only ever need to store std type in a variable then it is OK to use std.

Libs/MRML/Widgets/qMRMLLayoutManager.cxx
```diff
@@ -1330,3 +1330,81 @@ QWidget* qMRMLLayoutManager::viewWidget(vtkMRMLNode* viewNode) const
...
+void qMRMLLayoutManager::resumeRender()
+{
+  // Note: views that are instantiated between pauseRender() and resumeRender() calls will
+  // not be affected by resumeRender()
+
+  Q_D(qMRMLLayoutManager);
+
+  if (d->PausedViews.empty())
+    {
+    qCritical() << "No stored render pause state!";
```
lassoan: Add method name.

qCritical() << Q_FUNC_INFO << " failed: No stored...

Libs/MRML/Widgets/qMRMLLayoutManager.cxx
```diff
@@ -1330,3 +1330,81 @@ QWidget* qMRMLLayoutManager::viewWidget(vtkMRMLNode* viewNode) const
...
+    return;
+    }
+
+  QString sliceNodeType = "vtkMRMLSliceNode";
+  QString threeDNodeType = "vtkMRMLViewNode";
+
+  std::vector<qMRMLLayoutManagerPrivate::PausedView> pausedViews = d->PausedViews.back();
```
lassoan: In Qt classes use Qt containers (QVector instead of std::vector) unless there is a specific reason to use std containers.

Libs/MRML/Widgets/qMRMLLayoutManager_p.h
```diff
@@ -150,6 +153,16 @@ public slots:
   vtkMRMLChartViewNode*   ActiveMRMLChartViewNode;
   vtkMRMLTableViewNode*   ActiveMRMLTableViewNode;
   vtkMRMLPlotViewNode*    ActiveMRMLPlotViewNode;
+
+  struct PausedView
+  {
+    std::string Name;
+    std::string NodeType;
+    bool        Paused;
```
lassoan: Maybe WasPaused would be more clear.

Libs/MRML/Widgets/qMRMLLayoutManager.cxx
```diff
@@ -1330,3 +1330,81 @@ QWidget* qMRMLLayoutManager::viewWidget(vtkMRMLNode* viewNode) const
...
+
+  Q_D(qMRMLLayoutManager);
+
+  qMRMLLayoutManagerPrivate::PausedViewVector currentPausedViews;
+
+  QString sliceNodeType = "vtkMRMLSliceNode";
+  qMRMLLayoutViewFactory* sliceViewFactory = this->mrmlViewFactory(sliceNodeType);
+  foreach(const QString& viewName, sliceViewFactory->viewNodeNames())
+    {
+    ctkVTKAbstractView* view = (ctkVTKAbstractView*)this->sliceWidget(viewName)->sliceView();
```
lassoan: use qobject_cast<> instead of C-style cast

lassoan (commented):
Thank you, it looks good! Added a few comments, but they should all be easy to fix.

Libs/MRML/Widgets/qMRMLLayoutManager.cxx
Sunderlandkyl: That reminds me of something.

This works fine without any casting:
ctkVTKAbstractView* view = this->threeDWidget(viewName)->threeDView();

While this requires some sort of cast to store in ctkVTKAbstractView:
ctkVTKAbstractView* view = this->sliceWidget(viewName)->sliceView();

They both follow the same inheritance. Any idea why the casting is necessary?

lassoan: If the pointers are compatible but there is a compilation error then maybe the header file of the child class is not included.

Libs/MRML/Widgets/qMRMLLayoutManager_p.h
```diff
@@ -150,6 +153,16 @@ public slots:
   vtkMRMLChartViewNode*   ActiveMRMLChartViewNode;
   vtkMRMLTableViewNode*   ActiveMRMLTableViewNode;
   vtkMRMLPlotViewNode*    ActiveMRMLPlotViewNode;
+
+  struct PausedView
+  {
+    std::string Name;
```
jcfr: Prefer QString since we are in a Qt class

Libs/MRML/Widgets/qMRMLLayoutManager_p.h
```diff
@@ -50,6 +50,9 @@
 // VTK includes
 #include <vtkSmartPointer.h>
 
+// std includes
```
jcfr:
```
// STD includes
```

Libs/MRML/Widgets/qMRMLLayoutManager.cxx
```diff
@@ -1330,3 +1330,81 @@ QWidget* qMRMLLayoutManager::viewWidget(vtkMRMLNode* viewNode) const
...
+    {
+    qCritical() << "No stored render pause state!";
+    return;
+    }
+
+  QString sliceNodeType = "vtkMRMLSliceNode";
+  QString threeDNodeType = "vtkMRMLViewNode";
+
+  std::vector<qMRMLLayoutManagerPrivate::PausedView> pausedViews = d->PausedViews.back();
+  foreach(const qMRMLLayoutManagerPrivate::PausedView pausedView, pausedViews)
```
jcfr: `const qMRMLLayoutManagerPrivate::PausedView pausedView` -> `const qMRMLLayoutManagerPrivate::PausedView& pausedView`

Libs/MRML/Widgets/qMRMLLayoutManager_p.h
jcfr: I wonder if `QCache` could be used here

jcfr (commented):

I wonder if the layout should be paused when starting the associated scene invoke `StartBatchProcessEvent` and resumed after `EndBatchProcessEvent` is invoked

jcfr:
I also wonder if instead of the boolean `WasPaused`, we should have an integer. The rendering would be resumed on a view only if the counter reach 0. an incremented each time the pause function is called ...

Libs/MRML/Widgets/qMRMLLayoutManager_p.h
Sunderlandkyl: @jcfr could you elaborate a little?
Do you mean if we switch WasPaused to an integer we could use QCache instead to track the paused state of all the views?

jcfr: > QCache vs integer 
One does not prevent the other. I just wanted to point out that QCache (or collection of them) could be used to manage state.

> use of integer

While i may miss some of the details, since a view can either be paused/resumed, may be simply keep track of the number of call to `pause()/resume() and increment an integer would be simpler ?

Sunderlandkyl:
I like the idea of using an integer flag instead of the bool, it could probably be simplified even further if this was introduced at the CTK level instead of Slicer.

@lassoan What do you think about changing the pauseRender(bool) function in ctkVTKAbstractView to pauseRender() resumeRender() and using an integer flag?

lassoan:
PauseRender(bool) is a bit safer than Pause/ResumeRender(), as there is a much better chance you recover from a missed ResumeRender().

We cannot do this in the layout manager because previous state of all views cannot be stored in a single int. We would either need to return a complex state object or store the state internally.

It is true that state can be stored internally in an array of [viewName, pauseCount, view Class] if we assume that multiple pauseRender(bool) calls are always properly nested (only the function that paused a view first can resume it), which I think is a safe assumption.

jcfr:
> changing the pauseRender(bool) function in ctkVTKAbstractView to pauseRender() resumeRender() and using an integer flag

It could indeed be done at the `ctkVTKAbstractView` level, that way the implementation in Slicer would be  simpler.

The API could be:

```
// Pause (or resume rendering) and return pause counter
int ctkVTKAbstractView::pauseRender(bool)
```

with 

* counter incremented by one if bool true, decremented if false
* if counter reaches 0, `scheduleRender` is called
* if counter was already 0, equivalent to `scheduleRender`

With the state stored in the view object, I think you could get ride of `currentPausedViews` and implementation would be something like this ... 


```
void qMRMLLayoutManager::pauseRender(bool pause)
{
  foreach(const QString& viewName, this->mrmlViewFactory("vtkMRMLSliceNode"))
    {
    ctkVTKAbstractView* view = qobject_cast<ctkVTKAbstractView*>(this->sliceWidget(viewName)->sliceView());
    view->pauseRender(pause)
    }
  foreach(const QString& viewName, this->mrmlViewFactory("vtkMRMLViewNode"))
    {
    ctkVTKAbstractView* view = qobject_cast<ctkVTKAbstractView*>(this->sliceWidget(viewName)->sliceView());
    view->pauseRender(pause)
    }
}
```

I think we could get ride of `qMRMLLayoutManager::resumeRender()` or we keep it, we could have `qMRMLLayoutManager::pauseRender()` instead.


Thinking more about it, what about the following `ctkVTKAbstractView` API make sense: 

```
int ctkVTKAbstractView::setRenderPaused(bool)
int ctkVTKAbstractView::pauseRender()
int ctkVTKAbstractView::resumeRender()
```

jcfr:
Nice 

> I've added the pause render functions to qSlicerApplication 

* could we keep the signal `pauseRenderRequested/resumeRenderRequested` specific to the layout manager ? 

* and instead of having two signals could we have only one that has a boolean as argument ?

jcfr:
That said, if anticipate having multiple layout manager, I could see the function at the application level as they would look over all layout manager can call the pause function ... 

Also, since the signal are emitted after the state is effectively changed, I don't think they should have the word `Requested` in them.

A signal like `renderPaused(bool)` would be more appropriate i think ... 

Thinking further ... since we always expect the render to be synchronously paused ... i don't think these signals are useful.

Sunderlandkyl:
Signal renamed to renderPaused(bool).

PR #1018, opened 2018-09-07
BUG: Report error if --python-script associated with non-existent file

jcfr:
This commit ensures an exception is raised when the python script does not
exist. When associated with the --exit-after-startup option, a non-zero status
code is now returned when Slicer exit.


jcfr:
Cc: @lassoan @ihnorton

jcfr:
Thanks for the review 

> Would it be possible to throw the error from inside this->corePythonManager()->executeFile()

Definitively, that makes sense.

For reference:

https://github.com/commontk/CTK/blob/6c0eddaaf8b306c5cf553e8421bca7a5640e380a/Libs/Scripting/Python/Core/ctkAbstractPythonManager.cpp#L303-L333

PR #1019, opened 2018-09-07
ENH: Update IO manager adding convenience function "openAddMarkupsDialog"

jcfr:


jcfr:
Thanks for the review.

Integrated in [r27398](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27398)

PR #1013, opened 2018-09-04
Update ITK 4.13.1+

dzenanz:

PR #1020, opened 2018-09-10
COMP: SetProcessDPIAware() needs WinUser.h for building on Windows

dzenanz:
Spawned off from #1013.


jcfr:
Thanks @dzenanz 

Integrated in [27406](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27406)

PR #1021, opened 2018-09-12
BUG: Update VTK to improve VR, Python Wrap, OpenGL perf and threading management

jcfr:
```
  Fix Slicer issue 4510 (VTK OpenGL2 Backend: Cropping is broken with GPU Volume rendering if depth peeling is enabled)
  See https://issues.slicer.org/view.php?id=4510

  Fix Slicer issue 4548 (VTK OpenGL2 Backend: Volume rendering artifacts at volume edges when clipped)
  See https://issues.slicer.org/view.php?id=4548

  Fix Slicer issue 4598 (VTK OpenGL2 Backend: Enabling Volume Rendering module breaks viewers)
  See https://issues.slicer.org/view.php?id=4598

  Fix Slicer issue 4513 (VTK OpenGL2 Backend: Problem with volume rendering surface gradient computation)
  See https://issues.slicer.org/view.php?id=4513

Thanks: Allison Vacanti <allison.vacanti[at]kitware.com>
Thanks: Andras Lasso <lasso[at]queensu.ca>
Thanks: David Gobbi <david.gobbi[at]gmail.com>
Thanks: Davide Punzo <punzodavide[at]hotmail.it>
Thanks: Ken Martin <ken.martin[at]kitware.com>
Thanks: Kyle Sunderland <sunderlandkyl[at]gmail.com>
Thanks: Lucas Gandel <lucas.gandel[at]kitware.com>
Thanks: Sankhesh Jhaveri <sankhesh.jhaveri[at]kitware.com>
Thanks: Mathieu Westphal <mathieu.westphal[at]kitware.com>

... and many more.



Highlighted list of VTK changes
---------------------------

* Volume Rendering Fixes

  * https://gitlab.kitware.com/vtk/vtk/merge_requests/4338/commits (from Allison Vacanti <allison.vacanti[at]kitware.com>)
    Fix volume clip bug and add regression test.

      When clipping a volume using in-shader clipping planes, it was possible for the starting point of the ray cast to
      lie beyond the data volume. The raycast code is written such that the first sample is always taken before testing
      termination criteria, and in these cases we would always take a single sample outside of the volume, leading to
      artifacts.

      Fixed this behavior by checking that the starting position calculated by AdjustSampleRangeForClipping is indeed
      inside of the volume bounds and aborting the raycast if it is not.

      The existing TestGPURayCastClipping test would have caught this, except that the vase.vti volume used for testing
      has all 0's at the boundaries, so the rendering was correct even with the edge-clamp repetition outside of the volume
      (the faulty samples always computed RGBA=vec4(0)). I replaced the vase.vti of this test with a wavelet with finite
      boundary values that will catch this problem if there's a regression.

  * https://gitlab.kitware.com/vtk/vtk/merge_requests/4573/commits (from Allison Vacanti <allison.vacanti[at]kitware.com>)
    Fix box widget face highlighting.

      The actual data was updated correctly in the BoxWidget
      code, but the wrong object was being marked as modified.

      The rendering code got smarter at some point and now
      only rebuilds VBOs when the corresponding cell array
      has been modified. Marking the entire polydata as modified
      is no longer sufficient to update the on-device
      representation of the highlighted face.

      This patch fixes the BoxWidget so that the active face is
      used for highlighting instead of getting "stuck" on an
      arbitrary face.

  * https://gitlab.kitware.com/vtk/vtk/merge_requests/4568 (from Allison Vacanti <allison.vacanti[at]kitware.com>)
    Volume peel widget flicker

      Correctly apply jittering in ClampToSampleLocation. We were reapplying it inconsistently in the DDP. This
      pass was updated also.

      Rearrange cast initialization code in DDP. The start point was adjusted for clipping *after* clamping
      to a sample position. This would move the effective sample positions to an inconsistent locations if clipping planes
      were present.

  * https://gitlab.kitware.com/vtk/vtk/merge_requests/4627 (from Sankhesh Jhaveri <sankhesh.jhaveri[at]kitware.com>)
    Introduce volume API to specify intensity for clipped voxels

      See Slicer issue 4513 (https://issues.slicer.org/view.php?id=4513)

      Specifying a fixed intensity value for voxels in clipped space provides
      the ability to change lighting computations at the clipped face.

  * https://gitlab.kitware.com/vtk/vtk/merge_requests/4614 (from Sankhesh Jhaveri <sankhesh.jhaveri[at]kitware.com>)
    VolumeMapper: Invoke update shader event each frame

      This allows applications to modify uniforms on the dynamically created and compiled shader program.


* Performance

  * https://gitlab.kitware.com/vtk/vtk/merge_requests/4175/commits (from David Gobbi <David Gobbi <david.gobbi[at]gmail.com>>)
    Resolve "Multithreader creates many unnecessary threads"

      Creating threads is expensive, so creating threads that do nothing should be avoided.

      - Limit number of threads in vtkImageHistogram
      - Remove redundant UpdateExtent calculation.
      - Use vector to allocate arrays instead of new
      - 17279: Limit num threads to num pieces.


  * https://gitlab.kitware.com/vtk/vtk/merge_requests/4175/commits (from Ken Martin <ken.martin[at]kitware.com>)
    hardware picking code: improve picking performance and memory footprint

      Significant rework of the hardware picking code.

      Previously each mapper had to make sure that the rendered
      colors were correct for point/cell etc ID during hardware picking
      and this required large datastructures and texture uploads
      to the GPU on each pick.

      Now it collects the color buffers and gives the mappers a chance to
      update them. This allows us to use gl_VertexId and gl_PrimitiveId
      directly in the shader and then if picked, allow the mapper to
      adjust the color buffer as needed. This allows us to avoid
      rebuilding the VBO and textures each time and avoids the memory
      footprint related to that.

  * https://gitlab.kitware.com/vtk/vtk/merge_requests/4616/commits (from Ken Martin <ken.martin[at]kitware.com>)
    Fix issue where picking happened with a bad context

      Make sure our OpenGL context is current when hardware picking


  * https://gitlab.kitware.com/vtk/vtk/merge_requests/4416/commits (from Ken Martin <ken.martin[at]kitware.com>)
    VolumeRendering: update volume rendering to share jitter and depth textures

  * https://gitlab.kitware.com/vtk/vtk/merge_requests/4389/commits (from Ken Martin <ken.martin[at]kitware.com>)
    some opengl performance fixes

      mostly cache a few items to avoid glGet* invocations

      remove an unused ivar form the image slice mapper

      make it so that the image slice mapper does not force
      an IBO rebuild every frame as typically it is not needed
      and a waste of CPU-GPU

* Python Wrapping

  * https://gitlab.kitware.com/vtk/vtk/merge_requests/4545/commits (from David Gobbi)
    Wrap std::vector for basic vector types

      Wraps std::vector arguments for Python where the value type is std::string or a fundamental
      numeric type (excluding 'char' due to its ambiguity)


* OpenVR

  * https://gitlab.kitware.com/vtk/vtk/merge_requests/4540/commits (from Ken Martin <ken.martin[at]kitware.com>)
    remove openvr factory

  * https://gitlab.kitware.com/vtk/vtk/merge_requests/4447/commits (from Ken Martin <ken.martin[at]kitware.com>)
    fix an issue with flying controller labels

      Labels would flyu off the controller while
      flying through VR due to position calc being done
      prior to the world translation.

  * https://gitlab.kitware.com/vtk/vtk/merge_requests/4480/commits (from Ken Martin <ken.martin[at]kitware.com>)
    better code for handling panel rotations

* QVTKOpenGLWidget

  * https://gitlab.kitware.com/vtk/vtk/merge_requests/4398/commits (from Kyle Sunderland <sunderlandkyl[at]gmail.com>)
    Add Qt gestures events to QVTKOpenGLWidget

      Qt gestures in QVTKOpenGLWidget are handled by QVTKInteractorAdapter, which will update the
      interactor and invoke corresponding events

  * https://gitlab.kitware.com/vtk/vtk/merge_requests/4400/commits (from Mathieu Westphal <mathieu.westphal[at]kitware.com>)
    Correct QVTKOpenGLWidget double click management, Correct MakeCurrent behavior

  * https://gitlab.kitware.com/vtk/vtk/merge_requests/4317/commits (from Mathieu Westphal <mathieu.westphal[at]kitware.com>)
    Add stero support

    - Improve QVTKOpenGLWidget and add QVTKOpenGLWindow class based on QOpenGLWindow
    - Fix viewport size before FXAA pass on OSX
    - Invoke WindowResizeEvent on vtkWindow::SetSize() call
    - Fix memory exception with checkerboard stereo
    - Add WindowStereoTypeChangedEvent fired by vtkRenderWindow

  * https://gitlab.kitware.com/vtk/vtk/merge_requests/4642/commits (from Jean-Christophe Fillion-Robin <jcfr[at]kitware.com>
    Rename QVTKOpenGLSimpleWidget to QVTKOpenGLNativeWidget

      Improve QVTKOpenGLWidget docstring

  * https://gitlab.kitware.com/vtk/vtk/merge_requests/4608/commits (from Jean-Christophe Fillion-Robin <jcfr[at]kitware.com>
    vtkRenderer::PickProp: Fix crash handling empty selection

* Widgets

  * https://gitlab.kitware.com/vtk/vtk/merge_requests/4493/commits (from Sankhesh Jhaveri <sankhesh.jhaveri[at]kitware.com>)
    Invoke DeletePointEvent before deleting vtkSeedWidget seed

* Chart

  * https://gitlab.kitware.com/vtk/vtk/merge_requests/4531/commits (from Davide Punzo <punzodavide[at]hotmail.it>)
    add single click selection in chartXY


Complete list of VTK changes
----------------------------

$ git shortlog 097c78f34e..9db353bb7 --no-merge

Alessandro Genova (6):
      Add a test for the axis->setPrecision(0) issue on Windows
      Increase the font size of the axis label
      Fix the Precision = 0 bug on MSVC compilers
      Address compiler warning
      Cleanup comments and unused includes
      Resolve last compiler warning

Allison Vacanti (23):
      Retain renderer to recompute right rendered rectangle.
      Populate DataArraySelection objects for vtm files.
      Fix volume clip bug and add regression test.
      Fix bug in mathtext path rendering.
      Set voxel type for OSPRay AMR volumes.
      Switch between ospNewMaterial2 over (deprecated) ospNewMaterial as needed.
      Use ospShutdown when ospInit fails.
      Support hexes and wedges in the OSPRay volume mapper.
      Select appropriate ospNewLight[2?] signature depending on version.
      Find embree3 library when building ospray.
      Make OSPRayMaterialLibrary aware of versioned materials.
      Silence some overly-cautious unused variable warnings.
      Rearrange cast initialization code in DDP.
      Simplify raycast jitter bookkeeping.
      Correctly apply jittering in ClampToSampleLocation.
      Add test for interactive clipping and volume peeling.
      Fix box widget face highlighting.
      Reduce memory usage in vtkm cell converters.
      Add check for timestep in the XML writer.
      Support appended data mode in the vtkXMLHTGWriter.
      Remove out of date comment.
      Add a random hyper tree grid source with parallel support.
      Add missing override.

Andreas Buykx (3):
      Fix issue #17211: crash on pick of composite data
      Backport fix for issue #17211 to release.
      Avoid unnecessary allocation of vtkInformation

Andrew Bauer (19):
      Fixing issues with the HyperTreeGrid and improving documentation.
      Hypertreegrid cleanup.
      Update the HyperTreeGridSource to work in parallel.
      Fixing vtkImageData::FindPoint() bug
      Fixing typo in error message.
      Fixing bug with unstructured grid ghost cell generator.
      Fixing issue with writing structured XML field data
      Updating to C++11 standards and fixing memory access bug.
      Make a non-MPI dependent version of vtkUnstructuredGridGhostCellsGenerator
      Adding in destructor override specifier for several classes
      Passing ghost information for the vtkPassArrays filter.
      Fixing Lagrange tet derivative
      Fixing Lagrange wedge and hex cell Derivative methods.
      Fix issue with single cell for ghost cells generator.
      Fixing issue with SGI's MPT MPI library
      Fix issue with empty input dataset.
      Changes to allow aggregation of topologically regular grids
      Fix issues with vtkDIYAggregateDataSetFilter
      General code cleanup for the Tecplot reader

Aron Helser (9):
      Remove MomentInvariants filter, prep to make remote
      Remote Modules, add MomentInvariants, and ParallelMI
      Remote: change ParallelMoments to use branch
      Remote: parallel code included in single module
      MomentInvariants final remote repo location and hash
      Testing: force remote module on for buildbot test
      Remove test cmake options - don't force on remote module
      Fix leak introduced in code rework.
      Update MomentInvariants remote module for test fix.

Ben Boeckel (1):
      libproj: update the import source listing

Berk Geveci (6):
      Introduced new composite datasets for partitioned data.
      Fixed bug in the XML structured data writer.
      Added support for reading/writing time as field data.
      Added API to array selection for disabling arrays by algorithms.
      New simplified interface for developing reader.
      FIX: Added missing module dependency.

Bill Lorensen (2):
      Add NumberToString conversion class.
      ENH: Bump WikiExamples Tag

Boris Basic (3):
      Fix depth peeling issues with polygonal handles and QuadricLODActor
      Fix vtkMFCWindow compilation
      Fix issues and performance problems in vtkMergePoints and vtkSMPMergePoints

Brad King (21):
      ParallelMomentInvariants: Exclude from all pending fixes
      cmake: Set policy CMP0075 explicitly to avoid warnings
      CTestCustom: Drop unused Utilities/ matches
      pre-commit: Add support for ExternalData SHA512 objects
      git-gitlab-push: Add support for ExternalData SHA512 objects
      ExternalData: Switch from MD5 to SHA512 for new content links
      ExternalData: Add URLs to fetch from Girder on data.kitware.com
      ExternalData: Drop URLs using midas3.kitware.com
      ExternalData: Update URLs for vtk.org to use https
      ArchiveTestingDataOnGirder: Switch default to VTK collection
      Rename module vtklibproj4 to vtklibproj
      libproj: update vtklibproj module library target name
      Geovis/Core: Remove deprecated tests that fail with PROJ 4.9.3
      cmake: declare support for CMake versions through 3.12
      Documentation: Update links to CMake ExternalData module docs
      SourceTarball.bash: Use https to download ExternalData from vtk.org
      SourceTarball.bash: Factor out ExternalData extension-to-algo map
      SourceTarball.bash: Update for ExternalData SHA512 objects
      SetupForDevelopment: Fix run through Windows file association
      xdmf2: Set policy CMP0077 explicitly to avoid warnings
      CTestCustom: Suppress warnings internal to openmpi headers

Carson Brownlee (7):
      fixing unnecessary volume updates in ospray on updates to tf.
      Fix material params with textures and add principled materials.
      Fix material params with textures and add principled materials.
      fixing unnecessary volume updates in ospray on updates to tf.
      hookup scalar opacity unit distance control into ospray
      adding shared volumes for ospray
      let ospray to respect choice of data or transfer function range

Cory Quammen (46):
      Use slightly clearer name for variables
      Stop drawing line segments outside image buffer bounds
      Use vtkMath::Min/Max instead of std::min/max
      Fix logic when tolerance is 0
      Clamp tolerance to range [0.0, VTK_DOUBLE_MAX]
      Set the field type for block selections
      Change algorithm for computing selections
      Rename vtkSelectionOperator to vtkSelector
      Change type of InsidednessArrayName member variable
      Add missing 'override' keyword
      Fix warning caused by redundant member variable
      Reorganize vtkInternals constructors
      Guard against modulo 0
      Remove nearly duplicate function
      Change shape of selection list array
      Fix vtkLocationSelector
      Find cells using vtkDataSet API instead of vtkCellLocator
      Change internal classes to have only the members they need
      Prevent crash if insidedness array is null or empty
      Guard against null array
      Ensure that the glyphNormalMatrix is used in shader
      Fix -Wparentheses warning
      Resolve -Wformat-truncation warnings
      Fix unsafe use of strncpy
      Fix -Wstringop-truncation warning
      Fix -Wstringop-truncation warning
      Fix -Wrestrict warning
      Fix -Wstringop-overflow warnings
      Avoid warnings about string truncation caused by string ops
      Improve vtkPConnectivityFilter performance
      Make filter work with vtkDataSet
      Move vtkPConnectivityFilter to vtkFiltersParallelGeometry
      Reduce amount of point data exchanged in vtkPConnectivityFilter
      Move doxygen figures for vtkPConnectivityFilter to proper directory
      Make vtkPConnectivityFilter an implementation of vtkConnectivityFilter
      Provide more functionality when data is on one rank
      Test against different point array types
      Fix -Wstringop-truncation warnings in vtkNIFTIImageWriter.cxx
      Test for threshold selection extraction on varied multiblock dataset
      Add some null guards to prevent crashes
      Clarify some warning messages in vtkValueSelector
      Added Schwarz counter for static variable
      Make copy constructor and operator= private
      Standardize vtkAxis::PrintSelf()
      Prevent compiler from generating copy constructor and assignment
      Fix logic for AdjustLowerBoundForLogPlot

Dan Lipsa (25):
      Fix SegY3D reader.
      Use SegYReader for both 2D and 3D data.
      Fix RequestInformation
      Use X,Y coordinates instead of inline,crossline.
      Use X, Y for SegY 3D.
      Test changed because of switch to XY coordinate system.
      Simplify computation.
      Use crossline, inline, sampleCountPerTrace order to match OpendTect.
      Clean up 2D reader and apply options to both readers.
      Add an option to create a StructuredGrid for a 3D SegY dataset.
      Update Progress.
      Flip data if increment is negative.
      Fix ParaView parameters and remove printouts.
      Make StructuredGrid the default for 3D data.
      Fix SegY warnings.
      Fix uninitialized variables.
      Fix warnings.
      Add SEGY to StandAlong group and warning fixes.
      Override and vector init warnings.
      Use a standard 4 byte unsigned int instead of long.
      Add cstdint include.
      Fix leaks on FileName.
      Add check for long lines and increase available space.
      Add deleted copy constructor and assignment op. to silence warning.
      Add ability to read from short integers.

David E. DeMarle (16):
      work around a bug where multiple threads delete image together
      silently ignore objects that lack requested valuepass array
      Increment version to VTK 8.1.1
      fix a memory leak and vtkstyle this for member
      Plug a vtk2osp leak and avoid a copy while exposing more formats.
      A change to the texture has to cause a redraw.
      Take advantage of zero copy textures in ospray.
      Fix a huge memory leak.
      dont leak the light either
      don't leak ospray buffer either
      use osp release not delete with osp volumes to prevent crashes
      suppress and fix comp warnings
      paraview client w/ osp would crash on connection to non osp server
      fix cases in which lut ranges were not sane
      fix parallel projection zoom ghost effect
      avoid exceptions to prevent a crash on Mac

David Gobbi (39):
      17265: Fix OpenGL2 vtkImageResliceMapper OOB issue
      17279: Limit num threads to num pieces.
      Use vector to allocate arrays instead of new
      Remove redundant UpdateExtent calculation.
      Limit number of threads in vtkImageHistogram
      17265: Fix OpenGL2 vtkImageResliceMapper OOB issue
      Fix thin border around vtkImageResliceMapper
      Backport 'reslicemapper-border' to release branch.
      Fix misuse of purportedly boolean Border.
      Fix documentation, use K for slice index.
      Fix typo in comments, xform to sform
      Expand the description of vtkImageMapToWindowLevelColors
      17317: Truncate NumberOfIds in Resize().
      Make IntersectWithLine return 0 or 1.
      Constify the SetWindowInfo parameter
      Use const char pointer for AddIconType
      Use const char pointer for array name APIs.
      Use const char pointer for filenames.
      Fix constness of keysym argument
      Use const char pointer for field name
      Fix compilation issue due to Python3.7 API change
      Fix char pointer constness for wrapper built-in methods
      Safe wrapping of non-const char pointer arguments
      Use size_t for array size in wrappers
      Use memcpy, memcmp for efficient array wrapping
      Add zerocopy check to char pointer
      When parsing mangled types, check for 'St'
      Wrap std::vector<T> arguments for numeric types
      Wrap vectors of std::string in Python
      Add test for wrapping std::vector arguments
      Expand typedefs within the correct scope
      Expand typedefs for arg types in correct scope
      Improve scoping for vtkParseHierarchy
      Handle typedefs in template parameters.
      Fix warning by replacing strncpy with strcpy
      Add std::vector help to Python README.md
      Implement clamp with ternary for better optimization
      Fix wrapping multiple inheritance crash
      Wrapper build failed for pointers to vectors.

David Thompson (2):
      Bump the VTK-m that VTK uses forward; fixes for API changes.
      Add support for 2 new elements to Exodus files.

Davide Punzo (4):
      ENH: added single click selection in chartXY
      BUG: fix for selection vtkChart test failing
      STYLE: fix formatting vtkChartXY
      BUG: fix warning in vtkChart GetClickActionToButton

Florian CHEVASSU (2):
      Fix vtkDistributedPointCloudFilter synchronisation
      Fix bug in OptimizeBoundingBox

Francois Budin (1):
      Check that context exists before trying to pop context.

GitSetup Upstream (1):
      GitSetup 2018-06-26 (1ed3dc31)

HDF Upstream (1):
      hdf5 2018-06-04 (247b2ae5)

Haocheng LIU (10):
      Expose vtkm point elevation filter
      Remove an unnecessary include in vtkmPointElevation
      Expose vtkm histogram filter
      Bump the VTK-m that VTK uses forward
      Bump vtk-m forward
      Expose the vtkm Probe filter
      Expose vtkm NDHistogram filter into vtk
      Add casts to vtkError/Warning/Debug Macros
      Expose vtkmWarpScalar filter into vtk
      Expose vtkm's WarpVector filter to vtk

Jean-Christophe Fillion-Robin (3):
      vtkRenderer::PickProp: Fix crash handling empty selection
      Rename QVTKOpenGLSimpleWidget to QVTKOpenGLNativeWidget
      doc: Improve QVTKOpenGLWidget docstring

Jefferson Amstutz (3):
      only use OSPRay API handle types instead of opaque pointers
      vtk style fixes
      use target_include_directories() to propogate OSPRay header dependencies

Jerome Dubois (1):
      Added the ability to set character separator for vtkLabeledDataMapper

Joachim Pouderoux (9):
      Fix destructor override warning
      Fix a GRAM leak in QVTKOpenGLWindow due to creation of offscreen contexts
      Make sure offscreen surface is always destroyed
      Fix OpenFOAMReader that infer wronh wedge cell type on some polyhedron.
      NetCDFReader now provides time&data units as field data arrays.
      Cleaning pass
      Clean & refactorize
      Improve TestDistributedPointCloudFilter
      Add a null guard to prevent crash on vtkExtractSelection

John Tourtellott (1):
      Added GDAL reprojection filter from vtkMap

Jon Haitz Legarreta (1):
      DOC: Update the Gitlab banner to new VTK logo design.

Juliano Masson (2):
      Update face texcoord and vertex color support
      Update face texcoord and vertex color support

Julien Schueller (1):
      Cast in vtkGetLibraryPathForSymbol macro

KWSys Upstream (2):
      KWSys 2018-04-23 (aa0f06ec)
      KWSys 2018-04-24 (85c211d6)

Ken Martin (94):
      clean up the state API and fix Qt widget
      do not use glDGetDouble on OpenGL ES
      add an option for a floor to the vtkSkybox class
      fix an access violation happening in StructureGridGeom filter
      fix uninitialized index in DataArray.cxx
      fix case where depth buffer was underflowing
      fix crash when scene has scalar values
      remove failihg part of test
      improve the picking on the pointgaussian mapper
      significantly improve performance for many blocks
      fix an access violation happening in StructureGridGeom filter
      add better support for using mesa when needed on windows
      add support for writing larger unstructured grids
      make sure we configure CTestCustom.cmake.in
      fix handling of large textures
      dataset mapper was not forwarding relative coincident parameters
      remove old 32bit simulator for ios
      add an executable to do opengl detection
      fix issues with the array calc and string arrays
      fixed case where the array name differed from
      Remove vtkProbeOpebGLSupport executable from Mobile
      comparison was being done for the wrong case
      fix ios build issues
      remove remaining opengl32 status methods
      fix buffer overrun
      cleanup and improve sphere and stick mappers in opengl
      fix a bug in image difference producing bad results
      add back in some valid images that were lost
      add another valid image for text differences
      fix an unused var warning in this class
      rev wikiexamples to ge a new valid image
      test was one of the slowest so make faster
      fix a ABR error in the STLWriter for binary header
      update a test and add a valid image for hythloth
      add a suppression for doxygen generation
      add support for composite datasets and picking of them
      quiet all doxygen warnings
      some opengl performance fixes
      remove an unused variable to queit a warning
      typo was  flipping the blend state
      change two methods to be virtual
      fixes for point gaussian
      add in glBlendEquationSeparate and defaults
      update volume rendering to share jitter and depth textures
      quiet a qt cppcheck related warning
      add more doxygen warning supp for ike
      use shared noise texture and cleanup tests
      update wiki examples
      fix an issue with flying controiller labels
      fix a few issues with skybox
      improve picking performance and memory footprint
      fix a bug in polydatamapper2D when scalarvisibility changes
      fix some warnings exposed by newer compiler
      try suppressing cppcheck warning about unit members
      better code for handling panel rotations
      remove warning supp that should be in cppcheck file
      fix some reported memory leaks
      fix some memory leaks related to xdmf
      address a gcc warning related to youngsmaterialinterface
      gcc warns on strncpy that is short of the term char
      fix a few more strncpy gcc warnings
      fix one more spot where it was leaking
      fix two spots that were missing const qualifiers
      suppress a couple warnings that remain
      use hardwareselector to perform picking
      fix a few opengl issues reported by scanbuild
      fix memory leak in selector
      fix minor ospray compiler warning
      fix two divide by zero spots
      fix case where nullptr passed to memcpy
      fix a crash in the jpeg reader
      add a bunch more debugging with VTK_REPORT_OPENGL_ERRORS
      fix some dynamic analysis issues
      Try to gracefully handle degenrate cube axes
      remove openvr factory
      Add a couple reserves to reduce memory allocs
      Improve performance of SDDP
      Fix an issue where we were relying on Qt's isCurrent
      switch to map for camera poses in openvr
      Fix memory leak and read violation
      Fix a reported crash in SupportsOpenGL
      Fix incorrect micro second value
      add code for ospray to handle rotated normals
      Fix issue where picking happened with a bad context
      rev VTKExamples to 8c4ceec93b11b9b64db4e57ee2d644d95f8f8115
      Add order independent translucent pass
      fix issue with multiple contexts
      fix issue with input data changing
      add new header to list of headers for install
      improve handling of some segy files
      fix issue with glViewport on OSX
      Fix point picking used with cell scalars
      fix cmake warning from CMP0072
      shorten test from 10 seconds to 4 seconds

Kitware Robot (2):
      ExternalData: Convert content links from MD5 to SHA512
      ExternalData: Convert more content links from MD5 to SHA512

Kyle Edwards (1):
      Add find-module for double-conversion

Kyle Sunderland (1):
      Add Qt gestures events to QVTKOpenGLWidget

Lucas Gandel (6):
      Add WindowStereoTypeChangedEvent fired by vtkRenderWindow
      Fix memory exception with checkerboard stereo
      Invoke WindowResizeEvent on vtkWindow::SetSize() call
      Fix viewport size before FXAA pass on OSX
      Improve QVTKOpenGLWidget and add QVTKOpenGLWindow class based on QOpenGLWindow
      Make context current before resizing to VTK window size

Maik Froechtenicht (5):
      Fix for vtkTriangle::TrianglesIntersect() Issue #17092
      Undo copy vars for handover, format
      Fix for vtkTriangle::TrianglesIntersect() Issue #17092
      Fix for vtkTriangle::TrianglesIntersect() Issue #17092
      Fix for CoplanarTrianglesIntersect

Marcus D. Hanwell (2):
      Update incorrect file name in test
      Consistently use vtkNew, remove C-style cast

Mark Olesen (1):
      ENH: pass through OBJ groups as GroupIds (fixes #17225)

Mathieu Westphal (18):
      Set vtkInteractionStylePython-TestStyleRubberBandZoomPerspective to run serial
      Adding new StepTime related features to LPT
      Fix vtkContourHelper CellData copy over
      Correct QVTKOpenGLWidget double click management
      Correcting MakeCurrent behavior
      vtkOpenTURNS filters compatibility with OpenTURNS 1.11
      Removing backwards compatibility and hiding warnings
      in vtkScatterPlotMatrix, redraw active plot after animation
      Exposing cursor in QVTKOpenGLWidget and some doc fixes
      Fixing TestQVTKOpenGLSimpleWidgetWithDisabledInteractor
      mirror QVTKOpenGLWidget api in QVTKOpenGLSimpleWidget
      Implement a work around for a Qt Bug
      Set a property so KDE will never grab the QVTKOpenGLWidget
      Fix the OpenGL state glViewport in QVTKOpenGLSimpleWidget::recreateFBO
      Corrects !4556 so MakeCurrent check the correct isCurrent
      Revert "Corrects !4556 so MakeCurrent check the correct isCurrent"
      Add Reset methods to OpenGLState and use it on MacOS QVTKOpenGLWindow
      Adding a cmake_minimum_required on FindOpenMP.cmake

Matt McCormick (1):
      ArchiveTestingDataOnGirder: Add script to upload to data.kitware.com

Max Zeyen (1):
      CIEDE2000: update to defaults

Menno Deij - van Rijswijk (3):
      Fixed an indexing issue for polyhedron clipping.
      Fix for duplicate merge.
      Updated vtkTecplotReader to handle ZONETYPE

Michael Migliore (12):
      Add option in OpenFOAM reader to copy data to cell zones
      Add capacity to add a point outside of the current polyline
      Add tone mapping render pass
      Fix fxaa with float fbo
      Improve performance of surface filter for unstructured grid
      Fix build error on OpenGL ES
      Transfer user shader replacement from composite mapper to helpers
      Add isosurface support in OSPRay
      Fix Glyph3D Mapper LOD culling
      Fix sphere mapper when the sphere is translucent
      Allow BuildCellTextures overriding
      Add all parameters for Principled material of OSPRay

Nicolas Vuaille (7):
      Introduce Molecule Append filter
      Molecule to polydata conversion
      Improve pointSetToMolecule : optionally convert lines into bonds
      Fix point picking
      Fix molecule initialization
      Opacity for nan color
      Fix extract selection for non supported object

Norman Juchler (2):
      Fix for QVTKRenderWindowInteractor for better support of retina displays
      Style improvement for my fix in QVTKRenderWindowInteractor

ParaView Developers (1):
      Add vtkIntegrateAttributes

Peter Franz (2):
      Customize vtkAxis
      Add missing override statement to remove test warnings.

Proj Upstream (1):
      libproj 2018-05-09 (498671c3)

Robert Maynard (8):
      Update Accelerators/Vtkm to not produce any cpp check warnings
      VTK now properly detects VTK-m after new CMake infrastructure.
      Don't search for VTK-m if VTK-m targets already exist
      vtkmProbe includes the required std headers it needs
      VTK-m Accelerator supports the OpenMP backend
      VTK-m OpenMP support will be enabled if VTKSMP is OpenMP.
      Update VTK FindOpenMP to match the latest version from VTK-m.
      Remove CMake debug messages when enabling OpenMP for VTK-m.

Sankhesh Jhaveri (25):
      Module Geovis GDAL
      Fix compile warnings for module GeovisGDAL
      Volume 2D transfer function issue causing undefined variable shader errors
      Fixed raster reprojection filter for celldata and GDAL 2
      Include GDAL header directory for module vtkGeovisGDAL
      Fix unused parameter and unused variable warnings in vtkGeovisGDAL
      Add API for GDAL reader to provide projection string in PROJ.4 format
      Ability to directly provide the PROJ.4 string to vtkGeoProjection
      Fix crash due to unitialized pointer
      Free up GDAL created memory in a GDAL safe way
      Fix unused variable warnings in vtkGeovisGDAL module
      Fix compiler warning in GDAL vector reader
      Pass coincident topology parameters to glyph3d helper
      Deprecate vtkGeo* classes
      Add deprecation logic to tests for deprecated vtkGeo classes
      Remove unused test cases from vtkViewsGeovis
      Fix bug where re-enabling seed widget wouldn't move existing seeds
      Invoke DeletePointEvent before deleting vtkSeedWidget seed
      Make some orientation marker widget methods virtual
      VolumeMapper: Invoke update shader event each frame
      Fix vtkImageMapper's rendering of two component unsigned char data
      Introduce volume API to specify intensity for clipped voxels
      Added test for volume clipped intensity
      Add flag to compile with OpenGL ES API
      Disable vtkProbeOpenGLVersion when compiling against OpenGL-ES

Scott Wittenburg (2):
      Keep track of threads/workers and kill them explicitly
      vtkweb: Try to handle all possible vtkRenderWindow subtypes

Seacas Upstream (1):
      exodusII 2018-05-07 (e4638682)

Sean McBride (11):
      Added a #warning on code that I’m pretty sure is dead
      General code review of vtkMath: trivial & conservative
      Replace some int with vtkTypeBool (if public) or bool (if private)
      Added const to various vtkMath API
      Code review of vtkMath clamping
      Deprecated non-thread-safe colour conversion functions
      Introduce VTK's first use of C++11's std::lround() (in a test)
      Added missing overrides found by cppcheck 1.84
      Fixed misc issues found by cppcheck 1.84
      Fixed more cppcheck 1.84 warnings
      Fixed new clang warning about assignment to itself

Sebastien Jourdain (7):
      Add ThreadedImageWriter
      Improve ThreadedWriter test
      Try to improve TestThreadedWriter
      Remove ThreadedWriter timing failure
      DOC: Fix typo in HTG comments
      Add VERA out reader
      fix(vtkVeraOutReader): Memory leak

Shawn Waldon (24):
      Add vtkFrustumSelector a selection operator for frustum selection
      vtkExtractSelection2: compiles but only partly implemented
      vtkExtractSelection2: selection with frustums compiles
      Move selection operator to Filters/Extraction
      Add test for combining extraction nodes via expression
      Refactor vtkSelectionOperator to use vtkDataObject API
      Change selection operator to return a boolean success value
      Handle selecting points with containing cells
      Add missing include
      Give the selection array the expected name
      Fix compile error
      Update test to set association for cell selection
      Fix typos
      Ensure output array has correct name
      Handle INVERSE property on selection nodes
      Handle legacy 1-component 2 tuple threshold selection arrays
      Change TestExtraction to test both old and new filters
      vtkExtractSelection2: add support for extracting rows from a vtkTable
      Add test of extracting rows
      Fix dashboard errors and warnings
      Deprecate and rename old vtkExtractSelection filter
      Add API documentation
      Unmark New() as legacy
      libproj: change branch to import so old module system is still used

Shreeraj Jadhav (7):
      Fix For Wrong Field Name Passed To PointAverage Filter
      ComputeGradient Flag Not Checked For FastApprox=off
      Add ability to extract solid 3D output from vtkExtractCTHPart
      vtkImageBlend bug fix for compound mode
      FunctionParse Wrong Error Messages
      Clip Polyhedra Crashes for N-faced Elements
      Axes Grid Lines Disappear When Zoomed-In

Steven Hahn (3):
      Use public instead of protected member functions
      Use public instead of protected member functions
      Avoid integer overflow when clipping large volumes.

Sujin Philip (6):
      Update VTK-m
      Update to latest VTK-m master
      Fix some warnings in Accelerators/Vtkm
      Improve cell finding in vtkProbeFilter
      Fix vtkAMRSliceFilter
      Fix Filters/AMR Python Tests

T.J. Corona (7):
      Add tests for triangle intersection fixes.
      Add a cell validation filter.
      vtkLagrangeTriangle: Fix point ordering for edges.
      vtkCellValidator: fix warning about virtual destructor override.
      vtkTessellatorFilter: Warn and skip data arrays that exceed size limit
      Fix Lagrange wedge shape fns and derivatives
      Add documentation for Lagrange tetra, triangle

Tim Rowley (2):
      Add cell valued unstructured volume support to OSPRay mapper
      Add cell valued unstructured volume support to OSPRay mapper

Todd Martin (13):
      Add custom wrap compiler directive
      Fix definition name
      Add macro size hints
      Eliminate Windows name mangling workaround for GCCXML wrap
      Remove redundant code
      Add const qualifier to [in] arrays
      Add missing const qualifier to [in] arrays
      Add size hints and const to [in] arrays
      Fix declaration types
      Update source code documentation in header
      Add const qualifier to [in] arrays
      Fix const array of array declarations
      Change closestPoint to static array in EvaluatePosition()

Utkarsh Ayachit (59):
      Handle CFG files with other statements.
      put code blocks to simplify debugging
      Fix a shadowed var warning.
      extend grammar to include '-' in file names.
      updates to fix position file motions.
      Ensures output has as many rows as timesteps.
      Use `vtkCharArray` for mask array
      Add new doubleconversion 3rd party module.
      Update vtkNumberToString API to overcome build issues
      use correct package name when locating sustem doubleconversion.
      update thirdparty repo to point to the official fork.
      avoid segfault during app exit.
      Extend vtkSelection to support expressions.
      add missing return
      Add vtkValueSelector
      use vtkValueSelector in vtkExtractSelection2 (and its test)
      Add vtkLocationSelector to extract selected locations.
      Update TestExtractionExpression to test vtkLocationSelector.
      Update TestExtractionExpression to test threshold filter.
      fixes after API changes
      Misc vtkExtractSelection2 improvements
      simplify `vtkExtractSelection2::GetAttributeTypeOfSelection`
      exclude 3rd party warnings.
      vtkNumberToString doesn't need to delete assignment operator.
      add missing include.
      Don't use VTK_LEGACY with out set/get macros.
      vtkExtractSelection: remove unreachable code.
      fix missing `std::to_string` error
      fix warnings: remove testing of legacy filter
      fix missing `override` warnings
      warnings: fix `assignment operator could not be generated`
      docs: make `Return value: New reference.` explicit.
      fix warnings.
      exclude ospray warnings
      finally fix the implicit constructor/operator warnings.
      add MSVC 2013 exclusion.
      disable TestNumberToString on MSVC 2013.
      add tests to read/write multiblock with partial arrays
      Handle mismatch is chosen and available arrays.
      TestXMLWriteRead: fix cppcheck warnings.
      vtkPointSetToMoleculeFilter: fix cppcheck warnings
      init sys.argv in embedded Python environment.
      Add `vtkPythonScopeGilEnsurer` to vtkPythonAlgorithm.
      vtkAlgorithm.py: add missing import.
      vtkPythonAlgorithm: cleanup exception reporting code.
      python3: fix segault on exit when using GIL.
      Properly handle vtkSelectionNode::CONTAINING_CELLS()
      docs: fix incorrect docs for D3.
      Refactor vtkDataSetAttributes::FieldList
      Simplify vtkAppendPolyData.
      Make vtkAppendFilter respect copy flags for attributes.
      Simply `vtkAppendFilter` to use `FieldList`.
      Fix handling of global cell ids.
      qt: fix swapping buffer without making correct context current.
      add missing string include.
      add test for TestIntegrateAttributes.
      misc motionFX fixes
      add null guard to avoid segfault.
      improve vtkPDataSetReader reading of metadata

Will Schroeder (10):
      Fast 2D clipper for image data
      Avoid reading into undefined heap; optimizing comparison on unsigned char
      Input is now generalized to vtkPointSet
      Removed unnecessary variables
      Faster computation of point inside/outside
      New regression test images
      SMP Parallel algorithm for Voronoi tessellation
      Address Voronoi-related dashboard warnings
      Cleaned up confusing and overly complex code
      vtkFlyingEdges2D: Properly color multiple isocontour values

Yuri (1):
      Correction to when stat64 is used: FreeBSD doesn't use stat64.

double-conversion Upstream (2):
      doubleconversion 2018-03-29 (0e0b255d)
      doubleconversion 2018-05-11 (f3ac1585)

luz.paz (13):
      Typos: source and comments
      Removed duplicate line
      Add more misc. typos
      Added more typos
      Added more typos
      Follow-up typo
      More follow-up typos
      Last minute typos found
      Misc. typos
      Misc. typos
      Source typo fixes
      Misc. typos
      Misc. typo fixes
```


jcfr:
@lassoan This is ready for integration, if you can try building on Windows .. and confirm I suggest we merge.

lassoan:
Thanks a lot. I can start a Windows build in about 2 hours.

jcfr:
Great. In the mean time, I am also testing the build on the windows factory ..

jcfr:
Windows failed .. there are still issue with the wrapping of std::vector it seems. David Gobbi already fixed an other issue this morning .. but I will revert the changes I think.

lassoan:
I've fixed the std vector string * issue by replacing it by std vector string &. I think it is acceptable temporarily until the final fix in VTK.

jcfr:
> I think it is acceptable temporarily until the final fix in VTK.

This was fixed in VTK already (after I reported it this morning)... see commit "Wrapper build failed for pointers to vectors" here https://github.com/Slicer/VTK/commits/slicer-v9.0.0-2018-09-11-83dddfb943

lassoan:
I had to add "#include <vtkRenderWindowInteractor.h>" to CTK\Libs\Visualization\VTK\Widgets\ctkVTKDiscretizableColorTransferWidget.cpp, otherwise I got error that vtkRenderWindowInteractor class was undefined. Did you have the same issue?

jcfr:
My build didn't go that far on windows because it wasn't clean ... I then deleted the VTK* and CTK* and re-launched it. 

But is strange I didn't get the error on linux ..

jcfr:
> I had to add "#include <vtkRenderWindowInteractor.h>" to CTK\Libs\Visualization\VTK\Widgets\ctkVTKDiscretizableColorTransferWidget.cpp, otherwise I got error that vtkRenderWindowInteractor class was undefined.

I couldn't reproduce this.

But I got 

```
"D:\D\P\Slicer-0-build\Slicer.vcxproj" (default target) (36) ->
(CustomBuild target) ->
  D:\D\P\Slicer-0\Base\QTApp\qSlicerApplicationHelper.cxx(110): error C3861: 'SetProcessDPIAware': identifier
 not found [D:\D\P\Slicer-0-build\Slicer-build\Base\QTApp\qSlicerBaseQTApp.vcxproj] [D:\D\P\Slicer-0-build\Sl
icer.vcxproj]
  D:\D\P\Slicer-0\Base\QTApp\qSlicerApplicationHelper.cxx(110): error C3861: 'SetProcessDPIAware': identifier
 not found [D:\D\P\Slicer-0-build\Slicer-build\Base\QTApp\qSlicerBaseQTApp.vcxproj] [D:\D\P\Slicer-0-build\Sl
icer.vcxproj]

    1043 Warning(s)
```

I will then apply the patch proposed in https://github.com/Slicer/Slicer/pull/1020

jcfr:
Closing. Integrated in [r27405](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27405)

jcfr:
@lassoan  Since I need to use the factory to release an other software ... so far I only triggered the build + packaging on Windows and Linux. There are already available for download.

I will take care of macOS a little later.

Once I am done releasing the other software (SlicerSALT), I will also trigger the build of Slicer extensions.

lassoan:
Sounds good, thank you. I had to do a clean Slicer build, that's why I couldn't provide test results yet. I expect it'll work just fine but if not I'll let you know.

jcfr:
Sounds good. If you have a chance to download the windows nightly build, It would be nice to look into https://issues.slicer.org/view.php?id=4600 and provide input for @allisonvacanti  . Merci

PR #1023, opened 2018-09-19
BUG: Fix passing module paths containing single quote

jamesobutler:
Fixes https://issues.slicer.org/view.php?id=4592

This is a proposed fix and will need help merging into master by a main Slicer developer. Hopefully this can be included before the upcoming Slicer release.

@lassoan or @jcfr could you take a look at this? Thanks!

### DIFF
--- a/Base/QTCore/qSlicerCorePythonManager.h
+++ b/Base/QTCore/qSlicerCorePythonManager.h
@@ -50,6 +50,9 @@ public:
   /// Append \a paths to \a sys.path
   void appendPythonPaths(const QStringList& paths);
 
+  /// Convert a string to a safe python string literal using single quote (') for the literal
+  virtual QString toPythonStringLiteral(QString path);
+
   /// List of directories containing Python modules.
   virtual QStringList pythonPaths();
 
--- a/Base/QTCore/qSlicerCorePythonManager.h
+++ b/Base/QTCore/qSlicerCorePythonManager.h
@@ -34,27 +34,38 @@ class Q_SLICER_BASE_QTCORE_EXPORT qSlicerCorePythonManager : public ctkAbstractP
 {
   Q_OBJECT
 
+  /// List of directories containing Python modules.
+  Q_PROPERTY(QStringList pythonPaths READ pythonPaths)
+
 public:
   typedef ctkAbstractPythonManager Superclass;
   qSlicerCorePythonManager(QObject* parent=0);
   ~qSlicerCorePythonManager();
 
   /// Convenient function allowing to add a VTK object to the interpreter main module
-  void addVTKObjectToPythonMain(const QString& name, vtkObject * object);
+  Q_INVOKABLE void addVTKObjectToPythonMain(const QString& name, vtkObject * object);
 
   /// Append \a path to \a sys.path
   /// \todo Add these methods to ctkAbstractPythonManager
   /// \sa appendPythonPaths
-  void appendPythonPath(const QString& path);
+  Q_INVOKABLE void appendPythonPath(const QString& path);
 
   /// Append \a paths to \a sys.path
-  void appendPythonPaths(const QStringList& paths);
-
-  /// Convert a string to a safe python string literal using single quote (') for the literal
-  virtual QString toPythonStringLiteral(QString path);
+  Q_INVOKABLE void appendPythonPaths(const QStringList& paths);
 
   /// List of directories containing Python modules.
   virtual QStringList pythonPaths();
+
+  /// Convert a string to a safe python string literal.
+  /// Backslash, single-quote characters are escaped
+  /// and the string is enclosed between single quotes.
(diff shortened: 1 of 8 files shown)


lassoan:
Nice catch, thanks for proposing a fix.

It would be better to address the root cause of the issue - special characters in input string - rather than just slightly tuning quotation style.

Could you add a `toPythonStringLiteral` helper function (maybe a method or static function in qSlicerCorePythonManager) that would convert a string to a safe Python string literal? For now, you could encode only the `'` character by replacing it by `\'`. In the future, we could add more [escape sequence replacements](https://docs.python.org/2.0/ref/strings.html) as needed.

We can leave the CMakeLists.txt in that test as is, just add a comment there that explains why we had to use `\"` instead of `'`.

jamesobutler:
@lassoan I've pushed a second commit so you can see the changes I've made.  This should include the updates as you suggested.

lassoan:
Thank you, it is almost perfect. The only thing I would change is to return with a complete string literal (including quotation marks). So that you can use it like this:

    this->executeString(QString("import sys; sys.path.append(%1); del sys").arg(this->toPythonStringLiteral(path))); 

and

    QString("import imp;imp.load_source(%2, %1);del imp;") 
          .arg(pythonManager->toPythonStringLiteral(fileName))
          .arg(pythonManager->toPythonStringLiteral(moduleName)).toLatin1()

and

          QString("with open(%1, 'rb') as f:import imp;imp.load_module(%2, f, %1, ('.pyc', 'rb', 2));del imp") 
          .arg(pythonManager->toPythonStringLiteral(fileName))
          .arg(pythonManager->toPythonStringLiteral(moduleName)).toLatin1(),

Base/QTCore/qSlicerCorePythonManager.h
```diff
@@ -50,6 +50,9 @@ class Q_SLICER_BASE_QTCORE_EXPORT qSlicerCorePythonManager : public ctkAbstractP
   /// Append \a paths to \a sys.path
   void appendPythonPaths(const QStringList& paths);
 
+  // /// Convert a string to a safe python string literal
```
lassoan: Keep only 3 slashes instead of 5.
Describe which quote is used for the literal (`"` or `'`). Probably `'` would be better (later quote character can be made configurable).

jamesobutler:
I've pushed an update, but for some reason scripted modules in the additional module paths aren't getting loaded.  I keep getting this [error](https://github.com/jamesobutler/Slicer/blob/fc3fcceb2a6007899a19360646923ee5200eb58c/Base/QTCore/qSlicerScriptedUtils.cxx#L60-L61) from within `qSlicerScriptedUtils` though the output in that message looks fine.  Not sure what I'm doing wrong here.

lassoan:
Thank you! I've made some small changes and added tests. I'll double check that all tests pass and push the changes here so that you can review.

lassoan:
@jamesobutler please have a look at and test the latest commit. If it is OK for you then let me know and I merge the pull request.

jamesobutler:
Tested on the original computer that had the problem of a single quote in the user name and it appears additional module paths load correctly.  Thanks for improving this and adding the test cases!

With your updates I was trying to simulate the single quote in the path issue on my system by after installing an extension moving the contents of %AppData%\NA-MIC\Extensions-27422 to "C:\D\James's\" directory and updating the "Additional modules paths" in settings, but for some reason it can't load the qt-loadable-modules in those paths. Maybe just an issue with how I copied the files. See attached log with paths for SlicerOpenIGTLink extension.
[Slicer_27422_20180921_142049.log](https://github.com/Slicer/Slicer/files/2406542/Slicer_27422_20180921_142049.log)

lassoan:
Thanks for testing. SlicerOpenIGTLink may rely on DLLs in the bin folder (igtlio*.dll, OpenIGTLink.dll, igtlutil.dll), so maybe paths for these additional binaries are not set up correctly. Could you try if it works for non-superbuild extensions and Python-only extensions?

jamesobutler:
So I installed SegmentEditorExtraEffects (and its dependency MarkupsToModel) and moved them to the new path with the single quote and updated the additional module paths.

They Python-only SegmentEditorExtraEffects appears to have imported all correctly.  MarkupsToModel a qt-loadable-module with qSlicerMarkupsToModel*.dll and vtkSlicerMarkupsToModel*.dll also appears to have imported correctly.  So you might be right that it might be a problem specific to the DLLs in SlicerOpenIGTLink.

lassoan:
I've tested extension install and run with the user name _O'Test_ and it all worked well.

Integrated in https://github.com/Slicer/Slicer/commit/e40af17987ff03931930d3c9ba198a0738f37f86

I've set you as author in the commit in git, but it seems that git-svn does not propagate author information information from git to svn.

Thanks for your contribution!

PR #1024, opened 2018-09-26
Allow frames-per-second specification as alternative to video length.

torquil:
In the ScreenCapture extension, it has been necessary to specify the
output video length. The number of input frames is also specified.
From this the FPS value using by FFMPEG has been computed by Slicer.

However, in some cases it is necessary to specify the input FPS of
a series of images, e.g. X-ray fluoroscopy frames taken at particular
FPS. Since only a finite number of digits is allowed in the video
length specification, it is not always possible to specify a video
length that will give the exact input FPS for a given number of
input frames. Sometimes the exact FPS is important, e.g. in cases
where synchronization with other data is used.

This change allows the user to choose between video length or input
frame FPS to resolve this problem.


lassoan:
Thank you for your contribution. The feature has been merged. I've updated the GUI to keep both video length and FPS (and number of frames) displayed at the same time (and as one slider is moved, update the others accordingly) to make it simpler to use.

PR #1026, opened 2018-10-04
ENH: Report insufficient OpenGL support instead of crashing

lassoan:
Use same mechanism as ParaView to detect if VTK rendering capabilities are sufficient.
If not, then a popup is displayed, which allows the user to close the application, ignore the problem (continue to launch the application), or retry (do a delayed restart, to allow reconnection through remote desktop - see https://www.slicer.org/wiki/Documentation/Nightly/FAQ/General#Slicer_does_not_start).

OpenGL capabilities are displayed in Details section of the popup.

image

Details section shows OpenGL info and link to relevant FAQ:

image

Delayed restart in case user clicks "Retry":

image

Fixes https://issues.slicer.org/view.php?id=4252


Base/QTApp/qSlicerApplicationHelper.txx
```diff
@@ -93,6 +93,17 @@ void qSlicerApplicationHelper::postInitializeApplication(
   enableMainWindow = enableMainWindow && !app.commandOptions()->runPythonAndExit();
   bool showSplashScreen = !app.commandOptions()->noSplash() && enableMainWindow;
 
+
+  if (enableMainWindow && !app.testAttribute(qSlicerCoreApplication::AA_EnableTesting))
```
jcfr: should we check independently of the main window ?

jcfr:
Thanks @lassoan

jcfr:
> Use same mechanism as ParaView

It may be sensible to reference the ParaView file (and Paraview version) from which this was adapted from ..

jcfr:
@demarle @martinken Moving forward, I wonder if any of this code should be added to VTK proper instead of being duplicated in both Paraview and Slicer ... it seems like something any VTK based application would like to check.

pieper:
👍 thanks for adding this!

demarle:
I agree that should be at the VTK level.

martinken:
In terms of a Qt bit of code then sure that should be added to VTK in the GUI Qt module.

Base/QTApp/qSlicerApplicationHelper.txx
lassoan: I would not want to show a popup when somebody just uses Slicer command-line, potentially not rendering anything. It could make sense to log a warning though.

lassoan:
Qt part could remain application specific, but NewOffscreenRenderWindow() method and OS-specific logic that figures out ONSCREEN_RENDERING, HEADLESS_RENDERING_USES_OSMESA, HEADLESS_RENDERING_USES_EGL etc. capabilities could go into VTK.

jcfr:
> Qt part could remain application specific, but NewOffscreenRenderWindow() method and OS-specific logic that figures out ONSCREEN_RENDERING, HEADLESS_RENDERING_USES_OSMESA, HEADLESS_RENDERING_USES_EGL etc. capabilities could go into VTK.

Agreed.

@martinken @demarle  What do you think ?

Base/QTApp/qSlicerApplicationHelper.txx
jcfr: Gotcha. I was thinking of the case of application no having a main window but instantiating their own UI using python. That said this is still experimental and a corner case (I have  a topic I need to finalize ....)

lassoan: qSlicerApplicationHelper::checkRenderingCapabilities() is a public static method that slicelets can call.

jcfr: Indeed, we should be able to add a decorator for the method. The assumption is that no VTK initialization will be done before the python code is executed. 

All of  that said, there are way to address this and it should not block integration of this PR.

PR #1027, opened 2018-10-09
ENH: Disable TCL Bridge by default  See #4625

sjh26:
Disabling TCL since it is only used by EMSegment, which is
non-functional.  Disabling the TCL option will also disable
EMSegment


pieper:
LGTM

sjh26:
I have a build started to check this, but doubt there will be any issue.

jcfr:
Closing. Change integrated in [r27470](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27470)

PR #1029, opened 2018-10-16
BUG: Fix default display and storage node creation

cpinter:
The CreateDefaultDisplayNodes and CreateDefaultStorageNode functions did not use the CreateNewNodeByClass infrastructure so default nodes were not considered. This has been fixed in all node classes using such functions.

### DIFF
--- a/Libs/MRML/Core/Testing/vtkMRMLBSplineTransformNodeTest1.cxx
+++ b/Libs/MRML/Core/Testing/vtkMRMLBSplineTransformNodeTest1.cxx
@@ -12,10 +12,13 @@
 
 #include "vtkMRMLBSplineTransformNode.h"
 #include "vtkMRMLCoreTestingMacros.h"
+#include "vtkMRMLScene.h"
 
 int vtkMRMLBSplineTransformNodeTest1(int , char * [] )
 {
   vtkNew<vtkMRMLBSplineTransformNode> node1;
+  vtkNew<vtkMRMLScene> scene;
+  scene->AddNode(node1);
   EXERCISE_ALL_BASIC_MRML_METHODS(node1.GetPointer());
   return EXIT_SUCCESS;
 }
--- a/Libs/MRML/Core/Testing/vtkMRMLColorTableNodeTest1.cxx
+++ b/Libs/MRML/Core/Testing/vtkMRMLColorTableNodeTest1.cxx
@@ -27,7 +27,11 @@ using namespace vtkMRMLCoreTestingUtilities;
 int vtkMRMLColorTableNodeTest1(int argc, char * argv[])
 {
   vtkNew<vtkMRMLColorTableNode> node1;
-  EXERCISE_ALL_BASIC_MRML_METHODS(node1.GetPointer());
+  {
+    vtkNew<vtkMRMLScene> scene;
+    scene->AddNode(node1);
+    EXERCISE_ALL_BASIC_MRML_METHODS(node1.GetPointer());
+  }
 
   if (argc != 2)
     {
@@ -57,15 +61,15 @@ int vtkMRMLColorTableNodeTest1(int argc, char * argv[])
     colorNode->SetColor(2, "two", 0.0, 1.0, 0.0, 1.0);
     colorNode->NamesInitialisedOn();
 
-    vtkSmartPointer<vtkMRMLStorageNode> colorStorageNode =
-        vtkSmartPointer<vtkMRMLStorageNode>::Take(colorNode->CreateDefaultStorageNode());
-
     // add node to the scene
     vtkNew<vtkMRMLScene> scene;
     scene->SetRootDirectory(tempDir);
     CHECK_NOT_NULL(scene->AddNode(colorNode.GetPointer()));
 
     // add storage node to the scene
+    vtkSmartPointer<vtkMRMLStorageNode> colorStorageNode =
+        vtkSmartPointer<vtkMRMLStorageNode>::Take(colorNode->CreateDefaultStorageNode());
+
     scene->AddNode(colorStorageNode);
     colorNode->SetAndObserveStorageNodeID(colorStorageNode->GetID());
 
--- a/Libs/MRML/Core/Testing/vtkMRMLDiffusionImageVolumeNodeTest1.cxx
+++ b/Libs/MRML/Core/Testing/vtkMRMLDiffusionImageVolumeNodeTest1.cxx
@@ -12,10 +12,13 @@
 
 #include "vtkMRMLCoreTestingMacros.h"
 #include "vtkMRMLDiffusionImageVolumeNode.h"
+#include "vtkMRMLScene.h"
 
 int vtkMRMLDiffusionImageVolumeNodeTest1(int , char * [] )
 {
   vtkNew<vtkMRMLDiffusionImageVolumeNode> node1;
+  vtkNew<vtkMRMLScene> scene;
+  scene->AddNode(node1);
   EXERCISE_ALL_BASIC_MRML_METHODS(node1.GetPointer());
   return EXIT_SUCCESS;
 }
--- a/Libs/MRML/Core/Testing/vtkMRMLDiffusionTensorVolumeNodeTest1.cxx
+++ b/Libs/MRML/Core/Testing/vtkMRMLDiffusionTensorVolumeNodeTest1.cxx
@@ -12,10 +12,13 @@
 
 #include "vtkMRMLCoreTestingMacros.h"
 #include "vtkMRMLDiffusionTensorVolumeNode.h"
+#include "vtkMRMLScene.h"
 
 int vtkMRMLDiffusionTensorVolumeNodeTest1(int , char * [] )
 {
   vtkNew<vtkMRMLDiffusionTensorVolumeNode> node1;
+  vtkNew<vtkMRMLScene> scene;
+  scene->AddNode(node1);
   EXERCISE_ALL_BASIC_MRML_METHODS(node1.GetPointer());
   return EXIT_SUCCESS;
 }
--- a/Libs/MRML/Core/Testing/vtkMRMLDiffusionWeightedVolumeNodeTest1.cxx
+++ b/Libs/MRML/Core/Testing/vtkMRMLDiffusionWeightedVolumeNodeTest1.cxx
@@ -12,10 +12,13 @@
 
 #include "vtkMRMLCoreTestingMacros.h"
 #include "vtkMRMLDiffusionWeightedVolumeNode.h"
+#include "vtkMRMLScene.h"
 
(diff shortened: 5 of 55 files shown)


jcfr:
Thanks :+1: 

Should we add this into the release ? Or can it wait ?

cpinter:
It's not critical but if it's easy to add to the release it wouldn't hurt. Thanks

jcfr:
It is easy enough, I would like other to also review. @lassoan @pieper

pieper:
From a visual read-through all looks good.  I'm sure there isn't 100% coverage on these classes, so I suppose that if there were a typo it could cause an issue, but that looks unlikely.  I'd say the risk is small and the benefit of having this in the release is good.

cpinter:
I did a search in all files so coverage should be 100% in Slicer core. I double checked the changes, and I'm running the tests. If there are any failing ones because of this I'll amend the PR asap.

jcfr:
Thanks for the review @pieper 

@cpinter I will wait for your feedback before integrating

cpinter:
Some tests fail because the scene is not set before calling CreateDefaultStorageNode
(vtkMRMLScalarVolumeNode (000001AF65D48220): CreateDefaultStorageNode failed: scene is invalid)
If we want to keep the functionality the same then I can return the new node the same way as before if there is no scene, but it does not seem right, nor returning with NULL without an error message. Opinions?

pieper:
I can't think of a valid use case for creating default storage/display nodes without a scene - they require IDs to be meaningful.  Even outside of the Slicer use case, these calls only make sense if you create a scene first, so I guess the tests should be changed.  If this is too much of a change we can wait until after the release.

jcfr:
Few comments:
* `CreateDefaultStorageNode` is a factory method, it means that the user of the method takes ownership of the method. This PR changes the semantic. 
* Fow now, I think it make sense to support creating the storage node even without a scene

To move forward and keep the semantic, I then suggest the following implementation:

```
vtkMRMLStorageNode* vtkMRMLColorTableNode::CreateDefaultStorageNode()
{
  vtkMRMLScene* scene = this->GetScene();
  if (scene != NULL)
    {
    vtkMRMLStorageNode * storageNode = vtkMRMLStorageNode::SafeDownCast(scene->CreateNodeByClass("vtkMRMLColorTableStorageNode"));
    storageNode->Register(this);
    return storageNode;
    }
  else
    {
    return vtkMRMLColorTableStorageNode::New();
    }
  
}
```

After the release, we can then revisit.

cpinter:
> the user of the method takes ownership of the method. This PR changes the semantic

Why? As far as I know neither
`return vtkMRMLDoubleArrayStorageNode::New();`
nor
`return vtkMRMLStorageNode::SafeDownCast(scene->CreateNodeByClass("vtkMRMLDoubleArrayStorageNode"));`
takes ownership of the object. Am I wrong?

jcfr:
You are correct, these two functions do not take ownership per se.

Instead, these two functions return an object with a reference count greater than 1 and expect the **caller** to "take ownership". (e.g using `vtkSmartPointer::Take()` function )

cpinter:
That's right. So I think this PR changes nothing in terms of this then.

jcfr:
Ooop ... sorry for the noise. you are exactly right. This was an oversight. 

I guess something like this would do it then, I would prefer we keep the same behavior when there are no scene. This could change after the release.

```
vtkMRMLStorageNode* vtkMRMLColorTableNode::CreateDefaultStorageNode()
{
  if (this->GetScene() != NULL)
    {
    return vtkMRMLStorageNode::SafeDownCast(
        this->GetScene()->CreateNodeByClass("vtkMRMLColorTableStorageNode"));
    }
  else
    {
    return vtkMRMLColorTableStorageNode::New();
    }
  
}
```

lassoan:
I would add a warning if this->GetScene() == NULL to make sure that developers are aware when there is a chance that not the correct default node is instantiated (e.g., due to incorrect order of operations).

jcfr:
If a warning is added, someone should run the test and make sure they pass first. Otherwise, I suggest we wait after the release for adding the warning.

jcfr:
Considering that the nightly build should started, this change is unlikely to be integrated in the release.

cpinter:
The tests are not trivial to fix, unfortunately I bumped into a contradiction.
EXERCISE_ALL_BASIC_MRML_METHODS does various types of tests, one of these two will fail if we keep the changes I made:
1. ExerciseBasicStorableMRMLMethods: Storage node cannot be created without the node being in the scene, but if I set a scene, then
2. ExerciseBasicMRMLMethods: CHECK_NULL(node->GetScene()); fails because it expects no scene

I'm not sure what is the reasoning behind the check in 2 (the scene has to be null check). If that can be removed then there is one last issue:
3. The [ReadXMLAttributes call](https://github.com/Slicer/Slicer/blob/master/Libs/MRML/Core/vtkMRMLCoreTestingUtilities.cxx#L204) in ExerciseBasicMRMLMethods sets the ID directly without triggering ID cache update in the scene, so there is this error when the scene is cleared on destruction: "GetNodeByID: Node is in the scene, but its ID is missing from the NodeIDs cache: vtkMRMLNodeTest1".

All this would be preventable by Jc's suggestion which returns a simply created node without taking the default node into account in the absence of a scene.

Not quite sure what would be the best choice, because the latter would ignore the default node while we cannot even throw a warning message (although I can change the tests to ignore the warning but not sure if that's a good idea), but also changing many of the tests and the way these core tests are being run can have unintended consequences.
Thoughts?

jcfr:
Should we simply disable the creation of node using node constructor from both C++ and python ?  That would be consistent with the new behavior we are promoting.

lassoan:
> Should we simply disable the creation of node using node constructor

There are probably valid use cases for instantiating nodes directly, so probably we should not disable it.

pieper:
Yes, instantiating a node and adding it to the scene seems like the logical pattern (it's in lots of examples).  But when you are using the utility method to create default display nodes then you should expect them to have the proper SetAndObserve* behavior and that depends on the scene being valid.

On the other hand, now that I look at it it's odd that CreateDefaultStorageNode doesn't call SetAndObserveStorageNodeID.

We could deprecate (or leave) the current methods and introduce CreateAndObserveDefault{Display,Storage}Node methods.

cpinter:
It is indeed strange that there is no observation set in CreateDefaultStorageNode. Considering that CreateDefaultDisplayNodes does set the observation, I think this is an inconsistency that we should fix, so at least the two functions with the similar names do the same things. I know there might be issues in some places, but I'd rather add the observation in the CreateDefaultStorageNode functions to make them work the same way.

cpinter:
I integrated this commit in https://github.com/Slicer/Slicer/commit/0f8704d1528fbaf50a91d707ed6565a55c8a7f70
All tests pass except py_nomainwindow_SlicerOptionDisableSettingsTest, but @lassoan says it tends to fail depending on the exact way Slicer is built, and is most probably not related.

We can do the changes we're discussing (that affect the API) later. I keep the PR open as long as we're discussing these.

cpinter:
If there is nothing more to discuss right now, feel free to close it though :)

jcfr:
>  I keep the PR open as long as we're discussing these.

Good point.

This morning we also talked about adding a `vtkMRMLScene::Reset(vtkMRMLNode*)` that would allow to reset a node with the expected default properties.

Either way, I am wondering if we should we also replace the calls to `vtkNew<vtkMRMLNAME>` with the equivalent call using `AddNewNodeByClass` or by adding a call to the (to be added)  `vtkMRMLScene::Reset(vtkMRMLNode*)` function ? 


```
$ ack "vtkNew\<vtkMRML[a-zA-Z]+Node\>" --ignore-dir=Testing -l
Modules/Loadable/Models/qSlicerModelsModuleWidget.cxx
Modules/Loadable/Models/Logic/vtkSlicerModelsLogic.cxx
Modules/Loadable/Transforms/Logic/vtkSlicerTransformLogic.cxx
Modules/Loadable/CropVolume/qSlicerCropVolumeModuleWidget.cxx
Modules/Loadable/CropVolume/Logic/vtkSlicerCropVolumeLogic.cxx
Modules/Loadable/VolumeRendering/Logic/vtkSlicerVolumeRenderingLogic.cxx
Modules/Loadable/Annotations/GUI/qSlicerAnnotationModulePropertyDialog.cxx
Modules/Loadable/Annotations/VTKWidgets/vtkAnnotationGlyphSource2D.cxx
Modules/Loadable/Annotations/Logic/vtkSlicerAnnotationModuleLogic.cxx
Modules/Loadable/DoubleArrays/Logic/vtkSlicerDoubleArraysLogic.cxx
Modules/Loadable/Tables/Logic/vtkSlicerTablesLogic.cxx
Modules/Loadable/Markups/VTKWidgets/vtkMarkupsGlyphSource2D.cxx
Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
Modules/Loadable/Markups/Logic/vtkSlicerMarkupsLogic.cxx
Modules/Loadable/Volumes/Logic/vtkSlicerVolumesLogic.cxx
Modules/Loadable/Units/Logic/vtkSlicerUnitsLogic.cxx
Modules/Loadable/Colors/Logic/vtkSlicerColorLogic.cxx
Modules/Loadable/Segmentations/Logic/vtkSlicerSegmentationsModuleLogic.cxx
Modules/Loadable/SceneViews/Logic/vtkSlicerSceneViewsModuleLogic.cxx
Modules/Loadable/Data/qSlicerDataModuleWidget.cxx
Modules/CLI/ExtractSkeleton/ExtractSkeleton.cxx
Modules/CLI/ModelMaker/ModelMaker.cxx
Modules/CLI/ExecutionModelTour/ExecutionModelTour.cxx
Modules/CLI/PETStandardUptakeValueComputation/PETStandardUptakeValueComputation.cxx
Libs/MRML/Core/vtkMRMLFreeSurferModelOverlayStorageNode.cxx
Libs/MRML/Core/vtkMRMLTransformDisplayNode.cxx
Libs/MRML/Core/vtkMRMLLabelMapVolumeNode.cxx
Libs/MRML/Core/vtkMRMLScalarVolumeNode.cxx
Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
Libs/MRML/Logic/vtkMRMLSliceLogic.cxx
Libs/MRML/Logic/vtkMRMLColorLogic.cxx
Base/QTCore/qSlicerCoreApplication.cxx
Base/QTCore/qSlicerXcedeCatalogReader.cxx
Base/QTCore/qSlicerSlicer2SceneReader.cxx
Base/Logic/vtkSlicerGlyphSource2D.cxx
```

cpinter:
I've created a Discourse topic for this discussion in case we want to keep thinking about this later.
https://discourse.slicer.org/t/discussion-about-default-display-and-storage-nodes/4595
Closing this PR because otherwise it's integrated.

PR #1030, opened 2018-10-22
BUG: Fixed update of imported nodes in subject hierarchy

lassoan:
Changing property (name, visibility, etc.) of any nodes that were loaded from a scene did not cause update of the subject hierarchy item.

How to reproduce: go to Data module, load [this scene](https://1drv.ms/u/s!Arm_AFxB9yqHte9cVS_amqjC0F8Fzw), switch to "All nodes" tab, rename "LeafletSegmentation" to "abc", switch back to "Subject hierarchy" tab => ERROR! "LeafletSegmentation" node name is not updated.

The problem was that node modification observers were not added for imported nodes.


cpinter:
Looks good to me, thanks!

I didn't test this because the fix seems obviously OK, but if you want I can test it.

jcfr:
Closing. Fix in integrated in [r27510](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27510)

lassoan:
@cpinter thanks for the review

PR #1032, opened 2018-10-22
BUG: Fix 3D picking in displayable managers

cpinter:
Model and segmentation displayable managers use vtkCellPicker to pick a node in a certain RAS position. The current implementation uses the GetDataSet method to get the first picked dataset. However, that dataset might be one that is not managed by the caller DM. This is fixed by going through all the picked props and finding the first managed prop.

### DIFF
--- a/Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.h
+++ b/Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.h
@@ -149,8 +149,10 @@ public:
   static vtkLookupTable* CreateLookupTableCopy(vtkLookupTable* source);
 
   /// Helper function for determining what type of scalar is active.
-  static bool IsCellScalarsActive(vtkMRMLDisplayNode* displayNode,
-    vtkMRMLModelNode* model = 0);
+  /// \return True if attribute location in display node is vtkAssignAttribute::CELL_DATA
+  ///   or active cell scalar name in the model node is vtkDataSetAttributes::SCALARS.
+  ///   False otherwise.
+  static bool IsCellScalarsActive(vtkMRMLDisplayNode* displayNode, vtkMRMLModelNode* model = 0);
 
 protected:
 
@@ -216,9 +218,6 @@ protected:
   void RemoveDispalyedID(std::string &id);
 
   vtkMRMLSelectionNode* GetSelectionNode();
-
-  /// Find picked node and cell from mesh and set PickedNodeID and PickedPointID in Internal
-  void FindPickedDisplayNodeFromMesh(vtkPointSet* mesh, double pickedPoint[3]);
 
 private:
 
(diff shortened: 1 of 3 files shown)


Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.h
```diff
@@ -217,9 +216,6 @@ class VTK_MRML_DISPLAYABLEMANAGER_EXPORT vtkMRMLModelDisplayableManager
 
   vtkMRMLSelectionNode* GetSelectionNode();
 
-  /// Find picked node and cell from mesh and set PickedNodeID and PickedPointID in Internal
-  void FindPickedDisplayNodeFromMesh(vtkPointSet* mesh, double pickedPoint[3]);
```
jcfr: What do you think of adding an entry to https://www.slicer.org/wiki/Documentation/Nightly/Developers/Tutorials/MigrationGuide ?

Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.h
```diff
@@ -149,8 +149,7 @@ class VTK_MRML_DISPLAYABLEMANAGER_EXPORT vtkMRMLModelDisplayableManager
   static vtkLookupTable* CreateLookupTableCopy(vtkLookupTable* source);
 
   /// Helper function for determining what type of scalar is active.
-  static bool IsCellScalarsActive(vtkMRMLDisplayNode* displayNode,
-    vtkMRMLModelNode* model = 0);
+  static bool IsCellScalarsActive(vtkMRMLDisplayNode* displayNode, vtkMRMLModelNode* model = 0);
```
jcfr: It may be sensible to update the docstring

cpinter: Not sure what you mean. I just removed a line break from here, nothing changed.

Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.h
cpinter: Should I rather add it back?

Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.h
jcfr: It could have been the opportunity to improve the docstring. It also fine to leave things the way they are.

Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.h
jcfr: Since we are moving Slicer 5.0 .. breaking things is fine. I suggest we keep a record.

cpinter:
Integrated in https://github.com/Slicer/Slicer/commit/65550bc6efff2df65a7216c36cf9fb70312b7c43

Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.h
cpinter: I added a new section for 5.0 changes and added this as the only item so far

PR #1035, opened 2018-10-29
ENH: add CompressionLevel to storage node, use in NRRDWriter

ihnorton:
ref: https://discourse.slicer.org/t/how-to-unset-default-compress-option-during-save/4488/18


ihnorton:
Numbers for an HCP brain MRI dataset:
- level 9, 96s -> 495MB
- level 1, 53s -> 505MB
- level 0, 20s -> 2GB
- compression off, 4s -> 2GB

(that last point is interesting, makes me wonder if the NRRD zlib interface is less efficient than it could be - level 0 should be effectively a memcopy, but it is much slower than compression off)

Libs/vtkTeem/vtkTeemNRRDWriter.h
```diff
@@ -55,6 +55,10 @@ class VTK_Teem_EXPORT vtkTeemNRRDWriter : public vtkWriter
   vtkGetMacro(UseCompression,int);
   vtkBooleanMacro(UseCompression,int);
 
+  vtkSetMacro(CompressionLevel,int);
+  vtkGetMacro(CompressionLevel,int);
+  vtkBooleanMacro(CompressionLevel,int);
```
lassoan: not Boolean

Libs/vtkTeem/vtkTeemNRRDWriter.h
```diff
@@ -55,6 +55,10 @@ class VTK_Teem_EXPORT vtkTeemNRRDWriter : public vtkWriter
   vtkGetMacro(UseCompression,int);
   vtkBooleanMacro(UseCompression,int);
 
+  vtkSetMacro(CompressionLevel,int);
```
lassoan: Use vtkSetClampMacro(CompressionLevel,int,0,9);

Libs/MRML/Core/vtkMRMLStorageNode.cxx
```diff
@@ -46,6 +46,7 @@ vtkMRMLStorageNode::vtkMRMLStorageNode()
   this->URI = NULL;
   this->URIHandler = NULL;
   this->UseCompression = 1;
+  this->CompressionLevel = 1;
```
lassoan: Need to add CompressionLevel to read/write xml and print methods.

I'm also wondering if we should use DEFAULT, NONE, LOW, MEDIUM, HIGH enums for compression levels. If we ever expose this to users, it would be nicer to show fewer, more explicit options than 0-9. We would also have more flexibility in how we map the enums to actual compression levels in various storage nodes.

We should at least have a default compression level so that we can easily re-map default compression levels and don't litter storage nodes with explicit compression level values. If we keep compression level as int then we could use -1 as "default" level (which would not be written to XML file and would be translated to actual compression level dynamically, currently to 1).

ihnorton: Ok, I added print and XML methods, an enum, and -1 <-> default mapping.

ihnorton:
>  Almost 2x faster compression

I was hoping for quite a bit more actually, more based on experience with pure python/zlib (using pynrrd):

```
In [8]: %timeit -n 1 nrrd.write("/tmp/c2/dwi_9.nrrd", img, header=hdr, compression_level=9)
1 loop, best of 3: 53.8 s per loop

In [15]: %timeit -r1 -n1 nrrd.write("/tmp/c2/dwi_1.nrrd", img, header=hdr, compression_level=1)
1 loop, best of 1: 5.84 s per loop

In [16]: %timeit -r1 -n1 nrrd.write("/tmp/c2/dwi_0.nrrd", img, header=hdr, compression_level=0)
1 loop, best of 1: 701 ms per loop
```

ihnorton:
Actually, my current Slicer is built in debug mode, so might see better speed-up in a release build.

ihnorton:
I'm not planning to do a release build locally right now (building deps takes hours). I think even ~50% boost is enough to justify this, though, becauseit doesn't add much complication...

pieper:
Now that we are in post-release 'break things' mode I have no objections to merging this so Murat and others can test.  I don't see much risk for downsides.

lassoan:
@sunderlandkyl works on compressed video storage (mkv file reading/writing/replaying works very nicely by the way). He will add CompressionParameter attribute to vtkMRMLStorageNode. It'll be a string parameter that can be populated from presets provided by concrete storage nodes. This can handle complex presets, such as video compression settings, and can be utilized for simple zlib compression settings (low/medium/high).

Until we receive that pull request, I think we should merge this as is, so that we can test different compression levels. Kyle will update CompressionParameter storage to the new API in his pull request.

pieper:
Have you thought about exposing this in the Save dialog too?  Would mean changing the checkbox to a combobox.

Sunderlandkyl:
The plan that I discussed with @lassoan was that the save dialog would have additional column containing a combobox if UseCompression was checked.

It could also potentially be implemented as a single combobox, but then there is a potential disconnect between the two parameters (UseCompression + CompressionParameter) being represented with a single combobox.

I could see it being done either way.

pieper:
It would be nice if the GUI just follows what's in the node.  If we're keeping the compression boolean and the level then yes, we can show both in the GUI.

PR #1037, opened 2018-10-31
ENH: Allow observing color name change

cpinter:
This change makes it possible to update the scalar bar when the isodose level is changed. Isodose level is stored as the color name, and it can be changed in a color table view. Related issue: https://github.com/SlicerRt/SlicerRT/issues/8

I'm issueing a PR because there was a TBD there but also because I'm not sure what are the cases where color names might be changed in bulk, causing potential issues with lots of Modified events.


pieper:
LGTM 👍

jcfr:
@vovythevov  Do you foresee any issue with this change ?

cpinter:
Integrated in https://github.com/Slicer/Slicer/commit/94fb4a6340c25eeba6183e0eb9a54b7e1d43db8c

PR #1038, opened 2018-11-02
ENH: Add vtkMRMLStorageNode CompressionParameter and presets

Sunderlandkyl:
-Adds a framework for specifying different preset levels of compression in vtkMRMLStorageNode
-Changes in vtkStreamingVolumeCodec also allow codecs to specify their own presets
-Replaces previous CompressionLevel int

image

### DIFF
--- a/Base/QTGUI/Resources/UI/qSlicerNodeWriterOptionsWidget.ui
+++ b/Base/QTGUI/Resources/UI/qSlicerNodeWriterOptionsWidget.ui
@@ -6,7 +6,7 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>94</width>
+    <width>145</width>
     <height>17</height>
    </rect>
   </property>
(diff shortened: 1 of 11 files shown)


pieper:
Generally this looks great to me (I didn't have a chance to test or read too carefully).

One thing that occurs to me is that the "Show options" checkbox persists during a Slicer session, but is reset on restart.  

If we persist the state to settings, then users like Murat would be able to have easy access to these options for their large datasets (where picking a particular mode on a per-dataset level may save several minutes per save operation).

Sunderlandkyl:
The checkbox should now persist between sessions.

Sunderlandkyl:
@lassoan

lassoan:
Thank you! Merged - https://github.com/Slicer/Slicer/commit/900bc41a33cb099fef202a5edc28932e4da69d80.

I've changed some method names, so you may need to update extensions that use this.

PR #1041, opened 2018-11-13
BUG: Fixed volume rendering ROI reset

lassoan:
When volume rendered node's transform was changed, ROI was reset. This made it impossible to review clipped volume sequences or transform clipped volumes.

ROI reset is not needed anyway, because if user would like to keep the volume fully displayed when it is transformed, then it can be achieved by setting the same parent transform for the ROI node as for the volume node.


lassoan:
@cpinter could you check this? Resetting the FOV had bad side effects (clipping was lost when volume was moved or replayed using Sequences). The simplest solution would be to not reset the ROI automatically.

pieper:
The change makes sense to me.

cpinter:
I'm okay with this change. I didn't intentionally change event handling when I did the rework, so since there is no comment explaining why it's needed, it's probably not.

lassoan:
Integrated https://github.com/Slicer/Slicer/commit/167e336f544ed05c72425bf432ba24eed0f6c64e

jcfr:
Makes sense :+1:

PR #1045, opened 2018-11-20
ENH: Allow specifying study, series, content datetime for DICOM image export

lassoan:
It was not possible to export 4D data sets, because study, series, frame of reference UIDs were always generated and it was not possible to set content time (to distinguish between image frames at the same spatial position at different time points).

Now study, series, frame of reference UIDs can be specified optionally and study, series, content date&time can be set.

### DIFF
--- a/Modules/CLI/CreateDICOMSeries/CreateDICOMSeries.cxx
+++ b/Modules/CLI/CreateDICOMSeries/CreateDICOMSeries.cxx
@@ -114,113 +114,158 @@ int DoIt( int argc, char * argv[])
               << std::endl
               << std::flush;
 
+    // Set all required DICOM fields
+    std::ostringstream value;
+
+    // Image Position (Patient)
     typename Image3DType::PointType    origin;
     typename Image3DType::IndexType    index;
     index.Fill(0);
     index[2] = i;
     image->TransformIndexToPhysicalPoint(index, origin);
-
-    std::ostringstream value;
-    // Set all required DICOM fields
     value.str("");
     value << origin[0] << "\\" << origin[1] << "\\" << origin[2];
     itk::EncapsulateMetaData<std::string>(dictionary, "0020|0032", value.str() );
 
+    // Image Orientation (Patient)
+    value.str("");
+    value << oMatrix[0][0] << "\\" << oMatrix[1][0] << "\\" << oMatrix[2][0] << "\\";
+    value << oMatrix[0][1] << "\\" << oMatrix[1][1] << "\\" << oMatrix[2][1];
+    itk::EncapsulateMetaData<std::string>(dictionary, "0020|0037", value.str() );
+
+    // Slice Thickness
+    value.str("");
+    value << spacing[2];
+    itk::EncapsulateMetaData<std::string>(dictionary, "0018|0050", value.str() );
+
+    // Instance Number
     value.str("");
     value << i + 1;
     itk::EncapsulateMetaData<std::string>(dictionary, "0020|0013", value.str() );
 
-    itk::EncapsulateMetaData<std::string>(dictionary, "0008|0008", std::string("ORIGINAL\\PRIMARY\\AXIAL") );  // Image
-                                                                                                               // Type
-    itk::EncapsulateMetaData<std::string>(dictionary, "0008|0016", std::string("1.2.840.10008.5.1.4.1.1.2") ); // SOP
-                                                                                                               // Class
-                                                                                                               // UID
-    itk::EncapsulateMetaData<std::string>(dictionary, "0010|0030", std::string("20060101") );                  //
-                                                                                                               // Patient's
-                                                                                                               // Birthdate
-    itk::EncapsulateMetaData<std::string>(dictionary, "0010|0032", std::string("010100.000000") );             //
-                                                                                                               // Patient's
-                                                                                                               // Birth
-                                                                                                               // Time
-    itk::EncapsulateMetaData<std::string>(dictionary, "0010|0040", std::string("M") );                         //
-                                                                                                               // Patient's
-                                                                                                               // Sex
-    itk::EncapsulateMetaData<std::string>(dictionary, "0008|0020", std::string("20050101") );                  // Study
-                                                                                                               // Date
-    itk::EncapsulateMetaData<std::string>(dictionary, "0008|0030", std::string("010100.000000") );             // Study
-                                                                                                               // Time
-    itk::EncapsulateMetaData<std::string>(dictionary, "0008|0050", std::string("1") );                         //
-                                                                                                               // Accession
-                                                                                                               // Number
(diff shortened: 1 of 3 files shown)


cpinter:
What's the reason you fixed the old DICOM exporter module instead of doing something in the new DICOM exporter?

lassoan:
I am using the new exporter. I just needed improvements in the low-level infrastructure. The exporter plugin is here: https://github.com/SlicerRt/Sequences/commit/b9f29d13d7aa40287e86f01d552b7c1bfbe6de34

cpinter:
I see. Thanks for the explanation

Modules/CLI/CreateDICOMSeries/CreateDICOMSeries.xml
```diff
@@ -50,6 +50,13 @@
       <label>Study Date</label>
       <default>20060101</default>
     </string>
+    <string>
+      <name>studyTime</name>
+      <longflag>--studyTime</longflag>
+      <description><![CDATA[The time of the study [0008-0030]]]></description>
```
fedorov: Is there a good reason to introduce a new notation conventions for DICOM tags here? Why not `(0008,0030)` or `(0x0008,0x0030)`? Same for "The time of the study" - why not `StudyTime` or `Study Time`?

Modules/CLI/CreateDICOMSeries/CreateDICOMSeries.cxx
```diff
@@ -222,6 +226,49 @@ int DoIt( int argc, char * argv[])
...
+
+    bool keepOriginalUID = (!studyInstanceUID.empty() || !seriesInstanceUID.empty() || !frameOfReferenceInstanceUID.empty());
+    gdcmIO->SetKeepOriginalUID(keepOriginalUID);
+    if( studyInstanceUID.size() > 0 )
+      {
+      itk::EncapsulateMetaData<std::string>(dictionary, "0020|000d", studyInstanceUID);
+      }
+    else if (keepOriginalUID)
+      {
+      std::cerr << "studyInstanceUID must be specified if any of the UIDs are specified." << std::endl;
```
fedorov: Should there be a `return` here? If this is a warning, then maybe the message should not say "must"? Also, I do not understand the logic why SeriesInstanceUID would be required if StudyInstanceUID or FrameOfReferenceUID are specified? Can you comment?

lassoan: If you prefer, we can return with error.

Modules/CLI/CreateDICOMSeries/CreateDICOMSeries.xml
lassoan: I followed existing pattern. We can change all of them to standard style.

fedorov: Makes sense. If it is existing pattern, I am fine leaving them as is.

Modules/CLI/CreateDICOMSeries/CreateDICOMSeries.cxx
fedorov: It is just confusing that the message is not consistent with the action. I would either return error, but probably better to just print a warning, because I don't see a good reason why this should be a "must".

lassoan:
Some tags are pre-populated from DICOM import (if they are stored in Subject Hierarchy) but not all, and certainly SOP instance UID should match modality, so this exporter will only export simple time series correctly. It is a step forward and if someone has more time or specific need to export DCE-MRI, etc. correctly then there is a starting point.

fedorov:
Makes sense, I agree.

pieper:
👍 I agree we could always imagine a more perfect overall solution, but this is a step in the right direction.

Modules/CLI/CreateDICOMSeries/CreateDICOMSeries.cxx
lassoan: > I don't see a good reason why this should be a "must"

[itkGDCMImageIO](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/IO/GDCM/src/itkGDCMImageIO.cxx#L1154-L1189) implements an all-or-nothing approach with these 3 UIDs. You either specify them or ITK generates all of them.

lassoan:
Thanks for the reviews. I'll make the proposed changes and merge.

Modules/CLI/CreateDICOMSeries/CreateDICOMSeries.cxx
fedorov: I see. Thanks for investigating. I don't understand that logic, but it is unrelated to this PR.

lassoan:
I've cleaned up the code a bit (changed DICOM tag to standard style, etc), added automatic SOP class UID setting based on image modality, and returning with error when not all UIDs are specified.

fedorov:
Thank you Andras, this is a good improvement!

Modules/CLI/CreateDICOMSeries/CreateDICOMSeries.xml
```diff
@@ -111,16 +132,55 @@
...
       <name>rescaleSlope</name>
       <longflag>--rescaleSlope</longflag>
-      <description><![CDATA[Rescale slope [0028-1053]. Converts pixel values on disk to pixel values in memory. (Pixel value in memory) = (Pixel value on disk) * rescaleSlope + rescaleInterscept.  Default is 1.0. Data values are converted on write (the data is scaled and shifted so that the slope and interscept will bring it back to the current intensity range).]]></description>
+      <description><![CDATA[Rescale slope (0028,1053). Converts pixel values on disk to pixel values in memory. (Pixel value in memory) = (Pixel value on disk) * rescaleSlope + rescaleInterscept.  Default is 1.0. Data values are converted on write (the data is scaled and shifted so that the slope and interscept will bring it back to the current intensity range).]]></description>
       <default>1.0000</default>
     </double>
+    <string>
+      <name>contentDate</name>
+      <longflag>--contentDate</longflag>
+      <description><![CDATA[The date of the image content (0008,0023)]]></description>
```
fedorov: Small nitpick - would be nice if "The date of the image content" and other similar occurrences were replaced with the actual attribute names from DICOM.

lassoan: DICOM tag name is already in the "label" field (that is displayed in Create DICOM Series module GUI). This CLI module is just a backend anyway, user interacts with the DICOM module's GUI.

fedorov: Yes, that is true. It is not critical.

lassoan:
This was committed a year ago in http://svn.slicer.org/Slicer4/trunk@27563.

PR #1046, opened 2018-11-23
Sh terminology and transform improvements

cpinter:
Two sets of changes in subject hierarchy:

- Add new column for color between visibility and transform columns. It shows the color for segments, models, and markup fiducials (selected color)
- Double-clicking the color brings up terminology selector where terminology and color can be selected. Terminology information is stored in segments as before (segment attributes), and for models and markups it is stored in MRML node attributes. When selecting a terminology type, color is overwritten but name is not by default
- Terminology selector has been improved to accommodate setting terminologies to model and markup fiducial nodes, for which it's not mandatory to set terminology (unlike segments)
  - None type always appears in type list unless type search term is specified
  - Multi-select is enabled for categories, and by default all the categories are selected so that all the types show up in the types list and a search finds any type in the context
  - Expand button added to hide categories list
  - Both categories and anatomy lists are hidden by default
  - Type search box has the focus when opened and selection can be accepted by Enter press (so terminology can be selected very quickly by just typing search term then pressing Enter)
Fixes https://issues.slicer.org/view.php?id=4658
Re https://github.com/PerkLab/SlicerOpenAnatomy/issues/4

- Transform column now shows icon instead of the first few letters of the transform name. There is a separate icon for linear and deformable transforms
- Double-clicking the icon brings up a menu instead of a node selector that had to be clicked again
- Subject hierarchy reference highlights are now updated immediately after selecting a transforms, so the highlighted transform node is always correct
Re https://issues.slicer.org/view.php?id=4401


cpinter:
Color column:
image

Simplified terminology selector:
image

Transform column:
image

Modules/Loadable/Models/SubjectHierarchyPlugins/qSlicerSubjectHierarchyModelsPlugin.cxx
```diff
@@ -221,12 +227,142 @@ QString qSlicerSubjectHierarchyModelsPlugin::tooltip(vtkIdType itemID)const
...
+      terminologyMetaData[qSlicerTerminologyItemDelegate::TerminologyRole].toString().toLatin1().constData() );
+    }
+  if (terminologyMetaData.contains(qSlicerTerminologyItemDelegate::NameRole))
+    {
+    modelNode->SetName(
+      terminologyMetaData[qSlicerTerminologyItemDelegate::NameRole].toString().toLatin1().constData() );
+    }
+  if (terminologyMetaData.contains(qSlicerTerminologyItemDelegate::NameAutoGeneratedRole))
+    {
+    modelNode->SetAttribute("NameAutoGenerated",
```
lassoan: Without any context, "NameAutoGenerated" is a bit obscure (not clear who generates it automatically, from what, when). It would clarify things if we included SubjectHierarchy and/or Terminology in the attribute name. For example:
- SubjectHierarchy.NameAutoGenerated
- SubjectHierarchy.UseTerminologyForName
- Terminology.NodeNameAutoGenerated
- **Terminology.AutoUpdateNodeName**

I think I like the last one the most.

fedorov:
@cpinter thanks for mentioning me! Considering RSNA is around the cornder - how urgent is this?

cpinter:
This is related to the OpenAnatomy project, and I don't think this is super urgent. However, I'll need to keep working on things that build on this change, so it would ake life easier if this could get integrated. If you and the others agree with what you see in the three images, then I guess this is safe to integrate. You'll use 4.10 stable at RSNA right?

Modules/Loadable/Models/SubjectHierarchyPlugins/qSlicerSubjectHierarchyModelsPlugin.cxx
cpinter: Sounds good to me

cpinter: The name of the module is Terminologies (plural). Shouldn't we use that instead of singular?

lassoan: Yes, good point, we should use the exact module name.

fedorov:
Csaba, the scope of changes sounds good to me. I am not concerned about breaking functionality for RSNA, but more about not having time to test the changes in the branch before RSNA. I know you will be maintaining this so, I don't worry too too much about something being broken, it can be fixed later, so I am fine if you merge.

A couple of (unrelated, sorry) feature suggestions I have, since I was playing with Segmentations/Segment Editor/Data modules in preparation to RSNA, and I am afraid I may forget:

Data module is very convenient to browse segmentations, since it shows the hierarchy, and is particularly useful when segmentations are stored in separate DICOM objects, loading as separate segmentation nodes. It would be very helpful if Data module provided:
1. functionality to toggle visibility for selected items (it already provides multi-select and visibility toggle for individual items, so it is just an item in the context menu to streamline the process)
2. "jump to slice" in the context menu, as in Segment Editor.

cpinter:
@fedorov Thanks, I'll probably merge soon if nobody has conceptual issues with this. I think I did thorough testing, and honestly it will be much easier to test for everyone if it's in a nightly release. And of course I'll fix whatever issue arises.

The SH suggestions sound good to me.
1. Let me think a bit about this. Currently only delete is in the multi-select pop-up menu, an dit wouldn't be a big deal adding this. However there is also a visibility context menu (right-click the eye), and I'm not sure if it should go there. In the meantime you can use the "hide all segments but this" option (if I remember correctly)
2. I think we talked about this point with @lassoan recently, and it will be no problem implementing it.

Modules/Loadable/Models/SubjectHierarchyPlugins/qSlicerSubjectHierarchyModelsPlugin.cxx
lassoan: It would be also good to define these attribute names somewhere (in Terminologies module?) to reduce chance of typos and have a place where these attributes can be documented.

fedorov:
@cpinter 

> 1. Let me think a bit about this. Currently only delete is in the multi-select pop-up menu, an dit wouldn't be a big deal adding this. However there is also a visibility context menu (right-click the eye), and I'm not sure if it should go there. 

I tested, and either it's a bug, or omission - if I have multiple items selected in the Data module, I do indeed see "Show only this segment" in the context menu when clicked over eye, but with multiple items selected, it has no effect, see below.

> In the meantime you can use the "hide all segments but this" option (if I remember correctly)

Yes, this indeed is working, but it is in the "Segmentations" module. So in a way there is a discrepancy between Data, Segmentations and Segment editor:
* Data does not support "Jump" or "Show selected only"
* Segment Editor does not support multi-select, and does not allow to select across segmentations (and both make sense)
* Segmentations does not allow to select across segmentations (and this also makes sense)

I can definitely achieve the task, but I need to jump across the modules. Adding "show multiple" and "jump to slice" I think would be a natural and much needed shortcut.

nov-23-2018 15-14-33

pieper:
Looks like great work!  I agree testing in the nightly will be easiest.  👍

lassoan:
Yes, since the stable is relatively recent, we can take use the nightly for testing.

However, CI reportee a build error that may be related to the changes:

/opt/qt/5.11.2/gcc_64/include/QtCore/qvariant.h:484:9: error: ‘int QVariant::compare(const QVariant&) const’ is protected
     int compare(const QVariant &other) const;
         ^
/usr/src/Slicer/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyModel.cxx:1064:47: error: within this context
       if (item->data().compare(transformNodeId))

cpinter:
Thanks! I'll fix the error.

cpinter:
I did the fix and although CircleCI returned errors, they were about network timeouts not actual build errors, so I went ahead and integrated the change in
https://github.com/Slicer/Slicer/commit/25ca9915aea0d83cfcb9b0adf24d0ebd1d661538 and
https://github.com/Slicer/Slicer/commit/3a5d29d3db9e85115c81326342bfb24f94f1d4bf

cpinter:
I just noticed that I didn't include @lassoan 's request about the attribute names although I was pretty sure I did. I'll address this in a subsequent commit.

lassoan:
Thanks a lot, it will be great to have these features. This makes terminologies much more accessible to users. I think I'll use terminologies more (even if I just want to set the colors, it will be easier to set the terminology).

cpinter:
Exactly. The previous selector with the jet cockpit complexity could have been a barrier for many people. It is now much quicker to make a meaningful selection.

cpinter:
@fedorov FYI "Show only this segment" is already present in subject hierarchy, see
image

For this you need to right-click the eye icon (there is a general context menu when you right-click on the name column and a visibility context menu over the eye icon).

I have implemented toggle visibility on multi-select because I think it's very useful regardless.

I'm thinking whether I should add "Jump to slices" to the general or the visibility context menu. I'm leaning towards the latter. @lassoan what do you think?

cpinter:
@fedorov I implemented everything in the issue above (decided to do it as I'm working on SH anyway and these were easy): https://issues.slicer.org/view.php?id=4659
Whenever you have the chance please give it a go and let me know what you think.

I'll work on the issue of mandatory type modifiers and duplicate type items next.

fedorov:
@cpinter 

> FYI "Show only this segment" was already present in subject hierarchy, see

Yes, I know that. What I was saying in https://github.com/Slicer/Slicer/pull/1046#issuecomment-441313074 is if I select **multiple items**, activate the menu, and select "show only", it had no effect.

cpinter:
True. Thanks for the clarification.

PR #1047, opened 2018-11-25
BUG: Fixed virtual reality picking error

lassoan:
When the pick position was close or inside non-selectable object, picking often failed because VTK's cell picker found that actor.

Now if a model, segmentation, or volume rendering node is not selectable then it is completely excluded from picking.


lassoan:
If there are no objections then I'll merge this today, as without this picking in VR is very unreliable.

jcfr:
LGTM

pieper:
👍

lassoan:
Thanks for reviewing. Merged.

PR #1048, opened 2018-12-03
Support per view interaction mode

jcfr:


Modules/Loadable/Annotations/Logic/vtkSlicerAnnotationModuleLogic.h
```diff
@@ -29,10 +29,10 @@ class VTK_SLICER_ANNOTATIONS_MODULE_LOGIC_EXPORT vtkSlicerAnnotationModuleLogic
   virtual void PrintSelf(ostream& os, vtkIndent indent) VTK_OVERRIDE;
 
   // Start the place mode for annotations
-  void StartPlaceMode(bool persistent=false);
+  void StartPlaceMode(bool persistent=false, const char* viewNodeID = NULL);
```
jcfr: Update doxygen

Modules/Loadable/Annotations/Logic/vtkSlicerAnnotationModuleLogic.h
```diff
@@ -29,10 +29,10 @@ class VTK_SLICER_ANNOTATIONS_MODULE_LOGIC_EXPORT vtkSlicerAnnotationModuleLogic
...
 
   // Exit the place mode for annotations
-  void StopPlaceMode(bool persistent=false);
+  void StopPlaceMode(bool persistent=false, const char* viewNodeID = NULL);
```
jcfr: Update doxygen

Modules/Loadable/Annotations/Logic/vtkSlicerAnnotationModuleLogic.h
```diff
@@ -41,7 +41,7 @@ class VTK_SLICER_ANNOTATIONS_MODULE_LOGIC_EXPORT vtkSlicerAnnotationModuleLogic
   void AddNodeCompleted(vtkMRMLAnnotationNode* annotationNode);
 
   // Cancel the current annotation placement or remove last annotation node
-  void CancelCurrentOrRemoveLastAddedAnnotationNode();
+  void CancelCurrentOrRemoveLastAddedAnnotationNode(const char* viewNodeID = NULL);
```
jcfr: Update doxygen

Modules/Loadable/Markups/Logic/vtkSlicerMarkupsLogic.h
```diff
@@ -186,7 +186,7 @@ class VTK_SLICER_MARKUPS_MODULE_LOGIC_EXPORT vtkSlicerMarkupsLogic :
   /// place mode according to the persistent flag.
   /// Return true on successfully going into place mode, false otherwise.
   /// \sa SetActiveIDList
-  bool StartPlaceMode(bool persistent);
+  bool StartPlaceMode(bool persistent, const char* viewNodeID = NULL);
```
jcfr: Update doxygen

Modules/Loadable/Markups/Logic/vtkSlicerMarkupsLogic.cxx
```diff
@@ -338,9 +338,11 @@ void vtkSlicerMarkupsLogic::SetActiveListID(vtkMRMLMarkupsNode *markupsNode)
     if (activePlaceNodeClassName && strcmp(activePlaceNodeClassName, "vtkMRMLMarkupsFiducialNode") == 0)
       {
       selectionNode->SetActivePlaceNodeID(NULL);
-      vtkMRMLInteractionNode *interactionNode = vtkMRMLInteractionNode::SafeDownCast(this->GetMRMLScene()->GetNodeByID("vtkMRMLInteractionNodeSingleton"));
-      if (interactionNode && interactionNode->GetCurrentInteractionMode() == vtkMRMLInteractionNode::Place)
+      vtkSmartPointer<vtkCollection> interactionNodes = vtkSmartPointer<vtkCollection>::Take
```
jcfr: Here I reset the interaction mode for all interaction nodes.

Is it the right approach ?

Libs/MRML/Core/vtkMRMLAbstractViewNode.h
```diff
@@ -98,6 +99,11 @@ class VTK_MRML_EXPORT vtkMRMLAbstractViewNode
   vtkGetMacro(Visibility, int);
   vtkSetMacro(Visibility, int);
 
+  /// Get interaction node.
+  /// If no node has been explicitly set, it the singleton interaction node.
```
jcfr:
```
If no node has been explicitly set, it the singleton interaction node.
```
 
-> 

```
If no node has been explicitly set using SetInteractionNode() or SetInteractionNodeID(), return
the singleton interaction node.
```

jcfr:
Shortly I will update tests (or add new one) to exercise the methods added to  `qSlicerMouseModeToolBar`, `qMRMLSegmentEditorWidget` and `qSlicerMarkupsPlaceWidget`, then I will integrate.

@lassoan When you have a chance, I would appreciate a cursory check.

Modules/Loadable/Markups/Widgets/qSlicerMarkupsPlaceWidget.cxx
```diff
@@ -180,6 +180,25 @@ vtkMRMLMarkupsFiducialNode* qSlicerMarkupsPlaceWidget::currentMarkupsFiducialNod
...
+
+//-----------------------------------------------------------------------------
+void qSlicerMarkupsPlaceWidget::setInteractionNode(vtkMRMLInteractionNode* interactionNode)
+{
+  Q_D(qSlicerMarkupsPlaceWidget);
+  if (d->InteractionNode == interactionNode)
+    {
+    return;
+    }
+  this->qvtkReconnect(d->InteractionNode, interactionNode, vtkCommand::ModifiedEvent, this, SLOT(updateWidget()));
```
lassoan: We may need to call `setInteractionNode(NULL)` in the destructor to avoid memory leaks (there may be indirect mechanisms that unobserver the interaction node, so maybe it explicit reset in destructor may not be necessary).

jcfr: Since we use a weak pointer, there is need to  be explicit. 

Regarding the observation, the connection is removed when the widget is deleted. See https://github.com/commontk/CTK/blob/6cd82aab5f21db7c021e1eb502c9e5161e4a636d/Libs/Visualization/VTK/Core/ctkVTKConnection.cpp#L468-L473

That said, it is better to be explicit. :+1:

Libs/MRML/Core/vtkMRMLAbstractViewNode.h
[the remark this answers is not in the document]
jcfr: Here it is:
```
If no node has been explicitly set using SetInteractionNode() or SetInteractionNodeID(), return
the singleton interaction node.

The singleton interaction node is considered to be the default interaction node. Associating
a specific interaction node to one or a multiple views allows to control the interaction mode
associated with these views independently.

Since by default, the interaction node is a singleton, a new interaction node may be
created doing the following:

  vtkNew<vtkMRMLInteractionNode> interactionNode;
  interactionNode->SetSingletonOff()
  this->GetMRMLScene()->AddNode(interactionNode.GetPointer());

```

Libs/MRML/DisplayableManager/vtkMRMLAbstractDisplayableManager.cxx
```diff
@@ -690,7 +683,7 @@ vtkMRMLInteractionNode* vtkMRMLAbstractDisplayableManager::GetInteractionNode()
 //---------------------------------------------------------------------------
 vtkMRMLSelectionNode* vtkMRMLAbstractDisplayableManager::GetSelectionNode()
 {
-  return this->Internal->MRMLSelectionNode;
+  return vtkMRMLSelectionNode::SafeDownCast(this->GetMRMLScene()->GetNodeByID("vtkMRMLSelectionNodeSingleton"));
```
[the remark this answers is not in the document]
jcfr: I don't know if this makes sense for the selection node. The setting of active node is quite general and not specific to a particular view. May it could be associated with a layout instead ? 

Instead, I think we should move method like `vtkMRMLSelectionNode::AddNewPlaceNodeClassNameToList` to the interaction node. This will allow to customize the type of annotation that can be added in a specific views.

Modules/Loadable/Markups/Logic/vtkSlicerMarkupsLogic.cxx
cpinter: I'd say yes

Libs/MRML/DisplayableManager/vtkMRMLAbstractDisplayableManager.cxx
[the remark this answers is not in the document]
cpinter: Agreed, it seems to belong more to interaction than selection.

cpinter:
Looks great, thanks, @jcfr :+1:

jcfr:
Thanks for the review. Change have been integrated in [r27606](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27606)

In a follow up commit, I will:
* update all self test to use the loadable module base classes, and also add a new one checking that setting interaction node behave as expected.
* relocate `AddNewPlaceNodeClassNameToList`

PR #1049, opened 2018-12-05
BUG: Fix cutoff in DICOM browser toolbar

cpinter:
The DICOM browser's toolbar looks like this on Windows in 4.10:
image

It is not a good idea to set fixed number of pixels for anything in Qt5, and this negative margin does not seem to be safe any more.


pieper:
👍

cpinter:
Thanks for the quick reviews! Integrated in https://github.com/Slicer/Slicer/commit/5bcbf3a1c3aff71ef6297347ae48b06c2d7afa64

PR #1051, opened 2018-12-06
ENH: Add MRMLSliceNode property to allow show/hide of reformat widget outline

jcfr:


jcfr:
If the approach seems reasonable, I will then update the MRML files associated with the tests.

jcfr:
Closing. Integrated in [r27613](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27613)

PR #1053, opened 2018-12-07
ENH: Speedup (from 2x to 10x faster) and simplify python tests

jcfr:
* Run python self tests with a messageDelay of 100ms. Message delay is used
  in function "delayDisplay" to (1) let the event loop catch up to the state
  of the test and (2) to display a meaningful message when running test
  interactively from Slicer.

  When running tests on the dashboard, there is no need to wait too long
  before each test steps. This commit sets a default of 50ms when tests
  are executed from the module panel. It also provides a Slicer to change
  the value.

* Refactor tests to use ScriptedLoadableModule base classes

* Remove redundant test case AbdominalAtlasTest already coverted by
  AtlasTests

* Remove duplicated implementation of functions like takeScreenshot.


lassoan:
This would take up very valuable space in all module GUI and would be rarely used (you would not change the value very frequently). Could we moved this to application settings / Developer section?

jcfr:
Make sense. Will do.

jcfr:
Et voila, 

image

Associated tool-tip is: _Time to wait before resuming self-test execution and hiding messages displayed to the user_

jcfr:
Thanks for the review. Closing, integrated in [r27617](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27617) and [r27618](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27618)

PR #1054, opened 2018-12-11
ENH: Use subject hierarchy in Models module

cpinter:
- Only shows model and model hierarchy nodes (which are represented as SH folders)
- All plugins are disabled except for Models, Folder, Opacity (and Default)
- Apply color to branch action provided by the Folder plugin replaces the checkbox in model hierarchy
  - When the user changes the color of the folder then this option gets enabled automatically
  - Can be enabled/disabled using a checkable visibility action (right-click on eye icon or color)
  - If enabled, then visibility only applies to the folder, otherwise the default SH behavior is used (show/hide all items in whole branch)

### DIFF
--- a/Modules/Loadable/Models/Resources/UI/qSlicerModelsModuleWidget.ui
+++ b/Modules/Loadable/Models/Resources/UI/qSlicerModelsModuleWidget.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>615</width>
-    <height>664</height>
+    <height>790</height>
    </rect>
   </property>
   <property name="sizePolicy">
@@ -20,7 +20,16 @@
    <string>Models</string>
   </property>
   <layout class="QVBoxLayout" name="verticalLayout">
-   <property name="margin">
+   <property name="leftMargin">
+    <number>0</number>
+   </property>
+   <property name="topMargin">
+    <number>0</number>
+   </property>
+   <property name="rightMargin">
+    <number>0</number>
+   </property>
+   <property name="bottomMargin">
     <number>0</number>
    </property>
    <item>
@@ -28,11 +37,20 @@
      <property name="orientations">
       <set>Qt::Vertical</set>
      </property>
-     <property name="sizeGripInside" stdset="0">
+     <property name="sizeGripInside">
       <bool>false</bool>
      </property>
      <layout class="QVBoxLayout" name="verticalLayout_5">
-      <property name="margin">
+      <property name="leftMargin">
+       <number>0</number>
+      </property>
+      <property name="topMargin">
+       <number>0</number>
+      </property>
+      <property name="rightMargin">
+       <number>0</number>
+      </property>
+      <property name="bottomMargin">
        <number>0</number>
       </property>
       <item>
@@ -99,11 +117,11 @@
          </spacer>
         </item>
         <item>
-         <widget class="ctkSearchBox" name="ScrollToModelSearchBox">
+         <widget class="ctkSearchBox" name="FilterModelSearchBox">
           <property name="placeholderText">
-           <string>Scroll to...</string>
+           <string>Filter by name...</string>
           </property>
-          <property name="showSearchIcon" stdset="0">
+          <property name="showSearchIcon">
            <bool>true</bool>
           </property>
          </widget>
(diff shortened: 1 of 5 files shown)


cpinter:
In the past weeks I added many features to subject hierarchy to support this PR. Now we have a column for colors, we have an opacity plugin, etc.

This PR replaces the model hierarchy tree in the Models module. I tried to do everything so that it's as close as possible to the old way, but there are of course changes.
I made this video that demoes the new tree:
https://drive.google.com/open?id=1n1w5mTWJu0nN89XioGPe3Em2Yb3sra94
What I forgot to include in the video is the opacity. There is no column for opacity in the new SH tree (the 1.00s looked strange and it was buggy too - flickering etc), but instead there is a plugin for it, and it can be set as before in the Display section. This is how the plugin looks like:
image

Please let me know what you think and I'd appreciate if someone familiar with the fiber related features could test it.
Thanks!

cpinter:
And also why I made all this :)

Other than "fixing" the bugs in the original scene model, this basically gives better support for atlases:
- Terminology can be associated to models
- Reorder is possible
- Saving branch expanded state is possible

jcfr:
I am curious to see how more complex atlas (e.g SPL atlas) would load.

@mhalle Since you have been working on the open anatomy browser, would you have any comment ?

@jamesjcook This may be relevant for your application

jamesjcook:
Thanks for the Ping JC. After some moderate growing pains, and some help from the discourse and other devs, I'm using the subject hierarchy now for our model and tractography display protocol.

pieper:
makes sense to me - thanks for working on this 👍

jcfr:
I am also wondering if performance are improved with large atlases

lassoan:
Fixes https://issues.slicer.org/view.php?id=4626

cpinter:
The folder show/hide operation is much slower with the SH-based one, because by default it turns on slice intersections as well. Without slice intersections (with Apply color to branch turned on and slice intersection off for the folder) it is much faster though (0s vs 1s).

What do you think about an option in Application Settings for visibility of slice intersections? By default it could be off. However I'm not sure how to make it easy for the user to realize that there is a setting for this. I personally hated that in the Models module it was only possible to turn on slice intersections and that's why I implemented SH visibility so that it sets Visibility and also SliceIntersectionVisibility. Or maybe it's time for the separate 2D and 3D visibility columns? It wouldn't be hard to implement.
@lassoan @pieper @jcfr @mhalle @ihnorton

lassoan:
I like how it is configurable for Segmentations: you specify 2D and 3D display options in various flags and enums (visibility, color, projection mode, intersection thickness, etc.) and you have an additional global visibility flag to control overall visibility. If this was a backward-incompatible change, I think we could still do it (in Slicer5).

cpinter:
Now that 4.10.1 is out, I can integrate this change soon I think.

The only outstanding issue is the slowness of showing huge branches due to the slice intersections. @pieper @lassoan Have you had the chance to talk about this in Las Palmas?
I can either add an application setting for the slice intersections, or do a reform of the visibility controls in SH, whichever seems the best to everyone.

lassoan:
I think the best solution would be to have overall, slice, and 3D visibility options. The eye icon in the visibility column in SH would only change the overall visibility flag (therefore it would not turn on slice intersections).

cpinter:
So you'd add two more visibility columns? I guess we can have them hidden by default and have a checkbox for their visibility, similarly to the ID column.

lassoan:
For now, we could keep the single visibility column in the SH table (it would control the overall visibility), and add expose all 3 visibilities (overall, slice, 3D) in Models module.

Later we could reconsider how to make the 3 visibility flags conveniently accessible in SH.

pieper:
It would be nice if there were a way to automatically turn off the slice
intersections and/or put up a progress bar (maybe a dialog that has a
button that offers to cancel and turn them off) if we detect the render is
taking too long. But lacking that I think an option to globally turn them
on and off.  That would allow turning them off for fast interactions but
then turning them back on without losing any per-segment visibility
configurations.  Or if you feel it's easier to keep this at the node level
that's fine too.

cpinter:
The solution we're proposing:
* Change SH behaviour so that it does not set slice intersection visibility
* Change display nodes (base class and up) so that they have a 2D and 3D visibility
* Add a simple Visibility plugin in SH that offers toggling 2D and 3D visibility separately from the visibility context menu (right-click eye)

If this sounds good then I'll start doing it today.

jcfr:
Thanks for the detailed summary.

> Change display nodes (base class and up) so that they have a 2D and 3D visibility

What would be the plan to implement this ? 

Is the idea to leverage `SetViewNodeIDs` ?

cpinter:
ViewNodeIDs should work the same as before.

The idea is to do it similarly to how it works for segmentations
https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/Segmentations/MRMLDM/vtkMRMLSegmentationsDisplayableManager3D.cxx#L174

lassoan:
The display node's "overall visibility" would be decoupled from the "2D visibility" and "3D visibility", as it is already done for segmentations:
- The node is visible in 2D if and only if both overall visibility and 2D visibility are enabled.
- The node is visible in 3D if and only if both overall visibility and 3D visibility are enabled.

The advantage of this decoupling is that we can show/hide of a node without changing its 2D/3D visibility flags.

cpinter:
Would it be possible to merge this PR so that I have an easier time with doing the display node and DM rework? Models module would work pretty well in general during this one week (I hope) period, except that atlas folder show/hide would be slower.

cpinter:
@jcfr @pieper what do you think about integrating this? (see previous comment)

jcfr:
> Would it be possible to merge this PR
> I have an easier time with doing the display node and DM rework?

Sounds good to me.  I don't have the cycle to do thorough testing but the outlined plan seems very reasonable.

cpinter:
Thanks, Jc! If there are no objections I'll merge this over the weekend. In the meantime I'm making progress with the display nodes and the 2D/3D visibilities.

cpinter:
Integrated in https://github.com/Slicer/Slicer/commit/e9f25f8c980e6e9b933518bb472f89858b3f0fe3

PR #1055, opened 2018-12-11
BUG: ComputeJacobianWithRespectToPosition parameter mismatch

hjmjohnson:
Remove functions that do not override, and are not required

The use of ITK_OVERRIDE identified a parameter type mismatch.  Both ITKv4 and
ITKv5 take an InputPointType as the first parameter.

Remove the set of functions intended to override abstract base classes do not actually
match the signature of existing functions in ITKv5 or ITKv4.

### DIFF
--- a/Libs/MRML/Core/vtkITKTransformInverse.h
+++ b/Libs/MRML/Core/vtkITKTransformInverse.h
@@ -62,15 +62,25 @@ namespace itk
       {
       itkExceptionMacro( "Only storage methods are implemented for InverseBSplineTransform" );
       }
+
     virtual void ComputeJacobianWithRespectToPosition(
         const typename Superclass::InputPointType  &,
+#if ITK_VERSION_MAJOR < 5
         typename Superclass::JacobianType &) const ITK_OVERRIDE
+#else
+        typename Superclass::JacobianPositionType &) const ITK_OVERRIDE
+#endif
       {
       itkExceptionMacro( "Only storage methods are implemented for InverseBSplineTransform" );
       }
+
     virtual void ComputeInverseJacobianWithRespectToPosition(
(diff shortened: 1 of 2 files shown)


jcfr:
Thanks @hjmjohnson

hjmjohnson:
@jcfr the C++ ITK_OVERRIDE correctly caused a compiler error stating that the function was not overridden.  That's what prompted the recognition that the signatures did not match, and the functions were not used :).

hjmjohnson:
@lassoan Do you have any concerns about this PR?

lassoan:
The goal of this wrapper class is to allow file IO but prevent accidentally using any computations on the transform (since those are not implemented). If ComputeJacobianWithRespectToParameters method is not available anymore in ITK transform classes then this override should be removed (not just #ifdef-d out). It the method signature has changed then the code should be updated to the new signature to avoid potential misuse.

It would be really useful to allow lossless storage of inverted transforms natively in ITK (it would be just a flag that would indicate that the inverse of the transform should be used). Then it would not be necessary to have such wrapper classes.

hjmjohnson:
This should be ready to merge now.  Removed items that do not overload parent member functions (due to wrong signatures).

phcerdan:
ITKv5 changed the signature from JacobianType to JacobianPositionType.
I changed signatures depending on the version of ITK, instead of completely disabling it.

However, removing some methods from here might be a good choice, most are not implemented in ITK. But this could be done later.

I pushed into @hjmjohnson branch. I backed up @hjmjohnson code here: https://github.com/phcerdan/Slicer/tree/backup_remove-unused-methods

phcerdan:
Superseeded in #1096. Please close.

hjmjohnson:
Moved to #1096

PR #1056, opened 2018-12-11
ENH: Future proof null pointer behavior for ITKv5

hjmjohnson:
Enhancements in nullptr behavior in ITKv5 provide more clear
type checking and respect the nullptr identifier.  The 'long 0' value
known as NULL causes an abiguity for overload compilations of the ITKv5 smartpointers.

Using ITK_NULLPTR is both backwards and forwards compatible.


hjmjohnson:
This is backwards compatible with ITKv4.

jcfr:
Thanks. Locally tested and now integrated as [r27622](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27622)

PR #1057, opened 2018-12-11
ENH: Support ITKv5 new threading models

hjmjohnson:
ITKv5 reorganized the multi-threading capabilities of ITK into
a more modular organization and naming to support several modern
parallism paradigms.


hjmjohnson:
This is backwards compatible with ITKv4

phcerdan:
Superseeded in #1096

PR #1058, opened 2018-12-11
ENH: If building VTK 8 with C++11 enable CXX11_FEATURES.

hjmjohnson:


hjmjohnson:
This should be backwards compatible with current build system.

SuperBuild/External_VTK.cmake
```diff
@@ -92,6 +92,11 @@ if((NOT DEFINED VTK_DIR OR NOT DEFINED VTK_SOURCE_DIR) AND NOT Slicer_USE_SYSTEM
       #-DVTK_USE_RPATH:BOOL=ON # Unused
       )
   endif()
+  if("${Slicer_VTK_VERSION_MAJOR}" GREATER_EQUAL "8")
+    if (CMAKE_CXX_STANDARD GREATER_EQUAL 11)
+      list(APPEND EXTERNAL_PROJECT_OPTIONAL_CMAKE_CACHE_ARGS -DVTK_USE_CXX11_FEATURES:BOOL=ON)
```
jcfr: I couldn't find reference to the variable `VTK_USE_CXX11_FEATURES` or the string `_USE_CXX11_FEATURES` in either VTK or CMake.

hjmjohnson:
I got the 100% backwards!  (too many build trees)



~/Dashboard/src/VTK **((v7.1.1))**
```
johnsonhj@hj$ git grep CXX11_FEATURES
CMake/VTKConfig.cmake.in:set(VTK_USE_CXX11_FEATURES @VTK_USE_CXX11_FEATURES@)
CMake/VTKConfig.cmake.in:if (VTK_USE_CXX11_FEATURES)
CMakeLists.txt:  option(VTK_USE_CXX11_FEATURES "Build VTK using C++11 features (if available)"
CMakeLists.txt:  set(VTK_USE_CXX11_FEATURES OFF)
CMakeLists.txt:if (VTK_USE_CXX11_FEATURES)
```
I recommend that this NOT be added.  if we are moving to VTK8.  Will VTK7 still be supported for a while?

If we need both VTK7 and ITKv5 support then I'll push a corrected version where line 95 state EQUAL 7.

jcfr:
I see.

Since VTK7 is always built with C++98, Qt4 ... and will be dropped after Slicer 4.10.1 (last release is 4.10), we will not support ITKv5 in that case.

PR #1061, opened 2018-12-13
ENH: Create new view to embed DICOM browser in main widget

Sunderlandkyl:
Accomplished through the use of the new view factory qSlicerSingletonViewFactory.
Both C++ and Python classes can register any singleton QWidget by calling factory->setWidget(QWidget*) and factory->setTagName(std::string).

The DICOM browser layout is registered as a single view which spans the whole widget.
Navigating to the DICOM browser module will change to this new layout and hide the data probe, while navigating to a different widget, loading a series (with browser persistence off), or clicking the new "Close" button, will restore the previous layout and settings.

See discussion: https://discourse.slicer.org/t/dicom-browser-is-stuck-behind-main-window-after-dicom-import/4826/24

image

### DIFF
--- a/Base/QTGUI/qSlicerSingletonViewFactory.cxx
+++ b/Base/QTGUI/qSlicerSingletonViewFactory.cxx
@@ -36,8 +36,12 @@ public:
 
   virtual void init();
 
-  QSharedPointer<QWidget> Widget;
+  // Widget returned from createViewFromXML
+  QWidget* Widget;
+  // Tag used for widget in createViewFromXML
   QString TagName;
+  // Internal QWidget used to hold the widget until it is taken over by the layout manager
+  QSharedPointer<QWidget> InternalWidget;
 
 protected:
   qSlicerSingletonViewFactory* q_ptr;
@@ -49,12 +53,17 @@ protected:
 qSlicerSingletonViewFactoryPrivate
 ::qSlicerSingletonViewFactoryPrivate(qSlicerSingletonViewFactory& object)
   : q_ptr(&object)
+  , Widget(nullptr)
+  , InternalWidget(new QWidget())
 {
 }
 
 //-----------------------------------------------------------------------------
 qSlicerSingletonViewFactoryPrivate::~qSlicerSingletonViewFactoryPrivate()
-= default;
+{
+  Q_Q(qSlicerSingletonViewFactory);
+  q->setWidget(nullptr);
+}
 
 //-----------------------------------------------------------------------------
 void qSlicerSingletonViewFactoryPrivate::init()
@@ -78,19 +87,39 @@ qSlicerSingletonViewFactory::qSlicerSingletonViewFactory(QObject* parent)
 qSlicerSingletonViewFactory::~qSlicerSingletonViewFactory()
 = default;
 
-
 //-----------------------------------------------------------------------------
 QWidget* qSlicerSingletonViewFactory::widget()
 {
   Q_D(qSlicerSingletonViewFactory);
-  return d->Widget.data();
+  return d->Widget;
 }
 
 //-----------------------------------------------------------------------------
 void qSlicerSingletonViewFactory::setWidget(QWidget* widget)
 {
   Q_D(qSlicerSingletonViewFactory);
-  d->Widget = QSharedPointer<QWidget>(widget);
+  if (d->Widget == widget)
+    {
+    return;
+    }
+
+  if (d->Widget)
+    {
+    QObject::disconnect(d->Widget, &QWidget::destroyed, this, &qSlicerSingletonViewFactory::onWidgetDestroyed);
+    }
+
+  d->Widget = widget;
+  if (d->Widget)
+    {
+    d->Widget->setParent(d->InternalWidget.data()); // Hold the widget in the internal widget until the layout manager can take it
+    QObject::connect(d->Widget, &QWidget::destroyed, this, &qSlicerSingletonViewFactory::onWidgetDestroyed);
+    }
+}
+
+//-----------------------------------------------------------------------------
+void qSlicerSingletonViewFactory::onWidgetDestroyed()
+{
+  this->setWidget(nullptr);
 }
 
 //-----------------------------------------------------------------------------
@@ -119,7 +148,7 @@ QWidget* qSlicerSingletonViewFactory::createViewFromXML(QDomElement layoutElemen
 {
   Q_UNUSED(layoutElement);
   Q_D(qSlicerSingletonViewFactory);
-  if (this->widget()->isVisible())
+  if (d->Widget && d->Widget->isVisible())
     {
     qCritical() << "qSlicerSingletonViewFactory::createViewFromXML - Widget for view \"" << d->TagName << "\" is already in use within the current layout!";
     }
--- a/Base/QTGUI/qSlicerSingletonViewFactory.h
+++ b/Base/QTGUI/qSlicerSingletonViewFactory.h
@@ -57,6 +57,9 @@ public:
   /// Get the XML tag that identifies the view where the widget should be placed
   Q_INVOKABLE QString tagName();
 
+public slots:
+  virtual void onWidgetDestroyed();
+
 protected:
   QScopedPointer<qSlicerSingletonViewFactoryPrivate> d_ptr;
 
--- a/Modules/Scripted/DICOM/DICOM.py
+++ b/Modules/Scripted/DICOM/DICOM.py
@@ -38,6 +38,10 @@ This work is supported by NA-MIC, NAC, BIRN, NCIGT, and the Slicer Community. Se
     self.parent.icon = qt.QIcon(':Icons/Medium/SlicerLoadDICOM.png')
     self.parent.dependencies = ["SubjectHierarchy"]
 
+    self.detailsPopup = None
+    self.viewWidget = None
+    self.browserSettingsWidget = None
+    self.currentViewArrangement = 0
 
     # Tasks to execute after the application has started up
     slicer.app.connect("startupCompleted()", self.performPostModuleDiscoveryTasks)
@@ -45,6 +49,11 @@ This work is supported by NA-MIC, NAC, BIRN, NCIGT, and the Slicer Community. Se
   def setup(self):
     pluginHandlerSingleton = slicer.qSlicerSubjectHierarchyPluginHandler.instance()
     pluginHandlerSingleton.registerPlugin(slicer.qSlicerSubjectHierarchyDICOMPlugin())
+
+    self.viewFactory = slicer.qSlicerSingletonViewFactory()
+    self.viewFactory.setTagName("dicombrowser")
+    if slicer.app.layoutManager() is not None:
+      slicer.app.layoutManager().registerViewFactory(self.viewFactory)
 
   def performPostModuleDiscoveryTasks(self):
     """Since dicom plugins are discovered while the application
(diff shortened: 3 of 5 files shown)


Base/QTGUI/qSlicerSingletonViewFactory.cxx
```diff
@@ -0,0 +1,112 @@
...
+  return QStringList() << d->TagName;
+}
+
+//---------------------------------------------------------------------------
+QWidget* qSlicerSingletonViewFactory::createViewFromXML(QDomElement layoutElement)
+{
+  Q_D(qSlicerSingletonViewFactory);
+  if (layoutElement.tagName() == d->TagName)
+    {
+    //return d->DicomBrowserWidget;
```
lassoan: remove

Base/QTGUI/qSlicerSingletonViewFactory.h
```diff
@@ -0,0 +1,58 @@
...
+#include "ctkLayoutViewFactory.h"
+
+// QtGUI includes
+#include "qSlicerBaseQTGUIExport.h"
+
+class ctkDICOMBrowser;
+class qSlicerSingletonViewFactoryPrivate;
+
+class Q_SLICER_BASE_QTGUI_EXPORT qSlicerSingletonViewFactory : public ctkLayoutViewFactory
+{
```
lassoan: Add documentation, most importantly, give a high-level summary of why this is useful and how it is supposed to be used.

Modules/Scripted/DICOM/DICOM.py
```diff
@@ -369,10 +400,34 @@ def setup(self):
...
+    if savedViewArrangement != self.DICOM_BROWSER_LAYOUT_ID:
+
+      centralWidget = slicer.util.mainWindow().findChild("QWidget", "CentralWidget")
+      dataProbe = slicer.util.mainWindow().findChild("QWidget", "DataProbeCollapsibleWidget")
+
+      self.wasCentralWidgetAutoFillBackground = centralWidget.autoFillBackground
+      self.wasDataProbeVisible = dataProbe.isVisible()
+      self.savedViewArrangement = savedViewArrangement
+
+      centralWidget.setAutoFillBackground(False)
```
lassoan: Why autofill background needs to be disabled?

Modules/Scripted/DICOM/DICOM.py
```diff
@@ -329,7 +343,24 @@ def setup(self):
...
+    # Create widget and add a new DICOM browser layout
+    self.viewFactory = slicer.qSlicerSingletonViewFactory()
+    self.viewFactory.setTagName("dicombrowser")
+
+    self.detailsPopup = None
+    self.setDetailsPopup(self.getSavedDICOMDetailsWidgetType()())
+
+    layoutManager = slicer.app.layoutManager()
+    layoutManager.registerViewFactory(self.viewFactory)
+    self.DICOM_BROWSER_LAYOUT_ID = 2200
```
lassoan: It would be better to define this layout ID as a constant outside of the class so that you don't need to instantiate a class to access it.

Modules/Scripted/DICOM/DICOM.py
```diff
@@ -329,7 +343,24 @@ def setup(self):
...
-    self.detailsPopup = self.getSavedDICOMDetailsWidgetType()()
+    # Create widget and add a new DICOM browser layout
+    self.viewFactory = slicer.qSlicerSingletonViewFactory()
+    self.viewFactory.setTagName("dicombrowser")
+
+    self.detailsPopup = None
+    self.setDetailsPopup(self.getSavedDICOMDetailsWidgetType()())
+
+    layoutManager = slicer.app.layoutManager()
+    layoutManager.registerViewFactory(self.viewFactory)
```
lassoan: The layout should be registered right at application startup, because if the scene is saved with DICOM layout active then when we open the scene it would try to switch to that layout.

By the way - does the browser work correctly if someone just activates the DICOM layout but does not switch to the DICOM module?

Modules/Scripted/DICOM/DICOM.py
lassoan: Instead of manipulating the parent widget's AutoFillBackground property we could try to set a background for the DICOM widget: "When a widget has a style sheet with a valid background or a border-image, this property is automatically disabled." (http://doc.qt.io/qt-5/qwidget.html#autoFillBackground-prop)

Modules/Scripted/DICOM/DICOM.py
jcfr: ### the layout base id approach

For layout id, I am wondering if we should allocate 100 slots per extension or module or domain, and come up with hash function returning an integer and multiply it by 100:

```
unsigned int ComputeLayoutBaseId(const std::string& domain)
```

That way for any given module or extension, we could be sure to have a unique base ID and we would avoid clash.

layout specific to a module/extension would then be indexed relative to their layout base id.

### explicit registry

An other approach would be to allocate layout base id and keep track of assignment in a shared document.

Modules/Scripted/DICOM/DICOM.py
jcfr: When I work on the ShapePopulationViewer, I didn't bother generalizing the approach  ... I wonder if we should improve the layout manager so that it takes care of managing this ?

```
      "<layout type=\"horizontal\">"
      " <item autoFillBackground=\"false\">"
      "  <dicombrowser></dicombrowser>"
      " </item>"
      "</layout>"
```

lassoan: The widget (dicombrowser) knows if it needs autofillbackground or not, so if we really need to control this manually then the widget could have a get function or property that would tell if it requires background fill.

Do you know why dicombrowser is a special case that needs this autofill property manipulation? What happens if we don't set it? Can we set background color using style sheets instead?

Modules/Scripted/DICOM/DICOM.py
Sunderlandkyl: Yes, calling set layout and calling a layout that has the dicombrowser tag displays a functional DICOM browser widget.

lassoan: Isn't it a problem that autofillbackground is not set in this case?

Modules/Scripted/DICOM/DICOM.py
lassoan: Choosing to use integer layout identifiers was a mistake. We cannot correct it in Slicer-4.10.x but we should switch to string IDs in Slicer5. String IDs can may be prefixed with modulename+"." as we do it for singleton tags and node attributes. We may remain somewhat backward compatible by having SetLayoutID(int) method that maps known layout integer IDs to the new string IDs.

Sunderlandkyl:
Would it make sense to allow each instance of qSlicerSingletonViewFactory to create multiple different types of widgets?

We could call qSlicerSingletonViewFactory::adWidget(QWidget* name, QString tag):

Ex.
````
factory->addWidget(widget1, "widget1");
factory->addWidget(widget2, "widget3");
factory->addWidget(widget3, "widget3");
````

Rather than creating a factory for each one, a module may want to create a single factory holding all of the desired singleton view widgets.

File: Modules/Scripted/DICOM/DICOM.py (hunk 3 above)
Comment (Sunderlandkyl) [reply Modules/Scripted/DICOM/DICOM.py]: You're right, setting autoFillBackground to false doesn't seem to be required in this case.

**Comment by Sunderlandkyl** [comment]
No, I can't think of any specific use cases.

Next question: should the option be provided to continue to use the existing pop-up window? (Using a checkbox/QSettings)

**Comment by lassoan** [comment]
> should the option be provided to continue to use the existing pop-up window? (Using a checkbox/QSettings)

Since the current behavior is not good (stuck behind on MacOS) and we plan to make further improvements, which may be more complicated if we want to keep both modes working, I don't think we need the compatibility option.

**Comment by pieper** [comment]
> Since the current behavior is not good (stuck behind on MacOS) and we plan to make further improvements, which may be more complicated if we want to keep both modes working, I don't think we need the compatibility option

There was a user/use case where people wanted the dicom browser on one monitor in persistent mode while interacting in the main monitory.  I haven't been following this discussion closely but would that still be supported in your new design?

File: Base/QTGUI/qSlicerSingletonViewFactory.h (hunk 2 above)
Comment (jcfr) [reply Base/QTGUI/qSlicerSingletonViewFactory.h]: And make sure to define the meaning of "Singleton" in this context.

**Comment by lassoan** [comment]
Multi-monitor support would come with multi-monitor support of view layouts. Modules could also display a browser in a popup.

**Comment by jcfr** [comment]
The `qSlicerSingletonViewFactory` is useful in itself, if there are no consensus regarding the update of the DICOM module, I suggest we split the PR and move forward with the integration of  `qSlicerSingletonViewFactory`

**Comment by Sunderlandkyl** [comment]
Sounds good, I'll create a separate pull request for qSlicerSingletonViewFactory.

**Comment by Sunderlandkyl** [comment]
I've rebased this change onto the latest master.

File: Modules/Scripted/DICOMLib/DICOMWidgets.py [hunk 6]
```diff
@@ -81,6 +81,8 @@ class DICOMDetailsBase(VTKObservationMixin, SizePositionSettingsMixin):
   """
 
   widgetType = None
+  loadingFinished = qt.Signal()
```
jcfr: Are these signal used ? If yes, it may be worth documenting both `loadingFinished` and `closed` in the class description

lassoan:
I've pushed a commit on my branch that would be a step towards the final GUI that you talked about (with a subject hierarchy tree on the left): https://github.com/lassoan/Slicer/tree/dicom_browser_view

Sunderlandkyl:
Updated this pull request to include the changes that you made @lassoan, as well as moved the browser settings (database directory and table density) to the left panel. Also made some fixes to the qSlicerSingletonViewFactory class.

Here's what it looks like now:
image

lassoan:
Thank you, this will be awesome!

Few minor things (except the first one, only fix those that are not hard to implement):
- Tables and menu are shown twice
- Could we hide the node IDs in "Loaded data" tree?
- Could you collapse "Browser settings" by default?
- Change title of the collapsible section that contains the "Loaded data" tree from "DICOM database" to something else. Maybe just "Data"? Or put the data tree outside of collapsible sections?
- Move out "Show DICOM Browser" from the collapsible section.
- Make "Show DICOM Browser" a toggle button (it would remain pushed as long as the browser is visible and released when the DICOM browser gets hidden)
- Move "Horizontal" and "Browser persistent" to Browser settings section (they are not used frequently, so it would be better if they did not clutter the GUI)

jcfr:
I also wonder if the "close" button at the bottom should be removed

lassoan:
If the toggle button is made prominent enough then I agree that we don't really need the Close button.

Sunderlandkyl:
Great, thanks for the feedback!

The double table issue is already fixed, I just forgot to update the image, and the "Browser settings" groupbox is already closed by default.

I'll make the rest of the changes, including removing the close button, and update the pull request.

Sunderlandkyl:
Ok, changes made. Here is the new layout:

image

lassoan:
This looks awesome. I'm just wondering if we should change the label of the browser toggle button to "Show DICOM Database", since that is the window title of the browser. Also, since it is a toggle button, I think we can keep displaying the "Show" verb (regardless the button is pressed or not).

cpinter:
I like it! I always wonder if there is a toggle button that says show/hide, how people would interpret it. What you did Kyle is what I usually do as well, just not sure if it's the right way. I mean that when it's pushed, and it says Hide, then would the user think that it would hide, or that hide was activated.

cpinter:
Right, @lassoan had the same idea :)

Sunderlandkyl:
@lassoan "Show DICOM Database" title case or "Show DICOM database" sentence case.

lassoan:
It should probably have the same capitalization as the window title of the browser. Maybe "DICOM database" title and "Show DICOM database browser" button text.

Sunderlandkyl:
OK, sounds good. Title/button text changed.

lassoan:
This works really well! Merged r28447.

cpinter:
Great :) Looking forward to using it!

PR #1062, opened 2018-12-14
Improve sampledata downloadfromsource api

jcfr:
This changes are first step toward improving testing infrastructure discussed in https://discourse.slicer.org/t/improving-testing-data-management-for-self-test/5014/6


pieper:
LGTM 👍

Applications/SlicerApp/Testing/Python/JRC2013Vis.py
```diff
@@ -319,19 +304,15 @@ def test_Part2Head(self):
     #
     # first, get some data
     #
-    import urllib
-    downloads = (
-        ('http://slicer.kitware.com/midas3/download?items=8609', '3DHeadData.mrb', slicer.util.loadScene),
+    from SampleData import SampleDataSource, SampleDataLogic
+    dataSources = [
```
lassoan: Most of the time we download a single data source, so creating a list of sources is unnecessary complication. Could we change the code to this simpler pattern?

```
from SampleData import SampleDataSource, SampleDataLogic
SampleDataLogic().downloadFromSource(SampleDataSource(
  fileNames='3DHeadData.mrb', 
  uris='http://slicer.kitware.com/midas3/download?items=8609', 
  loadFileType='SceneFile'))
```

Applications/SlicerApp/Testing/Python/JRC2013Vis.py
```diff
@@ -476,19 +453,15 @@ def test_Part4Lung(self):
...
     #
-    import urllib
-    downloads = (
-        ('http://slicer.kitware.com/midas3/download?items=8612', 'LungData.mrb', slicer.util.loadScene),
+    from SampleData import SampleDataSource, SampleDataLogic
+    dataSources = [
+      SampleDataSource(
+        fileNames='LungData.mrb',
+        uris='http://slicer.kitware.com/midas3/download?items=8612',
+        loadFileType='SceneFile'
```
lassoan: I think we already have a mechanism to guess based on file extension. Could we use that by default and require specifying loadFileType only if the default is not good?

Applications/SlicerApp/Testing/Python/JRC2013Vis.py
jcfr: Agreed. Just wanted to get some feedback ... before simplifying.

What about that then:

```
SampleDataLogic().download(
  fileNames='3DHeadData.mrb', 
  uris='http://slicer.kitware.com/midas3/download?items=8609', 
  loadFileType='SceneFile')
```

jcfr: Also we call the function `downloadFromSource` (or `download`) but most of the time it is loaded in Slicer.

Should instead of the method be named
* `SampleData.loadFromSource` or `SampleData.downloadAndLoadFromSource`
* `SampleData.load` or `SampleData.downloadAndLoad`

lassoan: If the input of the method is not a `DataSource` object then we don't need to include that in the method name. So, `SampleData.load` would work well.

Probably `SampleData.loadData` or ``SampleData.loadFromURL` would be even better because it would not suggest that the input is some sample data.

`SampleData.load` could be used for loading sample data, such as `SampleData.load('MRHead')`.

lassoan: What do you think about specifying a helper function at the module level to further simplify syntax?

```
import SampleData
SampleData.loadFromURL(…)
```

jcfr: Or we could have methods like this directly in SampleData:

```
def loadFromURL(...):
  return SampleDataLogic().loadFromURL()

def downloadMRHead():
  return SampleDataLogic().downloadMRHead()

[...]
```

jcfr: That said, I just realized that `downloadMRHead` and a like are deprecated. And that the new modern approach is to use `downloadSample("MRHead")`, it means your suggestion would work very well

jcfr:
@lassoan I reviewed the approach. Note that I will squash the last commits before integrating ... 

In a nutshell:

* add support for guessing loadNodeTypes ... it is still suboptimal because I am not using any Slicer logic. 

* renamed `downloadFromDataSource` to `downloadFromURL`

* added convenience method directly into SampleData

* updated code to make use of the convenience methods. It now looks like this:
```
import SampleData
SampleData.downloadFromURL(...)
mrhead = SampleData.downloadSample('MRHead')[0]
```

Applications/SlicerApp/Testing/Python/SlicerBoundsTest.py
```diff
@@ -71,8 +71,8 @@ def test_Volume(self):
     """ Test the GetRASBounds & GetBounds method on a volume.
     """
     #self.delayDisplay("Starting test_Volume")
-    from SampleData import SampleDataLogic
-    volumeNode = SampleDataLogic().downloadAbdominalCTVolume()
+    import SampleData
+    volumeNode = SampleData.downloadSample('CTA abdomen\n(Panoramix)')[0]
```
jcfr: @lassoan  I don't link this name including a newline ... I think we should have a SampleName and SampleDescription.

Modules/Loadable/Markups/Testing/Python/NeurosurgicalPlanningTutorialMarkupsSelfTest.py
```diff
@@ -143,8 +143,8 @@ def run(self):
 
     # use the sample data module logic to load data for the self test
     self.delayDisplay("Getting Baseline volume")
-    from SampleData import SampleDataLogic
-    baselineVolume = SampleDataLogic().downloadWhiteMatterExplorationBaselineVolume()
+    import SampleData
+    baselineVolume = SampleData.downloadSample('BaselineVolume')[0]
```
jcfr: This SampleName is not very explicit .. should we rename it ?

PR #1065, opened 2018-12-19
ENH: Added option to load MRB files from Sample Data module

smrolfe:
In the Sample Data module, scene files (.mrb or .mrml) are not loaded into Slicer by default. This default could not be changed when registering a custom data set. Now a parameter has been added to specify if the files should be loaded for a custom data set.


jcfr:
Thanks @smrolfe :+1: 

The change makes sense.

jcfr:
I will integrate the change and also update the docstring.

smrolfe:
Great! thanks @jcfr

PR #1072, opened 2019-01-10
ENH: Add scripted module template with Qt Designer GUI

lassoan:
Added a scripted module template that creates module GUI from .ui file generated by Qt Designer.

Todo:
- [ ] bundle Qt Designer in installation package
- [ ] add button to Reload&Test section to load .ui file in Qt Designer


pieper:
Looks great 👍

lassoan:
Thanks for the feedback, I've merged this.

@jcfr if you could make designer.exe available in the install tree (with --designer option in the launcher) then it would be great. As is now, the "Edit UI" button only shows up if Slicer is launched from the build tree.

PR #1073, opened 2019-01-11
ENH: Update level tracing effect on mouse wheel event

vovythevov:
This allows the level tracing effect to be updated when the user scrolls
through the slices.

On the implementation:
Note that the method processViewNodeEvents is used instead of adding the
events with the MouseMoveEvent as the MouseMoveEvent eats the event
(abortEvent == True) which prevent the slice from scrolling.
If the MouseWheelForwardEvent and MouseWheelBackwardEvent are given their
own case (without abortEvent == True) weird interaction happen as the
effect preview is seemingly called before the slice has scrolled to the new
offset.


vovythevov:
Thanks for the review !
Intergrated in http://svn.slicer.org/Slicer4/trunk@27684

ihnorton:
Thanks for fixing this!

PR #1076, opened 2019-01-15
Constrain fiducials (WIP, do not merge)

pieper:
This works pretty well, and can be a placeholder until the more comprehensive markup overhaul is ready.

You might want to try it and see if we can figure out how to make the fiducial 'stick' to the mouse, but if not I think okay as-is and better than the default behavior.  I'd be tempted to include it as-is in the 4.10.2 patch release.

Note that I looked into making it possible to toggle this behavior, but didn't see how to get access to state from a QSetting into a vtkMRMLDisplayableManager (some options discussed in the commit message).  @jcfr or @lassoan any thoughts?

### DIFF
--- a/Base/QTGUI/Resources/UI/qSlicerSettingsViewsPanel.ui
+++ b/Base/QTGUI/Resources/UI/qSlicerSettingsViewsPanel.ui
@@ -312,6 +312,34 @@ p, li { white-space: pre-wrap; }
         </layout>
        </item>
        <item>
+        <layout class="QHBoxLayout" name="threeDHorizontalLayout">
+         <item>
+          <widget class="QLabel" name="ThreeDViewLabel">
+           <property name="text">
+            <string>Constrain fiducials to 3D surface:</string>
+           </property>
+          </widget>
+         </item>
+         <item>
+          <widget class="QCheckBox" name="ThreeDConstrainMarkupsCheckBox">
+           <property name="sizePolicy">
+            <sizepolicy hsizetype="Expanding" vsizetype="Fixed">
+             <horstretch>0</horstretch>
+             <verstretch>0</verstretch>
+            </sizepolicy>
+           </property>
+           <property name="toolTip">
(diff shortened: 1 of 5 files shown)


pieper:
Just tried this with multiple fiducials and it mostly works, but when adjusting a fiducial the most recently updated one may also move by a small amount, so there's still some debugging and state management to look into here.  The mrml and widget may be "fighting" over where the fiducial really is.

Might need to go back to using a [vtkPointPlacer](https://vtk.org/doc/nightly/html/classvtkPointPlacer.html) of some kind, either a standard one or a custom one.

lassoan:
Davide's complete markup redesign branch will be ready for integration in within 1-2 weeks. If you can make any fixes or improvements in 4.10.2 then it's fine but I would only work on the redesigned infrastructure.

It would be great if we could incorporate the results of your prototyping into the new infrastructure. The new design is hugely improved, mainly by eliminating several extra layers that added complexity, instability, and decreased performance.

pieper:
Is @punzo ready to share anything yet so we can have a look?  Being able to implement constraints like this will be a good test of the new architecture.

Punzo:
Hi Steve, sure, I haven't finished (I have still to integrate to Slicer the line, angle and curve widgets, already implemented at VTK level and the projection in the 2 views), but I can move my branch in https://github.com/Slicer/Slicer and continue to work on it.

pieper:
Thanks @Punzo !

Punzo:
@lassoan @pieper I have created the PR in the Slicer repos:

https://github.com/Slicer/VTK/pull/20
https://github.com/Slicer/Slicer/pull/1079

I'll continue  to work  on https://github.com/Slicer/Slicer/pull/1079.

muratmaga:
We have putting together some sample data for our projects during PW. If you need to a large set of fiducials (markups from 126 subjects, 55 fiducial in each subject), feel free to test it with:
https://github.com/SlicerMorph/SampleData/blob/master/mouse_skull_LMs.zip?raw=true

jcfr:
Closing. With the new markup infrastructure, fiducial are contained to 3D surface by default.

@pieper If I missed something, feel free to re-open.

pieper:
Thanks @jcfr yes, the new infrastructure works well for this as far as I can tell.

PR #1077, opened 2019-01-16
Webwidget streamline python use

jcfr:
Improve use of python from javascript.

@pieper Instead of a systematically re-initialization the `QWebChannel`, this PR re-refactor the code introducing `qSlicerExtensionsInstallWidgetPrivate::initializeWebChannelTransport(QByteArray& webChannelScript)` allowing both the base and derives classes to customize the WebChannelTransport initialization call back.

See also this documentation mentioning that only one WebChannel can be initialized only once per page. See http://doc.qt.io/qt-5/qwebenginepage.html#setWebChannel

This means that to evaluate python from a webpage loaded through a `qSlicerWebWidget`, it is now as simple as doing `window.slicerPython.evalPython("print('hello from JS through python')")`

### DIFF
--- a/Applications/SlicerApp/Testing/Python/WebEngine.py
+++ b/Applications/SlicerApp/Testing/Python/WebEngine.py
@@ -152,15 +152,17 @@ class WebEngineTest(ScriptedLoadableModuleTest):
     webWidget.show()
     self.delayDisplay('Showing widget')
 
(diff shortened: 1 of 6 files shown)


pieper:
Looks good to me Jc - thanks!

jcfr:
Thanks for the review.

Integrated in [r27941](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27941)

PR #1078, opened 2019-01-16
ENH: Add qSlicerSingletonViewFactory

Sunderlandkyl:
The new view factory can be used by both C++ and Python classes to register any singleton QWidget by calling factory->setWidget(QWidget*) and factory->setTagName(std::string).

Since the widget is a singleton, the factory can only be used to add a single such view to a given layout.

Spun off from https://github.com/Slicer/Slicer/pull/1061

### DIFF
--- a/Base/QTGUI/qSlicerSingletonViewFactory.cxx
+++ b/Base/QTGUI/qSlicerSingletonViewFactory.cxx
@@ -1,0 +1,130 @@
+/*==============================================================================
+
+Copyright (c) Laboratory for Percutaneous Surgery (PerkLab)
+Queen's University, Kingston, ON, Canada. All Rights Reserved.
+
+See COPYRIGHT.txt
+or http://www.slicer.org/copyright/copyright.txt for details.
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+This file was originally developed by Kyle Sunderland, PerkLab, Queen's University
+and was supported through CANARIE's Research Software Program, and Cancer
+Care Ontario.
+
+==============================================================================*/
+
+// QtGUI includes
+#include "qSlicerSingletonViewFactory.h"
+
+#include <QDebug>
+#include <QMap>
+#include <QSharedPointer>
+#include <QWidget>
+
+//-----------------------------------------------------------------------------
+class qSlicerSingletonViewFactoryPrivate
+{
+  Q_DECLARE_PUBLIC(qSlicerSingletonViewFactory);
+public:
+  qSlicerSingletonViewFactoryPrivate(qSlicerSingletonViewFactory& object);
+  virtual ~qSlicerSingletonViewFactoryPrivate();
+
+  virtual void init();
+
+  QSharedPointer<QWidget> Widget;
+  QString TagName;
+
+protected:
+  qSlicerSingletonViewFactory* q_ptr;
+};
+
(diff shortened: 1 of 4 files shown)


Base/QTGUI/qSlicerSingletonViewFactory.cxx
```diff
@@ -0,0 +1,122 @@
...
+class qSlicerSingletonViewFactoryPrivate
+{
+  Q_DECLARE_PUBLIC(qSlicerSingletonViewFactory);
+public:
+  qSlicerSingletonViewFactoryPrivate(qSlicerSingletonViewFactory& object);
+  virtual ~qSlicerSingletonViewFactoryPrivate();
+
+  virtual void init();
+
+  QWidget* Widget;
```
lassoan: Widget should be initialized to NULL in the constructor.

Base/QTGUI/qSlicerSingletonViewFactory.h
```diff
@@ -0,0 +1,68 @@
...
+// QtGUI includes
+#include "qSlicerBaseQTGUIExport.h"
+
+class ctkDICOMBrowser;
+class qSlicerSingletonViewFactoryPrivate;
+
+/// This class provides an interface for C++ and Python classes to register a singleton view replacement widget.
+/// New view widgets can be registered by registering a new qSlicerSingletonViewFactory, and then setting the widget and tag using setWidget(QWidget*) and
+/// setTagName(QString).
+/// This factory contains a single pointer to an instance of the widget, so only one view can be created within a given layout.
```
lassoan: What happens if two instances are used in a layout? Could we check this and report an error if it happens?

Base/QTGUI/qSlicerSingletonViewFactory.h
```diff
@@ -0,0 +1,68 @@
...
+public:
+  typedef ctkLayoutViewFactory Superclass;
+  qSlicerSingletonViewFactory(QObject* parent=0);
+  virtual ~qSlicerSingletonViewFactory();
+
+  /// Reimplemented to support custom element names
+  virtual QStringList supportedElementNames()const;
+
+  /// Set the singleton widget instance that will be used to create the view
+  Q_INVOKABLE virtual void setWidget(QWidget* widget);
```
lassoan: Who is responsible for deleting the widget? It should be at least clarified in the documentation. We may consider allowing the factory taking responsibility for deleting the widget if that simplifies the implementation.

lassoan (commented):
Thank you it looks good overall. There are just a few things to clarify - see inline comments.

Base/QTGUI/qSlicerSingletonViewFactory.h
Sunderlandkyl: If two or more of the same singleton view are included in the same layout, then only the last one will be filled. Previous views will be blank.

Probably could log an error if this happens.

Base/QTGUI/qSlicerSingletonViewFactory.h
Sunderlandkyl: I changed the regular pointer to a QSharedPointer.
Factory is now responsible for deleting the widget.

Base/QTGUI/qSlicerSingletonViewFactory.h
Sunderlandkyl: If two of the same views are created within the same layout, the following error message is now displayed:
````
qSlicerSingletonViewFactory::createViewFromXML - Widget for view "TagName" is already in use within the current layout!
````

Sunderlandkyl:
Any other changes/comments?

lassoan:
Thank you, this looks good to me. I'll merge it now.

lassoan:
Merged, thank you!

PR #1017, opened 2018-09-06
ENH: Add generic storage node for Markups

vovythevov:
Markups did not have a generic storage node that worked for all for types
of markups. This new storage nodes aims to fill this need. It is also meant
to be easily extensible in potential markup subclasses.

The new generic storage exports markups to JSON for ease of interface with
the web, python and other applications.

The existing Markups Fiducial reader is left as-is and is still the default
for fiducials.

CC: @jcfr

PR #1079, opened 2019-01-17
WIP+ENH: Widget rework

Punzo:
DO NOT MERGE

@lassoan @pieper @jcfr 

here the ongoing work for the markups internal rework and adding line/angle/curve interactive widgets

video showing the current state (i.e. integration of the vtkSlicerPointsWidget into fiducials):
https://www.dropbox.com/s/p8v1m7mgopsnrp4/Widget-rework.mkv?dl=0

the new infrastructure provide better performance (no drop-down at 1000 points) and many other interactive features (i.e., support for middle, right, double clicks, shift, scale rotate actions, mouse hover with selection of the points, etc...)

implementation summary:
A) For each vtkMRMLMarkupsFiducialNode there is a vtkWidget for each view.
B) Each vtkMRMLMarkupsFiducialNode have N ControlPoint structs (N for fiducials, 2 for rulers, 3 for angles and N for curves). The ControlPoint struct is the one that have the information of the position, Selected, Visibility, locked, etc.. for each point of the widget.
C) Each widget have a representation.
D) Each representation has 3 actors for the points: "normal", "selected" and "active" (the last is always only one point) for fiducials. The others widget will have also the actor for the line. + actors for texts.
E) The ControlPoint structs in the vtkMRMLMarkupsFiducialNode are shared directly to all the representations.
F) The update of the actors in the representations is handled by the display managers (by using the info stored in the ControlPoint structs) when the info in the ControlPoint structs in the vtkMRMLMarkupsFiducialNode are modified by any widget (the display manager simply call BuildRepresentation method of the representation at the point modified events).
E) at level of vtk widgets, the widgets have more features, and they have also been simplified (just one representation, the handles have been removed).

To Do:
- Fiducials widget rework done (it uses the new vtkSlicerPointsWidget)
- Integrate the others widgets: line, angle, curve (already implemented at VTK level).
- Implement the widget projection in the 2d views (both total and distance gradient fading, similar to models).
-  Additional to do: move the point infos contained in the ControlPoint struct in VTK objects (vtkPointSet).

Minimal modification to VTK are required as:
This needs https://github.com/Slicer/VTK/pull/20

The new vtk widgets, for now, are in the folder Modules/Loadable/Markups/VTKWidgets. Two reasons for having the widget in Slicer:

1) the widget have a pointer to the MarkupsNode to share the coordinates and others proprieties of the points (i.e. removing a lot of manual updates)
2) Faster prototyping  and dev.


lassoan:
@punzo Thanks for your awesome work! I start review and testing and if everything works out well (there are no significant regression in existing features) then we will integrate this on Friday so that developers at the project week can all test, fix, and enhance it.

Punzo:
> @Punzo Thanks for your awesome work! I start review and testing and if everything works out well (there are no significant regression in existing features) then we will integrate this on Friday so that developers at the project week can all test, fix, and enhance it.

Yes I plan to finish to integrate all the widgets and add the projection on 2D views within Friday

vovythevov:
@Punzo: This is awesome !
I don't know if that's any help to you or not, but I had a WIP implementation of planes and splines that works in this PR: https://github.com/Slicer/Slicer/pull/1010.
There is also this PR to add an easily extensible JSON storage node for Markups that I think would be useful for your case: https://github.com/Slicer/Slicer/pull/1017.

Punzo:
Hi @lassoan, now the Points and Line widgets should be stable. Tomorrow I'll add the angle and curve widgets. If you'd like please start review.

I may be deleted a commit of yours regarding the VTK branch? sorry if yes. I didn't checked before doing a force push. 

Regarding the line widgets I have a doubt regarding the picking of the line (done with a vtkPropPicker), see comment in code: https://github.com/Slicer/Slicer/pull/1079/files#diff-9266791f752bc6b3e38c5326ba760dc1R352
The picking done with vtkCellPickers is not great (I need to pick every time I move the mouse). vtkPropPicker gives kind of better performances but it creates a lot of rendering flickering. I tired also the vtkScenePicker (much faster), but in this case the flickering are horrible.

I am thinking to use the tree locator to pick the line by using the intermediate points. If I use enough intermediate points, they should be enough to have a good enough picking with fast performances.   

P.S.: to pick the points I don't use a picker, but the locator which has all the points coordinates in a tree, much faster. See ActivateNode in e.g. https://github.com/Slicer/Slicer/pull/1079/files#diff-9266791f752bc6b3e38c5326ba760dc1R348

lassoan:
Line picking can be done so efficiently that we could even add our own picker or implement it without a picker class.

I guess this will be more critical for curves, where there can be thousands of line segments for each curve.

I don't think render view should flicker because of picking. That may be a VTK bug.

lassoan:
I've already built the branch and started testing. A few issues:

- [ ] If "Persistent" option is enabled then after placing a point, the new fiducial sticks to the mouse pointer. It is kind of nice to see the future label next to the mouse cursor, so this could be a feature. But this does not happen consistently (when persistent option is disabled or when placing the first point). See https://1drv.ms/v/s!Arm_AFxB9yqHtstcwAp4GHR1k3D83A
- [ ] After loading a scene, markup points don't show up until I add a new markup point.

pieper:
Overall looking very promising 👍 

But I also got a crash at one point but I haven't been able to reproduce it (now running in the debugger).  Will let you know if I can get a stack trace.

Punzo:
Thanks for testing!!!

> Line picking can be done so efficiently that we could even add our own picker or implement it without a picker class.
I guess this will be more critical for curves, where there can be thousands of line segments for each curve.
I don't think render view should flicker because of picking. That may be a VTK bug.

ok for the moment I'll leave the picking with the vtkPropPicker. it would be good to understand why the render flicker. However with already 20 line widgets in the view, the perfoamnces starts to drop. Probably the best for the mouse hover would be the vtkScenePicker (mouse hover was it is its purpose).

Andras' comments:
1) this helps with lines and especially curves, because with the preview point it draws also the line. So the idea was to show the preview point only after the first (so there will be the line). If we want to change behavior I am fine.
2) ok, I have to say I still didn't test the save and load, I am glad that it didn't crash (:
3) for it works, but maybe I didn't understand, here a video https://www.dropbox.com/s/0wxf5rfh20el6dk/selecting.mkv?dl=0
4) this also worked for me. maybe a bug only in some conditions.
5) for me at 391 the performances are still good. It make sense to have the scale only for line/(angle?) and curve widgets. I'll disable for points.
6) for some reasons I thought the actor were scaled. I'll put fixed coordinates. Thanks for catching this.
7) In general vtkPickingManager solves this picking issue when there are more widgets. there are two problems: 1) the picking of points is not done with a picker (to be fast); 2) in 2d view also the picking of lines is not done with the vtkPickingManager for bad performances reasons (see https://github.com/Slicer/Slicer/pull/1079/files#diff-9266791f752bc6b3e38c5326ba760dc1R352). Maybe one solution is to use the vtkScenePicker for both points and lines, I'll check.
8) ok I can fix this easly
9) I already do, I disable the hover (and therefore any interaction, because the hover "initialize" the click actions) for the locked points. In the case of the line, the hover over of the line is disabled if both points are locked.
10) yeah not sure how to solve this: if a line has only one selected point and we "activated" the other point, the line should have the selected or not selected color? 
11) yes that is the current behavior, e.g. https://www.dropbox.com/s/ajm8kt7969wrpxt/activate.mkv?dl=0. we can increase the tolerance both in 2d and 2d view is you think is better. Something like 1.5 the radius??
12) ok, I'll check how to fix
13) ok, I'll check how to fix
14) ok, I'll check how to fix this as well.

Punzo:
@lassoan  I have fixed 2-5-6-8-11-12-13-14 and replied to 1-3-4-9. 10 I am not sure which would be the best behavior.  7 is problematic and connected to the picking issue. Now I'll finish to integrate the angle and curve widgets. After that I'll go back to 7 and the picking issues.

lassoan:
- re 1. I like that we can see the point label when placing the point, but it should be consistent. We should then see the point when placing the first point, too. Also, I should be able to place point in any of the views (currently, when I go to a new view, the green highlighted temporary point remains in the original view - if I move fast enough, and no temporary point is shown in the new view until I place a point there).
- re 5. Scaling could be useful for some widgets but probably it should be disabled by default for most of them. Also, we will use right-click for context menu, so we'll have to use a different shortcut.
- re 9. There is an important distinction between locking a point position (it is still hightlghted when hovering over and you can right-click on it to bring up a context menu), and disabling interactions (no highlighting when hovering over the widget). In Slicer master it is implemented by widget and point lock. If widget is locked then it disables al interactions. If point is locked then it cannot be moved but you can still interact with it.
- re 10. The main problem was that the line highlight color randomly changed (depending on which markup was highlighted). I think the behavior would be consistent if the line segment was selected when both endpoints were selected.
- re 11. Tolerance should be defined relative to screen size percentage and should be configurable in application settings (it depends on the user, screen size, mouse precision how accurately a point can be hit)
- re all picking issues: Probably we should implement a custom picker, probably separate from 2D and 3D views.

Punzo:
re 1) ok I'll see to show always the preview point  and remove the constrain to the view during the placing.

re 5) yes I disabled for points widgets (changed with a pickAction, the same of the double clicks). Line and other widgets also will have pickAction on rightClick, but I'll add ALT+right-click for scaling. 
P.S.: the rotation is ALT+lef-click

re 9) ok I like it, I can implement the same.

re 10) ok

re 11) now the tolerance is defined in terms of the HandleSize (Scale value of the Markups GUI). the default value is 0.5, so it means that the tolerance will be HandleSize + (HandleSize * 0.5). The screen dimensions is automatically accounted for with a scale factor as calculated in https://github.com/Slicer/Slicer/pull/1079/commits/56ba97a19e7a771c796cd882b02e52fb15f29387#diff-b8365d1f0f752b023b89aa3472f5a64bR377

re 12) ok, probably we will need to talk a about this more. But let's talk about it tomorrow? 

For today I'll finish to integrate the angle and curve widgets and doing these other fixes 1-5-9-10-11.

lassoan:
I've tested problem 3 and 4 and they are not reproducible anymore with the latest version.

lassoan:
I've added some more findings, starting from 15.

Punzo:
comment regarding 11: so I changed the the current behaviors as explained in https://github.com/Slicer/Slicer/pull/1079#issuecomment-457239387  here how it works: https://www.dropbox.com/s/qaa6tfqpw0knkqu/pickingPointsTolerance.mkv?dl=0.

if the tolerance is too much now, we can reduce the tolerance value from 0.5 to 0.3. But the current method works pretty well when zooming or scaling the widget.

You asked also to be able to set the value in the settings, but the markups settings are disabled by default.

Punzo:
@lassoan fixed 1-9-10-15-16-17.

I Have no idea regarding 18. 

Still to Do:
-Add curve widget (doing now)
-fix 7 and picking
-add 2d actor view projections (both total and slice distance gradient as in model)

lassoan:
Awesome progress, thank you! It shows that the design (and of course you :) ) are good that issues can be fixed so quickly.

Re 18: don't worry about it for now, we'll check it with Jc or other Kitware guys.

lassoan:
There are many differences between open and closed curves (what they are used for, what metrics can be computed from them, how they are crated and displayed), it would be very rare that someone would want to quickly toggle between them, and we already separate widgets that have similar implementation but different role (e.g., line and curve). So, I think we should have an abstract curve widget base class and concrete OpenCurve and ClosedCurve widget classes.

jcfr:
This is somewhat inline with what exist in VTK: `vtkPolyLinewidget` and `vtkSplineWidget2`, both having their dedicated representation deriving from `vtkCurveRepresentation`

lassoan:
Curve works very nicely!

- re 2: After loading a scene, markups still not show up.
- re 19: The crash happens in vtkIncrementalOctreePointLocator::FindClosestPointInSphere line 455. checkNode is NULL and checkNode->IsLeaf() method is called. this->OctreeRootNode is NULL. this->DataSet is NULL.

lassoan:
Many further improvements could be done, but I think the current implementation would be ready to merge - except the flashing viewers when the mouse is moved.

Picking is also extremely slow on Windows in debug-mode builds (takes several second to highlight a point, even if there are just a few ten points and a few angle widgets).

Are there other pickers that we could use? Even trivial computations could be used (no need for octree, etc. because we just need to picks lines and points).

Punzo:
Ok I'll use two different widgets for open and closed curve thanks for the feedback!

lassoan:
CPU-based picking is extremely fast for widgets because typically we need to check distance from a handful of points or lines. If hardware-based picking requires a rendering step then it is probably much slower.

For each widget point, the widget should know if it is visible or not otherwise it could not hide labels for occluded points. If the widget knows visibility of each point then that can be used for picking, too. To determine if a widget point is visible, probably some hardware acceleration is needed (e.g., get Z buffer from the renderer).

pieper:
Hardware picking should have practically no overhead at all if it's done right.  Basically it's just another output from the fragment shader stored in another frame buffer.  This is trivial with current GPU hardware.  It's already being done in vtkjs and should be added to vtkcxx if it isn't already.

This would allow us to scale to arbitrarily large numbers of widgets and would support dragging on the surfaces of volumes and models (which is required for SlicerMorph but very useful for other purposes too).

Punzo:
> CPU-based picking is extremely fast for widgets because typically we need to check distance from a handful of points or lines. If hardware-based picking requires a rendering step then it is probably much slower.

well I actually tried also the vtkCellPicker and the vtkPointPicker in combination with the vtkPickingManager. No flickering, but the performances were even slower. Probably it will be good to investigate this together during the project week.

Punzo:
Here I do again a summary of the To Do:
- [] add closed curve widget (I'll do soon)
- [] fix 2 (I'll do soon)
- [] fix 7 and picking
- [] fix 18
- [] fix 19 (I am checking this now)
- [] add 2d actor view projections (both total and slice distance gradient as in model)

lassoan:
Can you make the code switchable between vtkCellPicker and hardware-accelerated picker (using #ifdef)?

Do you set the picker to only handle the associated widget's actors? It must be very fast then. However, we can make it even faster by creating a custom picker that uses point and line positions directly (a cell picker would use all the facets of the sphere to determine if we are near the actor, while we know that it would be enough to just check the distance from a single point).

Punzo:
> Can you make the code switchable between vtkCellPicker and hardware-accelerated picker (using #ifdef)? 

sure

> Do you set the picker to only handle the associated widget's actors? It must be very fast then. However, we can make it even faster by creating a custom picker that uses point and line positions directly (a cell picker would use all the facets of the sphere to determine if we are near the actor, while we know that it would be enough to just check the distance from a single point).

yes the pick only the actor of that representation. When it is only one widget is pretty fast (i.e. one actor to pick). The issue it seems to be when there starts to be about 10 widgets (10 actors to pick)

lassoan:
> Hardware picking should have practically no overhead at all if it's done right. Basically it's just another output from the fragment shader stored in another frame buffer. This is trivial with current GPU hardware. It's already being done in vtkjs and should be added to vtkcxx if it isn't already. This would allow us to scale to arbitrarily large numbers of widgets and would support dragging on the surfaces of volumes and models (which is required for SlicerMorph but very useful for other purposes too).

I think we talk about different things.

After we already grabbed a point then of course we can use hardware pickers to move along rendered surfaces.

Our problem now that it is very slow when we move the mouse (without pressing a button) mouse and need to highlight the nearest widget. Using a hardware picker for this would be an overkill: you would need to create specialized actors for an extra rendering step (because you want to pick when you are _near_ a widget, not just when the mouse is actually over a potentially very small actor), instead of just iterating through a few points (even checking distance from a few hundred points would take negligible CPU time).

lassoan:
re 21: Lines are correctly displayed in release-mode builds but don't show up in debug-mode builds

Release mode (good):

image

Debug mode (missing line in slice view, not good):

image

lassoan:
Re projections: We had some limitations when we displayed widgets with models: we had to project the final polydata. With widgets, we can project just the centerline and then apply a tube filter to make rendered lines look nicer.

Punzo:
ok strange I am in debug mode as well, but on linux it shown them

> Re projections: We had some limitations when we displayed widgets with models: we had to project the final polydata. With widgets, we can project just the centerline and then apply a tube filter to make rendered lines look nicer.

ok, probably the best it will be to talk about picking and projections by a videocall??

Punzo:
would it the best doing a pure vtk example with many actors and trying the several pickers? also we can test if the flickering is a vtk bug or ours.

Punzo:
@lassoan fixed 2-19-22. 

Still to be fixed 7/20(picking issues)-18(fix rendering when depth peeling)-21(fix rendering lines in windows debug). 

To Do: add closed curve (I will do this tomorrow) and projection on 2d views.

Punzo:
Hi @lassoan @pieper @jcfr, probably you are very busy, but if there is anything to discuss regarding this PR with me by videocall please let me know. Thanks!

pieper:
Hey Davide - Yes, busy here!  But productive.

Question for you: are your vtk changes compatible (mergeable) with
Slicer/VTK?

Punzo:
> Question for you: are your vtk changes compatible (mergeable) with Slicer/VTK?

You need this https://github.com/Slicer/VTK/pull/20

Punzo:
P.S.: @pieper @lassoan I made a vtk example to test the picking while moving the mouse: https://github.com/Punzo/vtkPickerExample

It is different by of case, because our case has one picker for each representation and then the picking is  managed by the pickingManager, while in the example there is only one picker. However, with one picker, using a vtkGenericOpenGLRenderWindow does not show any flickering and it is very fast. 
Maybe we should move in some way the picking to at the level of MRMLDM instead of widgets?

lassoan:
> Maybe we should move in some way the picking at the level of MRMLDM instead having a picker for each widgets?

This is a very good idea. Since each displayable manager knows exactly what widgets are there and how to manage them, it should be possible to use a single picker, which may make things much faster.

There will be still other pickers in the scene, some of them even using picking manager (such as ROI widget or later the cursor widget, etc.), so we cannot assume that there is only one picker, but the number of pickers will be much smaller if we handle all markups with a single one.

smrolfe:
@Punzo @lassoan @pieper @jcfr 

Hi Davide, 
I've been testing your new module at project week and wanted to document a few issues I ran into with our data sets. 

1. The points are not always visible in the 3D viewer. When I first place them (using either slice or volume view) they show up in the slice view,  but in the 3D viewer they appear briefly, then disappear. The points are then visible\hidden depending on the rotation of the view. Plotting a 2D slice view in 3D also causes them to be displayed.  I'm attaching a screen shot where two landmark points are visible in the slice views but invisible in the 3d viewer. The second image shows the same data, where the volume was rotated and the landmarks are visible.

image

image

2. The glyph scaling between the slice view and 3D view is very different. If the points are visible on the 2D slice, they are too large to be useful in the 3D view. This can be fixed using the slider, but there is no way to look at them simultaneously, which is useful for landmark placement. Can the glyph size be fixed to the window size? You can see this issue on the screenshots I've attached.

I'm also interested in adding the ability to stick a landmark to an object (model or rendered volume) and constraining it's movement to that surface. Could you let me know if you have any plans to implement this or if you have any thoughts on how it should be done?

Punzo:
1) thanks for reporting the bug. I'll check it out.
2)  the points are scaled on the dimensions of the each view. Probably we should just scale them on the total dimensions of the Slicer windows.
3) Steve implemented that recently on the previous infrastructure. For the moment I don't have the time to look at deeply, but it could be possible to do also in the new infrastructure by using a vtkVolumePicker in the widget and constrain the dragging of the points with the picker coordinates in the representation method TranslateNode.

pieper:
Thanks @Punzo 👍   For point 3, we've been discussing it here in Las Palmas and I'm happy to try porting over the previous experiment to the new infrastructure unless someone else wants to have a go.

Punzo:
> Thanks @Punzo +1 For point 3, we've been discussing it here in Las Palmas and I'm happy to try porting over the previous experiment to the new infrastructure unless someone else wants to have a go.

ok perfect, thanks!

smrolfe:
Thanks @pieper and @Punzo!

smrolfe:
Hi Davide,
I'm using your most recent commit. I compiled on my remote machine and could try compiling locally to see if that might be a factor. 

For 2, this issue is related to the image resolution. Our data sets are in typically in the range of tens of microns. I can send some example data if that would be helpful.

smrolfe:
Hi Davide, 
Here's a data set to replicate issue #2: https://github.com/SlicerMorph/SampleData/blob/master/LMExample-Scene.mrb?raw=true

Punzo:
Hi Sara, thanks for the datasets.

I still could not reproduce 1, but I could reproduce 2. 2 was more a zoom issue when we have volumes
with spacing != 1. The fix for 2 was simple, if there is an active volume I get the mean of the 3 spacing values and I multiply the scale of the widget for that. Please test the last commit.  

image

smrolfe:
Hi Davide, 
Thanks, that's great news! I'm compiling the new commit now and will update you after testing.

spujol:
Hi Davide,

Is the bounding box ROI in 3D with rotations on the roadmap?

jcfr:
As of [r27951](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27951), Slicer/VTK was updated to integrate @Punzo  changes

lassoan:
Added a few more small issues that we've found. All the remaining issues should be easy to fix or not critical, so probably we can integrate this branch next week.

Punzo:
> Hi Davide, Is the bounding box ROI in 3D with rotations on the roadmap?

Hi @spujol. Unfortunately no. On the other hand, it should not be too difficult. we have already such interactions for transforms, so at VTK level probably is already there.

> As of r27951, Slicer/VTK was updated to integrate @Punzo changes

Thanks Jc!

> Added a few more small issues that we've found. All the remaining issues should be easy to fix or not critical, so probably we can integrate this branch next week.

ok Thanks for testing. Probably it will be worth to have a videocall next week, to see what is still to do.

Punzo:
> Adding markup point does not trigger vtk.vtkCommand.ModifiedEvent event (changed compared to previous behavior)

adding a markup now fires a PlacePointEvent. We can change as before or fire both

Punzo:
@Punzo is that the same rotatable ROI we were discussing? 

I think yes (if I understood @spujol request). 

At the moment the ROI for clipping the volume rendering is not rotatable by mouse interactions.

lassoan:
> does that mean somebody figured the picking performance issues?

Point picking performance seems to be comparable to what was before, so it should not block integration. For next week we will use cell picker (or disable curve/line picking completely). We will switch to single picker per view, which will improve performance. To get optimal performance, we need to debug hardware-based picking.

lassoan:
> adding a markup now fires a markup PlacePointEvent. We can change as before or fire both

We should probably keep specific place event, but we need to invoke modified event as well.

jdx-john:
> > does that mean somebody figured the picking performance issues?
> 
> Point picking performance seems to be comparable to what was before, so it should not block integration. For next week we will use cell picker (or disable curve/line picking completely). We will switch to single picker per view, which will improve performance. To get optimal performance, we need to debug hardware-based picking.

OK. I suppose even if it is slow, the functionality might be retained (but off by default), if design allows swapping it out for a better solution neatly? We'd certainly like to be able to test/demo it in our software even if under carefully controlled lab conditions :)

fedorov:
About ROI, if the new infrastructure does not support it, it implies existing functionality in Slicer will remain in place until it does, right?

Punzo:
> About ROI, if the new infrastructure does not support it, it implies existing functionality in Slicer will remain in place until it does, right?

yes I didn't remove any "old" annotation (ROI, ruler, etc...) (Annotation module). 

The only thing that is not backward compatible is the previous Points Markups classes which are replaced by the new ones (together with the new line, angle and curve). In practice, this PR modify the Markups module.

lassoan:
> The only thing that is not backward compatible is the previous Points Markups

The internal design is completely different but the public interface of markups fiducial will remain largely unchanged (at least for now, to make the transition smoother).

smrolfe:
Hi @Punzo,
I have tested the new code and it looks great. With a little adjusting I can find a scale that works in 2D/3D. Thanks for resolving this.

image

lassoan:
@Punzo Transforms do not seem to have any effects on the markups. Can you confirm? Is this a known limitation? This must be fixed before integration.

smrolfe:
Hi @Punzo,
I have an update on the issue 1 (landmark visibility) now that I'm back from PW and have tested the new markups on my local machine. I no longer see this issue with the MRHead sample data, but it is reproducible for higher resolution images, occurring more frequently on larger data sets. 

Could you let me know if you can replicate this behavior on the image I sent previously? It has a smaller file size, but it's still occurring occasionally. The zoom level also appears to be a factor. 

I've attached some examples screenshots of a volume/model in dual 3D views with slightly different rotations.
image
image
image

Punzo:
> @Punzo Transforms do not seem to have any effects on the markups. Can you confirm? Is this a known limitation? This must be fixed before integration.

I just tested it now. Indeed does not work. I am not familiar with the transform integration. Do you have any hints why does not work?

> Hi @Punzo,
I have an update on the issue 1 (landmark visibility) now that I'm back from PW and have tested the new markups on my local machine. I no longer see this issue with the MRHead sample data, but it is reproducible for higher resolution images, occurring more frequently on larger data sets.
Could you let me know if you can replicate this behavior on the image I sent previously? It has a smaller file size, but it's still occurring occasionally. The zoom level also appears to be a factor.
I've attached some example screenshots of a volume/model in dual 3D views with slightly different rotations.

OK thanks for the further testing! I'll try to reproduce it and try to fix asap.

lassoan:
I've added landmark visibility issue as 31. Created an [example scene](https://1drv.ms/u/s!Arm_AFxB9yqHtvpBBO0RbHNtzKP3Sw) to reproduce the problem easily. Load it and zoom in the 3D view and the markups disappear while other scene content is displayed correctly as shown in this [demo video](https://1drv.ms/v/s!Arm_AFxB9yqHtvpCNKMc_5tvI9fulw).

lassoan:
@punzo Is `if (this->GetNumberOfPoints() > 2)` in vtkMRMLMarkupsCurveNode::AddPoint a copy-paste error?

lassoan:
@Punzo FYI, I'm working on adding transforms support. The issue was that local and world (transformed) positions were not distinguished clearly enough. In the ControlPoint structure, we should only store local coordinates.

Punzo:
> @Punzo FYI, I'm working on adding transforms support. The issue was that local and world (transformed) positions were not distinguished clearly enough. In the ControlPoint structure, we should only store local coordinates.

ok thanks for fixing it.

> @Punzo Is `if (this->GetNumberOfPoints() > 2)` in vtkMRMLMarkupsCurveNode::AddPoint a copy-paste error?

yes definitively, thanks for catching it!

lassoan:
@Punzo Why this hack is needed in vtkSliceViewInteractorStyle::OnEnter()?

  double FieldOfView[3];
  sliceNode->GetFieldOfView( FieldOfView );
  sliceNode->SetFieldOfView( FieldOfView[0] * 1.1, FieldOfView[1] * 1.1, FieldOfView[2] * 1.1 );
  sliceNode->UpdateMatrices();
  sliceNode->SetFieldOfView( FieldOfView[0], FieldOfView[1], FieldOfView[2]);

Punzo:
> @Punzo Why this hack is needed in vtkSliceViewInteractorStyle::OnEnter()?
double FieldOfView[3];
sliceNode->GetFieldOfView( FieldOfView );
sliceNode->SetFieldOfView( FieldOfView[0] * 1.1, FieldOfView[1] * 1.1, FieldOfView[2] * 1.1 );
sliceNode->UpdateMatrices();
sliceNode->SetFieldOfView( FieldOfView[0], FieldOfView[1], FieldOfView[2]);

It seemed the only way to refresh the widgets when interacting with a widget in another view. 
Essentially, without this workaround, after interacting with the widgets in a view, the widget in the other views do not to react anymore at any mouse event.
The same for the 3D View. Not sure if there is a better way than this workaround

lassoan:
Since we have full control over the widgets, we can add an API that performs the necessary action directly in the widget. Most likely this hack causes issue 23.

Punzo:
> Most likely this hack causes issue 23.

yes it is the cause.

lassoan:
re 28: Markup transformations are now fixed.

lassoan:
re 31: Markup disappearance is now fixed, too

jamesobutler:
@Punzo These new widgets look great! 👏  I’m looking forward to using them in Slicer 😄

Punzo:
> @Punzo These new widgets look great!  I’m looking forward to using them in Slicer

Thanks! @lassoan is currently finishing to polish them and bug fixing in https://github.com/lassoan/Slicer/commits/widget-rework2 .

jcfr:
@Punzo @lassoan and everyone who contributed directly or indirectly with development time, idea, review and testing. Thanks for moving this forward and making it possible :+1: :smile:

Punzo:
Hi @lassoan I was testing https://github.com/lassoan/Slicer/commits/widget-rework2

It is very nice: code cleaned a lot, performances much better (especially the picking) and the 2d line actor gradient coloring is pretty nice.

I notice few bugs while I was doing some fast testing on linux. Maybe you can reproduce as well on windows and you know them, but I thought it will good to report:
- [ ] middle click interaction (translate widget operation) works only on 2d views and if done on a line it will jump the widget to have the nearest fiducial under the mouse. In 3d views the translate widget operation does not work.
- [ ] alt + right click (scale widget operation), not working either on 2d or 3d neither on points or lines
- [ ] alt + left click (rotate widget operation), not working either on 2d or 3d neither on points or lines
- [ ] ctrl + left click (add point operation after placement) on closed curve does not work
- [ ] double middle and right clicks were removed? They can come handy to map future operations.
- [ ] when removing points after placement (DEL) and we go from 2 to 1 (and from 1 to 0) markup points the line actors should be hided or updated (to show nothing) 
- [ ] 2d line actors of the line and angle widgets do not have the gradient coloring, bug or wanted?
- [ ] picking points of angle widget in 3d view does not work (I could not move the angle widget in the 3d view at all).
- [ ] picking lines of spline widgets (open and closed curve) seems to pick the line on a straight line instead of following the spline
- [ ] sometimes the sampling for the interpolation points of the spline for open and closed curve seems quite large, see figure below:
image
 the vtkBezierInterpolator had a very good adaptive algorithm, was there some issue that you replaced it? It will be nice to have very smooth splines both for aesthetics (which can be very important), but also it is very important to have smooth spline when calculating physical quantities, such as the curvature, along the curvilinear coordinate of the spline. 

A part these few bugs, it works very well! I'll wait you create a new PR for testing further.

lassoan:
I've created a spreadsheet to keep track of the open issues:
https://1drv.ms/x/s!Arm_AFxB9yqHtvxAgdAfdLS_GHsOxw

I would like to fix all priority 1 and preferably priority 2 issues before submitting the pull request.

Punzo:
Another point:

- [ ] bad performances of middle click interaction/translate widget operation (and therefore also rotate and scaling) at already ~100 points. 

I think this is due because we set all the points position one by one in https://github.com/lassoan/Slicer/blob/3c6aed7f7a62f50a1af4b0d5766ee063483b3770/Modules/Loadable/Markups/VTKWidgets/vtkSlicerAbstractWidget.cxx#L800-L815
and there is something strange going on on the Start/End Modify methods. In fact, Start/End Modify methods filter the events and there is only one event propagating but the performance are very slow. If we instead disable the events completely (with DisableModifiedEventOn/Off) and fire only one generic modified event the performance are interactive even at 1000 points.

Maybe there is some performance issue at the tracking level of the Start/End Modify methods.

Another option may be to add a method:
markupsNode->SetNControlPointPositionsWorldFromArray(N, worldPos[N][3])
which will set all the position of N points in one shot and calling only one generic modified event (or a new event, something like: allPositionsModified) and it will be used only in these occasions of multiple points modifications.

lassoan:
Thanks for testing. I have not checked widget translate/rotate operations yet. They should probably not use pointwise Set events but update all points at once.

This DisableModified/StartModify performance difference is very interesting, we need to debug this, because this mechanism is very widely used.

Punzo:
> This DisableModified/StartModify performance difference is very interesting, we need to debug this, because this mechanism is very widely used.

yeap, I tried to track down the issue and full understand it, but I didn't managed to solve it.

lassoan:
Superseded by https://github.com/Slicer/Slicer/pull/1094

PR #1080, opened 2019-01-21
BUG: Fix distance encoded projection in Slice view

Sunderlandkyl:
- Implicit plane function for calculating distance uses SliceXYToRAS, however the distance is scaled in the Z direction
- Fixed by normalizing the plane normal used for distance encoded projection


lassoan:
Looks good, thank you!

To be consistent, could you please make the same fix in vtkMRMLSegmentationsDisplayableManager2D.cxx?

Sunderlandkyl:
Sure, done!

lassoan:
Merged, thank you!

PR #1081, opened 2019-01-24
Add dicomweb client

pieper:
This works for me on mac, but I believe it's right for other platforms too.  @jcfr can you have a look?


SuperBuild.cmake
```diff
@@ -118,7 +118,14 @@ if(Slicer_BUILD_DICOM_SUPPORT)
...
 
 if(Slicer_BUILD_DICOM_SUPPORT AND Slicer_USE_PYTHONQT_WITH_OPENSSL)
+  list(APPEND Slicer_DEPENDENCIES python-pydicom1)
   list(APPEND Slicer_DEPENDENCIES python-pydicom)
+  list(APPEND Slicer_DEPENDENCIES python-certifi)
+  list(APPEND Slicer_DEPENDENCIES python-idna)
+  list(APPEND Slicer_DEPENDENCIES python-chardet)
+  list(APPEND Slicer_DEPENDENCIES python-urllib3)
+  list(APPEND Slicer_DEPENDENCIES python-requests)
+  list(APPEND Slicer_DEPENDENCIES python-dicomweb-client)
```
jcfr: Could you dispatch the indirect dependencies in their respective external project ? That way only `python-dicomweb-client` would be listed here.

SuperBuild/External_python-certifi.cmake
```diff
@@ -0,0 +1,45 @@
+set(proj python-certifi)
```
jcfr: Not sure if this is needed, indeed we already provide certs within Slicer:
* https://github.com/Slicer/Slicer/tree/master/Base/QTCore/Resources/Certs
* https://github.com/Slicer/Slicer/blob/a3cb76715f1158d5e0c1b17982ee7e1836e3fcc5/CMake/SlicerBlockCTKAppLauncherSettings.cmake#L131-L135

SuperBuild.cmake
pieper: Thanks for the review Jc 👍 

What do you mean by dispatch?  I didn't find anyplace where these dependencies were listed, I only discovered them by trying to import the package.

SuperBuild/External_python-certifi.cmake
pieper: This package is a dependency when importing urllib3, so trying to use our own certificates would probably mean changing requests and urllib3, which is probably not a good idea.

SuperBuild.cmake
jcfr: The idea is to list the direct dependency here: https://github.com/Slicer/Slicer/pull/1081/files#diff-248ac0f903bd7e21b07e0b618e2bba3cR4

For pydicom, they can be extracted from here: https://github.com/clindatsci/dicomweb-client/blob/bfbaff8cfdc74d27fb30b9ea8cc462d1e317135d/setup.py#L58-L63

pieper: Ok, I see what you mean, yes, that would be cleaner.

SuperBuild/External_python-certifi.cmake
jcfr: It is an optional dependency, see https://github.com/urllib3/urllib3/blob/799f50d70720accd2343905ce8225062b8b099fe/setup.py#L66-L78

Also, by default urllib3 do not create https connection, it has to be explicitly done.  See https://github.com/urllib3/urllib3/blob/791e9b4a97d71e6fccb11b4d1c67b01cc0848776/docs/user-guide.rst#certificate-verification

This means that instead of using `certifi.where()`, the path to the Slicer cert bundle should probably be used.
See https://github.com/urllib3/urllib3/blob/791e9b4a97d71e6fccb11b4d1c67b01cc0848776/docs/advanced-usage.rst#custom-ssl-certificates

Now, I don't what this means for pydicom web client, this would have to be verified

pieper: I know I couldn't even import the package without certifi.

dicomweb-client uses urllib3 via requests, and apparently it does verify by default.  But it also looks like we could pass in a path to the certificates.

http://docs.python-requests.org/en/master/user/advanced/#ssl-cert-verification

It looks like if dicomweb-client could be imported then it would be possible to pass in the cert path to the client.

https://github.com/clindatsci/dicomweb-client/blob/12ea4e6c8b03298109feb6c52cbbe0455f12855c/src/dicomweb_client/api.py#L207-L208

This looks like something we could sort out with Markus next week when we're all together.

pieper: Looks like certifi is a hard requirement for requests, even if it's not for urlib3: 

https://github.com/requests/requests/blob/master/setup.py#L48

If I remove certify then requests won't import:
```
>>> import dicomweb_client
  File "/q5r/python-install/lib/python2.7/site-packages/dicomweb_client/__init__.py", line 3, in <module>
    from dicomweb_client.api import DICOMwebClient
  File "/q5r/python-install/lib/python2.7/site-packages/dicomweb_client/api.py", line 16, in <module>
    import requests
  File "/q5r/python-install/lib/python2.7/site-packages/requests/__init__.py", line 112, in <module>
    from . import utils
  File "/q5r/python-install/lib/python2.7/site-packages/requests/utils.py", line 24, in <module>
    from . import certs
  File "/q5r/python-install/lib/python2.7/site-packages/requests/certs.py", line 15, in <module>
    from certifi import where
ImportError: No module named certifi
>>> 
```

@jcfr , as we discussed at Project Week, while it would be best not to have two copies of the certificates, I don't see an obvious way around it in this case.  I'll go ahead and commit this and then we can change later if we think of a workaround.

pieper: One option could be to provide a fork of certifi that points to the Slicer cacert pem path, but that strikes me as more confusing than having two bundles.

jamesobutler:
I was looking at old open pull requests and noticed this was already merged. @pieper you can probably close this PR.

Integrated as [27955](https://github.com/Slicer/Slicer/commit/99142bce0076570388373f5746d022c5b7fa33e7), [27956](https://github.com/Slicer/Slicer/commit/dabd7138c6ed0dba8806066a69c88f32e0762fb7), [27957](https://github.com/Slicer/Slicer/commit/3b5e1167baf37b875f6b8190a787dc9594331441), [27958](https://github.com/Slicer/Slicer/commit/7c5c6ee991df04b733545efe28dfead16448bc76)

pieper:
Yes, this was added.

PR #942, opened 2018-04-27
SegmentEditor used too much space after returning to None effect

che85:
# Before changes

| Paint Effect        | None           | 
| ------------- |:-------------:|
| before01| before02       |

# After changes

| Paint Effect        | None           | 
| ------------- |:-------------:|
| after01 | after02   |

PR #1085, opened 2019-01-30
ENH: Friendlier segment editor widget for better intergration in layouts

vovythevov:
This PR aims to simplify integration of the Segment Editor widget in other widget.
Here is an illustration of what it looks like with and without this PR.
illustration

The code used to generate this is here: 
[SegmentEditorUIFix.zip](https://github.com/Slicer/Slicer/files/2814355/SegmentEditorUIFix.zip)


vovythevov:
Also, here is the SegmentEditor before/after the change:

segmenteditor_illustration

cpinter:
When we tried to do similar fixes in the past, there were always some bugs that came out when stress-testing that prevented applying that specific fix. So please try every option, in every effect, including the "Show details" button in the effect panels. Thanks!

che85:
That's right, but after setting up my new development environment I didn't get to continue working on it. 

Definitely looks good to me! Qt can be such a pain when it comes to sizing widgets.

jamesobutler:
So I tested this and it does appear to fix the empty space above and below the undo/redo buttons.

However, when selecting an active effect, the options groupbox becomes visible and increases the size of the widget and upon switching back to "None" the widget maintains its size, but the table is the driving force and pushes the effect groupbox down the page farther.  @che85 was trying to solve this in #942.  This is most noticeable for an effect with a large options groupbox such as "Logical Operators".  It is annoying when switching between effects for the effects groupbox to be moving up/down in the SegmentEditorWidget.  I would suggest that the table be a fixed size so that the effects groupbox is always in the same position in the widget.  This shouldn't be a problem because the table will show a vertical scroll bar once there are more contents than can be shown in the max size.

vovythevov:
Thanks everyone for the feedback ! I agree, layouts are finicky :)

I amended my commit and I think that provides a satisfying user experience. The table widget is now properly resizable (up to a minimum of a 100 pixel. That particular value is arbitrary, feel free to change it

). The effect buttons resizes itself properly when switching to and from each effect without affecting the size of the table.
The only thing is that there is a small jitter when switching from an effect to None. I am not sure where it comes from.

Here is a (low quality) gif to illustrate it all:
segmenteditordemo_lowquality

jamesobutler:
> The table widget is now properly resizable (up to a minimum of a 100 pixel. That particular value is arbitrary, feel free to change it).

I totally support that change for it to be resizeable instead of my suggested fixed size.  The changes to keep the effect groupbox in the same location upon switching effects looks great! Good work!  

I'm also not sure why adjusting the bottom widget and then changing to none causes that jitter in the layout policy. I'll see if I can help out. This is pretty close to being complete.

jamesobutler:
I have been messing around with this layout stuff for awhile without any success.  When going to the "None" effect, the options groupbox and masking groupbox are hidden which results in the segments table in the expandable widget expanding instead of maintaining its currently expanded size and the widget downsizing.

The flickering of the layout also exists in parts when simply collapsing/expanding the masking groupbox. So maybe worth trying to figure out that as a first step.

lassoan:
Csaba and I have spent days with trying to improve the layout behavior. There have been many other "almost good" solutions, which were better in several aspects but had some unacceptable issues. I don't think there is a simple fix, just by manipulating existing widget options.

I suspect that some of the CTK widgets that we use don't behave correctly and that throws off Qt's sizing strategy. We may fix the widgets, stop using certain widgets, or reorganize GUI layout to resolve this, but most likely this will require 1-2 weeks of work.

vovythevov:
Hi guys,

Thanks for all the interest on this. I agree, with you @lassoan, I think that issues are coming from the CTK widget (particurarly the ctkExpadingWidget) and this is shaping up to be a much bigger change that I anticipated. Sorry for opening pandora's box :(

If anything can be salvaged from this PR though, I think we could at least remove the spacer item from the widget and move it into the python widget (respectively [from here](https://github.com/Slicer/Slicer/pull/1085/files#diff-7f2770d07a8b11d16935a41020155195L472) and [to here](https://github.com/Slicer/Slicer/pull/1085/files#diff-c1bcf66514042746b8d6f481d166944fR63)).

I will let you decide what do with this PR, please feel free to close it, spin if off into an issue, cherry-pick part of it or whatever you think is appropriate.

jcfr:
I am jumping into the conversation and i am probably missing some points but I was wondering if the [ctkDynamicSpacer](https://github.com/commontk/CTK/blob/master/Libs/Widgets/ctkDynamicSpacer.h#L33) would be of any help.

```
/// Description
/// A spacer widget that has a dynamic size policy controllable via its slot
/// activate(bool). It can be usefully when you don't want a rigid layout. 
```

PR #1087, opened 2019-02-07
BUG: DicomWidgets - Do not assume that the extensions manager exists

vovythevov:
Since Slicer may not be built with the extension manager enabled,
qSlicerCoreApplication::extentionsManagerModel may not exists (it's
surrounded by #ifdef Slicer_BUILD_EXTENSIONMANAGER_SUPPORT).

### DIFF
--- a/Modules/Scripted/DICOMLib/DICOMWidgets.py
+++ b/Modules/Scripted/DICOMLib/DICOMWidgets.py
@@ -412,6 +412,9 @@ class DICOMDetailsBase(VTKObservationMixin, SizePositionSettingsMixin):
     """The dicom browser will emit multiple directoryImported
     signals during the same operation, so we collapse them
     into a single check for compatible extensions."""
+    if not hasattr(slicer.app, 'extensionsManagerModel'):
+      # Slicer may not be built with extension manager support
+      return
     if not self.extensionCheckPending:
       self.extensionCheckPending = True
       def timerCallback():


lassoan:
Wouldn't it be more appropriate to just disable "DICOM/PromptForExtensions" flag in application settings? Or do you already do that?

vovythevov:
@lassoan: I missed it, thanks for pointing that out !
No, I did not try that. Wouldn't the turning off ```DICOM/PromptForExtensions``` bypass the extensions check entirely ? It seems ```checkExtensions()``` checks for all the extensions and only return the ones that  [have not been installed](https://github.com/Slicer/Slicer/pull/1087/files#diff-72abf59758d07b4b52e73040fc7f4565L504) through the extension manager. Wouldn't it be possible for someone to have an extensions [installed](https://github.com/Slicer/Slicer/pull/1087/files#diff-72abf59758d07b4b52e73040fc7f4565L477) (maybe manually) and no extension manager ?


Alternatively, we could do something like this:
```python
promptForExtensionsEnabled = settingsValue('DICOM/PromptForExtensions', True, converter=toBool)
if not hasattr(slicer.app, 'extensionsManagerModel'):
  promptForExtensionsEnabled = False
```

lassoan:
In a custom application you probably don't want the popup to be displayed at all.

I like the code snippet above that you suggested (to force promptForExtensionsEnabled = False if there is no extension manager).

jcfr:
@vovythevov Here is the change between and after my update. It avoids the use of `QTimer` if the extension manager model is not available.

How does that look ?

```diff
diff --git a/Modules/Scripted/DICOMLib/DICOMWidgets.py b/tmp/DICOMWidgets.py
index 061c13583..af16e6cfc 100644
--- a/Modules/Scripted/DICOMLib/DICOMWidgets.py
+++ b/tmp/DICOMWidgets.py
@@ -412,15 +412,15 @@ class DICOMDetailsBase(VTKObservationMixin, SizePositionSettingsMixin):
     """The dicom browser will emit multiple directoryImported
     signals during the same operation, so we collapse them
     into a single check for compatible extensions."""
+    if not hasattr(slicer.app, 'extensionsManagerModel'):
+      # Slicer may not be built with extension manager support
+      return
     if not self.extensionCheckPending:
       self.extensionCheckPending = True
       def timerCallback():
         # Prompting for extension may be undesirable in custom applications.
         # DICOM/PromptForExtensions key can be used to disable this feature.
         promptForExtensionsEnabled = settingsValue('DICOM/PromptForExtensions', True, converter=toBool)
-        if not hasattr(slicer.app, 'extensionsManagerModel'):
-          # Slicer may not be built with an extension manager
-          promptForExtensionsEnabled = False
         if promptForExtensionsEnabled:
           self.promptForExtensions()
         self.extensionCheckPending = False
```

jcfr:
Closing. Integrated in [r28033](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28033)

PR #1089, opened 2019-02-08
ENH: adds hook for loading directory archetypes

pieper:
When a directory is added via the Add Data dialog,
it may contain a collection of files that should be
treated as a single MRML data type, such as a series
of image files that get loaded as a volume.

This change allows a qSlicerFileReader subclass to define
a method that can examine the contents of a directory and
filter out any files that should be loaded as a group.  One
of these files then serves as the archetype and the io properties
can be configured such that the files are loaded correctly.


pieper:
WIP: do not commit yet.

PR #1092, opened 2019-02-27
BUG: Fix deprecated use of QPixmap::grabWidget

jamesobutler:
This fixes a deprecation warning `QPixmap::grabWidget is deprecated, use QWidget::grab() instead`.
To replicate the warning, use ScreenCapture module to capture a single image of all views.

This use case was accidentally overlooked when https://github.com/Slicer/Slicer/commit/70c23d09a3fa6e970b6ae1c24e055b123e015e11 was committed which fixed other instances.  

Once Slicer is no longer trying to support both Qt4 and Qt5, then QWidget::grab() can be used directly.  Currently [ctkWidgetUtils.grabWidget](https://github.com/commontk/CTK/blob/cccef11c938cc946789ddb25ccc613287a33c56e/Libs/Widgets/ctkWidgetsUtils.cpp#L54-L57) is used which uses QPixmap::grabWidget if Qt4 and QWidget::grab if Qt5.

### DIFF
--- a/Modules/Scripted/ScreenCapture/ScreenCapture.py
+++ b/Modules/Scripted/ScreenCapture/ScreenCapture.py
@@ -851,7 +851,7 @@ class ScreenCaptureLogic(ScriptedLoadableModuleLogic):
       # no view is specified, capture the entire view layout
 
       # Simply using grabwidget on the view layout frame would grab the screen without background:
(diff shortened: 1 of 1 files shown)


jcfr:
Thanks for the patch :+1: 

Closing. Integrated in [r27969](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27969)

PR #1094, opened 2019-03-02
ENH: Reworked markups fiducials and added line, angle, and curves

lassoan:
Features:
- New markups: angle, line, open curve, closed curve
- Speed improvement: thousands of markup points can be visualized and interactively manipulated (previously, the limit was about 50-100 points)
- Live preview: preview of markup position and label is shown during placement in all views, current point is highlighted in all views
- Slide on surface: in 3D views, markups are kept on visible surfaces (previously, placement was done on visible surface, but adjustments were simply constrained to the view plane)
- Improved keyboard shortcuts:
  - LeftClick to jump to control point
  - LeftClick-and-drag: translate control point
  - Delete: delete highlighted markup
  - Ctrl-LeftClick: insert point along line
  - MiddleClick-and-drag: translate widget (experimental)

Fixes:
- Labels in 3D can no longer be partially obscured by 3D geometry (sometimes labels are visible when the corresponding markup is not visible anymore, but this will be fixed soon)
- Point picking in 3D views is unstable

Infrastructure improvements:
- Focus management is centralized (interactor style queries all the displayable managers if they can process the event and then forwards the event for processing to the most suitable one). This allows much faster event management than using VTK's picking manager.
- Interaction events are converted to vtkEventData type (originally introduced for virtual reality), which can carry all important metadata with them (display and world position, key modifiers, associated view and data nodes, etc.). This allows processing all kinds of interaction events (mouse, keyboard, virtual reality, any new interaction devices, etc.) using the same framework and modules have easy access to lots of metadata in all interaction events.
- Maximum number of markups can be specified for markups fiducials (e.g., if it is set to 1 then the markup node can be used for getting a single point from the user).
- Any point can be placed again (no user interface for this yet) - this can be used for predefining markup names and then place them.
- Keyboard shortcuts are configurable (no user interface yet) - flexible module/application specific keyboard mapping is feasible (VTK's event translator does not support event translation with modifiers, widget-state-dependent shortcuts, etc.)
- No more synchronization needed between MRML and widget states: widget states are directly stored in MRML, including point positions (as vtkPolyData) - in local and world coordinate system. This removed lots of synchronization overhead and potential for bugs.

Note that this first set of changes is just a beginning of a number of new features and improvements.

**See list of known bugs and limitations in this spreadsheet:**
https://1drv.ms/x/s!Arm_AFxB9yqHtvxAgdAfdLS_GHsOxw


Punzo:
Great! Thanks @lassoan 

I have tested and I have added only one additional point to the spreadsheet.

Finally I have a question: since we cleaned a lot the display manager, do we still propagate the click events similar as in the links below? (I could not see in the code any usage of the events)

https://github.com/Punzo/Slicer/blob/d82d002f210cdd424deb29852af72bef7f9bcb81/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.h#L160-L166

https://github.com/Punzo/Slicer/blob/d82d002f210cdd424deb29852af72bef7f9bcb81/Modules/Loadable/Markups/MRMLDM/vtkMRMLMarkupsDisplayableManager2D.cxx#L170-L218

If not then we should probably clean https://github.com/Slicer/VTK/pull/20 from the SelectionButton solution, but then still we need something else to propagate clicks events!

jcfr:
Outstanding :+1: :tada:  Well done @lassoan and @Punzo 

The timing is perfect. Indeed this week, @phcerdan will work on updating Slicer code base to use newer version of CMake, update ITK version, remove VTK7/Qt4 support, update developer instructions, .... This also means all outstanding PR from @hjmjohnson will be review and integrated :smile: 

> See list of known bugs and limitations in this spreadsheet:

After opening the link, there are no Active issue (I only see content in the Completed sheet). If all identified issues have been addressed (including last [comment](#issuecomment-469211423) of @Punzo ), I think we should move forward with the integration :racing_car:

lassoan:
OK! I'm fixing some test errors and after that I merge the changes. We tried to keep the API backward compatible, but probably there will be still some compatibility issues. We'll keep an eye on them and fix them as they are being reported.

cpinter:
I really like the vtkInteractionEventData stuff!

Will vtkCurveGenerator be then removed from SlicerMarkupsToModel?

jcfr:
> but probably there will be still some compatibility issues

whenever possible, I suggest we keep track of the update in https://www.slicer.org/wiki/Documentation/Nightly/Developers/Tutorials/MigrationGuide

> issues

One minor nitpick: After creating an open or close curve annotation, it is not possible to create a second annotation. the existing one is always being edited. I wonder what would be the best way to handle this.

lassoan:
> After creating an open or close curve annotation, it is not possible to create a second annotation.

You need to go to Markups module to create a new one. This behavior is similar to behavior of fiducials (and it is useful for example when you want to make adjustments to the views while placing points) but we can certainly change it. Would you have any suggestion?

Punzo:
> One minor nitpick: After creating an open or close curve annotation, it is not possible to create a second annotation. the existing one is always being edited. I wonder what would be the best way to handle this.

Maybe we can add a shortcut (SHIFT+left click?) to create a new one when we are in placing mode.

lassoan:
Merged in r27974.

jcfr:
> > After creating an open or close curve annotation, it is not possible to create a second annotation.

> You need to go to Markups module to create a new one. This behavior is similar to behavior of fiducials (and it is useful for example when you want to make adjustments to the views while placing points) but we can certainly change it. Would you have any suggestion?

That makes sense. 

I wonder if we should expose an annotation selector somewhere in the toolbar.

lassoan:
> since we cleaned a lot the display manager, do we still propagate the click events similar as in the links below? (I could not see in the code any usage of the events)

All _user interface events_ (such as clicks) are translated to _widget events_ (such as jump to markup) and invoked on the markups display node.

We can decide to observe them in the markups data node and invoke it from there, too.

The advantage of this (compared to directly invoking low-level click events) is that we can re-map keyboard shortcuts for custom applications or advanced features. I've added generic widget events to vtkSlicerAbstractWidget, such as WidgetCustomAction1, WidgetCustomAction2, etc. that can be mapped to any user interface event (e.g., we can map shift+double-click to WidgetCustomAction1; or map left-click to WidgetCustomAction1 and assign left-click's original widget event to some other user interface event).

jcfr:
Would it be possible to condlidate and/or update and/or move the `VTK Widgets improvements` and `Improving Markups` pages to the Completed section ? See https://www.slicer.org/wiki/Documentation/Labs/ViewInfrastructureImprovements

lassoan:
Consolidation with https://www.slicer.org/wiki/Documentation/Labs/VTKWidgets is done - see https://www.slicer.org/wiki/Documentation/Labs/Improving_Markups.

For now, I would not merge with https://www.slicer.org/wiki/Documentation/Labs/ViewInfrastructureImprovements page. I'll learn more about the view infrastructure improvements from the effort of making LandmarkRegistration module work with the update widget.

PR #1097, opened 2019-03-06
COMP: Update CMake minimum required version from 3.5 to 3.13 in internal projects

phcerdan:
Following on the CMake update in 2ba0c4d367130d1ef7fb18508fee3e435ccae0b0


jcfr:
Thanks @phcerdan :+1: 

Integrated in [r27993](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27993)

PR #1101, opened 2019-03-08
Remove support for building Slicer with C++98, Qt4 and VTK7

jcfr:


jcfr:
Integrated in [r28000](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28000) to [r28007](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28007)

PR #1105, opened 2019-03-11
BUG: Fix errors for model slice display manager when not in slice plane

Sunderlandkyl:
- If there is no input or if the input has no points, the vtkTransformPolyDataFilter will display an error message on every update: "No input data".
- To prevent the error, if the input is empty then the actor should not be visible since there is nothing to display.

### DIFF
--- a/Libs/MRML/DisplayableManager/vtkMRMLModelSliceDisplayableManager.cxx
+++ b/Libs/MRML/DisplayableManager/vtkMRMLModelSliceDisplayableManager.cxx
@@ -516,6 +516,18 @@ void vtkMRMLModelSliceDisplayableManager::vtkInternal
 #endif
     pipeline->Cutter->SetInputConnection(pipeline->ModelWarper->GetOutputPort());
 
+#if VTK_MAJOR_VERSION >= 9 || (VTK_MAJOR_VERSION >= 8 && VTK_MINOR_VERSION >= 2)
+    // If there is no input or if the input has no points, the vtkTransformPolyDataFilter will display an error message
+    // on every update: "No input data".
+    // To prevent the error, if the input is empty then the actor should not be visible since there is nothing to display.
+    pipeline->GeometryFilter->Update();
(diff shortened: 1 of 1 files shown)


Libs/MRML/DisplayableManager/vtkMRMLModelSliceDisplayableManager.cxx
```diff
@@ -516,6 +516,18 @@ ::UpdateDisplayNodePipeline(vtkMRMLDisplayNode* displayNode, const Pipeline* pip
 #endif
     pipeline->Cutter->SetInputConnection(pipeline->ModelWarper->GetOutputPort());
 
+#if VTK_MAJOR_VERSION >= 9 || (VTK_MAJOR_VERSION >= 8 && VTK_MINOR_VERSION >= 2)
```
jcfr: I just realized that we should have cleanup the code in https://github.com/Slicer/Slicer/commit/1e4cc6293144ced711b625a912f472e59043b143. We will keep this in this topic and proceed to the cleanup after this is integrated.

jcfr: All of that said, the change makes  sense :+1: 

Cc: @lassoan

jcfr (approved):

lassoan:
Thank you, this is a good workaround for these unnecessary messages. We should merge this and in the meantime I've started a [discussion on the VTK forum about removal of these messages](https://discourse.vtk.org/t/logging-error-messages-for-empty-input-data/453).

jamesobutler:
Thanks for trying to fix this! 

Xref https://github.com/lassoan/SlicerSegmentEditorExtraEffects/issues/11

jcfr:
Closing. Integrated in [r28017](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28017)

PR #1110, opened 2019-03-13
BUG: Fix MRML scene undo/redo mechanism

lassoan:
MRML scene undo/redo mechanism has been disabled for a long time and this commit keeps it that way. However, this commit contain a number of fixes for problems that initial testing of the undo/redo mechanism revealed. While the undo/redo infrastructure is still not ready for general use, these fixes allow enabling it for selected nodes in custom applications.

- Add flag to enable/disable undo/redo tracking for nodes. This allows enabling undo/redo only for selected nodes, that are implemented carefully to work well with undo/redo.
- Enforce maximum stack size on undo and redo stacks.
- Node IDs that are referenced within the node stack and are not within the scene are considered to be in use for the purpose of generating unique node ids (this makes sure no node ID conflicts can occur between new nodes and nodes that were restored from the undo stack).
- Add missing SaveStateForUndo calls into Markups widgets.
- Always save the full scene state - saving only state of selected nodes can break node references and make the state of the scene inconsistent. Even if a Slice core module would think that as a result of a user action only specific nodes are updated, due to behavior of custom modules, it is possible that a simple node change triggers events that change state of other nodes, which would not be possible to revert without saving the state of all nodes.
- Add flag to be able to disable automatic display node creation for Markups.
- Removed SaveStateForUndo calls from logic and MRML classes. Undo/Redo should be handled at the application/GUI level and not at MRML library level (at the library level it is not known when is a good time to save a new state).

Co-authored-by: Kyle Sunderland <sunderlandkyl@gmail.com>

### DIFF
--- a/Libs/MRML/Core/vtkMRMLNode.h
+++ b/Libs/MRML/Core/vtkMRMLNode.h
@@ -319,6 +319,14 @@ public:
   vtkSetMacro(Selectable, int);
   vtkBooleanMacro(Selectable, int);
 
+  /// Specifies if the state of this node stored in the scene's undo buffer.
+  /// False by default to make sure that undo can be enabled selectively,
+  /// only for nodes that are prepared to work correctly when saved/restored.
+  /// Nodes with different UndoEnabled value must not reference to each other,
+  /// because restoring states could lead to unresolved node references.
+  vtkGetMacro(UndoEnabled, bool);
+  vtkSetMacro(UndoEnabled, bool);
+  vtkBooleanMacro(UndoEnabled, bool);
 
   /// Propagate events generated in mrml.
   virtual void ProcessMRMLEvents ( vtkObject *caller, unsigned long event, void *callData );
@@ -893,6 +901,7 @@ protected:
   int Selectable;
   int Selected;
   int AddToScene;
+  bool UndoEnabled;
 
   int  SaveWithScene;
 
--- a/Libs/MRML/Core/vtkMRMLScene.cxx
+++ b/Libs/MRML/Core/vtkMRMLScene.cxx
@@ -117,9 +117,8 @@ vtkMRMLScene::vtkMRMLScene()
   this->UniqueNames.clear();
 
   this->Nodes =  vtkCollection::New();
-  this->UndoStackSize = 100;
+  this->MaximumNumberOfSavedUndoStates = 20;
   this->UndoFlag = false;
-  this->InUndo = false;
 
   this->NodeReferences.clear();
   this->ReferencedIDChanges.clear();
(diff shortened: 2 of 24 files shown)


Libs/MRML/Core/vtkMRMLNode.h
```diff
@@ -319,6 +319,14 @@ class VTK_MRML_EXPORT vtkMRMLNode : public vtkObject
   vtkSetMacro(Selectable, int);
   vtkBooleanMacro(Selectable, int);
 
+  /// Specifies if the state of this node stored in the scene's undo buffer.
+  /// False by default to make sure that undo can be enabled selectively,
+  /// only for nodes that are prepared to work correctly when saved/restored.
+  /// Nodes with different UndoEnabled value must not reference to each other,
+  /// because restoring states could lead to unresolved node references.
```
jcfr: Does this last sentence conflict with the notes from the commit message:

> Always save the full scene state - saving only state of selected nodes can break node references and make the state of the scene inconsistent. Even if a Slice core module would think that as a result of a user action only specific nodes are updated, due to behavior of custom modules, it is possible that a simple node change triggers events that change state of other nodes, which would not be possible to revert without saving the state of all nodes.

jcfr: In the commit message, it may be sensible to discuss the relationship with UndoEnabled node property.

Libs/MRML/Core/vtkMRMLScene.cxx
```diff
@@ -2701,56 +2815,66 @@ void vtkMRMLScene::Redo()
...
+        if (node && node->GetUndoEnabled())
+          {
+          undoMap[node->GetID()] = node;
+          }
         }
       }
     }
 
   //std::hash_map<std::string, vtkMRMLNode*>::iterator iter;
   //std::hash_map<std::string, vtkMRMLNode*>::iterator curIter;
```
jcfr: Remove these comments ?

Libs/MRML/Core/vtkMRMLScene.h
```diff
@@ -362,9 +371,15 @@ class VTK_MRML_EXPORT vtkMRMLScene : public vtkObject
   void SaveStateForUndo();
 
   /// Save current state of the node in the undo buffer
+  /// Deprecated! Use SaveStateForUndo() instead.
```
jcfr: You could use http://www.doxygen.nl/manual/commands.html#cmddeprecated

Libs/MRML/Core/vtkMRMLScene.cxx
```diff
@@ -1285,14 +1284,61 @@ vtkMRMLNode* vtkMRMLScene::AddNewNodeByClass(
...
+    return NULL;
+    }
+
+  nodeToAdd->SetID(nodeID.c_str());
+  if (!nodeBaseName.empty())
+    {
+    nodeToAdd->SetName(nodeBaseName.c_str());
+    }
+
+  return this->AddNode(nodeToAdd);
```
jcfr: I wonder if we should explicitly update the `UniqueIDs` map so that `GetUniqueIDIndex` doesn't accidentally override an ID explicitly provided.

jcfr: I just realized this is already taken care of in the modified `vtkMRMLScene::GetUniqueIDIndex`.

jcfr:
Closing. Integrated in [r28038](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28038)

PR #1109, opened 2019-03-13
STYLE: Replace VTK_OVERRIDE with override

phcerdan:
This commit replaces occurences of the VTK_OVERRIDE macro with override. This one-liner
was used to perform the update:

git grep -l "VTK_OVERRIDE" | fgrep -v CMakeLists.txt |fgrep -v .cmake | xargs sed -i '' -e "s/VTK_OVERRIDE/override/g"

PR #1108, opened 2019-03-12
STYLE: Replace ITK_NULLPTR with nullptr

phcerdan:
This commit replaces occurences of the ITK_NULLPTR macro with nullptr. This one-liner 
was used to perform the update:

```
git grep -l "ITK_NULLPTR" | fgrep -v CMakeLists.txt |fgrep -v .cmake | xargs sed -i '' -e "s/ITK_NULLPTR/nullptr/g"
```
It was adapted from https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/ITKv5Preparation/ReplaceITK_NULLPTRMacroNames.sh

Updated information can be found in the wiki:
 https://www.slicer.org/wiki/Documentation/Nightly/Developers/Tutorials/MigrationGuide/ObsoleteCodeRemoval#C.2B.2B11:_Update_source_code_to_use_nullptr

PR #1107, opened 2019-03-12
STYLE: Replace ITK_OVERRIDE with override

phcerdan:
This commit replaces occurences of the ITK_OVERRIDE macro with override. This one-liner
was used to perform the update:

```
git grep -l "ITK_OVERRIDE"  | fgrep -v CMakeLists.txt |fgrep -v .cmake | xargs sed -i '' -e "s/ITK_OVERRIDE/override/g"
```

It was adapted from https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/ITKv5Preparation/ReplaceITK_OVERRIDEMacroNames.sh

PR #1102, opened 2019-03-08
ENH: Modernize: use nullptr instead of NULL or 0

phcerdan:
Update associated with this PR are documented on the wiki. See [C++11: Update source code to use nullptr](https://www.slicer.org/wiki/Documentation/Nightly/Developers/Tutorials/MigrationGuide/ObsoleteCodeRemoval#C.2B.2B11:_Update_source_code_to_use_nullptr)

PR #1113, opened 2019-03-14
BUG: Fix ITK filters to use ITKv5 dynamic multi threading

phcerdan:
Fix failing tests with errors similar to:
```
Description: itk::ERROR: TransformDeformationFieldFilter(0x7faf052005c0): Subclass should override this method!!!
If old behavior is desired invoke this->DynamicMultiThreadingOff(); before Update() is called. The best place is in class constructor.
```

We have two options:
 - Keep using the old ITKv4 system, this just requires adding `this->DynamicMultiThreadingOff()` to the constructor.
 - Update to the dynamic multi threading system.

Check the ITK migration guide:
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/ITK5MigrationGuide.md

If the ThreadedGenerateData function does not use `threadId` (such
as in this case), we could just change the name of the function:
`ThreadedGenerateData` to: `DynamicThreadedGenerateData`
and delete the unused `threadId` argument.

This will allow a dynamic thread usage.

Note that a bug in itkGrowCutSegmentationImageFilter was discovered.
Issue opened in Mantis: https://issues.slicer.org/view.php?id=4679
And some inline comments about it. This should be tackled in another PR.

### DIFF
--- a/Modules/CLI/ResampleDTIVolume/itkDiffusionTensor3DResample.h
+++ b/Modules/CLI/ResampleDTIVolume/itkDiffusionTensor3DResample.h
@@ -75,7 +75,7 @@ public:
   void SetOutputParametersFromImage( InputImagePointerType Image );
 
 // /Get the time of the last modification of the object
-  unsigned long GetMTime() const ITK_OVERRIDE;
+  unsigned long GetMTime() const override;
 
   itkSetMacro( DefaultPixelValue, OutputDataType );
   itkGetMacro( DefaultPixelValue, OutputDataType );
@@ -95,13 +95,13 @@ protected:
 
   void DynamicThreadedGenerateData( const OutputImageRegionType & outputRegionForThread) override;
 
-  void BeforeThreadedGenerateData() ITK_OVERRIDE;
+  void BeforeThreadedGenerateData() override;
 
-  void GenerateOutputInformation() ITK_OVERRIDE;
+  void GenerateOutputInformation() override;
 
-  void AfterThreadedGenerateData() ITK_OVERRIDE;
+  void AfterThreadedGenerateData() override;
 
-  void GenerateInputRequestedRegion() ITK_OVERRIDE;
+  void GenerateInputRequestedRegion() override;
 
 private:
   typename InterpolatorType::Pointer      m_Interpolator;
--- a/Modules/CLI/ResampleDTIVolume/itkDiffusionTensor3DResample.h
+++ b/Modules/CLI/ResampleDTIVolume/itkDiffusionTensor3DResample.h
@@ -93,7 +93,7 @@ public:
 protected:
   DiffusionTensor3DResample();
 
-  void ThreadedGenerateData( const OutputImageRegionType & outputRegionForThread, ThreadIdType threadId ) ITK_OVERRIDE;
+  void DynamicThreadedGenerateData( const OutputImageRegionType & outputRegionForThread) override;
 
   void BeforeThreadedGenerateData() ITK_OVERRIDE;
 
--- a/Modules/CLI/ResampleDTIVolume/itkDiffusionTensor3DResample.txx
+++ b/Modules/CLI/ResampleDTIVolume/itkDiffusionTensor3DResample.txx
@@ -165,7 +165,7 @@ void
 DiffusionTensor3DResample<TInput, TOutput>
 ::AfterThreadedGenerateData()
 {
-  m_Interpolator->SetInputImage( NULL );
+  m_Interpolator->SetInputImage( nullptr );
 }
 
 /**
--- a/Modules/CLI/ResampleDTIVolume/itkDiffusionTensor3DResample.txx
+++ b/Modules/CLI/ResampleDTIVolume/itkDiffusionTensor3DResample.txx
@@ -83,10 +83,9 @@ DiffusionTensor3DResample<TInput, TOutput>
 template <class TInput, class TOutput>
 void
 DiffusionTensor3DResample<TInput, TOutput>
-::ThreadedGenerateData( const OutputImageRegionType &outputRegionForThread,
-                        ThreadIdType itkNotUsed(threadId) )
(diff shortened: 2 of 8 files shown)


lassoan:
Nice. Could you please make the comment more verbose (basically describe the same that in the pull request).

Modules/CLI/ResampleDTIVolume/itkDiffusionTensor3DResample.txx
```diff
@@ -29,6 +29,7 @@ DiffusionTensor3DResample<TInput, TOutput>
   m_OutputDirection.SetIdentity();
   m_OutputSize.Fill( 0 );
   m_DefaultPixelValue = static_cast<OutputDataType>( ITK_DIFFUSION_TENSOR_3D_ZERO );
+  this->DynamicMultiThreadingOff(); // Keep using ITKv4 threading
```
phcerdan: @lassoan this inline comment?

phcerdan:
```cpp
// Keep using the ITKv4 threading system.
// This class could be updated to use the ITKv5 dynamic threading system in the future
// Check the ITK migration guide: https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/ITK5MigrationGuide.md
this->DynamicMultiThreadingOff();
```

Too much?

phcerdan: Ok, I updated directly to use ITKv5 threading system. Check updated message.

phcerdan:
I add to replace override, NULL, etc, to avoid conflicts with the ongoing PRs about. It would be nice to have them merged as soon as possible to avoid future conflicts.

#1102  #1107 #1108 #1109 
@jcfr

jcfr:
Let s wait tomorrow after integrate the ITKv5 fix and having test passing.

Modules/CLI/ResampleDTIVolume/itkSeparateComponentsOfADiffusionTensorImage.h
```diff
@@ -61,14 +61,11 @@ class SeparateComponentsOfADiffusionTensorImage
...
   SeparateComponentsOfADiffusionTensorImage();
 
-  void ThreadedGenerateData( const OutputImageRegionType & outputRegionForThread, ThreadIdType threadId ) ITK_OVERRIDE;
+  void DynamicThreadedGenerateData( const OutputImageRegionType & outputRegionForThread) override;
 
-  void GenerateOutputInformation() ITK_OVERRIDE;
-
-  void GenerateInputRequestedRegion() ITK_OVERRIDE;
-
-private:
```
jcfr: Is removing `private` desired ?

Modules/CLI/ResampleDTIVolume/itkSeparateComponentsOfADiffusionTensorImage.h
```diff
@@ -61,14 +61,11 @@ class SeparateComponentsOfADiffusionTensorImage
 protected:
   SeparateComponentsOfADiffusionTensorImage();
 
-  void ThreadedGenerateData( const OutputImageRegionType & outputRegionForThread, ThreadIdType threadId ) ITK_OVERRIDE;
+  void DynamicThreadedGenerateData( const OutputImageRegionType & outputRegionForThread) override;
 
-  void GenerateOutputInformation() ITK_OVERRIDE;
-
-  void GenerateInputRequestedRegion() ITK_OVERRIDE;
```
jcfr: Note: I will integrate the commit the replacement of `ITK_OVERRIDE`. That said i will keep the change in a different commit for later integration

Modules/CLI/ResampleDTIVolume/itkSeparateComponentsOfADiffusionTensorImage.h
phcerdan: Yes, it was intended, there were no members or functions after the private keyword. It should have no effect.

Modules/CLI/ResampleDTIVolume/itkDiffusionTensor3DResample.h
```diff
@@ -93,7 +93,7 @@ class DiffusionTensor3DResample
 protected:
   DiffusionTensor3DResample();
 
-  void ThreadedGenerateData( const OutputImageRegionType & outputRegionForThread, ThreadIdType threadId ) ITK_OVERRIDE;
+  void DynamicThreadedGenerateData( const OutputImageRegionType & outputRegionForThread) override;
```
jcfr: In the commit message, the conditional time was used `we could just change the name of the function: ThreadedGenerateData to: DynamicThreadedGenerateData and delete the unused threadId argument.`

Does it means that this change is not needed to fix the actual bug ?

phcerdan: No, it's a grammatical error, it should be can instead of could. I applied that statement to a few cases in this commit.

jcfr: Thanks for clarifying.

Then, is it correct to say that (1) changing from  `ThreadedGenerateData` to `DynamicThreadedGenerateData` and (2) calling `DynamicMultiThreadingOff` are both required to fix the failing tests ?

phcerdan: That's correct. They are the two possible strategies, depending on the case (I mean, they cannot be applied together to the same case). And both were applied in this commit.

phcerdan:
Also, please note that the error in:

661 - py_LandmarkRegistration (Failed)
-- Solved in https://github.com/pieper/LandmarkRegistration/pull/27
Removing VTKv6 test (unsupported VTK version)

jcfr:
Closing. Integrated in [r28020](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28020)

PR #1114, opened 2019-03-15
ENH: add slice distance color mapping in 2d view for markups widgets

Punzo:
This adds the slice distance color mapping for all the line actors of the new markups widgets. 

It is possible to create the color mapping in two ways: 1) giving a procedural color node (by setting the reference in the display node of the markups); 2) using a default mapping which is possible to modify by changing the parameters of the display node: color, opacity, distance fading, saturation and hue offset. If the line color mapping node is set, it will overwrite the "default" one (i.e. 1 overwrites 2).

The PR fix also some style issues: removed many unnecessary includes and added parenthesis double spaces (it was not uniform). 

Still TO DO:
- [x] 1) Projection of the point for all the widget (project behind slice and in front of slice);
- [ ] 2) highlight widget-slice intersection with a thick circle or X where the contour intersects the slice exactly . 

Update: (2) will not be done in this PR.
Please @lassoan review anytime. Thanks!

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.cxx
@@ -15,11 +15,17 @@
 
 ==============================================================================*/
 
-// MRMLMarkups includes
+// MRML includes
 #include "vtkMRMLMarkupsDisplayNode.h"
+#include <vtkMRMLProceduralColorNode.h>
 
 // VTK includes
+#include <vtkCommand.h>
+#include <vtkDiscretizableColorTransferFunction.h>
+#include <vtkIntArray.h>
+#include <vtkNew.h>
 #include <vtkObjectFactory.h>
+#include <vtkPiecewiseFunction.h>
 
 // STL includes
 #include <sstream>
@@ -41,6 +47,9 @@ const char *vtkMRMLMarkupsDisplayNode::GlyphTypesNames[GlyphMax+2] =
   "StarBurst2D",
   "Sphere3D"
 };
+
+const char* vtkMRMLMarkupsDisplayNode::LineColorNodeReferenceRole = "lineColor";
+const char* vtkMRMLMarkupsDisplayNode::LineColorNodeReferenceMRMLAttributeName = "lineColorNodeRef";
 
 //----------------------------------------------------------------------------
 vtkMRMLNodeNewMacro(vtkMRMLMarkupsDisplayNode);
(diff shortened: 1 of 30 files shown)


lassoan:
Thank you, it works nicely. I've made some simplifications and style updates and rebased on latest master and forced push.

I've just noticed two issues.

1. When a control point projection is enabled then we should allow moving the point by drag-and-drop. If the user does not click on the point (press, release, without moving between) just presses left button, holds it down and moves the mouse then we should move the point in parallel with the slice plane, but without jumping the current slice to that location or snapping the point to current slice plane.

2. Arc of the angle widget is not visible if points are placed in certain slice views (e.g., in yellow slice view) .

@Punzo, it would be great if you could have a look.

Punzo:
> Thank you, it works nicely. I've made some simplifications and style updates and rebased on latest master and forced push.

ok, thanks for updating the code!

1) ok, but without making the point "Active" (i.e., green), right?

2) confirmed, working on it.

Punzo:
(2) done in https://github.com/Slicer/Slicer/pull/1114/commits/788dfb77215c5092bcc1aeed51ae9472268a95f3

working on (1)

lassoan:
Out-of-plane points would be moved the same way (temporarily making it active = green). The behavior would be changed for every visible control points: we would not snap to the current slice plane by default. Maybe we could have some keyboard/mouse shortcut to move the point to the current slice (which is the current behavior, but is very destructive when points are farther from the current slice plane).

Punzo:
> Out-of-plane points would be moved the same way (temporarily making it active = green). The behavior would be changed for every visible control points: we would not snap to the current slice plane by default. Maybe we could have some keyboard/mouse shortcut to move the point to the current slice (which is the current behavior, but is very destructive when points are farther from the current slice plane).

Ok!

Punzo:
@lassoan done in https://github.com/Slicer/Slicer/pull/1114/commits/769d3febc4f590b3a634c7afd4485ab7de5afb42

Now when a node is active, the projection node is green. 
The projection node now can be interacted (i.e., delete node, etc...). 
When click and dragging a projection node, it will be translated on the "original" position.
I added a key shortcut "p" to snap the point into the projection point (i.e. at the current slice). 

I also squashed all the commits in one. 

Please have a look. Thanks!

Punzo:
P.S.: I noticed a bug (not related to this PR): when we modify the display node of the markups, some events/interactions (e.g., delete a node) are disabled up to the markup node is modified with one of the interactions that works (e.g., click and drag).

lassoan:
Thank you, merged. 

https://github.com/Slicer/Slicer/commit/9165a4ce95f129b5a9854def6276cc1be991047d

PR #1115, opened 2019-03-15
ENH: Change display nodes so that they have a 2D and 3D visibility

cpinter:
- The node is visible in 2D if and only if both overall visibility and 2D visibility are enabled
- The node is visible in 3D if and only if both overall visibility and 3D visibility are enabled

The advantage of this decoupling is that we can show/hide of a node without changing its 2D/3D visibility flags.

Additional changes:
- Update subject hierarchy to use new 2D/3D visibility options
  - SH now does not toggle 2D (formerly slice intersection) visibility when toggling general visibility
  - Add Visibility plugin that allows toggling separate 2D/3D visibility for branches and items from the visibility context menu (right-click on eye)
  - Remove slice intersection visibility option from the Models SH plugin (because the new Visibility plugin supports it now)
- Fix tree view height calculation in Models module


Libs/MRML/Core/vtkMRMLDisplayNode.h
```diff
@@ -252,17 +278,29 @@ class VTK_MRML_EXPORT vtkMRMLDisplayNode : public vtkMRMLNode
   vtkBooleanMacro(Clipping, int);
 
   /// Set the slice intersection visibility of the display node.
-  /// \sa SliceIntersectionVisibility, GetSliceIntersectionVisibility(),
+  /// Function to manage \sa Visibility2D for backwards compatibility
```
jcfr: You could use `\deprecated`

Libs/MRML/Core/vtkMRMLDisplayNode.h
```diff
@@ -252,17 +278,29 @@ class VTK_MRML_EXPORT vtkMRMLDisplayNode : public vtkMRMLNode
...
-  /// \sa SliceIntersectionVisibility, SetSliceIntersectionVisibility(),
+  /// Function to manage \sa Visibility2D for backwards compatibility
+  /// \sa Visibility2D, SetSliceIntersectionVisibility(),
   /// SliceIntersectionVisibilityOn(), SliceIntersectionVisibilityOff()
-  vtkGetMacro(SliceIntersectionVisibility, int);
+  /// \deprecated GetSliceIntersectionVisibility
+  int GetSliceIntersectionVisibility();
+  /// Set the slice intersection visibility of the display node.
+  /// Function to manage \sa Visibility2D for backwards compatibility
+  /// \sa Visibility2D, SetSliceIntersectionVisibility(),
```
jcfr: Don't think `Visibility2D` visibility will link to the private member.

cpinter:
Awesome, thanks @jcfr for reviewing it after all!

pieper:
👍 LGTM - thanks for leaving in the old methods with the deprecation warning.

Libs/MRML/Core/vtkMRMLDisplayNode.h
cpinter: It's actually there already, just on the bottom of the dox, right above the function definition.

Libs/MRML/Core/vtkMRMLDisplayNode.h
cpinter: I left this in because I'm curious whether it is properly referenced or not. I don't see any good reason why not (private members are also listed in doxygen), and it's useful to link to those. I'll remove it later if they are not link.

cpinter:
Integrated in https://github.com/Slicer/Slicer/commit/1feb9fa64feeb52837c7a088aeae5b3d4b5b3ad0

PR #1116, opened 2019-03-16
ENH: Modernize: add override in virtual functions of derived classes

phcerdan:
After ITK_OVERRIDE and VTK_OVERRIDE have been replaced for `override`
we run clang-tidy to add overrides on virtual functions of derived
classes.

Example:

```diff
-virtual void ShowProgress()
+void ShowProgress() override
```

This allows compiler warnings and errors if the method
does not override the base class after a change in the base class.

Check the Slicer Wiki for an updated version of the documentation on
how to run clang-tidy.
https://www.slicer.org/wiki/Documentation/Nightly/Developers/Tutorials/MigrationGuide#C.2B.2B11:_Update_source_code_to_use_override

```
run-clang-tidy.py -checks=-*,modernize-use-override  -header-filter=.* -fix
```

Multiple `override` are added to the same method (probably because
running clang-tidy in parallel)

```cpp
void PrintSelf(ostream& os, vtkIndent indent) override override override;
```

Clean them with (15 here is just a high enough number):

```
for n in `seq 1 15`;
do
    ack "override override" -l --print0 | xargs -0 -n 1 sed -i \
    's/override override/override/g'
done
```

When all of them are fixed, you should see in the terminal:
```
sed: no input files
```


phcerdan:
It includes a single fix of VTK_OVERRIDE in Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidgetRepresentation.h. That was probably introduced in a later patch before the replacement.

jcfr:
Thanks @phcerdan :+1:

jcfr:
Closing. Integrated in [r28035](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28035)

PR #1120, opened 2019-03-19
Cli node references

cpinter:
It was possible to define references in CLIs before, but they had special meaning (set parent transform, set subject hierarchy parent), and they were defined "in reverse", i.e. in the referenced object.
This change adds support for generic, forward references, which allow specifying any type of reference between two nodes. For example in a registration module, the output transform can specify node references to the moving and fixed images.

Registration modules have been updated to use this reference feature to reference the input images from the output transformations. This change is needed for proper DICOM spatial registration export (https://github.com/SlicerRt/SlicerRT/issues/42)

The following PR is needed before this can be integrated: https://github.com/Slicer/SlicerExecutionModel/pull/109

### DIFF
--- a/Base/QTCLI/vtkSlicerCLIModuleLogic.cxx
+++ b/Base/QTCLI/vtkSlicerCLIModuleLogic.cxx
@@ -2455,13 +2455,17 @@ void vtkSlicerCLIModuleLogic::ApplyTask(void *clientdata)
     // same spot of the subject hierarchy as the reference.
     //        (subject, predicate, object) = (reference, setParent, volume/model)
     //
+    // 3. If a node has a forward reference (reference as a child element not an attribute, which
+    // specifies reference from the place where it was defined as opposed to 1. which points
+    // backwards), then a node reference is added from that node to the specified node with a given
+    // role.
+    //        (subject, predicate, object) = (node, addNodeReference, referencedNode)
+    //
     for (pgit = pgbeginit; pgit != pgendit; ++pgit)
       {
       // iterate over each parameter in this group
-      std::vector<ModuleParameter>::const_iterator pbeginit
-        = (*pgit).GetParameters().begin();
-      std::vector<ModuleParameter>::const_iterator pendit
-        = (*pgit).GetParameters().end();
+      std::vector<ModuleParameter>::const_iterator pbeginit = (*pgit).GetParameters().begin();
+      std::vector<ModuleParameter>::const_iterator pendit = (*pgit).GetParameters().end();
       std::vector<ModuleParameter>::const_iterator pit;
 
       for (pit = pbeginit; pit != pendit; ++pit)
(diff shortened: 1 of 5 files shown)


pieper:
Sounds good to me 👍

cpinter:
Thanks, Steve! Can we go ahead with https://github.com/Slicer/SlicerExecutionModel/pull/109 as well?

pieper:
Yes, both patches look good to me (haven't built, but looks good).  Thanks for the contributions! 👍

fedorov:
Just for the reference (no pun intended!), the original intent of adding "reference" to CLI was not motivated by transformations, but to initialize window/level for the resampled volume consistently with the input to the resampling tools. I managed to find the email below from 2009 when I emailed @hjmjohnson about adding this feature. I hope that feature is still working!

cpinter:
As far as I know this piece of code processes references in Slicer CLIs:
https://github.com/Slicer/Slicer/blob/master/Base/QTCLI/vtkSlicerCLIModuleLogic.cxx#L2449-L2534

And all I can see here are the transforms and the subject hierarchy parents. If the window/level feature stll works, then I don't know where it's implemented.

lassoan:
It would be great to copy window/level indeed (I don't think it was implemented in Slicer4). Generating default volume name based on input volume name would be useful, too.

@cpinter How do you think such features could be added in the future? Using special role names?

cpinter:
It's a good idea to use references for actions on a node using another node, because it can be donw without cluttering the CLI infrastructure, and handling these on the Slicer level would be easy. For example:
- Forward reference from the resampled volume to the input volume with role called "RequestWindowLevelCopy" or simiar
- OnNodeReferenceAdded could react on this request, then even remove the reference if it makes sense

cpinter:
@jcfr Three out of four tests fail in SEM on my Windows machine with or without these changes. The [dashboard](http://slicer.cdash.org/index.php?project=SlicerExecutionModel&date=2019-03-19) only shows linux tests, but those are fine. Are you aware of this? It seems unlikely that it's specific to my machine.

```
      Start  5: GenerateCLPTest-Slicer3Example-Configure
 5/15 Test  #5: GenerateCLPTest-Slicer3Example-Configure ....   Passed    6.18 sec
      Start  6: GenerateCLPTest-Slicer3Example-Build
 6/15 Test  #6: GenerateCLPTest-Slicer3Example-Build ........   Passed    3.18 sec
      Start  7: GenerateCLPTest-Slicer3Example-Test
 7/15 Test  #7: GenerateCLPTest-Slicer3Example-Test .........***Failed    0.10 sec
      Start  8: GenerateCLPTest-PointExample1-Configure
 8/15 Test  #8: GenerateCLPTest-PointExample1-Configure .....   Passed    6.20 sec
      Start  9: GenerateCLPTest-PointExample1-Build
 9/15 Test  #9: GenerateCLPTest-PointExample1-Build .........   Passed    3.05 sec
      Start 10: GenerateCLPTest-CLPExample1-Configure
10/15 Test #10: GenerateCLPTest-CLPExample1-Configure .......   Passed    6.13 sec
      Start 11: GenerateCLPTest-CLPExample1-Build
11/15 Test #11: GenerateCLPTest-CLPExample1-Build ...........   Passed    3.34 sec
      Start 12: GenerateCLPTest-CLPExample1-Test
12/15 Test #12: GenerateCLPTest-CLPExample1-Test ............***Failed    0.11 sec
      Start 13: GenerateCLPTest-CLPTestMultiple-Configure
13/15 Test #13: GenerateCLPTest-CLPTestMultiple-Configure ...   Passed    6.15 sec
      Start 14: GenerateCLPTest-CLPTestMultiple-Build
14/15 Test #14: GenerateCLPTest-CLPTestMultiple-Build .......   Passed    3.29 sec
      Start 15: GenerateCLPTest-CLPTestMultiple-Test
15/15 Test #15: GenerateCLPTest-CLPTestMultiple-Test ........***Failed    0.24 sec
```

Other than this, the forward references are not used on the CLP level, only in the application, so I'm not sure how I could test it here.

cpinter:
@jcfr @pieper @lassoan I cherry-picked my BRAINS commit from their master after it was merged to Slicer's BRAINS fork's master, but then I realized that it did not build, and that Slicer uses the slicer-2019-03-07-v5.0.0-2af1e31 branch not the master. What should I do? Create a new branch based on the slicer-2019-03-07-v5.0.0-2af1e31 branch? If yes, what should it be called? Thanks!

cpinter:
I talked to @lassoan and did the following:
- Cherry-picked that commit in the existing slicer-2019-03-07-v5.0.0-2af1e31 branch
- Removed the commit from the master

Now I'll update Slicer's SuperBuild.cmake to use the latest hash from the branch (and do the same for SEM) and push this PR's commits and that one.

cpinter:
Integrated in
https://github.com/Slicer/Slicer/commit/4ad9ed17f3531237d9a378241ee106cd326a6175
https://github.com/Slicer/Slicer/commit/62a9f9aeb1e90fb03264147635be111d2669eb20
https://github.com/Slicer/Slicer/commit/8301c8c58341cd86946572ca21485c6369ac8e2f
https://github.com/Slicer/Slicer/commit/fea81f73f0ad6a2f77c0849fea71b78d7f78ba5e

PR #1123, opened 2019-04-03
COMP: Fix ITK5 build errors in ResampleDTIVolume

cpinter:


Modules/CLI/ResampleDTIVolume/itkDiffusionTensor3DMatrix3x3Transform.h
```diff
@@ -96,7 +100,11 @@ class DiffusionTensor3DMatrix3x3Transform :
   VectorType                  m_Translation;
   VectorType                  m_Offset;
   PointType                   m_Center;
-  MutexLock::Pointer          m_Lock;
+#if ITK_VERSION_MAJOR >= 5
```
jcfr: Should we just keep the case of with ` #if ITK_VERSION_MAJOR >= 5 ` ?

cpinter: Sorry I don't understand what we should keep

jcfr: Instead of (re-)introducing the support for both ITKv5 and ITKv4, I suggest we simply update the file to use the ITKv5 convention.

In this case, it means that instead of having:

```
#if ITK_VERSION_MAJOR >= 5
  std::mutex                  m_Lock;
#else
  MutexLock                   m_Lock;
#endif
```

we would use only 

```
std::mutex                  m_Lock;
```

Is this PR related to https://discourse.slicer.org/t/slicerrt-does-not-build-due-to-missing-itk-remote-module/6214 ? 

And is the template code only used in extension ?

cpinter: No I got this build error when building a SlicerCustomApp. Not sure why it doesn't fail in a normal Slicer build.
I can update the PR to only use ITKv5, but then what happens if the user configures Slicer with Slicer_ITK_VERSION_MAJOR=4?

jcfr: Since we removed the support for ITKv4, it should be a non issue. See https://discourse.slicer.org/t/2019-03-05-hangout/6007/2?u=jcfr

I suggest to update the custom app following https://www.slicer.org/wiki/Documentation/Nightly/Developers/Tutorials/MigrationGuide#Transition_from_ITK4_to_ITK5

cpinter: OK.
I don't understand why the custom app build fails, it uses neither ResampleDTIVolume, or any ITK directly. Or rather, I don't understand why Slicer build was successful...

jcfr:
I suspect the code is building fine because ITK is built with `ITKDeprecated` module enabled:

https://github.com/Slicer/Slicer/blob/33586edd007f15eddc131e3afaf15276c185752b/SuperBuild/External_ITK.cmake#L85

In you custom app, is this module enabled ?

jcfr:
> suspect the code is building fine because ITK is built with ITKDeprecated module enabled:

I am checking if this is effectively the case.

jcfr:
It wasn't the issue because `vtkITK` explicitly require this module

jcfr:
> it uses neither ResampleDTIVolume, or any ITK directly.

Was the ResampleDTI module disabled in CMake ? 

By default, it is enabled in custom app:

https://github.com/KitwareMedical/SlicerCustomAppTemplate/blob/16d94b3cf74bb76c7421ea728d6e73db269166b3/%7B%7Bcookiecutter.project_name%7D%7D/CMakeLists.txt#L81-L85

jcfr:
Closing. Integrated in [r28081](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28081)

Similar changes have been applied to `vtkSlicerApplicationLogic` and `vtkSlicerCLIModuleLogic`. See [r28082](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28082)

cpinter:
Thanks a lot, @jcfr !

> Was the ResampleDTI module disabled in CMake

Not at first (the app uses the CropVolume module, and it uses this CLI). When I disabled it it built, but then CropVolume was not available. This CLI should be also enabled in Slicer because of CropVolume isn't it?

jcfr:
Yes. If there are any issues of this aoet, bring them up and we will
address them.

PR #1122, opened 2019-04-03
ENH: Enable gesture interactions in Slice and 3D views

Sunderlandkyl:
Adds Qt gesture interactions for pinch, rotate and pan gestures

Pinch:
- Slice view: Zoom in and out at mouse location
- 3D view: Zoom in and out at focal point

Rotate:
- Slice view: If slice intersection is visible, rotate parallel slices around current slice
- 3D view: Roll camera around current forward direction

Pan:
- Slice view: Translate slice plane
- 3D view: Translate camera position and focal point

___

Supersedes: https://github.com/Slicer/Slicer/pull/970

Requires a few changes in VTK to ensure that gestures are correctly captured:
https://github.com/Sunderlandkyl/VTK/commit/3ed40faaf3c785084f6eb13d562b30087d3010d6

Example video here:
https://www.dropbox.com/s/3zif5a4kvrad8xj/gesture_demo_2.mp4?dl=0

___

TODO:
- [x] Test gestures on touchscreen

PR #1124, opened 2019-04-09
ENH: Use common interactor style for slice and 3D views

lassoan:
Common interactor class is responsible for delegating interaction events to displayable managers (and if none of the displayable managers can process an event then emit an interaction event).

Translation from interaction event to camera manipulation events is not hardcoded anymore in the interactor style but it is moved to vtkMRMLCameraWidget.

Added support for double-click events in the interactor.

Added menu (right-click) and action (double-click) events to widgets. They are not handled yet in Slicer core.


lassoan:
@Sunderlandkyl I've integrated your https://github.com/Slicer/Slicer/pull/1122 PR into this one. Could you please test if the Mac touchpad gestures still work?

lassoan:
Merged in r28095

cpinter:
Nice! We can maybe make the right-click interaction pick a node and show a subject hierarchy context menu for the picked node. It should be quite easy to do.

pieper:
Looks great! 👍 

+100 for subject hierarchy context menu in the viewers.

lassoan:
We are very close to being able to show SH menu. I have all the pieces just need to put it together.

jcfr:
@lassoan Could update the migration guide to document the class renames ? See https://www.slicer.org/wiki/Documentation/Nightly/Developers/Tutorials/MigrationGuide#Slicer_backward_incompatible_changes

PR #1126, opened 2019-04-10
ENH: Pause rendering in views while user interactions are processed

Sunderlandkyl:
This prevents desynchronized update of displayable managers during user interaction (ie. slice intersection widget or segmentations lagging behind during slice translation)

When the user interacts with a view, PauseRenderEvent is invoked on the instance of vtkMRMLApplicationLogic.
Observers in qSlicerCoreApplication call pauseRender to temporarily stop rendering in all views.
When the user interaction has been processed ResumeRenderEvent is invoked, allowing rendering to continue.

### DIFF
--- a/Libs/MRML/Logic/vtkMRMLApplicationLogic.h
+++ b/Libs/MRML/Logic/vtkMRMLApplicationLogic.h
@@ -189,7 +189,9 @@ public:
 
   /// List of custom events fired by the class.
   enum Events{
-    RequestInvokeEvent = vtkCommand::UserEvent + 1
+    RequestInvokeEvent = vtkCommand::UserEvent + 1,
+    PauseRenderEvent = vtkCommand::UserEvent + 101,
+    ResumeRenderEvent
   };
   /// Structure passed as calldata pointer in the RequestEvent invoked event.
   struct InvokeRequest{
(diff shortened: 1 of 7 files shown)


jcfr:
Thanks :+1: 

Would it be possible to update the wiki to document this as well as the recent changes to the interactor style ? See https://www.slicer.org/wiki/Documentation/Nightly/Developers/DisplayableManagers

cc: @lassoan

pieper:
Thanks for working on this Kyle - the swimming widgets and segmentations have been bugging me.

lassoan:
Note that this is not related to the widgets rework, just happened to realize now that we can fix this. It is due to the forced rendering that we added to improve update speed on Mac (without that change, rendering did not happen on Mac until the message queue was empty, causing long delays when browsing slices). If needed, we may pause rendering at other GUI interactions, too (e.g., to the slice browsing slider).

pieper:
> Note that this is not related to the widgets rework

Right, I had noticed this on segmentations too.  Good to have a clean architecture that fixes everything uniformly.

Libs/MRML/Logic/vtkMRMLApplicationLogic.h
```diff
@@ -219,6 +221,11 @@ class VTK_MRML_LOGIC_EXPORT vtkMRMLApplicationLogic
   /// Uses current scene's URL property, so the URL must be up-to-date when calling this method.
   void SaveSceneScreenshot(vtkImageData* screenshot);
 
+  /// Invokes "PauseRenderEvent"
```
lassoan: It is an internal detail that the application logic implements pause/resume rendering by emitting an event. Here we should describe _what_ the method does (maybe include details such as when it should be used, etc.). We may describe _how_ it does it in the implementation (in the .cxx file), if it needs explanation.

Sunderlandkyl:
@lassoan Updated comments.

lassoan:
Thank you @Sunderlandkyl , I've added a missing PauseRender call (probably accidentally removed when the extra comment was added) and added pause/resume render to the slice browser slider, too. Committed r28098.

jcfr:
>  e.g., to the slice browsing slider

This is great.

PR #1128, opened 2019-04-17
BUG: Ensure scripted module tests fail if any module fail to be instantiated

jcfr:
This commit ensures that scripted module tests fail if any of the
associated modules fail to be instantiated.


jcfr:
While working on updating https://github.com/QIICR/PETTumorSegmentation to support qt5 and python3, I realized that scripted modules with python syntax error were not causing the test to fail.

Cc: @fedorov @pieper

jcfr:
Closing. Integrated in [r28152](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28152)

PR #1129, opened 2019-04-17
BUG: Ensure tests fail if an internal error is caught

jcfr:


jcfr:
While working on updating https://github.com/QIICR/PETTumorSegmentation to support qt5 and python3, I realized that execution of self-test leading to an internal error were displaying a popup while running the test. This commit ensure the application exit with an error code when running with `--testing` flag.

Cc: @fedorov @pieper

jcfr:
Closing. Integrated in [r28151](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28151)

PR #1131, opened 2019-04-23
BUG: Do not report error when logging message in ScriptedLoadableModuleWidget.cleanup

jcfr:
Since message logged from ScriptedLoadableModuleWidget.cleanup() happen
while the application is being destroyed, addEntry is not available at
that time because the error log model instance has already been deleted.


jcfr:
Closing. Integrated in [r28164](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28164) and [r28165](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28165)

PR #1132, opened 2019-04-24
ENH: BuildSystem: Add Slicer_BUILD_WEBENGINE_SUPPORT option

jcfr:
This new option allows to build custom Slicer applications without
WebEngine libraries. It reduces the install size by ~120MB

If Slicer_BUILD_WEBENGINE_SUPPORT option is OFF, the following happens:

  - exclude the following classes:

    - qSlicerWebWidget, qSlicerWebDownloadWidget and qSlicerWebPythonProxy
    - qMRMLExpandingWebView and qMRMLExpandingWebViewPlugin
    - qMRMLChartView, qMRMLChartWidget and qMRMLChartViewControllerWidget

  - views:

    - skip registration of qMRMLLayoutChartViewFactory in layout manager
       and return sensible values for method like viewNode(), viewWidget()

    - skip download of jqPlot only used in qMRMLChartView class

    - vtkMRMLChart nodes are still available and registered.

  - disable tab showing shortcuts in  qSlicerActionsDialog

  - datastore module is not available

  - extension manager:

    - enabling Slicer_BUILD_EXTENSIONMANAGER_SUPPORT while having
      Slicer_BUILD_WEBENGINE_SUPPORT off is supported.

    - installing from file is still supported.

    - in "Manage Extensons" tab, clicking on an extension link will NOT
      show a web-browser inside the manager.

    - in "Install Extensions" tab, nothing will be shown.

### DIFF
--- a/Base/QTApp/qSlicerApplicationHelper.cxx
+++ b/Base/QTApp/qSlicerApplicationHelper.cxx
@@ -121,6 +121,9 @@ void qSlicerApplicationHelper::preInitializeApplication(
   // Enable automatic scaling based on the pixel density of the monitor
   QApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
 
+  // Enables resource sharing between the OpenGL contexts used by classes like QOpenGLWidget and QQuickWidget
+  QApplication::setAttribute(Qt::AA_ShareOpenGLContexts);
+
   // Allow a custom application name so that the settings
   // can be distinct for differently named applications
   QString applicationName("Slicer");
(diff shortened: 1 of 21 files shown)


jcfr:
It turns out that QWebEngine is required by `qMRMLChartView`

lassoan:
Plots are replacing chart views, so it should not be an issue to not have charts. We could just disable charts if WebEngine is disabled.

jcfr:
> Plots are replacing chart views, so it should not be an issue to not have charts. We could just disable charts if WebEngine is disabled.

This is what I am aiming for.

jcfr:
This updated topic allows to build without webengine support

jcfr:
* [ ] fix crash on startup
* [ ] test packaging

jcfr:
Closing. Integrated in [r28172](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28172)

PR #1136, opened 2019-05-09
ENH: Added Frenet-Serret frame and spline driven image slicer

lassoan:
These classes will be used by Markups.

Source is at https://github.com/midas-journal/midas-journal-838/tree/master/Filters, which is a VTK remote module. I've tried to enable this remote module in Slicer's VTK but nothing happened (maybe I did it incorrectly) and thought that it could be useful if we can easily modify the files, so I went ahead and just copied the source files. But if it's easy to enable it as a VTK remote module then we may consider that instead of merging this pull request.

### DIFF
--- a/Libs/vtkAddon/CMakeLists.txt
+++ b/Libs/vtkAddon/CMakeLists.txt
@@ -72,6 +72,11 @@ set(vtkAddon_SRCS
   vtkStreamingVolumeCodecFactory.h
   vtkRawRGBVolumeCodec.cxx
   vtkRawRGBVolumeCodec.h
+  # From https://github.com/lorensen/midas-journal-838:
+  vtkFrenetSerretFrame.cxx
+  vtkFrenetSerretFrame.h
+  vtkSplineDrivenImageSlicer.cxx
+  vtkSplineDrivenImageSlicer.h
 )
 
 if(Slicer_VTK_RENDERING_USE_OpenGL2_BACKEND)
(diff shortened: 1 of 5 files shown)


pieper:
I agree that using this code via a VTK remote module would be preferable, but I have never tried that with Slicer's VTK so I'm not sure why it didn't work for you.  

I assume you just did the steps described [here](https://vtk.org/Wiki/VTK/Remote_Modules#To_add_a_Remote_Module_to_a_VTK_build)?  It seems like it should be possible to make that work for us.  

If the remote option really doesn't work then I don't see any problem including it in Slicer as a workaround.

lassoan:
Yes, I've tried those steps. The ...remote.cmake file was already there, so I just enabled the remote module, reconfigured and rebuilt VTK but nothing happened. No error and no module.

@jcfr Do you know if this is supposed to work? I can try again on a clean build if you think it should work.

lassoan:
I've found the issue why the remote module was ignored. It was an incorrect regexp in VTK source code. It was [fixed](https://github.com/Kitware/VTK/commit/0a90fe94a47d46e516bc36086ed619427acf968b) but the fix is not included in Slicer's VTK. Could we backport this fix?

jcfr:
> Could we backport this fix?

Definitively, I will update the Slicer/VTK fork shortly

jcfr:
> > Could we backport this fix?

> Definitively, I will update the Slicer/VTK fork shortly

Done in [r28219](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28219)

lassoan:
VTK remote modules now work well. This PR is superseded by https://github.com/Slicer/Slicer/pull/1139.

PR #1139, opened 2019-05-11
ENH: Enable SplineDrivenImageSlicer VTK remote module

lassoan:
It will be used by Markups module for getting curve normal direction.

### DIFF
--- a/SuperBuild/External_VTK.cmake
+++ b/SuperBuild/External_VTK.cmake
@@ -47,16 +47,22 @@ if((NOT DEFINED VTK_DIR OR NOT DEFINED VTK_SOURCE_DIR) AND NOT Slicer_USE_SYSTEM
       )
   endif()
 
+  # Markups module needs vtkFrenetSerretFrame, which is available in
+  # SplineDrivenImageSlicer remote module.
+  list(APPEND EXTERNAL_PROJECT_OPTIONAL_CMAKE_CACHE_ARGS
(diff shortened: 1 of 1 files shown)


SuperBuild/External_VTK.cmake
```diff
@@ -47,16 +47,22 @@ if((NOT DEFINED VTK_DIR OR NOT DEFINED VTK_SOURCE_DIR) AND NOT Slicer_USE_SYSTEM
...
     -DVTK_USE_GUISUPPORT:BOOL=ON
     -DVTK_USE_QVTK_QTOPENGL:BOOL=ON
     -DModule_vtkTestingRendering:BOOL=ON
     )
-    list(APPEND EXTERNAL_PROJECT_OPTIONAL_CMAKE_CACHE_ARGS
-      -DVTK_QT_VERSION:STRING=5
-      -DVTK_Group_Qt:BOOL=ON
-      -DQt5_DIR:FILEPATH=${Qt5_DIR}
-      )
+  list(APPEND EXTERNAL_PROJECT_OPTIONAL_CMAKE_CACHE_ARGS
```
jcfr: I suggest we fix indent in CMake modules in a dedicated commit. 

For some background, this was left indented when Qt4 support was removed with the intent to make it easier to "blame" and browse from  history of changes.

lassoan:
Thanks for the review merged into r28226 (without the indentation change).

PR #1140, opened 2019-05-16
ENH: Added quantification features to markups and fixed display scaling

lassoan:
First step towards implementing markup measurements, and some fixes.

Store measurements in markups in vtkMRMLMeasurement class instead of plain text fields. vtkMRMLMeasurement can also store measurements computed by other modules (such as Segment Statistics).

Fixed scaling of markups. Size can be set relatively to screen size or setting an absolute markup size.

Improve GUI:
- Make markup point labels hidden by default for all nodes except fiducials.
- Make qSlicerMarkupsPlaceWidget compatible with all markups


lassoan:
Committed r28258.

jcfr:
Outstanding! Thanks for integrating. Out at a hackathon for the week, I
didn't have a chance to look at the details. I will give a try tomorrow.

PR #1143, opened 2019-05-27
ENH: Improved automatic scalar volume display window/level computation

lassoan:
The old method (based on histogram bimodal analysis) rarely provided optimal results. The displayed image was often too bright.

A simple fixed percentile based method used now instead. Displayed intensity range now includes voxel intensities between 0.1th and 99.9th percentile.

Addresses issue discussed at https://discourse.slicer.org/t/feedback-requested-how-to-improve-mouse-interaction-in-views/6420/15


Base/QTApp/qSlicerApplicationHelper.cxx
```diff
@@ -106,6 +106,7 @@ void qSlicerApplicationHelper::preInitializeApplication(
   // support volume rendering and other VTK functionality that reads from the
   // framebuffer; see https://gitlab.kitware.com/vtk/vtk/issues/17095.
   QSurfaceFormat format = QVTKOpenGLWidget::defaultFormat();
+  format.setProfile(QSurfaceFormat::CompatibilityProfile);
```
jcfr: Could this be added into its own comment along with referencing the discussion on VTK discourse ?

It looks we should introduce an helper function and update all tests to use that. See https://github.com/Slicer/Slicer/blob/2f8f88fa961375b5c902fdc7ec5ca9e6d5200640/Base/QTGUI/Testing/Cxx/qSlicerMouseModeToolBarTest1.cxx#L42-L47

I agree that ideally we should improve VTK to initialize the surface appropriately but that is its own "can of worms".  to move forward, adding an helper function allowing to perform initialization for all tests would make sense i think.

lassoan:
Compatibility mode was included in this commit by mistake (I thought removed it but apparently not). I'll address it in a separate pull request and add an application settings key.

lassoan:
Merged r28267.

Updated migration guide:
https://www.slicer.org/wiki/Documentation/Nightly/Developers/Tutorials/MigrationGuide/Slicer#Slicer_4.11:_Display_window.2Flevel_.28brightness.2Fcontrast.29_adjustment

OpenGL compatibility profile will be addressed in a separate pull request (WIP).

PR #1145, opened 2019-05-29
BUG: fix linking of threeD views

pieper:
The new vtkMRMLCameraWidget captures the InteractionEvents
and so the qMRMLThreeDView never invoked the
CameraInteractionEvents that cause the views to be linked.

With this change the camera widget invokes an interaction
event on the render window interactor when it has been
processed so that the ViewLinkLogic knows to broadcast
the camera view.

### DIFF
--- a/Libs/MRML/DisplayableManager/vtkMRMLCameraWidget.cxx
+++ b/Libs/MRML/DisplayableManager/vtkMRMLCameraWidget.cxx
@@ -345,6 +345,16 @@ bool vtkMRMLCameraWidget::ProcessInteractionEvent(vtkMRMLInteractionEventData* e
       processedEvent = false;
     }
 
+  if (processedEvent)
+    {
+    // invoke interaction event for compatibility with pre-camera-widget
+    // behavior of vtk event processing.  This enables events to pass
+    // through the qMRMLThreeDView to the cameraNode
+    // for broadcast to other cameras
+    vtkRenderWindowInteractor* interactor = this->Renderer->GetRenderWindow()->GetInteractor();
+    interactor->InvokeEvent(vtkCommand::InteractionEvent);
+    }
+
   return processedEvent;
 }
 


pieper:
@lassoan can you see if this looks like the right fix?  There are many levels of events going on here but this change seems to adapt the new widget infrastructure to the existing view broadcast logic.

pieper:
Issue reported here: https://discourse.slicer.org/t/camera-linking-broken/6951/3

lassoan:
Good catch! I've just added start/end interaction event support a few days ago and I was wondering if I should add vtkCommand::InteractionEvent, too, but I was not sure if it was used. Apparently, it is used!

The change mostly looks good to me, the only thing that would be nice to check if there is any side effect of invoking the events outside of start/end interaction, for all kinds of events. You could consider firing it only from ProcessRotate, ProcessTranslate, ProcessScale, ProcessSpin methods (or at least check if we are in an interactive state with something like `interactorStyle->GetState() != VTKIS_NONE`).

pieper:
Thanks @lassoan - 👍   

I tried out the interaction modes and everything works as I expect.  I believe the interaction event state is handled correctly elsewhere in the code based on the Interacting member of the camera node, but if you think additional logic is needed here we can add it.

lassoan:
The current implementation is probably the simplest possible and if you don't experience any unintended side effects then we can leave it as is. We can revise it later if we find that something is not optimal. Maybe just add a comment there why this event is needed.

Also test after saving the scene and loading it. I've experienced that camera linking has issues on a loaded/imported scene (as I remember, cameras were duplicated or not correctly linked to views).

pieper:
I tested with three 3D views and there are definitely still issues with linking, but this bug fix is still an improvement so I went ahead and committed it.  I believe this addresses a SlicerMorph issue, so I hope it's robust enough for that and I don't see that it makes anything worse. 

At some I suggest we look at how events are being passed here - it's very indirect and can no doubt be simplified.

PR #1146, opened 2019-05-31
ENH: Add vtkMRMLTextNode and "Texts" module

Sunderlandkyl:
vtkMRMLTextNode and vtkMRMLTextStorageNode migrated to Slicer core from SlicerOpenIGTLink.

The Texts module adds a new designer plugin widget: "qMRMLTextWidget".
qMRMLText provides an text box with properties for readOnly, and continuousUpdate.

image


Sunderlandkyl:
- Added vtkMRMLTextNode and vtkMRMLTextStorageNode tests.
- Added warning messages when saving/cancelling an edit in qMRMLTextWidget could overwrite/discard data:
image 
image

Modules/Loadable/Texts/Widgets/qMRMLTextWidget.h
```diff
@@ -0,0 +1,131 @@
...
+
+  /// Get the table node the columns are edited of.
+  Q_INVOKABLE vtkMRMLTextNode* mrmlTextNode() const;
+  Q_INVOKABLE vtkMRMLNode* mrmlNode() const;
+
+  /// Returns true if the text editor is read only
+  bool readOnly();
+
+  /// Returns true if the text editor is continuously updated from the text node and vice versa
+  bool continuousUpdate();
```
jcfr: What do you think of using `autoUpdate` ?  This would be consistent with the  `updateWidget` and `updateFinished` method names

Modules/Loadable/Texts/Widgets/qMRMLTextWidget.h
```diff
@@ -0,0 +1,131 @@
...
+  void updateWidget();
+
+  /// Method invoked when the contents of the text node is modified
+  void onTextNodeModified();
+
+  /// Method invoked when the contents of the text edit is changed
+  void onTextChanged();
+
+  /// Method invoked when the "Edit" button is clicked
+  void onEditClicked();
```
jcfr: onEditClicked -> onEditButtonClicked ? Same comment for Cancel and Save

That said, I wonder if it makes sense to have these signals. Having signal like `updateRequested` and `updateCanceled` would be more generic  and would also nicely complement the existing `updateFinished` signal

Modules/Loadable/Texts/Widgets/qMRMLTextWidget.h
```diff
@@ -0,0 +1,131 @@
...
+  void onCancelClicked();
+
+  /// Method invoked when the "Save" button is clicked
+  void onSaveClicked();
+
+signals:
+  /// This signal is emitted if updates to the widget have finished.
+  void updateFinished();
+
+  /// This signal is emitted if the node changes
```
jcfr:
```
\sa setMRMLNode, mrmlNode()
```

Modules/Loadable/Texts/Widgets/qMRMLTextWidget.h
```diff
@@ -0,0 +1,131 @@
...
+
+public:
+  typedef qSlicerWidget Superclass;
+  qMRMLTextWidget(QWidget *parent=nullptr);
+  ~qMRMLTextWidget() override;
+
+  Q_PROPERTY(bool readOnly READ readOnly WRITE setReadOnly)
+  Q_PROPERTY(bool continuousUpdate READ continuousUpdate WRITE setContinuousUpdate)
+
+  /// Get the table node the columns are edited of.
```
jcfr: Comment should be updated

Modules/Loadable/Texts/Widgets/qMRMLTextWidget.h
```diff
@@ -0,0 +1,131 @@
...
+  void onSaveClicked();
+
+signals:
+  /// This signal is emitted if updates to the widget have finished.
+  void updateFinished();
+
+  /// This signal is emitted if the node changes
+  void mrmlNodeChanged(vtkMRMLNode*);
+
+  /// This signal is emitted if the read only propery is changed
```
jcfr: propery -> property

Modules/Loadable/Texts/Widgets/qMRMLTextWidget.h
```diff
@@ -0,0 +1,131 @@
...
+  void readOnlyChanged(bool);
```
jcfr: Instead of having a signal in the widget, should we instead have a a property in the text node ?

Libs/MRML/Core/vtkMRMLTextNode.h
```diff
@@ -0,0 +1,104 @@
...
+
+class  VTK_MRML_EXPORT vtkMRMLTextNode : public vtkMRMLStorableNode
+{
+public:
+  enum
+  {
+    ENCODING_US_ASCII = 3,
+    ENCODING_ISO_8859_1 = 4,
+    ENCODING_LATIN1 = ENCODING_ISO_8859_1 // alias
+                      // see other codes at http://www.iana.org/assignments/character-sets/character-sets.xhtml
```
jcfr: It would be nice to also support UTF8 and add test for it.

jcfr (changes requested):
This is great ! Thanks for working on this. 

Having first class support for text base document like json, xml, txt is very nice.

I wonder if this should be used as a base for other node stored as text or json (e.g markup)

Modules/Loadable/Texts/Widgets/qMRMLTextWidget.h
Sunderlandkyl: Sounds good! autoUpdate() is much better than continuousUpdate().

Modules/Loadable/Texts/Widgets/qMRMLTextWidget.h
Sunderlandkyl: I think that they would serve two different purposes, although having a readOnly property in the text node does seem like a useful feature.

- Read only in the qMRMLTextWidget allows text node content display without allowing editing (Whether or not the node is read only).

Libs/MRML/Core/vtkMRMLTextNode.cxx
```diff
@@ -0,0 +1,165 @@
...
+}
+
+//-----------------------------------------------------------------------------
+vtkMRMLTextNode::~vtkMRMLTextNode()
+{
+  this->SetText(NULL);
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLTextNode::SetText(const char* text)
```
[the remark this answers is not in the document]
Sunderlandkyl: Changed const* char to std::string.

Modules/Loadable/Texts/Widgets/qMRMLTextWidget.h
```diff
@@ -0,0 +1,131 @@
...
+
+  Q_PROPERTY(bool readOnly READ readOnly WRITE setReadOnly)
+  Q_PROPERTY(bool continuousUpdate READ continuousUpdate WRITE setContinuousUpdate)
+
+  /// Get the table node the columns are edited of.
+  Q_INVOKABLE vtkMRMLTextNode* mrmlTextNode() const;
+  Q_INVOKABLE vtkMRMLNode* mrmlNode() const;
+
+  /// Returns true if the text editor is read only
+  bool readOnly();
```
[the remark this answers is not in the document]
Sunderlandkyl: Done!

Modules/Loadable/Texts/Widgets/qMRMLTextWidget.h
Sunderlandkyl: Changed methods to XButtonClicked(). Added new signals (updateRequested, editingChanged(bool)).

Modules/Loadable/Texts/Widgets/qMRMLTextWidget.h
Sunderlandkyl: Done!

Modules/Loadable/Texts/Widgets/qMRMLTextWidget.h
Sunderlandkyl: Left ReadOnly out of the node for now.

Re a discussion with @lassoan, If it was implemented, it would probably be like the Locked parameter in markup and table nodes.

Libs/MRML/Core/vtkMRMLTextNode.h
Sunderlandkyl: Added a test for UTF-8, and decided to just use the VTK_ENCODING_X enum to represent encoding.

Modules/Loadable/Texts/Widgets/qMRMLTextWidget.h
Sunderlandkyl: Done!

Modules/Loadable/Texts/Widgets/qMRMLTextWidget.h
Sunderlandkyl: Done!

Modules/Loadable/Texts/Widgets/qMRMLTextWidget.h
Sunderlandkyl: Done!

Libs/MRML/Core/vtkMRMLTextNode.cxx
```diff
@@ -0,0 +1,148 @@
...
+const int LENGTH_BEFORE_STORAGE_NODE = 256;
+
+//----------------------------------------------------------------------------
+vtkMRMLNodeNewMacro(vtkMRMLTextNode);
+
+//-----------------------------------------------------------------------------
+vtkMRMLTextNode::vtkMRMLTextNode()
+  : Text("")
+  , Encoding(VTK_ENCODING_US_ASCII)
+  , ForceStorageNode(false)
```
lassoan: It would be better to use the variable name ForceCreateStorageNode with enum value: CreateStorageNodeAuto, CreateStorageNodeAlways, CreateStorageNodeNever (to make the behavior more predictable).

Libs/MRML/Core/vtkMRMLTextNode.cxx
```diff
@@ -0,0 +1,148 @@
...
+}
+
+//-----------------------------------------------------------------------------
+vtkMRMLTextNode::~vtkMRMLTextNode()
+{
+  this->SetText("");
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLTextNode::SetText(std::string text, int encoding/*VTK_ENCODING_NONE*/)
```
lassoan: Performance could be slightly better with const std::string &text (which might matter if the string is long and updates are frequent). If you don't see any drawback then use const ref instead.

Libs/MRML/Core/vtkMRMLTextNode.cxx
```diff
@@ -0,0 +1,148 @@
...
+{
+  if (!this->Scene)
+    {
+    return "";
+    }
+
+  if (!this->ForceStorageNode)
+    {
+    int length = this->Text.length();
+    if (length < LENGTH_BEFORE_STORAGE_NODE)
```
lassoan: MAX_STRING_LENGTH_FOR_SAVE_WITHOUT_STORAGE_NODE

Libs/MRML/Core/vtkMRMLTextNode.h
```diff
@@ -0,0 +1,96 @@
...
+  ///
+  /// Copy the node's attributes to this object
+  virtual void Copy(vtkMRMLNode* node) override;
+
+  ///
+  /// Get node XML tag name (like Volume, Model)
+  virtual const char* GetNodeTagName() override { return "Text"; };
+
+  /// Set text node contents
+  void SetText(const std::string text, int encoding=VTK_ENCODING_NONE);
```
lassoan: It is interesting that this compiled (const is in the header but not in the implementation). Anyway, it probably should be const std::string&.

Libs/MRML/Core/vtkMRMLTextNode.h
```diff
@@ -0,0 +1,96 @@
...
+  vtkGetMacro(Text, std::string);
+
+  ///
+  /// Set encoding of the text
+  /// For character encoding, please refer IANA Character Sets
+  /// (http://www.iana.org/assignments/character-sets/character-sets.xhtml)
+  /// Default is VTK_ENCODING_US_ASCII
+  vtkSetClampMacro(Encoding, int, VTK_ENCODING_NONE, VTK_ENCODING_UNKNOWN);
```
lassoan: Probably we should invoke TextModifiedEvent even if encoding was changed.

Libs/MRML/Core/vtkMRMLTextNode.h
```diff
@@ -0,0 +1,96 @@
...
+  /// in the scene (in XML), without using a storage node.
+  virtual vtkMRMLStorageNode* CreateDefaultStorageNode() override;
+
+  /// Determines the most appropriate storage node class for the
+  /// provided file name and node content.
+  virtual std::string GetDefaultStorageNodeClassName(const char* filename=nullptr) override;
+
+  enum
+  {
+    TextModifiedEvent = 51000,
```
lassoan: Document when this event is invoked. Also document if Modified event is not invoked for certain Set methods.

Modules/Loadable/Texts/Widgets/qMRMLTextWidget.cxx
```diff
@@ -0,0 +1,382 @@
...
+    {
+    // not changed
+    return;
+    }
+
+  // Reconnect the appropriate nodes
+  this->qvtkReconnect(d->CurrentTextNode, node, vtkMRMLTextNode::TextModifiedEvent, this, SLOT(onTextNodeModified()));
+  d->CurrentTextNode = node;
+  d->Editing = false;
+  this->updateRequested();
```
lassoan: Direction of update is not clear. Use updateWidgetFromMRML (updateWidgetFromMRMLRequested).

Modules/Loadable/Texts/Widgets/qMRMLTextWidget.cxx
```diff
@@ -0,0 +1,382 @@
...
+{
+  Q_D(qMRMLTextWidget);
+  return d->isEditing();
+}
+
+//------------------------------------------------------------------------------
+void qMRMLTextWidget::setReadOnly(bool readOnly)
+{
+  Q_D(qMRMLTextWidget);
+  if (this->ReadOnly == readOnly)
```
lassoan: What happens if you set read-only while the widget is being edited? Do you save the results? (probably should)

Modules/Loadable/Texts/Widgets/qMRMLTextWidget.cxx
```diff
@@ -0,0 +1,382 @@
...
+}
+
+//------------------------------------------------------------------------------
+void qMRMLTextWidget::onCancelButtonClicked()
+{
+  Q_D(qMRMLTextWidget);
+
+  if (d->TextEditModified)
+    {
+    QMessageBox msgBox;
```
lassoan: Proper QMessageBox deleting is tricky. You may need to make it auto-close if you don't destroy explicitly. Search for other similar messageboxes and copy their implementation.

Modules/Loadable/Texts/Widgets/qMRMLTextWidget.cxx
```diff
@@ -0,0 +1,382 @@
...
+//------------------------------------------------------------------------------
+void qMRMLTextWidget::onSaveButtonClicked()
+{
+  Q_D(qMRMLTextWidget);
+
+  int saveAction = QMessageBox::Save;
+  if (d->TextNodeModified)
+    {
+    QMessageBox msgBox;
+    msgBox.setText("The original text has been modified.");
```
lassoan: Mention that the text has been modified _since editing was started_.

Modules/Loadable/Texts/Widgets/qMRMLTextWidget.h
```diff
@@ -0,0 +1,139 @@
...
+  Q_PROPERTY(bool autoUpdate READ isAutoUpdate WRITE setAutoUpdate)
+
+  /// Get the text node
+  Q_INVOKABLE vtkMRMLTextNode* mrmlTextNode() const;
+  Q_INVOKABLE vtkMRMLNode* mrmlNode() const;
+
+  /// Returns true if the text editor is read only
+  bool isReadOnly();
+
+  /// Returns true if the text editor is continuously updated from the text node and vice versa
```
lassoan: Add a note on how auto-update works while editing is in progress.

Modules/Loadable/Texts/Widgets/qMRMLTextWidget.h
```diff
@@ -0,0 +1,139 @@
...
+  ~qMRMLTextWidget() override;
+
+  Q_PROPERTY(bool readOnly READ isReadOnly WRITE setReadOnly)
+  Q_PROPERTY(bool autoUpdate READ isAutoUpdate WRITE setAutoUpdate)
+
+  /// Get the text node
+  Q_INVOKABLE vtkMRMLTextNode* mrmlTextNode() const;
+  Q_INVOKABLE vtkMRMLNode* mrmlNode() const;
+
+  /// Returns true if the text editor is read only
```
lassoan: Add some more info about this. How the appearance and behavior is different if read-only or not.

Modules/Loadable/Texts/Widgets/qMRMLTextWidget.h
```diff
@@ -0,0 +1,139 @@
...
+
+  /// Method invoked when the "Cancel" button is clicked
+  void onCancelButtonClicked();
+
+  /// Method invoked when the "Save" button is clicked
+  void onSaveButtonClicked();
+
+signals:
+  /// This signal is emitted if updates to the widget have been requested.
+  void updateRequested();
```
lassoan: rename to update...From...Requested

Modules/Loadable/Texts/qSlicerTextsModule.cxx
```diff
@@ -0,0 +1,129 @@
...
+{
+  return QIcon(":/Icons/SlicerTexts.png");
+}
+
+
+//-----------------------------------------------------------------------------
+QStringList qSlicerTextsModule::categories() const
+{
+  return QStringList()
+    << "";
```
lassoan: Put it in the same category as Tables (Informatics).

lassoan (commented):
Very nice, thank you. I just added a couple of small comments that should be mostly trivial to address.

Modules/Loadable/Texts/Widgets/qMRMLTextWidget.cxx
Sunderlandkyl: This is how QMessageBox and ctkMessageBox are handled in qMRMLSceneViewMenu and qMRMLAnnotationTreeView.

Is there another style of implementation that would be better?

Sunderlandkyl:
I made most of the changes (With the exception of the QMessageBox change), and added some more lines in the test for the new storage node options (CreateStorageNodeAlways, CreateStorageNodeAuto, CreateStorageNodeNever).

Sunderlandkyl:
Message boxes removed in favor of a warning label under the save button:
image

lassoan:
Great, thank you. Maybe we could make it more subtle? For example, add a warning icon to the "Save" button and add the "The original text has been modified..." text to the button's tooltip. This way, we would not need to change the size of the widget and the warning icon may be more visible then the text below.

Sunderlandkyl:
Changed to icon+tooltip on save button:

image

lassoan:
Thank you, this looks good. I'll merge this next week after the CARS tutorial is finished, just to make sure we don't introduce any regressions in the last minute.

lassoan:
I've reviewed and tested this and pushed some fixes/improvements. There are just a few things that would be nice to address before merging:
- [ ] Clear widget contents when scene is closed.
- [ ] Add word wrap option (Boolean flag in the text edit widget and a checkbox in the module widget's advanced section)
- [ ] Double-check that Qt designer plugin works (for me the text edit widget did not show up in QtDesigner, but maybe I just did not rebuild enough)

Sunderlandkyl:
Text nodes in SH currently show node contents in tooltip:
image

"Edit text" in SH context menu:
image

lassoan:
Looks good!
What does "Edit text" do? If the same as "Edit properties..." (open Texts module with the text node selected) then it is probably redundant.

Sunderlandkyl:
Yeah it does. Didn't even notice that "Edit properties..." was there.

Sunderlandkyl:
Removed "Edit text" and added encoding info to the SH tooltip.

lassoan:
@Sunderlandkyl 

> added encoding info to the SH tooltip.

I'm not sure if this is needed. Encoding is only needed to interpret the byte stream and convert it to displayable string. If we already display the string in the tooltip it does not really matter what encoding was used (especially not for users). If the text is empty or very short then displaying encoding information might be distracting.

Let me know if you could address the other things (scene close, etc.) and it is ready for testing. Thank you!

@jcfr 

> I wonder if this should be used as a base for other node stored as text or json (e.g markup)

This is a good idea, but we need to think about it a bit more. This pull request already contains many changes, so I would merge this and get back to this idea later.

Sunderlandkyl:
Sorry, the other issues have been addressed.

What do you think the ideal tooltip should be?

1.
```
Text (Encoding: ASCII)
// Node contents
```
2.
```
Text
// Node contents
```
3.
```
// Node contents
```

1 and 2 are more consistent with other node types, while is much simpler.

lassoan:
I like both 1 and 3 (1 is more consistent with other nodes; 3 is more to the point). Let's just keep the current implementation then. We can tune later based on real-world experience. I'll merge this now.

lassoan:
Merged in r28328. Thanks a lot for the great work!

You can update SlicerOpenIGTLink now to not use its own text node in Slicer-4.11.

Sunderlandkyl:
Great, thanks Andras!

jcfr:
Currently working on GitHub transtion, data migration, ... I added a note to the authorship page to add you back as author with Andras as co-author and committer. See https://www.slicer.org/wiki/Documentation/Labs/TransitionToGit#Authorship

PR #1147, opened 2019-05-31
ENH: add option for black or white ruler

pieper:
Works for both slice and threeD views.

Saves and restores with mrml scene.

Is propagated with other attributes when views are linked.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLAbstractViewNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLAbstractViewNode.cxx
@@ -446,10 +446,10 @@ const char* vtkMRMLAbstractViewNode::GetRulerTypeAsString(int id)
 int vtkMRMLAbstractViewNode::GetRulerTypeFromString(const char* name)
 {
   if (name == nullptr)
-  {
+    {
     // invalid name
     return -1;
-  }
+    }
   for (int i=0; i<RulerType_Last; i++)
     {
     if (strcmp(name, GetRulerTypeAsString(i))==0)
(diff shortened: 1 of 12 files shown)


Libs/MRML/Core/vtkMRMLAbstractViewNode.cxx
```diff
@@ -457,6 +462,39 @@ int vtkMRMLAbstractViewNode::GetRulerTypeFromString(const char* name)
...
+      // invalid id
+      return "";
+    }
+}
+
+//-----------------------------------------------------------
+int vtkMRMLAbstractViewNode::GetRulerColorFromString(const char* name)
+{
+  if (name == nullptr)
+  {
```
lassoan: Indentation

pieper: I agree it's a bit odd, but this matches the style for GetRulerTypeAsString above so I figured it's better to stick with precedent.

pieper: Ah, I see,  you were talking about GetRulerColorFromString - yes, that's an indent issue I copied from GetRulerTypeFrom string - I'll fix them both.

pieper:
Yellow makes sense - I'll push an update in a minute.

jcfr:
Nitpick: would be great if could squash the commit, GitHub now keep track of it and it make it easy to review the  history in the future and understand the changes as it was accepted in.

pieper:
Yellow option added, squashed, and committed as r28283

pieper:
Thanks @jcfr - I force-pushed the squashed commit here for future reference.

lassoan:
Awesome, thank you!

PR #1149, opened 2019-06-07
Enabling PUBLIC/PRIVATE/INTERFACE keyword

adamrankin:
Macro slicerMacroBuildLoadableModule now supports PUBLIC/PRIVATE/INTERFACE keyword for target_link_libraries

### DIFF
--- a/CMake/SlicerMacroBuildLoadableModule.cmake
+++ b/CMake/SlicerMacroBuildLoadableModule.cmake
@@ -218,8 +218,9 @@ macro(slicerMacroBuildLoadableModule)
   set_target_properties(${lib_name} PROPERTIES LABELS ${lib_name})
 
   target_link_libraries(${lib_name}
-    ${Slicer_GUI_LIBRARY}
     ${LOADABLEMODULE_TARGET_LIBRARIES}
+    PUBLIC
+      ${Slicer_GUI_LIBRARY}
     )
 
   # Apply user-defined properties to the library target.
--- a/CMake/SlicerMacroBuildLoadableModule.cmake
+++ b/CMake/SlicerMacroBuildLoadableModule.cmake
@@ -218,6 +218,8 @@ macro(slicerMacroBuildLoadableModule)
   set_target_properties(${lib_name} PROPERTIES LABELS ${lib_name})
 
   target_link_libraries(${lib_name}
+    # The two PUBLIC keywords are not a duplication, they allow developers to 
+    # include PRIVATE/INTERFACE keywords in their library list
     PUBLIC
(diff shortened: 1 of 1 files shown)


jcfr:
Thanks for moving this forward.

I suggest we look at this with the idea of leveraging `target_include_directories` as well and simplifying the overall handling of include directories.

It may be worth creating a Labs page documentation the list of CMake macros/functions that should be updated along with changes to SlicerConfig/CTKConfig

lassoan:
I'm not against it at all, just curious, why this change was needed?

adamrankin:
When linking to an upstream extension that links a static lib (SlicerVideoCameras uses opencv_videoio), downstream extensions also try to link to opencv_videoio without knowing the target details.

opencv_videoio should be private to SlicerVideoCameras...Logic

adamrankin:
Ah, but my change doesn't work...

adamrankin:
If you can have `target_link_libraries(<target> PUBLIC PRIVATE lib1...)` then my latest commit should work

adamrankin:
I can confirm that this patch works as expected.

CMake/SlicerMacroBuildLoadableModule.cmake
```diff
@@ -218,8 +218,10 @@ macro(slicerMacroBuildLoadableModule)
   set_target_properties(${lib_name} PROPERTIES LABELS ${lib_name})
 
   target_link_libraries(${lib_name}
-    ${Slicer_GUI_LIBRARY}
-    ${LOADABLEMODULE_TARGET_LIBRARIES}
+    PUBLIC
+      ${LOADABLEMODULE_TARGET_LIBRARIES}
+    PUBLIC
+      ${Slicer_GUI_LIBRARY}
```
jcfr: This could be simplified to

```
target_link_libraries(${lib_name}
  PUBLIC
    ${LOADABLEMODULE_TARGET_LIBRARIES}
    ${Slicer_GUI_LIBRARY}
    )
```

Any particular reason, you used `PUBLIC` twice ?

adamrankin: That particular formulation fails if LOADABLEMODULE_TARGET_LIBRARIES has the PRIVATE keyword within it.

An alternative is to reorder it and put GUI_LIBRARY first.

lassoan: Add this info in a comment to prevent other developers in the future remove the "unnecessary" duplication.

jcfr:
Thanks @adamrankin  and @lassoan 

Closing. Integrated in [r28442](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28442)

PR #1150, opened 2019-06-11
ENH: Update SampleData to use checksum

jcfr:
This commit ensures a file is re-download if a previous download is
incomplete or if the the expected checksum is updated.

See https://discourse.slicer.org/t/improving-testing-data-management-for-self-test/5014


jcfr:
So that we can more confidently update the infrastructure on Kitware side, I will move forward with the integration.

jcfr:
Integrated in [r28290](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28290)

PR #1151, opened 2019-06-11
ENH: Allow Slicer install on Windows for non-admin users

lassoan:
This makes it easier to install Slicer for users that do not have administrator access.
Default install location is changed accordingly to C:\Users\<username>\AppData\Local (which is writable by regular users).

See this discussion for background: https://discourse.slicer.org/t/automatic-install-of-python-packages/7078/8

### DIFF
--- a/CMake/SlicerApplicationOptions.cmake
+++ b/CMake/SlicerApplicationOptions.cmake
@@ -114,3 +114,29 @@ if(NOT DEFINED ${Slicer_MAIN_PROJECT_APPLICATION_NAME}_SOURCE_DIR)
   set(${Slicer_MAIN_PROJECT_APPLICATION_NAME}_SOURCE_DIR ${CMAKE_SOURCE_DIR})
 endif()
 mark_as_superbuild(${Slicer_MAIN_PROJECT_APPLICATION_NAME}_SOURCE_DIR)
+
+#-----------------------------------------------------------------------------
+# Set default installation folder and admin account requirement for Windows
+#-----------------------------------------------------------------------------
+if(WIN32)
+
+  if(NOT DEFINED Slicer_CPACK_NSIS_INSTALL_ROOT)
+    # Set root install directory (displayed to end user at installer-run time)
+    # to C:\Users\<username>\AppData\Local by default to allow installation
+    # without having administrative privileges.
+    set(Slicer_CPACK_NSIS_INSTALL_ROOT "$LOCALAPPDATA" CACHE STRING
+      "Default installation location. Accepted values include $LOCALAPPDATA, $APPDATA, $PROGRAMFILES, $PROGRAMFILES64.")
+    mark_as_advanced(Slicer_CPACK_NSIS_INSTALL_ROOT)
+  endif()
+  mark_as_superbuild(Slicer_CPACK_NSIS_INSTALL_ROOT)
+  message(STATUS "Configuring ${Slicer_MAIN_PROJECT_APPLICATION_NAME} install root [${Slicer_CPACK_NSIS_INSTALL_ROOT}]")
+
+  if(NOT DEFINED Slicer_CPACK_NSIS_INSTALL_REQUIRES_ADMIN_ACCOUNT)
+    set(Slicer_CPACK_NSIS_INSTALL_REQUIRES_ADMIN_ACCOUNT OFF CACHE STRING
+      "Require administrator account to install the application. Must be enabled if Slicer_CPACK_NSIS_INSTALL_ROOT is only writable by administrators.")
+    mark_as_advanced(Slicer_CPACK_NSIS_INSTALL_REQUIRES_ADMIN_ACCOUNT)
+  endif()
+  mark_as_superbuild(Slicer_CPACK_NSIS_INSTALL_REQUIRES_ADMIN_ACCOUNT)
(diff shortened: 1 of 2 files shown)


jcfr:
Nice. When building app for commercial customer, i think we want to keep the option to install in program files. 

Setting variable like this would allow this from [KitwareMedical/SlicerCustomAppTemplate](https://github.com/KitwareMedical/SlicerCustomAppTemplate)
```
if(NOT DEFINED)
 set(Slicer_CPACK_NSIS_INSTALL_ROOT "$LOCALAPPDATA")
endif()
```

I also wonder what should be set to `CPACK_NSIS_DEFINES` to keep the ability to install into program files

lassoan:
I've added Slicer_CPACK_NSIS_INSTALL_ROOT and Slicer_CPACK_NSIS_INSTALL_REQUIRES_ADMIN_ACCOUNT optional CMake variables that can be overridden to change install location and requirement of using an admin account.

CMake/SlicerApplicationOptions.cmake
```diff
@@ -114,3 +114,29 @@ if(NOT DEFINED ${Slicer_MAIN_PROJECT_APPLICATION_NAME}_SOURCE_DIR)
...
+    # without having administrative privileges.
+    set(Slicer_CPACK_NSIS_INSTALL_ROOT "$LOCALAPPDATA" CACHE STRING
+      "Default installation location. Accepted values include $LOCALAPPDATA, $APPDATA, $PROGRAMFILES, $PROGRAMFILES64.")
+    mark_as_advanced(Slicer_CPACK_NSIS_INSTALL_ROOT)
+  endif()
+  mark_as_superbuild(Slicer_CPACK_NSIS_INSTALL_ROOT)
+  message(STATUS "Configuring ${Slicer_MAIN_PROJECT_APPLICATION_NAME} install root [${Slicer_CPACK_NSIS_INSTALL_ROOT}]")
+
+  if(NOT DEFINED Slicer_CPACK_NSIS_INSTALL_REQUIRES_ADMIN_ACCOUNT)
+    set(Slicer_CPACK_NSIS_INSTALL_REQUIRES_ADMIN_ACCOUNT OFF CACHE STRING
```
jcfr: option type should be `BOOL`

CMake/SlicerApplicationOptions.cmake
```diff
@@ -114,3 +114,29 @@ if(NOT DEFINED ${Slicer_MAIN_PROJECT_APPLICATION_NAME}_SOURCE_DIR)
...
+      "Require administrator account to install the application. Must be enabled if Slicer_CPACK_NSIS_INSTALL_ROOT is only writable by administrators.")
+    mark_as_advanced(Slicer_CPACK_NSIS_INSTALL_REQUIRES_ADMIN_ACCOUNT)
+  endif()
+  mark_as_superbuild(Slicer_CPACK_NSIS_INSTALL_REQUIRES_ADMIN_ACCOUNT)
```
jcfr: Let's be explicit:

```
mark_as_superbuild(Slicer_CPACK_NSIS_INSTALL_REQUIRES_ADMIN_ACCOUNT:BOOL)
```

CMake/SlicerApplicationOptions.cmake
```diff
@@ -114,3 +114,29 @@ if(NOT DEFINED ${Slicer_MAIN_PROJECT_APPLICATION_NAME}_SOURCE_DIR)
...
+
+  if(NOT DEFINED Slicer_CPACK_NSIS_INSTALL_ROOT)
+    # Set root install directory (displayed to end user at installer-run time)
+    # to C:\Users\<username>\AppData\Local by default to allow installation
+    # without having administrative privileges.
+    set(Slicer_CPACK_NSIS_INSTALL_ROOT "$LOCALAPPDATA" CACHE STRING
+      "Default installation location. Accepted values include $LOCALAPPDATA, $APPDATA, $PROGRAMFILES, $PROGRAMFILES64.")
+    mark_as_advanced(Slicer_CPACK_NSIS_INSTALL_ROOT)
+  endif()
+  mark_as_superbuild(Slicer_CPACK_NSIS_INSTALL_ROOT)
```
jcfr:
```
mark_as_superbuild(Slicer_CPACK_NSIS_INSTALL_ROOT:PATH)
```

CMake/SlicerApplicationOptions.cmake
```diff
@@ -114,3 +114,29 @@ if(NOT DEFINED ${Slicer_MAIN_PROJECT_APPLICATION_NAME}_SOURCE_DIR)
...
+#-----------------------------------------------------------------------------
+# Set default installation folder and admin account requirement for Windows
+#-----------------------------------------------------------------------------
+if(WIN32)
+
+  if(NOT DEFINED Slicer_CPACK_NSIS_INSTALL_ROOT)
+    # Set root install directory (displayed to end user at installer-run time)
+    # to C:\Users\<username>\AppData\Local by default to allow installation
+    # without having administrative privileges.
+    set(Slicer_CPACK_NSIS_INSTALL_ROOT "$LOCALAPPDATA" CACHE STRING
```
jcfr: let's set type to `PATH`. See https://cmake.org/cmake/help/latest/command/set.html#set-cache-entry

jcfr:
After applying these few tweaks. LGTM

jcfr:
Closing. Change integrated in [r28293](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28293)

PR #1152, opened 2019-06-11
SampleData: Fix support for custom downloader

lopezmt:
This is the first step to make it easier for the user to get started with SlicerSALT as suggested in https://github.com/Kitware/SlicerSALT/issues/119

### DIFF
--- a/Modules/Scripted/SampleData/SampleData.py
+++ b/Modules/Scripted/SampleData/SampleData.py
@@ -303,7 +303,7 @@ class SampleDataWidget(ScriptedLoadableModuleWidget):
           rowIndex += 1
           columnIndex = 0
         if source.customDownloader:
-          b.connect('clicked()', source.customDownloader)
+          b.connect('clicked()', lambda s=source: s.customDownloader(s))
         else:
           b.connect('clicked()', lambda s=source: self.logic.downloadFromSource(s))
 
@@ -412,6 +412,7 @@ class SampleDataLogic(object):
       loadFiles=loadFiles,
       loadFileProperties=loadFileProperties,
       checksums=checksums,
+      customDownloader=customDownloader,
       ))
 
   def __init__(self, logMessage=None):
@@ -421,6 +422,8 @@ class SampleDataLogic(object):
     self.developmentCategoryName = 'Development'
     self.registerBuiltInSampleDataSources()
     self.registerDevelopmentSampleDataSources()
+    if slicer.app.testingEnabled():
+      self.registerTestingDataSources()
 
   def registerBuiltInSampleDataSources(self):
     """Fills in the pre-define sample data sources"""
@@ -481,6 +484,10 @@ class SampleDataLogic(object):
       loadFileType=['VolumeFile', 'SegmentationFile'],
       checksums=['SHA256:c0743772587e2dd4c97d4e894f5486f7a9a202049c8575e032114c0a5c935c3b', 'SHA256:3243b62bde36b1db1cdbfe204785bd4bc1fbb772558d5f8cac964cda8385d470']
       )
+
+  def registerTestingDataSources(self):
+    """Register sample data sources used by SampleData self-test to test module functionalities."""
+    self.registerCustomSampleDataSource(**SampleDataTest.CustomDownloaderDataSource)
 
   def downloadFileIntoCache(self, uri, name, checksum=None):
     """Given a uri and and a filename, download the data into
@@ -794,8 +801,11 @@ class SampleDataTest(ScriptedLoadableModuleTest):
   https://github.com/Slicer/Slicer/blob/master/Base/Python/slicer/ScriptedLoadableModule.py
   """
 
+  customDownloads = []
+
   def setUp(self):
     slicer.mrmlScene.Clear(0)
+    SampleDataTest.customDownloads = []
 
   def runTest(self):
     for test in [
@@ -810,7 +820,8 @@ class SampleDataTest(ScriptedLoadableModuleTest):
       self.test_downloadFromSource_loadNodes,
       self.test_downloadFromSource_loadNodesWithLoadFileFalse,
       self.test_isSampleDataSourceRegistered,
-      self.test_categoryVisibility
+      self.test_categoryVisibility,
+      self.test_customDownloader,
     ]:
       self.setUp()
       test()
@@ -964,3 +975,29 @@ class SampleDataTest(ScriptedLoadableModuleTest):
     self.assertFalse(widget.isCategoryVisible('BuiltIn'))
     widget.setCategoryVisible('BuiltIn', True)
     self.assertTrue(widget.isCategoryVisible('BuiltIn'))
+
+  class CustomDownloader(object):
+    def __call__(self, source):
+      SampleDataTest.customDownloads.append(source)
+
+  CustomDownloaderDataSource = {
+    'category': "Testing",
+    'sampleName': 'customDownloader',
+    'uris': 'http://down.load/test',
+    'fileNames': 'cust.om',
+    'customDownloader': CustomDownloader()
+  }
+
+  def test_customDownloader(self):
+    if not slicer.app.testingEnabled():
+      return
+    slicer.util.selectModule("SampleData")
+    widget = slicer.modules.SampleDataWidget
+    button = slicer.util.findChild(widget.parent,'customDownloaderPushButton')
+
+    self.assertEqual(self.customDownloads, [])
+
+    button.click()
+
+    self.assertEqual(len(self.customDownloads), 1)
+    self.assertEqual(self.customDownloads[0].sampleName, 'customDownloader')


Modules/Scripted/SampleData/SampleData.py
```diff
@@ -315,6 +315,14 @@ def logMessage(self, message, logLevel=logging.INFO):
     logging.log(logLevel, message)
     slicer.app.processEvents(qt.QEventLoop.ExcludeUserInputEvents)
 
+  def setCategoryVisible(self,categoryName, visible):
+    for i in range(self.layout.count()):
+      widget = self.layout.itemAt(i).widget()
+      if hasattr(widget, "name") and widget.name == '%sCollapsibleGroupBox' % categoryName:
+        if visible:
```
jcfr:
```
widget.setVisible(visible)
```

Modules/Scripted/SampleData/SampleData.py
```diff
@@ -315,6 +315,14 @@ def logMessage(self, message, logLevel=logging.INFO):
     logging.log(logLevel, message)
     slicer.app.processEvents(qt.QEventLoop.ExcludeUserInputEvents)
 
+  def setCategoryVisible(self,categoryName, visible):
```
jcfr: `(self,categoryName, visible):` -> `(self, categoryName, visible):`

jcfr: add docstring

```
"""Update visibility of a SampleData category given its name.
"""
```

jcfr: I suggest you introduce a

```
def isRegisteredDataSource(category):
  return category in slicer.modules.sampleDataSources.keys()
```

and then use it in the function

Modules/Scripted/SampleData/SampleData.py
```diff
@@ -315,14 +315,21 @@ def logMessage(self, message, logLevel=logging.INFO):
...
-          widget.hide()
+  def setCategoryVisible(self, categoryName, visible):
+    """Update visibility of a SampleData category given its name.
+    """
+    if self.isRegisteredDataSource(categoryName):
+      for i in range(self.layout.count()):
+        widget = self.layout.itemAt(i).widget()
+        if hasattr(widget, "name") and widget.name == '%sCollapsibleGroupBox' % categoryName:
+          widget.setVisible(visible)
+          return
```
jcfr: I suggest to simplify even further  to something like this:

```python
  def setCategoryVisible(self, categoryName, visible):
    """Update visibility of a SampleData category given its name.
    """
    if not self.isRegisteredDataSource(categoryName):
      return
    slicer.util.findChild(self.parent, '%sCollapsibleGroupBox' % categoryName).setVisible(visible)
```

lopezmt: This is a really useful function! is there any documentation somewhere that lists all the function in slicer.util?

lassoan: See https://slicer.readthedocs.io/en/latest/developer_guide/slicer.html

lopezmt:
This is even shorter, is it still better to split the test and the action or is it fine in one block like that?

```python 
def setCategoryVisible(self, categoryName, visible):
    """Update visibility of a SampleData category given its name.
    """
    if self.isRegisteredDataSource(categoryName):
      slicer.util.findChild(self.parent, '%sCollapsibleGroupBox' % categoryName).setVisible(visible)
```

Modules/Scripted/SampleData/SampleData.py
lopezmt: This is even shorter, is it still better to split the test and the action or is it fine in one block like that?

```python 
def setCategoryVisible(self, categoryName, visible):
    """Update visibility of a SampleData category given its name.
    """
    if self.isRegisteredDataSource(categoryName):
      slicer.util.findChild(self.parent, '%sCollapsibleGroupBox' % categoryName).setVisible(visible)
```

jcfr: It is usually more readable to split things.
```
# check, early exit if needed
# then action
```

jcfr:
* Consider squashing all three commit together and then force-pushing the topic.
* Prefer imperative when writing commit message. See https://www.slicer.org/wiki/Documentation/Nightly/Developers/Style_Guide#Commits
* use `category` instead of `categoryName` (to be consistent with remaining of the code, I incorrectly suggested this earlier)
* add `isCategoryVisible(category)`
* add `test_widget_categoryVisiblity`

lopezmt:
I have never created a test function to test a function, can you tell me what needs to be check to consider that a function is tested?

jcfr:
Thanks for adding tests :+1:  

I suggest to use `isRegisteredCategory` instead of `isRegisteredDataCategory` 

> customDownloader works and remains working?

To follow up on what we discussed earlier today, consider changing this function:

```
  def __init__(self, logMessage=None):
    if logMessage:
      self.logMessage = logMessage
    self.builtInCategoryName = 'BuiltIn'
    self.developmentCategoryName = 'Development'
    self.registerBuiltInSampleDataSources()
    self.registerDevelopmentSampleDataSources()
```

into 

```
  def __init__(self, logMessage=None):
    if logMessage:
      self.logMessage = logMessage
    self.builtInCategoryName = 'BuiltIn'
    self.developmentCategoryName = 'Development'
    self.registerBuiltInSampleDataSources()
    self.registerDevelopmentSampleDataSources()
    if slicer.app.testingEnabled():
      self.registerTestingDataSources()
```

Then implement `registerTestingDataSources`  to register data source with custom downloader and add a a test function `test_dataSourceWithCustomDownloader` that would check it works as expected. The custom downloader could simply raise an RuntimeException and the you would use https://docs.python.org/3.6/library/unittest.html#unittest.TestCase.assertRaises

jcfr:
The support for show/hide of category has been integrated in [r28314](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28314)

The remaining of the topic has been rebased and updated against the latest Slicer.

@lopezmt For other development, considering updating your Slicer checkout.

Modules/Scripted/SampleData/SampleData.py
```diff
@@ -483,6 +490,16 @@ def registerDevelopmentSampleDataSources(self):
       checksums=['SHA256:c0743772587e2dd4c97d4e894f5486f7a9a202049c8575e032114c0a5c935c3b', 'SHA256:3243b62bde36b1db1cdbfe204785bd4bc1fbb772558d5f8cac964cda8385d470']
       )
 
+  def registerTestingDataSources(self):
+    """Fills in the sample data sources displayed only if developer mode is enabled."""
+
+    self.registerCustomSampleDataSource(
+      category=self.testingCategoryName, sampleName='customDownloader',
```
jcfr: Look like `self.testingCategoryName` is invalid

You could simply use a string like `"Testing"`

Modules/Scripted/SampleData/SampleData.py
```diff
@@ -965,3 +983,13 @@ def test_categoryVisibility(self):
     self.assertFalse(widget.isCategoryVisible('BuiltIn'))
     widget.setCategoryVisible('BuiltIn', True)
     self.assertTrue(widget.isCategoryVisible('BuiltIn'))
+
+  def test_customDownloader(self):
```
jcfr:
```
slicer.util.selectModule("SampleData")
```

Modules/Scripted/SampleData/SampleData.py
```diff
@@ -965,3 +983,13 @@ def test_categoryVisibility(self):
...
+    logic = SampleDataLogic()
+    widget = slicer.modules.sampledata.widgetRepresentation().self()
```
jcfr: -> `widget = slicer.modules.SampleDataWidget`

Modules/Scripted/SampleData/SampleData.py
```diff
@@ -415,13 +415,20 @@ def registerCustomSampleDataSource(category='Custom',
       customDownloader=customDownloader,
       ))
 
+    
+  @staticmethod
+  def CustomDownloadTest(source):
+    raise Exception('CustomDownloadTesting')
```
jcfr: Use  `RuntimeError`, `Exception` is too general and the test could instead catch an unrelated exception

jcfr: `CustomDownloadTesting` -> `CustomDownloadTesting: This exception confirms that setting a custom download works`

Modules/Scripted/SampleData/SampleData.py
```diff
@@ -415,13 +415,20 @@ def registerCustomSampleDataSource(category='Custom',
       customDownloader=customDownloader,
       ))
 
+    
+  @staticmethod
+  def CustomDownloadTest(source):
```
jcfr: Compare source parameter is the registered source and raise an `AssertError` if not the expected one.

Modules/Scripted/SampleData/SampleData.py
```diff
@@ -483,6 +490,16 @@ def registerDevelopmentSampleDataSources(self):
       checksums=['SHA256:c0743772587e2dd4c97d4e894f5486f7a9a202049c8575e032114c0a5c935c3b', 'SHA256:3243b62bde36b1db1cdbfe204785bd4bc1fbb772558d5f8cac964cda8385d470']
       )
 
+  def registerTestingDataSources(self):
+    """Fills in the sample data sources displayed only if developer mode is enabled."""
```
jcfr: ``Fills in the sample data sources displayed only if developer mode is enabled.` -> `Register sample data sources used by SampleData self-test to test module functionalities.`

Modules/Scripted/SampleData/SampleData.py
```diff
@@ -965,3 +983,13 @@ def test_categoryVisibility(self):
     self.assertFalse(widget.isCategoryVisible('BuiltIn'))
     widget.setCategoryVisible('BuiltIn', True)
     self.assertTrue(widget.isCategoryVisible('BuiltIn'))
+
+  def test_customDownloader(self):
+    logic = SampleDataLogic()
```
jcfr: Remove `logic`

jcfr:
I address the comments. There is one last issue, the exception raised can not be catched from the test because it happen a signal is triggered ... to workaroud this, we could add an ivar to the test object, associated the test object with CustomDownloader object and and ensure the test object is updated

jcfr:
Closing. Integrated in [r28318](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28318)

@lopezmt  Since I forgot to specify (--add-author-from) when committing, I added  a not the this wiki page to remind us to give you credit when transitioning the repository to GitHub.

PR #1153, opened 2019-06-19
Add SlicerModuleEnablement CMake module, build JsonCpp if VolumeRendering is enabled

jcfr:


lassoan:
Looks good to me, except the word "enablement". I think enablement means something else and very infrequently used in source code anyway. Something like SlicerCheckModuleEnabled or SlicerModuleEnabledCheck would be good (similarly to SlicerCheckCMakeHTTPS.cmake and many other CMake functions).

jcfr:
Thanks for the review, I will update per your suggestion. ( I initially got the idea from [ITKModuleEnablement](https://github.com/InsightSoftwareConsortium/ITK/blob/4da2717e6fd43ca9eb19bc42601238889906502f/CMake/ITKModuleEnablement.cmake))

lassoan:
Functions in ITKModuleEnablement not just query the status of  modules but also make modules enabled/disabled, so "enablement" is correct there (although it is still such an uncommon word that I would avoid using it).

jcfr:
Indeed. Makes sense.

Closing, changes integrated in [r28299](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28299), [r28300](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28300),  and [r28301](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28301),

PR #1156, opened 2019-06-19
BUG: Update slicer.util.reloadScriptedModule to remove layout item after cleanup

jcfr:


jcfr:
cc: @lassoan @pieper

Base/Python/slicer/util.py
```diff
@@ -639,17 +650,17 @@ def reloadScriptedModule(moduleName):
     except AttributeError:
       pass
 
-  # remove spacer items
+  # call cleanup function for the existing widget
+  if hasattr(slicer.modules, widgetName):
+    w = getattr(slicer.modules, widgetName)
+    w.cleanup()
```
lassoan: Is cleanup called again when module is about to be unloaded?
Should we call something like this to avoid double cleanup?

    slicer.app.moduleManager().disconnect('moduleAboutToBeUnloaded(QString)', w._onModuleAboutToBeUnloaded)

jcfr: Good point, the signal should indeed be disconnected.

jcfr:
Closing. Change integrated in [r28315](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28315) to [r28317](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28317)

PR #1157, opened 2019-06-20
BUG: Return failure exit code if exception raised during scripted module cleanup

jcfr:


lassoan:
For testing it is good that we capture errors in module destruction. However, in production code, it could be bad if the return code is overwritten if any module shuts down incorrectly. Maybe we should only do it if Slicer runs in --testing mode?

jcfr:
> Maybe we should only do it if Slicer runs in --testing mode?

Agreed. Thanks for the review.

I will update `qSlicerCoreApplication::onAboutToQuit` to override the return code only if testing mode is enabled and add an other test case to verify the behavior.

jcfr:
Closing. Integrated in [r28319](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28319) and [r28320](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28320)

PR #1160, opened 2019-06-26
BUG: Fix regex matching for SimpleITK Python library fixup

blowekamp:
The SimpleITK Python library now is named
"_SimpleITK.cpython-36m-darwin.so", where the ".cpython-36m-darwin" is
new. The regular expression has been updated so that the library will
be corretly matched and fixed up.

### DIFF
--- a/CMake/SlicerCPackBundleFixup.cmake.in
+++ b/CMake/SlicerCPackBundleFixup.cmake.in
@@ -76,7 +76,7 @@ function(gp_item_default_embedded_path_override item default_embedded_path_var)
 
   set(Slicer_USE_SimpleITK "@Slicer_USE_SimpleITK@")
   if(Slicer_USE_SimpleITK)
-    if (item MATCHES  "site-packages/(SimpleITK.*)/_SimpleITK\\.so" )
+    if (item MATCHES  "site-packages/(SimpleITK.*)/_SimpleITK[^/]*\\.(so|dylib)$" )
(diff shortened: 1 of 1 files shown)


blowekamp:
@pieper @jcfr

jcfr:
Thanks for the fix, makes sense. I am integrating now.

jcfr:
Closing. Fix integrated in [r28347](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28347)

jcfr:
Thanks for your patience :pray:

fedorov:
The problem is still there, looks like the regex failed, since the paths are not fixed (if I understand correctly):

image

jcfr:
We discussed this with @blowekamp and we identify the remaining issue, he will submit an update shortly.

jcfr:
Thanks for the follow up.

Closing. Integrated in [r28348](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28348)

fedorov:
It is still not working, I have a nightly with r28349.

blowekamp:
The 6-30 nightly package appears to be working for me. @fedorov Have you checked out todays build?

jcfr:
Closing. Fernando reported it was now working.
See https://discourse.slicer.org/t/simpleitk-import-error/7294/22?u=jcfr

fedorov:
Yes, it is working in the today's nightly. The problem was observed in the yesterday's nightly.

PR #1161, opened 2019-06-28
BUG: Fix regex matching for SimpleITK Python library fixup

fedorov:
Fixes issue remaining following #1160

cc: @blowekamp


jcfr:
Thanks for the follow up.

Closing. Integrated in [r28348](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28348)

PR #1163, opened 2019-06-30
COMP: Support building with VS2017 allowing to build against TBB built for VS2015

jcfr:
See https://discourse.slicer.org/t/build-failure-windows/7322


jcfr:
Closing. Integrated in [r28352](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28352)

PR #1165, opened 2019-07-01
COMP: Enable OpenSSL install for VS2017 and VS2019

lassoan:
Add OpenSSL binaries for VS2017 and VS2019. The same binaries are used as for VS2015.

This allows building Slicer with OpenSSL enabled, using VS2017 and VS2019 (after tbb libraries are copied to the expected location - see https://github.com/Slicer/Slicer/pull/1163).


lassoan:
Merged http://svn.slicer.org/Slicer4/trunk@28354

PR #1166, opened 2019-07-01
ENH: Use CMake 3.0 default method to find OSX SDK root

blowekamp:
CMake 3.0 uses the `xcode-select -p` to determine the current OSX SDK
root path. This is robust to custom XCode installed paths unlike the custom
Slicer code.

### DIFF
--- a/CMake/SlicerInitializeOSXVariables.cmake
+++ b/CMake/SlicerInitializeOSXVariables.cmake
@@ -32,74 +32,13 @@
 #
 if(APPLE)
 
-  # Waiting universal binaries are supported and tested, complain if
-  # multiple architectures are specified.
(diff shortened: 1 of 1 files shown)


pieper:
Looks like a good improvement to me 👍 

@blowekamp do you ever run into this issue? https://issues.slicer.org/view.php?id=4681

blowekamp:
@pieper This patch helps the referenced issue some: https://github.com/Slicer/Slicer/pull/1167 But you still need to set `SDKROOT` environment variable with running make/ninja.

jcfr:
Closing. Integrated in [r28355](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28355)

PR #1167, opened 2019-07-01
BUG: OSX -isysroot not being used in external non-cmake projects

blowekamp:
CMake 3 replaced OSX specific CMAKE_CXX_HAS_ISYSROOT with the
generalized CMAKE_CXX_SYSROOT_FLAG. The patch restores setting
-isysroot for non-CMake based external projects.


jcfr:
Thanks for the fix.

For reference, see kitware/cmake@a0a0877a1eb111d37ac4766a368206e73458489e

jcfr:
Closing. Integrated in [r28356](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28356)

PR #1170, opened 2019-07-09
Remove harden question from sh

cpinter:
BUG: Remove SH harden transform question if changing transform on single item

When a subject hierarchy item had a transform and the user wanted to replace it with another transform, a message popped up asking if the user wants to harden it and then apply, or simply replace. Since hardening hardly ever seem to be needed, this question was removed.
The question, however, still pops up if the parent item of a branch in which there is at least one transformed item is to be transformed.

BUG: Fix crash when closing application after certain operations in SH

When a volume was cloned and moved in the tree (not sure if both are needed), then Slicer crashed on exit.


lassoan:
Looks good, thank you!

cpinter:
Integrated in
https://github.com/Slicer/Slicer/commit/c00081932b43884b4ff9c32768b023dc792e3afd
and
https://github.com/Slicer/Slicer/commit/3e0ea48e078f2f0b9d501e2ada70d292e96d1de4

Thanks for the reviews!

PR #1171, opened 2019-07-10
ENH: Allow handling events by interaction context

cpinter:
The goal was to fix the issue where the active component is decided based on what's the last device whose event is processed. E.g. a VR controller is in a markup which is supposed to be highlighted, but the other controller is not, and its move event overwrites the active component.

To solve this issue, it was needed to store the interaction context (mouse, VR left, etc.) in the interaction event data, and process accordingly.

### DIFF
--- a/Libs/MRML/DisplayableManager/vtkMRMLInteractionEventData.cxx
+++ b/Libs/MRML/DisplayableManager/vtkMRMLInteractionEventData.cxx
@@ -47,6 +47,7 @@ vtkMRMLInteractionEventData::vtkMRMLInteractionEventData()
   this->Rotation = 0.0;
   this->LastRotation = 0.0;
   this->WorldToPhysicalScale = 1.0;
+  this->InteractionContextName = "";
 }
 
 //---------------------------------------------------------------------------
@@ -287,6 +288,18 @@ vtkCellPicker* vtkMRMLInteractionEventData::GetAccuratePicker() const
 void vtkMRMLInteractionEventData::SetAccuratePicker(vtkCellPicker* picker)
 {
   this->AccuratePicker = picker;
+}
+
+//---------------------------------------------------------------------------
+void vtkMRMLInteractionEventData::SetInteractionContextName(const std::string& contextName)
+{
+  this->InteractionContextName = contextName;
+}
+
+//---------------------------------------------------------------------------
+const std::string& vtkMRMLInteractionEventData::GetInteractionContextName()
+{
+  return this->InteractionContextName;
 }
 
 //---------------------------------------------------------------------------
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.cxx
@@ -16,6 +16,7 @@
 ==============================================================================*/
 
 // MRML includes
+#include "vtkMRMLInteractionEventData.h"
 #include "vtkMRMLMarkupsDisplayNode.h"
 #include <vtkMRMLProceduralColorNode.h>
 
@@ -80,8 +81,8 @@ vtkMRMLMarkupsDisplayNode::vtkMRMLMarkupsDisplayNode()
   this->PropertiesLabelVisibility = true;
   this->PointLabelsVisibility = false;
 
-  this->ActiveComponentType = ComponentNone;
-  this->ActiveComponentIndex = -1;
+  // Set active component defaults for mouse (identified by empty string)
+  this->ActiveComponents[""] = ComponentInfo();
 
   this->LineThickness = 0.2;
 
@@ -295,8 +296,14 @@ void vtkMRMLMarkupsDisplayNode::PrintSelf(ostream& os, vtkIndent indent)
   vtkMRMLPrintBooleanMacro(SliceProjectionOutlinedBehindSlicePlane);
   vtkMRMLPrintVectorMacro(SliceProjectionColor, double, 3);
   vtkMRMLPrintFloatMacro(SliceProjectionOpacity);
-  vtkMRMLPrintFloatMacro(ActiveComponentType);
-  vtkMRMLPrintFloatMacro(ActiveComponentIndex);
+  {
+  os << indent << "ActiveComponents:   ";
+  for (std::map<std::string, ComponentInfo>::iterator it = this->ActiveComponents.begin(); it != this->ActiveComponents.end(); ++it)
+    {
+    os << it->first << ": " << it->second.Type << ", " << it->second.Index;
+    }
+  os << "\n";
+  }
   vtkMRMLPrintFloatMacro(LineThickness);
   vtkMRMLPrintFloatMacro(LineColorFadingStart);
   vtkMRMLPrintFloatMacro(LineColorFadingEnd);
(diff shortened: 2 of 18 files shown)


Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.cxx
```diff
@@ -295,8 +296,14 @@ void vtkMRMLMarkupsDisplayNode::PrintSelf(ostream& os, vtkIndent indent)
   vtkMRMLPrintBooleanMacro(SliceProjectionOutlinedBehindSlicePlane);
   vtkMRMLPrintVectorMacro(SliceProjectionColor, double, 3);
   vtkMRMLPrintFloatMacro(SliceProjectionOpacity);
-  vtkMRMLPrintFloatMacro(ActiveComponentType);
-  vtkMRMLPrintFloatMacro(ActiveComponentIndex);
+  {
```
jcfr: What about adding a `vtkMRMLPrintMapMacro`

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.h
```diff
@@ -66,26 +70,44 @@ class  VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkMRMLMarkupsDisplayNode : public
...
                                    void * /*callData*/ ) override;
 
-  /// Convenienve function for getting the displayable markups node
+  /// Convenience function for getting the displayable markups node
   vtkMRMLMarkupsNode* GetMarkupsNode();
 
-  /// Active item (item that the mouse is hovered over).
-  /// This propoerty is computed on-the-fly and saved to file.
-  vtkGetMacro(ActiveComponentType, int);
+  /// Active item (item that the mouse or other interaction context is hovered over).
```
jcfr:
```
  /// Get active component type
  /// If any, return the activated widget component type for the given context. If no context is provided, "mouse" context is assumed.
  /// A widget component is activated in given context after it is "hovered over".
  /// This property is computed on-the-fly and saved to file.
  int GetActiveComponentType(std::string context="");
```

jcfr: It may be sensible to clarify what the `and saved to file` means

jcfr: component is used in the API, but item is sometimes used in the documentation. I tried to consistently use `component`

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.h
```diff
@@ -66,26 +70,44 @@ class  VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkMRMLMarkupsDisplayNode : public
...
+    ComponentInfo()
+      {
+      this->Type = ComponentNone;
+      this->Index = -1;
+      }
+    int Type;
+    int Index;
+    };
+
+  /// Index of active item (item that the mouse or other interaction context is hovered over).
```
jcfr: -> `Index of active component`

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.h
```diff
@@ -66,26 +70,44 @@ class  VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkMRMLMarkupsDisplayNode : public
...
-  /// Index of active item (item that the mouse is hovered over).
-  /// This propoerty is computed on-the-fly and saved to file.
-  vtkGetMacro(ActiveComponentIndex, int);
+  /// Set active component type and index for interaction context (empty by default, meaning mouse)
+  void SetActiveComponent(int componentType, int componentIndex, std::string context="");
 
-  /// Set active component type and index.
-  void SetActiveComponent(int componentType, int componentIndex);
+  /// Query if there is an active component for any interaction context
+  /// \return List of interaction context names that have active components
```
jcfr: docstring `\return List of interaction context names that have active components` should not be added here. It belongs to the next function

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.h
```diff
@@ -271,9 +297,9 @@ class  VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkMRMLMarkupsDisplayNode : public
   vtkMRMLMarkupsDisplayNode( const vtkMRMLMarkupsDisplayNode& );
   void operator= ( const vtkMRMLMarkupsDisplayNode& );
 
-  // current active point or widget part (hovered by the mouse)
-  int ActiveComponentType;
-  int ActiveComponentIndex;
+  /// Current active point or widget part type and index (hovered by the mouse or other interaction context)
```
jcfr: `Current active point or widget part type` -> `Current active point or widget component type`

jcfr:
Well done, this will really help for working with VR and 3D view at the same time :+1:

cpinter:
Thanks for the review, @jcfr! I'll fix them asap.

Libs/MRML/DisplayableManager/vtkMRMLInteractionEventData.cxx
```diff
@@ -289,6 +290,18 @@ void vtkMRMLInteractionEventData::SetAccuratePicker(vtkCellPicker* picker)
   this->AccuratePicker = picker;
 }
 
+//---------------------------------------------------------------------------
+void vtkMRMLInteractionEventData::SetInteractionContextName(const std::string& v)
```
jcfr: try to avoid single letter variable. It makes it hard to search through the code.

`v` -> `contextName`

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.cxx
```diff
@@ -80,8 +81,8 @@ vtkMRMLMarkupsDisplayNode::vtkMRMLMarkupsDisplayNode()
   this->PropertiesLabelVisibility = true;
   this->PointLabelsVisibility = false;
 
-  this->ActiveComponentType = ComponentNone;
-  this->ActiveComponentIndex = -1;
+  // Set active component defaults for mouse (identified by empty string)
+  this->ActiveComponents[""] = ComponentInfo();
```
jcfr: What do you think of introducing a static variable called `DefaultContextName` ?

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.cxx
cpinter: I could create one, but it wouldn't be usable in this case due to the custom structure it has as values
(os << it->first << ": " << it->second.Type << ", " << it->second.Index;)

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.h
cpinter: I'll fix the mentions of "item".

> It may be sensible to clarify what the and saved to file means
Actually I'm not sure what it means. Need to ask @lassoan

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.h
cpinter: Good catch

Libs/MRML/DisplayableManager/vtkMRMLInteractionEventData.cxx
cpinter: I was just following the conventions in the file (there are ~10 functions with v as parameter name - I think for "value"). I'll change it to contextName

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.cxx
jcfr: Good point, it is not worth the indirection then.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.cxx
cpinter: Python wrapping fails when I want to use a const static string variable as a default value for a function parameter (like int GetActiveControlPointForContext(std::string context=DEFAULT_CONTEXT_NAME);)
Any advice?

jcfr: @dgobbi  Out of curiosity, do you know if this is expected to work in the latest VTK ? If yes, then we will backport the corresponding change into the Slicer/VTK fork  (we are still stuck with VTK of before the major build system overhaul)

dgobbi: I just tested with VTK master and it seems to work as expected, but I don't remember any specific change that enabled it.  What is the error?

cpinter: The error is `vtkMRMLMarkupsDisplayNodePython.cxx(451): error C2065: 'DEFAULT_CONTEXT_NAME': undeclared identifier`
It is a static const std::string defined in the header. The problem is that in the Python cxx there is no such declaration and when it's used in the function headers it is undefined.

dgobbi: Show me the file where `DEFAULT_CONTEXT_NAME` is declared and I'll figure out why it isn't visible in this context, the fact that the type is string is probably a red herring.  The wrappers are probably just not properly scoping a constant.

cpinter: Thank you, @dgobbi !
Here are the files: https://gist.github.com/cpinter/ba47dd9075a5afa5aae012b08c31e927

cpinter:
I integrated the change (without the constant) in https://github.com/Slicer/Slicer/commit/b258ba0f5b36d867ffafe052e10d2fdf2ae6bbba in the spirit of being able to move forward with the corresponding SlicerVR changes and the following Slicer core work.

I'm not closing the PR until we figured out the details about the file saving and the python wrapping of the constant.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.cxx
dgobbi: The declaration `static const std::string DEFAULT_CONTEXT_NAME;` is not wrapped since the wrappers only handle POD constants (no objects).  But I can modify the wrappers to handle the default value in cases like this.

jcfr: @dgobbi Thanks for looking into this :pray:  supporting this would be neat.

@cpinter In the mean time, we could also add a static function `GetDefaultContextName()` like what is done in [vtkMRMLSubjectHierarchyConstants](https://github.com/Slicer/Slicer/blob/master/Libs/MRML/Core/vtkMRMLSubjectHierarchyConstants.h)

cpinter: Added in https://github.com/Slicer/Slicer/commit/110e94795d768d772f6ff10536b1521e7a50e63f

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.h
cpinter: https://github.com/Slicer/Slicer/commit/ae9ae617f00b04fe0c19334ce1ce3dcc4453f07d#r34289815

cpinter:
Any objections against closing this PR?

lassoan:
No objections from me.

PR #1174, opened 2019-07-11
Ensure CLI libraries are fixed up on macOS

jcfr:
See https://discourse.slicer.org/t/slicer-tractography-roi-selection-not-working/7537


jcfr:
Closing. Integrated in [r28380](28380) and [r28381](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28381)

PR #1176, opened 2019-07-18
BUG: Fix MRMLNodeComboBox_Segmentation is not found in Segment Editor module

Sunderlandkyl:
When using the "Edit selected" button in Segmentations to switch to the Segment Editor, the Segmentations module widget attempts to find the node selector, however the name is incorrect.
Name was: MRMLNodeComboBox_Segmentation
Should be: SegmentationNodeComboBox

Fixed by looking for the correct name in Segmentations.


cpinter:
LGTM! Thanks

lassoan:
Thank you!
Committed r28396.

PR #1177, opened 2019-07-19
ENH: Update qMRMLSegmentsTableView to use model/view architecture

Sunderlandkyl:
Previously, qMRMLSegmentsTableView utilized a QTableWidget internally
that was populated statically whenever the segments were changed. Now
the internal table is a QTableView controlled by two new models:
qMRMLSortFIlterSegmentsProxyModel and qMRMLSegmentsModel.

qMRMLSegmentsModel populates the model with items based on the contents
of the segments contained within the segmentation node.

qMRMLSortFIlterSegmentsProxyModel provide the ability to filter the
list of segments based on multiple values. The provided filters are:
* NameFilter - Only shows segments whose name contain the filter string
* TextFilter - Only shows segments that match the filter string either
               in the segment name, or in the value of one of the tags
* Various status filters - Multiple boolean filters that will show
                           any segments that match one of the status

Added status tags for each segment:
* Not started
* In Progress
* Completed
* Flagged

When a segment is modified by a qSlicerSegmentEditorAbstractEffect,
and the state of the segment is "Not started", it is set to "In
progress". The segment state can also be changed by clicking on the
segment table in the "Status" column. This will iterate through
"In Progress", "Completed", and "Flagged" states. To return to the
"Not Started" state, the user can select the "Clear selected segment"
context menu option in the segment table to erase the selected segment
contents.


Sunderlandkyl:
The updated segment table looks like this:
image

The column at the far right represents the current segment status.

lassoan:
Looks nice! What is the difference between the "Edit" and "Flag" states?

Sunderlandkyl:
I think the purpose of the flag state is somewhat like a "Flag for review" indicator.
There is currently no difference between the states from a logical standpoint.

cpinter:
Looks nice! Can you make the buttons next to the filter line edit smaller? So that they have the same size icons as the table, and hopefully save some space (by having the same hight as the line edit) for the already very crowded segment editor widget. Thanks!

Sunderlandkyl:
Buttons should now be the same size as the filter line edit, and icons are now set to 16x16 to match the size in the table.
image

lassoan:
I've tested yes and it works great!

It would be nice if you could find higher resolution icons, especially for the flag. It would be also better to have an edit icon that is not the same as the editor module icon. Try to find an icon set that has flag, checkmark, edit, and some empty frame or small circle.

It would be useful to be able to show/hide filters row. Maybe it could be a checkbox in the context menu and/or in the Segmentations... button drop-down menu.

I would also consider only advancing the state of the icon is left-clicked (do not jump from completed to start state) to avoid the chance that you accidently throw back a completed or flagged segment to the crowd of unstarted segments. If possible, make right-click set previous state, or maybe even better, bring up a context menu that shows all states so that you can directly set any state.

lassoan:
@Sunderlandkyl I've accidentally merged this as is (and I don't revert it, because it is already very good), but please consider my comments in a new pull request:

- It would be nice if you could find higher resolution icons, especially for the flag. It would be also better to have an edit icon that is not the same as the editor module icon to avoid confusion. Maybe try to find an icon set that has flag, checkmark, edit, and some empty frame or small circle.
- It would be useful to offer the possibility to users to show/hide filters row. Maybe it could be a checkbox in the context menu and/or in the "Segmentations..." button's drop-down menu.
- Add option directly set any state using right-click menu on the "Segment status" column.

lassoan:
Integrated in r28398 and r28400.

PR #1178, opened 2019-07-22
ENH: Improve options for segment status in qMRMLSegmentsTableView

Sunderlandkyl:
* Changed status icons to icons from material.io
* Add option to change segment status from context menu
* Left clicking on flagged status now circles back to completed
* Add option to show/hide filter bar

image


Sunderlandkyl:
Setting the state via context menu currently only works on the segment that was right-clicked.

Would it make more sense to be able to set multiple segment statuses at the same time if multiple are selected?

Sunderlandkyl:
Ok, the status options now set the status on all selected segments.
Also, context menu options that require a selected segment no longer appear in the context menu if there are no segments selected.

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentsTableView.h
```diff
@@ -131,6 +131,9 @@ public slots:
   /// Segment's center is determined as the center of bounding box.
   void jumpSlices();
 
+  /// TODO
```
lassoan: Why is this a TODO?

Sunderlandkyl: Because I was supposed to fill in the description before making the pull request. I'll fix that now.

Sunderlandkyl:
@lassoan The crash is fixed + function documentation has been added

lassoan:
Thank you very much! Merged https://github.com/Slicer/Slicer/commit/6a6e49ea52eaff7a9faf86f8486db40762bbb7bc

PR #1179, opened 2019-07-22
Update to itk5

pieper:
@jcfr for this I referenced upstream repos and not our Slicer org forks.  Were there any patches still outstanding that aren't in the upstream?  If you agree, I think we should change the Welcome branches of to say that they are for legacy use (e.g. for 4.10) but won't be used in the future unless specific needs arise.

p.s. there's no particular rush, but [this extension](https://github.com/SNRLab/LineMarkerRegistration/pull/3) needs a fix from itk5 to build on mac.

### DIFF
--- a/SuperBuild.cmake
+++ b/SuperBuild.cmake
@@ -244,8 +244,8 @@ set(BRAINSTools_options
   USE_BRAINSFit:BOOL=ON
   USE_BRAINSROIAuto:BOOL=ON
   USE_BRAINSResample:BOOL=ON
-  USE_BRAINSDemonWarp:BOOL=ON
   # BRAINSTools comes with some extra tool that should not be compiled by default
+  USE_BRAINSDemonWarp:BOOL=OFF
   USE_AutoWorkup:BOOL=OFF
   USE_ReferenceAtlas:BOOL=OFF
   USE_ANTS:BOOL=OFF
@@ -271,7 +271,6 @@ set(BRAINSTools_options
   BRAINS_DEBUG_IMAGE_WRITE:BOOL=OFF
   USE_BRAINSTransformConvert:BOOL=ON
   USE_DWIConvert:BOOL=${Slicer_BUILD_DICOM_SUPPORT} ## Need to figure out library linking
-  USE_BRAINSDemonWarp:BOOL=ON
   USE_BRAINSRefacer:BOOL=OFF
   )
 
--- a/SuperBuild.cmake
+++ b/SuperBuild.cmake
@@ -275,8 +275,8 @@ set(BRAINSTools_options
   )
 
 Slicer_Remote_Add(BRAINSTools
-  GIT_REPOSITORY ${EP_GIT_PROTOCOL}://github.com/Slicer/BRAINSTools.git
-  GIT_TAG 53c15d6beac5b8d65689054da89deb69e61c7d32 # slicer-2019-03-07-v5.0.0-2af1e31
+  GIT_REPOSITORY ${EP_GIT_PROTOCOL}://github.com/BRAINSIA/BRAINSTools.git
+  GIT_TAG b75f58ada5456e75f068ecdc640d5da30c7703de # master corresponding to ITKv5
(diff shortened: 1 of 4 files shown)


SuperBuild/External_ITK.cmake
```diff
@@ -27,13 +27,15 @@ if(NOT DEFINED ITK_DIR AND NOT Slicer_USE_SYSTEM_${proj})
 
   ExternalProject_SetIfNotDefined(
     Slicer_${proj}_GIT_REPOSITORY
-    "${EP_GIT_PROTOCOL}://github.com/Slicer/ITK"
+    # "${EP_GIT_PROTOCOL}://github.com/Slicer/ITK"
+    "${EP_GIT_PROTOCOL}://github.com/Kitware/ITK"
```
dzenanz: https://github.com/InsightSoftwareConsortium/ITK is the source of truth. `github.com/Kitware/ITK` is one of the mirrors.

pieper: Makes sense, yes, thanks.

@dzenanz do you know if Slicer/ITK had any remaining patches compared to ITKv5?

dzenanz: I don't think so. @phcerdan, @sjh26 or @jcfr might know that with more certainty.

jcfr:
There is one patch. See https://github.com/Slicer/ITK/commits/slicer-v5.0.rc01-2019-01-30-eb54c22

SuperBuild.cmake
```diff
@@ -244,8 +244,8 @@ set(BRAINSTools_options
   USE_BRAINSFit:BOOL=ON
   USE_BRAINSROIAuto:BOOL=ON
   USE_BRAINSResample:BOOL=ON
-  USE_BRAINSDemonWarp:BOOL=ON
   # BRAINSTools comes with some extra tool that should not be compiled by default
+  USE_BRAINSDemonWarp:BOOL=OFF
```
jcfr: Does this mean we should set `USE_ANTS` to `ON` ?

pieper: Good question - @hjmjohnson should we enable ANTS in Slicer's BRAINSTools build as a replacement for DemonWarp?

hjmjohnson: Probably not.  That is a large burden for a very advanced feature that I don't think many people will appreciate.

BRAINSFit has conditional code that if ANTs exists, then one hard-coded non-linear warping method becomes available, but it is hard-coded with settings for MRI brains at 1mm voxels, and it does not have exposed parameters.  

BRAINSDemonsWarp is a mediocre non-linear warping tool that I would rather not expend energy maintaining, but it is easy and SEM compliant.  It should not take to much effort to get it compiling again.

ANTs is not SEM compliant, but it is far superior for non-linear warping.

lassoan: We could add an extension for ANTs (with a Python CLI that calls ANTs binaries) if people would like to use it in Slicer.

pieper:
Quite a few extensions are broken when building against ITKv5 release (my linux build is marked "(empty)", but many breakages are also present in the current SlicerPreview built against Slicer/ITK and the errors appear very similar, at least at a glance but there are definitely a few issues related to SpatialObjects and other issues that changed between the rc and the release.  

http://slicer.cdash.org/index.php?project=SlicerPreview&date=2019-08-05

I think it will be easier to encourage extension developers to fix their builds against the ITK release rather than the release candidate.  So I suggest we merge this pull request.

@jcfr is the one commit you added to Slicer/ITK likely to be merged into an ITK release?  If yes I suggest we switch to that release and use the upstream repo.  If no then I suggest we update Slicer/ITK to v5 release and re-add our patch on top.

dzenanz:
https://github.com/Slicer/ITK/commit/4f7f4bf295e7a34a7bac8474ca2fc26bccc53c12 is not likely to be integrated into ITK. Why do you want it to be able to build as an external module? @jcfr Why not just enable it for building when initially configuring ITK?

jcfr:
Closing. Integrated in [r28440](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28440)

PR #1180, opened 2019-07-23
ENH: Improve loading speed when importing segmentation from labelmap

Sunderlandkyl:
Before this change, SetTerminologyToSegmentationFromLabelmapNode would call vtkSegment::SetTag, invoking a vtkSegment::SegmentModified event on the Segmentation node for each segment.
This event was not handled with a StartModify/EndModify, so it caused a major slowdown.

Improvements:
* Added new class "MRMLNodeModify" to vtkMRMLNode.h that calls StartModify on construction and EndModify on destruction
* Added MRMLNodeModify(segmentationNode) to vtkSlicerSegmentationsModuleLogic to prevent unnecessary SegmentModified events

### DIFF
--- a/Libs/MRML/Core/vtkMRMLNode.h
+++ b/Libs/MRML/Core/vtkMRMLNode.h
@@ -953,4 +953,27 @@ private:
   std::map<int, int> CustomModifiedEventPending; // event id, pending value (number of events grouped together)
 };
 
+/// \brief Simple class that calls StartModify on instantiation and EndModify on destruction.
+class VTK_MRML_EXPORT MRMLNodeModify
+{
+public:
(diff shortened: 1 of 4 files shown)


Sunderlandkyl:
Importing the brain atlas (https://slicer.kitware.com/midas3/item/121588) from labelmap previously took ~2:00. With this change, it takes ~0:10.

lassoan:
Great, this speed improvement is very impressive!

The modification blocker is very useful, too, and it is very nicely implemented (works well for null-pointer input, uses weak pointer to be more robust, etc). I added "blocker" to its name to follow Qt signal blocker naming convention (I did not find similar blocker in VTK).

lassoan:
Merged in r28402.

PR #1181, opened 2019-07-24
ENH: Change qMRMLSegmentsTableView filter field to ctkSearchBox

Sunderlandkyl:
ctkSearchBox provides additional features on top of QLine edit, such as a button to clear the search field.

image

### DIFF
--- a/Modules/Loadable/Segmentations/Widgets/Resources/UI/qMRMLSegmentsTableView.ui
+++ b/Modules/Loadable/Segmentations/Widgets/Resources/UI/qMRMLSegmentsTableView.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>224</width>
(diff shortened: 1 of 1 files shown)


Sunderlandkyl:
I left the placeholder text empty, and hid the icon. Should either of those things to be enabled?

lassoan:
Adding "Filter" or "Search" as text would be a useful hint for users.

If we have space (and I think we do) then we can keep the icon, it make things a bit nicer.

See for example in Excel:

image

Sunderlandkyl:
Changes made.
Dropped the "Filter:" label as it was redundant with the placeholder text.

New appearance:
image

lassoan:
Looks good, thank you! If there are no other comments then I'll merge this later today.

cpinter:
This is great, thank you!

lassoan:
Committed r28403.
Thank you!

PR #1182, opened 2019-07-26
Update qSlicerScalarVolumeDisplayWidget.cxx

rprueckl:
Corrected setting the state of LockWindowLevelButton according to the correct property of the displayNode. The bug had no consequences, so just a formality.


lassoan:
Merged in https://github.com/Slicer/Slicer/commit/d0cbc4fded5c2536a590b2f7672601d11480c7d6

PR #1183, opened 2019-07-26
BUG: Fix segment previous/next shortcut and move segment up/down

Sunderlandkyl:
* The prev/next shortcuts in the segment editor now move the table selection up and down the list of visible segments, skipping all filtered segments
* The move segment up/down will move segments up/down, skipping all filtered segments

Filtering parameters are now stored in two new vtkMRMLSegmentationNode attributes:
* SegmentListFilterEnabled - Determines if the filter bar should be shown or not
* SegmentListFilterOptions - Serialized string containing the parameters used to filter the segments (text, status, etc)

Other changes:
* Cleaned up handling of segment status, made it more expandable in the future
* Moved ClearSegment methods to vtkSlicerSegmentationsModuleLogic to allow access in Python
* New segments will now set their id based on the text filter. This allows prevents created segments from being hidden by the filter (only applies to text filtering, not status filtering).

### DIFF
--- a/Libs/MRML/Core/vtkMRMLSegmentationNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLSegmentationNode.cxx
@@ -71,6 +71,8 @@ vtkMRMLSegmentationNode::vtkMRMLSegmentationNode()
   this->SegmentCenterTmp[2] = 0.0;
   this->SegmentCenterTmp[3] = 1.0;
 
+  this->SegmentListFilterEnabled = false;
+
   // Create empty segmentations object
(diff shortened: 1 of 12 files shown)


lassoan:
Awesome, thank you.

Could you make new segments appear when some state filter hides "not started" segments? (by setting its state to the first visible state)

Sunderlandkyl:
Ok, changes made.

lassoan:
Thank you very much, merged in https://github.com/Slicer/Slicer/commit/d4e9638187d03a3a5a96959750c559db76c73236

PR #1185, opened 2019-07-30
ENH: Rename overwrite-related masking options in Segment Editor

cpinter:
Instead of
  Overwrite other segments: All segments / Visible segments / None
now it says
  Modify other segments: Overwrite all / Overwrite visible / Overlap
The intention is to reduce confusion and make the consequence of the selection clearer.

### DIFF
--- a/Modules/Loadable/Segmentations/Widgets/Resources/UI/qMRMLSegmentEditorWidget.ui
+++ b/Modules/Loadable/Segmentations/Widgets/Resources/UI/qMRMLSegmentEditorWidget.ui
@@ -454,7 +454,7 @@
          <string/>
         </property>
         <property name="text">
-         <string>Overwrite other segments:</string>
+         <string>Modify other segments:</string>
         </property>
        </widget>
       </item>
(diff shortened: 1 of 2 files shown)


cpinter:
image

lassoan:
Thank you, looks good to me. This will be much more clear to users (and to us, too).

Would "Allow overlap" be more clear than just "Overlap"?

cpinter:
Well, maybe. But this is what we settled on when we discussed it with you and Kyle, so I'd wait for some more opinions before I make further changes. Maybe the final consensus will be different.
@jcfr @pieper @fedorov do you have anything to add?

pieper:
I guess I'd have to vote for 'Allow overlap' here.

fedorov:
I don't think I ever used this functionality. I would need to look in more detail. If @pieper and @jcfr have strong feelings one way or another, I will defer to them.

cpinter:
OK, Allow overlap it is! I'll change it and integrate it now unless there are further requests or objections.

cpinter:
Integrated in https://github.com/Slicer/Slicer/commit/b0a932cc4b3a9d674c64f37a13a3b47144274b69
Thanks for the feedback!

PR #1186, opened 2019-07-30
ENH: Optimize UI of Models and Markups modules

cpinter:
In the current Segmentations module, there is a visibility and opacity setting in the same row. This commit makes the same changes in the Models and Markups modules.
In addition, in the Markups module, the Display section is now open by default, with many of its settings moved to the Advanced section. This way the essential display options are not hidden (having a collapsed section on the very top was a bit confusing), but they do not take up too much space either.

### DIFF
--- a/Modules/Loadable/Markups/Resources/UI/qSlicerMarkupsModule.ui
+++ b/Modules/Loadable/Markups/Resources/UI/qSlicerMarkupsModule.ui
@@ -126,7 +126,7 @@
       <string>Display</string>
      </property>
      <property name="collapsed">
-      <bool>true</bool>
+      <bool>false</bool>
      </property>
      <layout class="QVBoxLayout" name="verticalLayout_3">
       <item>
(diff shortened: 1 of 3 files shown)


cpinter:
Models:
image
Markups:
image

lassoan:
Thanks for working on this! What layout do you use in the display section? It looks unusual that left side of widgets (checkboxes, sliders, etc.) are not aligned. I think by default we use form layout, which provides this alignment.

cpinter:
It's all grid layout, but the "Visibility: ..." row is in its own HBox layout so that all four widgets can be there. I could have the Visibility label in the grid layout and the other three in a HBox in the second column if that helps.

lassoan:
> I could have the Visibility label in the grid layout and the other three in a HBox in the second column if that helps.

Yes, I think that would look more consistent.

cpinter:
image

image

lassoan:
Thank you, it looks good!

cpinter:
Since there was no further comment or objection, I integrated this in https://github.com/Slicer/Slicer/commit/778d7fdec3cd92feae0a9354e31a7f42799fd32d

Thanks!

PR #1187, opened 2019-07-30
BUG: Fix segment selection change during undo

Sunderlandkyl:
Calling takeRow() and insertRow() caused an intermediate state where
the selected segment was removed from the table and caused the
selection in the table to change.

To fix this the order is only updated when it is required, and signals
from the model are now blocked while the order of the segments is being
updated, and the reorderItems method now emits layoutAboutToBeChanged
and layoutChanged to ensure that the table is up to date.


Sunderlandkyl:
I tried to update the indices directly using QAbstractItemModel::changePersistentIndexList(), but I wasn't able to get it to function correctly. If someone knows a better method for rearranging the items in a model, then feel free to let me know.

The other alternative that I tested was to use the qMRMLSortFilterSegmentsProxyModel to display the items in the correct order.

lassoan:
Thanks a lot, this works well now. Committed r28417.

PR #1189, opened 2019-08-01
BUG: Fix show only selected segments option in segment table

Sunderlandkyl:
Show only selected segments was previous applied to only the displayed segments in the table.
Fixed by applying the desired visibility to the total list of segments, excluding those that are specifically hidden in the table (through setHideSegments()).
Updated SegmentationWidgetsTest1 to test showOnlySelectedSegments().


lassoan:
Thank you, merged.

PR #1190, opened 2019-08-01
ENH: Improve segment table filtering performance

Sunderlandkyl:
Previously, each time the filter parameters were changed, the
segmentation node would be modified, triggering other expensive
operations (rendering, etc).

Now changing filter parameters will restart a 500ms one-shot timer that
will cause the node to only be updated once it has elapsed. Continuous
updates such as typing will not be saved until there is a 500ms gap in
the changes.


lassoan:
Thank you, merged.

PR #1192, opened 2019-08-06
COMP: Fix implicit conversion to string for CMAKE_CXX_MP_NUM_PROCESSORS

jamesobutler:
This PR fixes the CMake warning showing up during configuration (see [here](http://slicer.cdash.org/viewConfigure.php?buildid=1663462)). 
```log
CMake Warning (dev) at CMake/SlicerBlockCXXRequiredFlags.cmake:68 (set):
  implicitly converting '32' to 'STRING' type.
Call Stack (most recent call first):
  CMakeLists.txt:418 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

This also fixes the issue where MP_NUM_PROCESSORS wasn't actually getting set and it was being left empty.


lassoan:
Thank you very much!
Committed r28427.

PR #1196, opened 2019-08-13
COMP: Package JsonCpp if VolumeRendering module is enabled

jcfr:
Following r28285 (ENH: Add new node types to support custom
shader code for volume rendering), JsonCpp is a requirement
for VolumeRendering module.

This commit is a follow-up of r28301 (COMP: Build JsonCpp if
VolumeRendering module is enabled)

Reported-by: Csaba Pinter <csaba.pinter@queensu.ca>


jcfr:
Thanks for testing. 

Closing. Integrated in [r28437](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28437)

PR #1197, opened 2019-08-13
ENH: Adding SWIG executable for use in downstream projects (and/or ex…

adamrankin:
…tensions)

### DIFF
--- a/SuperBuild/External_Swig.cmake
+++ b/SuperBuild/External_Swig.cmake
@@ -103,4 +103,9 @@ ExternalProject_Execute(${proj} \"configure\" sh ${EP_SOURCE_DIR}/configure
   ExternalProject_GenerateProjectDescription_Step(${proj}
     VERSION ${SWIG_TARGET_VERSION}
     )
+
+  mark_as_superbuild(
+    VARS SWIG_EXECUTABLE:FILEPATH
+    LABELS "FIND_PACKAGE"
+    )
 endif()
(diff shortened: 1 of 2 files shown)


adamrankin:
@Slicer/slicer-core Is anyone available to review?

SuperBuild/External_Swig.cmake
```diff
@@ -68,6 +68,8 @@ if(NOT SWIG_DIR AND NOT Slicer_USE_SYSTEM_${proj})
 "#------------------------------------------------------------------------------
 # Added by '${CMAKE_CURRENT_LIST_FILE}'
 
+mark_as_superbuild(SWIG_EXECUTABLE)
```
jcfr: This shouldn't be added to the `_env_script`, it should instead be added at the end of the file. See [here](https://github.com/Slicer/Slicer/blob/1b7888e90e233729e4beb4941e8b33e622373655/SuperBuild/External_VTK.cmake#L260) for an example

jcfr: This should probably be `Slicer_SWIG_EXECUTABLE`

adamrankin: The variable Slicer_SWIG_EXECUTABLE doens't exist at this point, should I
```cmake
set(Slicer_SWIG_EXECUTABLE ${SWIG_EXECUTABLE})
```
?

jcfr:
@adamrankin I updated the topic, could you quickly try if it works for you ?

As a side note, Swig is currently built only if SimpleITK is enabled. Considering we will most likely move SimpleITK as an extension in the next few months (see [roadmap](https://www.slicer.org/wiki/Documentation/Labs/Slicer5-roadmap#Overall_Goals)), we may have to revisit the approach to ensure your extension can build.

adamrankin:
Yup, new approach works. When SimpleITK gets taken out, I'll copy `external_swig.cmake`

jcfr:
Closing. Fixed in [r28445](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28445)

PR #1200, opened 2019-08-20
BUG: Fix issues with DICOM browser widget

Sunderlandkyl:
Fix qSlicerDICOMModuleGenericTest failure
* qSlicerDICOMModuleGenericTest is run without the main window, so layout manager does not not exist.
* Now checks to make sure the layout manager exists before creating the connection

Fix issue where data is loaded and the widget closes, yet the layout remains the same
* See https://issues.slicer.org/view.php?id=4705
* Certain datasets invoke start/end batch processing
* This caused an unnecessary update of the previous/current layout within the DICOM module
* Solved by only changing the previous/current layout if it is different from the current layout

Allow the subject hierarchy tree to expand within the module layout
* Remove the extra spacer from the module layout, allowing the tree widget to expand

### DIFF
--- a/Modules/Scripted/DICOM/DICOM.py
+++ b/Modules/Scripted/DICOM/DICOM.py
@@ -176,6 +176,9 @@ This work is supported by NA-MIC, NAC, BIRN, NCIGT, and the Slicer Community. Se
 
 
   def onLayoutChanged(self, viewArrangement):
+    if viewArrangement == self.currentViewArrangement:
+      return
+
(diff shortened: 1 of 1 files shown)


cpinter:
Looks good, thank you!

jcfr:
Thanks for the fix :+1: 

Closing. Integrated in [r28451](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28451)

PR #1204, opened 2019-09-03
ENH: Improve qMRMLSegmentsModel handling of SegmentAdded/SegmentRemoved

Sunderlandkyl:
Previously, adding/removing a segment within a Start/EndModify would cause the selected segments to become deselected in the table view.
Now, if a SegmentAdded or SegmentRemoved event is received and the callData is empty, existing rows will be compared against the list of SegmentIDs in the segmentation.
* During onSegmentAdded, missing segments will be added to the table
* During onSegmentRemoved, segments that are not in the segmentation will be removed from the table


lassoan:
Thanks a lot, it works great. Merged in http://svn.slicer.org/Slicer4/trunk@28479.

PR #1205, opened 2019-09-05
ENH: Close DICOM browser when switching modules

Sunderlandkyl:
With the previous behavior, the DICOM browser will remain open when switching modules as long as "Browser persistent" is enabled.
This commit causes the DICOM browser to be closed when switching modules, regardless of the "Browser persistent" setting.

"Browser persistent" only affects whether or not the DICOM browser is closed after loading data from the database.


Sunderlandkyl:
@lassoan

lassoan:
Thank you! Committed r28486.

PR #1206, opened 2019-09-10
Update vtkSlicerAnnotationModuleLogic.cxx

rprueckl:
BUG: Fix calls to SetHierarchyAnnotationsVisibleFlag and SetHierarchyAnnotationsLockFlag that caused memory leaks.


lassoan:
Good catch, thank you. Could you please replace  `vtkCollection* children = vtkCollection::New();` by ` vtkNew<vtkCollection> children;` instead of adding manual deletion?

lassoan:
@cpinter I don't merge this to avoid merge conflicts, but make sure this error is not present in your subject hierarchy visibility rework.

lassoan:
Just checked this now and realized that this fix did not make it in to the master. Merged it now (https://github.com/Slicer/Slicer/commit/31d4ac17f1e2bc769e388985a5e29a17e0e8d65f).

PR #1207, opened 2019-09-10
BUG: Remove creation of DICOM browser widget before module is opened

Sunderlandkyl:
Previously the details popup was incorrectly created when the module was loaded, and again when it was opened.
Early creation of the was unnecessary and caused issues with the automated tests.

This commit removes the instantiation of the DICOM popup widget until the DICOM module is opened.


Sunderlandkyl:
@lassoan

pieper:
Sounds like a good fix - hope it clears up the dashboard issues. 👍

lassoan:
This is great, thank you, merged (r28492).

cpinter:
Glad it's this simple :) Yes this way there are no update related popups with any settings until opening the browser explicitly

jcfr:
:tada:  Thanks for working on this

PR #1209, opened 2019-09-11
ENH: Add X-ray volume rendering preset

lassoan:
Added non-shaded X-ray-like volume rendering preset. User has to set view background color manually.

image


bpaniagua:
This is so pretty.

PR #1211, opened 2019-09-12
BUG: Fix failure in SegmentationWidgetsTest1

Sunderlandkyl:
Changes in segment filtering parameters are written to the segmentation node after a short delay in order to prevent unnecessary rendering while typing in the filter bar.
This caused the section 4 of the test to fail because the table widget was deleted before the delay had elapsed and the changes to the segmentation filter could be written to the node.

This commit fixes the test by specifically setting the segmentation node filter parameters both once the test has completed (to avoid interfering with other test sections), and at the start of section 4 (to ensure that changes in the filtering parameters from other tests do not cause a failure in this part of the test again)


cpinter:
Thank you, @Sunderlandkyl ! I integrated this in https://github.com/Slicer/Slicer/commit/3fcf9e90ee21d5ea489a9ec59b9ff1f4956b1a3d

jcfr:
Thanks @Sunderlandkyl for taking care of the test :pray:

PR #1212, opened 2019-09-17
BUG: Fix traceback in SegmentEditorHollowEffect medial surface

Sunderlandkyl:
During the migration to Python 3 (a7f519bd884a2e30cc340d0abf24a9a7013b5b0f), kernelSizePixel[x]/2 was changed to int(kernelSizePixel[x],2).
This commit changes it to the call to: int(kernelSizePixel[x]/2, which prevents the error and matches the original implementation.


lassoan:
Thank you, indeed, somehow we missed this. Merged.

PR #1210, opened 2019-09-12
ENH: Refactoring the MRML scene to contain a SH node as a member vari…

adamrankin:
Moving consistency logic from vtkMRMLSubjectHierarchyNode::GetSubjectHierarchyNode to a new function vtkMRMLSubjectHierarchyNode::ResolveSubjectHierarchy and calling from end of scene import, node added, and SH plugin handler

Subject hierarchy plugin handler no longer cache's the SH node, but requests it from the scene instead

PR #1213, opened 2019-09-17
BUG: Restore the observation of the plugin handler

adamrankin:
Incorrectly removed the plugin handler's observation on the subject hierarchy node, restoring

See #1210

### DIFF
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginHandler.cxx
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginHandler.cxx
@@ -115,6 +115,11 @@ qSlicerSubjectHierarchyPluginHandler::qSlicerSubjectHierarchyPluginHandler(QObje
 //-----------------------------------------------------------------------------
 qSlicerSubjectHierarchyPluginHandler::~qSlicerSubjectHierarchyPluginHandler()
 {
+  if (m_MRMLScene != nullptr)
+    {
+    m_MRMLScene->RemoveObserver(m_CallBack);
+    }
+
   QList<qSlicerSubjectHierarchyAbstractPlugin*>::iterator pluginIt;
   for (pluginIt = this->m_RegisteredPlugins.begin(); pluginIt != this->m_RegisteredPlugins.end(); ++pluginIt)
     {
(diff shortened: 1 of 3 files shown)


adamrankin:
@cpinter Plugin's are now correctly found and used for SH items

cpinter:
Perfect, thank you so much for the quick fix!!

adamrankin:
Well, please check it over because there may be a cleaner way of doing things. I just restored the old mechanism.

cpinter:
Integrated in https://github.com/Slicer/Slicer/commit/073a38a0b8e5d0c855eb93b00eec3cd02d8ecf44

@adamrankin I had to amend the commit so the first line of the message fits the character limit. If you run c:\d\Slicer\Utilities\SetupForDevelopment.sh from the Git bash, then git will check the requirements when committing.

cpinter:
I think it's good. I don't have a better idea than what the original mechanism was for the observations.

Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginHandler.cxx
```diff
@@ -379,6 +384,23 @@ qSlicerSubjectHierarchyAbstractPlugin* qSlicerSubjectHierarchyPluginHandler::sel
   return candidatePlugins[0];
 }
 
+//----------------------------------------------------------------------------
+void qSlicerSubjectHierarchyPluginHandler::observeSubjectHierarchyNode(vtkMRMLSubjectHierarchyNode* shNode)
```
jcfr: I suggest to rename the function to `AddSubjectHierarchyNodeObservers`

adamrankin: Happy too. @cpinter ?

cpinter: I don't mind. But please have it start with lower case, because it's a Qt class.
Btw my opinion is that since this is integrated, creating another PR would be needed, which I think is unreasonable for such a change. The name of the function is not misleading, and it does not really do against conventions, because again, this is not a VTK class.

jcfr: > because again, this is not a VTK class.

Sorry for the noise, you are right there is no need to change it.

cpinter: Thanks, @jcfr no problem

PR #1217, opened 2019-10-01
COMP: Introducing conditional searching of CTKAppLauncherLib

RafaelPalomar:
Since Slicer_USE_CTKAPPLAUNCHER option exists, it makes sense that,
when Slicer_USE_CTKAPPLAUNCHER=OFF, CTKAppLauncherLib is not
searched. Currently, find_package(CTKAppLauncherLib REQUIRED) is
called regardless of the state of Slicer_USE_CTKAPPLAUNCHER. This can
cause configuration issues when configuring Slicer.

With the following configuration:

		-DSlicer_SUPERBUILD=OFF
		-DBUILD_TESTING=OFF
		-DSlicer_BUILD_EXTENSIONMANAGER_SUPPORT=OFF
		-DSlicer_BUILD_CLI_SUPPORT=OFF
		-DSlicer_BUILD_CLI=OFF
		-DCMAKE_CXX_STANDARD=11
		-DSlicer_REQUIRED_QT_VERSION=5
		-DSlicer_BUILD_DICOM_SUPPORT=OFF
		-DSlicer_BUILD_ITKPython=OFF
		-DSlicer_BUILD_QTLOADABLEMODULES=OFF
		-DSlicer_BUILD_QT_DESIGNER_PLUGINS=OFF
		-DSlicer_USE_CTKAPPLAUNCHER=OFF
		-DSlicer_USE_PYTHONQT=OFF
		-DSlicer_USE_QtTesting=OFF
		-DSlicer_USE_SimpleITK=OFF
		-DSlicer_VTK_RENDERING_BACKEND=OpenGL2
		-DSlicer_VTK_VERSION_MAJOR=8
		-DSlicer_INSTALL_DEVELOPMENT=ON
		-DCMAKE_INSTALL_RPATH=/usr/lib64/Slicer-4.11:/usr/lib64/ctk-0.1:/usr/lib64/Slicer-4.11/qt-loadable-modules:/usr/lib64/ITK-5.1.0
		-DCMAKE_BUILD_WITH_INSTALL_RPATH=ON
		-DSlicer_USE_SYSTEM_LibArchive=ON
		-DTeem_DIR=/usr/lib64
		-DjqPlot_DIR=/usr/share/jqPlot

the following error raises:

CMake Error at Base/QTCore/CMakeLists.txt:45 (find_package):
  By not providing "FindCTKAppLauncherLib.cmake" in CMAKE_MODULE_PATH this
  project has asked CMake to find a package configuration file provided by
  "CTKAppLauncherLib", but CMake did not find one.

  Could not find a package configuration file provided by "CTKAppLauncherLib"
  with any of the following names:

    CTKAppLauncherLibConfig.cmake
    ctkapplauncherlib-config.cmake

  Add the installation prefix of "CTKAppLauncherLib" to CMAKE_PREFIX_PATH or
  set "CTKAppLauncherLib_DIR" to a directory containing one of the above
  files.  If "CTKAppLauncherLib" provides a separate development package or
  SDK, be sure it has been installed.

### DIFF
--- a/Base/QTCore/CMakeLists.txt
+++ b/Base/QTCore/CMakeLists.txt
@@ -42,7 +42,9 @@ endif()
 #
 # CTKAppLauncherLib
 #
-find_package(CTKAppLauncherLib REQUIRED)
+if(Slicer_USE_CTKAPPLAUNCHER)
+  find_package(CTKAppLauncherLib REQUIRED)
+endif()
 
 #
 # See CMake/SlicerMacroBuildBaseQtLibrary.cmake for details


pieper:
This one also looks good to me - thanks for all the fixes! 👍

jcfr:
Thanks for the contribution, it turns out that `CTKAppLauncherLib` is different from `CTKAppLauncher`. It was created to provide a consistent interface for reading environment variables from settings file on all platforms. 

Since it is used on macOS (even when no launcher are distributed), I think we should always keep it.

jcfr:
This is also why `CTKAppLauncherLib` is not listed as a conditional dependency. See https://github.com/Slicer/Slicer/blob/76d0fbdd4d8f94f00f910faf551df54faafa783c/SuperBuild.cmake#L93

jcfr:
@RafaelPalomar If you would like to build against installed version of ITK/VTK/... while allowing the Slicer superbuild system to download dependencies like `CTKAppLauncherLib`, you could do the following:

```
[...]
-DSlicer_SUPERBUILD=ON
-DSlicer_USE_SYSTEM_ITK:BOOL=ON
-DSlicer_USE_SYSTEM_VTK:BOOL=ON
[...]
```

RafaelPalomar:
@pieper, @jcfr, thank you for your insights. I did not know about the difference between CTKAppLauncher and CTKAppLauncherLib. To me, what it would make most sense would be preserve CTKAppLauncherLib as is and let CTKAppLauncher be optional. Let me have a look at how would that be to do this and maybe come with some comments or suggestions on this matter.

RafaelPalomar:
@jcfr , @pieper  I think this change should not be applied based on the comments from @jcfr. It is still possible to build slicer without the CTKAppLauncher while preserving CTKAppLauncherLib for the functionality related to saving settings. 

I close it here. Let me know if there is any reason to re-open it

PR #1222, opened 2019-10-01
COMP: Enable install of development files

RafaelPalomar:
Install of development files is disabled (i.e.,
Slicer_INSTALL_DEVELOPMENT is always set to OFF). This commits allows
the Slicer_INSTALL_DEVELOPMENT option to be set to ON/OFF.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -340,9 +340,9 @@ mark_as_superbuild(Slicer_USE_SimpleITK_SHARED)
 # Install no development files by default, but allow the user to get
 # them installed by setting Slicer_INSTALL_DEVELOPMENT to true.
 #-----------------------------------------------------------------------------
-#option(Slicer_INSTALL_DEVELOPMENT "Install Slicer extension development files." OFF)
-#mark_as_advanced(Slicer_INSTALL_DEVELOPMENT)
-set(Slicer_INSTALL_DEVELOPMENT OFF)
+option(Slicer_INSTALL_DEVELOPMENT "Install Slicer extension development files." OFF)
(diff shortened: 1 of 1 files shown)


pieper:
I'm not sure that actually works, but if you have tried it I don't see any harm.

RafaelPalomar:
This is not enough to make an installable Slicer with development files, but it is a first step. I have advanced further in this direction and plan to pull-request more changes that will bring Slicer closer to work in this aspect. I think most of the changes needed are harmless to other Slicer configurations.

jcfr:
Thanks for much for working on this :pray: 

> it would make more sense if I bring all changes at once

:+1: 

Do not hesitate to have a branch with multiple commits, we can still squash the relevant changes before integrating.

jcfr:
Few additional comments:

* The dealbreaker is that neither VTK or ITK support building external module against an install tree. 
   That said, I still think we should move forward. We could still apply specific workaround to the ITK/VTK fork we maintained in the Slicer organization.

* Slicer currently use an older VTK where we backport fixes. This is done like this because recent VTK completely revamped their build system.  If you feel like it, working on updating Slicer to build against latest VTK would be an outstanding contribution.

RafaelPalomar:
@lassoan, @jcfr, thank you both for reviewing this. I will pull-request a branch with all changes needed to get a working install tree with development files. We can use it as a base for discussion and hopefully an eventual merge.

jcfr:
> I will pull-request a branch with all changes needed to get a working install tree with development files. We can use it as a base for discussion and hopefully an eventual merge.

Sounds great. Thanks for understanding and for working on this.

In the mean time, feel free to close PR that will be superseded.

PR #1223, opened 2019-10-02
BUG: Fix crash when setting circular parents in subject hierarchy

cpinter:
https://issues.slicer.org/view.php?id=4713

### DIFF
--- a/Libs/MRML/Core/vtkMRMLSubjectHierarchyNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLSubjectHierarchyNode.cxx
@@ -2584,8 +2584,13 @@ void vtkMRMLSubjectHierarchyNode::RemoveAllItems(bool removeDataNode/*=false*/)
 }
 
 //----------------------------------------------------------------------------
-void vtkMRMLSubjectHierarchyNode::SetItemParent(vtkIdType itemID, vtkIdType parentItemID)
+void vtkMRMLSubjectHierarchyNode::SetItemParent(vtkIdType itemID, vtkIdType parentItemID, bool enableCircularCheck/*=true*/)
 {
+  if (itemID == parentItemID)
+    {
+    vtkErrorMacro("SetItemParent: Cannot set an item as its own parent: " << itemID);
+    return;
+    }
   vtkSubjectHierarchyItem* item = this->Internal->SceneItem->FindChildByID(itemID);
   if (!item)
     {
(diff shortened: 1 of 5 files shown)


Libs/MRML/Core/vtkMRMLSubjectHierarchyNode.cxx
```diff
@@ -2599,6 +2599,21 @@ void vtkMRMLSubjectHierarchyNode::SetItemParent(vtkIdType itemID, vtkIdType pare
     return;
     }
 
+  // Check if the new parent is not a child of the item
+  if (!skipCircularCheck)
+    {
+    vtkSubjectHierarchyItem* currentParentItem = parentItem;
+    while ( (currentParentItem = currentParentItem->Parent) != nullptr )
```
lassoan: If there is already a loop then this check will hang. As it is a safety check, it would be probably nice to report if there is a loop already. We can detect an existing loop by comparing "item" to "parentItem" as well.

Libs/MRML/Core/vtkMRMLSubjectHierarchyNode.cxx
```diff
@@ -2584,7 +2584,7 @@ void vtkMRMLSubjectHierarchyNode::RemoveAllItems(bool removeDataNode/*=false*/)
 }
 
 //----------------------------------------------------------------------------
-void vtkMRMLSubjectHierarchyNode::SetItemParent(vtkIdType itemID, vtkIdType parentItemID)
+void vtkMRMLSubjectHierarchyNode::SetItemParent(vtkIdType itemID, vtkIdType parentItemID, bool skipCircularCheck/*=false*/)
```
lassoan: bool skipCircularCheck=false => double-negative is a bit harder to decode

why not use an argument name like _safeMode_ or _enableCircularCheck_

cpinter: I didn't like the safeMode attribute when I saw it in Kyle's PR either, I think it's so generic it can be misunderstood. I'll call it enableCircularCheck

lassoan: OK!

Libs/MRML/Core/vtkMRMLSubjectHierarchyNode.cxx
cpinter: I think the new version should do it. Let me know.

cpinter:
Integrated in https://github.com/Slicer/Slicer/commit/a7cbe158067e6c610732ae4c4751cfeb9a9bc9bd

PR #1225, opened 2019-10-09
BUG: Fix bugs and performance issues in Segmentations

Sunderlandkyl:
* Move and reformat "Binary labelmap layers" section of Segmentations module
* Only update layer section in Segmentations while it is open
* Add total segment count to layer section
* Show wait cursor while importing labelmap nodes to segmentation
* Improve performance when modifying segments through segment editor
* Fix bugs in Joint Smoothing and AutoComplete editor effects that allowed resulting segments to exist on multiple layers
* Improve 3D segment pipeline to improve speed when segment visibility is changed
* Preferred display representation is no longer set to be closed surface by default
* Fix bug that allowed segments to be added to the subject hierarchy twice

### DIFF
--- a/Modules/Loadable/Segmentations/EditorEffects/Python/AbstractScriptedSegmentEditorAutoCompleteEffect.py
+++ b/Modules/Loadable/Segmentations/EditorEffects/Python/AbstractScriptedSegmentEditorAutoCompleteEffect.py
@@ -291,9 +291,9 @@ class AbstractScriptedSegmentEditorAutoCompleteEffect(AbstractScriptedSegmentEdi
     previewNode.GetSegmentation().GetSegmentIDs(segmentIDs)
     for index in range(segmentIDs.GetNumberOfValues()):
       segmentID = segmentIDs.GetValue(index)
-      previewSegment = previewNode.GetSegmentation().GetSegment(segmentID)
-      previewSegmentLabelmap = previewSegment.GetRepresentation(vtkSegmentationCore.vtkSegmentationConverter.GetSegmentationBinaryLabelmapRepresentationName())
-      self.scriptedEffect.modifySegmentByLabelmap(segmentationNode, segmentID,previewSegmentLabelmap,
+      previewSegmentLabelmap = slicer.vtkOrientedImageData()
+      previewNode.GetBinaryLabelmapRepresentation(segmentID, previewSegmentLabelmap)
+      self.scriptedEffect.modifySegmentByLabelmap(segmentationNode, segmentID, previewSegmentLabelmap,
         slicer.qSlicerSegmentEditorAbstractEffect.ModificationModeSet)
       if segmentationDisplayNode is not None and self.isBackgroundLabelmap(previewSegmentLabelmap):
         # Automatically hide result segments that are background (all eight corners are non-zero)
--- a/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorSmoothingEffect.py
+++ b/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorSmoothingEffect.py
@@ -335,15 +335,20 @@ If segments overlap, segment higher in the segments table will have priority. <b
     imageToWorldMatrix = vtk.vtkMatrix4x4()
     mergedImage.GetImageToWorldMatrix(imageToWorldMatrix)
 
+    # TODO: Temporarily setting the overwite mode to OverwriteVisibleSegments is an approach that should be change once additional
+    # layer control options have been implemented. Users may wish to keep segments on separate layers, and not allow them to be separated/merged automatically.
+    # This effect could leverage those options once they have been implemented.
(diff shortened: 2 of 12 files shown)


Sunderlandkyl:
@lassoan This should fix most of the bugs mentioned here: https://github.com/Slicer/Slicer/pull/1221#issuecomment-539490163, except the test is still failing on MacOS.

I'll continue to look at that issue.

Sunderlandkyl:
Performance wise, the only remaining factor seems to be copying the segments for undo/redo.

Modules/Loadable/Segmentations/EditorEffects/Python/AbstractScriptedSegmentEditorAutoCompleteEffect.py
```diff
@@ -293,8 +293,14 @@ def onApply(self):
       segmentID = segmentIDs.GetValue(index)
       previewSegment = previewNode.GetSegmentation().GetSegment(segmentID)
       previewSegmentLabelmap = previewSegment.GetRepresentation(vtkSegmentationCore.vtkSegmentationConverter.GetSegmentationBinaryLabelmapRepresentationName())
-      self.scriptedEffect.modifySegmentByLabelmap(segmentationNode, segmentID,previewSegmentLabelmap,
+
+      oldOverwriteMode = self.scriptedEffect.parameterSetNode().GetOverwriteMode()
+      self.scriptedEffect.parameterSetNode().SetOverwriteMode(slicer.vtkMRMLSegmentEditorNode.OverwriteAllSegments)
```
lassoan: Why is this necessary?

Auto-complete effect ignores hidden segments, but now with this change, those segments would be modified, which is probably not what the user wants.

Modules/Loadable/Segmentations/EditorEffects/Python/AbstractScriptedSegmentEditorAutoCompleteEffect.py
```diff
@@ -474,7 +480,12 @@ def preview(self):
         color = segmentationNode.GetSegmentation().GetSegment(segmentID).GetColor()
         newSegment.SetColor(color)
         previewNode.GetSegmentation().AddSegment(newSegment, segmentID)
+
+      oldOverwriteMode = self.scriptedEffect.parameterSetNode().GetOverwriteMode()
+      self.scriptedEffect.parameterSetNode().SetOverwriteMode(slicer.vtkMRMLSegmentEditorNode.OverwriteAllSegments)
```
lassoan: Why is this needed?

Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorSmoothingEffect.py
```diff
@@ -342,8 +342,12 @@ def smoothMultipleSegments(self):
       smoothedBinaryLabelMap.ShallowCopy(stencil.GetOutput())
       smoothedBinaryLabelMap.SetImageToWorldMatrix(imageToWorldMatrix)
       # Write results to segments directly, bypassing masking
-      slicer.vtkSlicerSegmentationsModuleLogic.SetBinaryLabelmapToSegment(smoothedBinaryLabelMap,
-        segmentationNode, segmentId, slicer.vtkSlicerSegmentationsModuleLogic.MODE_REPLACE, smoothedBinaryLabelMap.GetExtent())
+      oldOverwriteMode = self.scriptedEffect.parameterSetNode().GetOverwriteMode()
+      self.scriptedEffect.parameterSetNode().SetOverwriteMode(slicer.vtkMRMLSegmentEditorNode.OverwriteAllSegments)
```
lassoan: Why this is needed?

Modules/Loadable/Segmentations/Resources/UI/qSlicerSegmentationsModule.ui
```diff
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>348</width>
-    <height>719</height>
+    <height>844</height>
```
lassoan: this is a bit big... maybe collapse sections in the .ui file would reduce this?

Sunderlandkyl: I've reduced the size as much as possible in the designer. If it is still too large, I can collapse some sections.

Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorSmoothingEffect.py
Sunderlandkyl: I investigated more, and the other overwrite mode changes in the AutoComplete effect were not required (after some bug fixing).

I've changed the overwrite mode to OverwriteVisibleSegments. I'm not sure if I understand how the layer mode would differ from the overwrite mode.

lassoan:
Thank you, merged!

PR #1224, opened 2019-10-08
COMP: Install python packages using PyPI wheels

jamesobutler:
This updates the build system to install python packages using the official Wheels from PyPI instead of creating Eggs. This change is possible as Slicer now uses the same compiler as is used to generate the PyPI wheels or is at least binary compatible. This PR maintains the version used for each python package.  

The biggest change is probably getting the Numpy wheel from PyPI instead of building it from source and applying patches. The correct Numpy wheel based on platform/python version etc will be handled by the fetch process from PyPI.

PR #1228, opened 2019-10-10
COMP: Install pip, setuptools, wheel using whl files

jamesobutler:
This updates the build process to install pip, setuptools, and wheel python packages using whl files.

python-ensurepip was added as a project that runs the [`ensurepip`](https://docs.python.org/3.6/library/ensurepip.html) python module to specifically install the default pip and setuptools packages for the used python version. In this case Slicer python 3.6.7 installs "pip-10.0.1-py2.py3-none-any.whl" and "setuptools-39.0.1-py2.py3-none-any.whl" which has been included in the python-build directory.

This means setuptools is no longer installed using https://github.com/Slicer/setuptools.  A patch was applied in that fork related to solving a race condition problem (see slicer issue [4394](https://issues.slicer.org/view.php?id=4394)), but I think egg race condition improvements have already been implemented in the upstream as https://github.com/pypa/setuptools/commit/e2dde5bd846016ac229af7322eec363e011737a1, originally included in release 40.1.0.

I have then updated python-pip, python-setuptools and python-wheel to install/upgrade the version specified in their individual project.  In this PR I'm not yet updating the versions of any of the python packages.

### DIFF
--- a/SuperBuild.cmake
+++ b/SuperBuild.cmake
@@ -124,7 +124,6 @@ endif()
 if(Slicer_USE_PYTHONQT)
   list(APPEND Slicer_DEPENDENCIES
     python-pythonqt-requirements  # This provides the "packaging.version.parse()" function
-    python-pip
     )
 endif()
 
--- a/SuperBuild/External_python-ensurepip.cmake
+++ b/SuperBuild/External_python-ensurepip.cmake
@@ -1,14 +1,7 @@
-set(proj python-pythonqt-requirements)
+set(proj python-ensurepip)
 
 # Set dependency list
-set(${proj}_DEPENDENCIES python python-ensurepip python-pip python-setuptools python-wheel)
-
-set(requirements_file ${CMAKE_BINARY_DIR}/${proj}-requirements.txt)
-file(WRITE ${requirements_file} [===[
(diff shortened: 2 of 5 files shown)


jamesobutler:
Could I get a quick review @jcfr @pieper? Thanks!

pieper:
This makes sense to me in general terms - can't say I know enough about the details to evaluate.  Generally 👍

lassoan:
DICOM and pydicom packages are missing - see https://discourse.slicer.org/t/cannot-import-pydicom/8766

jamesobutler:
The dicom packages problem was solved in #1231.  I've noticed that setuptools project is also having similar problems where external dependencies isn't currently being respected in this PR as setuptools is installed with python-ensurepip, but not updated in the python-setuptools project.

jamesobutler:
This PR is ready for review again. I've made slight changes and updated against latest master and through local testing I can confirm that I've solved the previous issues.

@lassoan Would you be able to review and merge for me?

jamesobutler:
Or any other Slicer developer with merge access to help out with this would be great!

lassoan:
Hasn't this pull request merged already?

jamesobutler:
No, this is different from #1224. This is similar to that one, but this PR actually handles the installing of wheels for some other supporting packages like pip, setuptools and wheel.

lassoan:
Thank you, it's been merged in r28580. Please have a look at the dashboard on Saturday or later to check if it works as expected.

PR #1229, opened 2019-10-12
ENH: DICOM browser rework

lassoan:
User features:
- Improved DICOM import: it runs in the background (GUI is not blocked), continuous, detailed progress information is displayed; and it is about 3x faster than before
- DICOM patient/study/series can be loaded by double-clicking on the item
- Improved DICOM module GUI: optimized column widths, default sorting (based on last insertion time, study date, and series number) in DICOM browser
- DICOM database version upgrade is not enforced when updating Slicer (user can choose between upgrade or creating a new database at a new location)
- Fixed starting and stopping of DICOM listener on Windows.
- Remove orphan empty directories after deleting files from the database
- Fixed crash when query/retrieve was used from Python (where database cannot be passed to the Q/R widget as a shared pointer)

Developer features:
- Moved many DICOM features into CTK (export to file system, DICOM metadata display, etc.)
- Separated DICOM widget from logic: DICOM database is now owned by Slicer application using a shared pointer. This allows DICOM import and loading without using GUI classes (without instantiating DICOM browser widget). See example in DICOMUtils.loadSeriesByUID.
- DICOM database object is always created (even if no valid DICOM database location is specified) when the application is started. If no valid DICOM database location is specified then the database will not be open.
- Made recent activity widget selection trigget selection of the series in DICOM browser
- Fixed tests (JRC2013Vis and RSNAVisTutorial, SubjectHierarchyGenericSelfTest)

image


lassoan:
I've committed this (r28545) to get dashboard builds on all platforms. Any comments and suggestions are still welcome.

pieper:
Yes, good - it's a big change but we'll only really find issues if people start using it.

fedorov:
@lassoan thank you for working on this - much needed improvements!

Today I put it to test to index a folder of about 60Gb. With the pre-improvements DICOM browser, after I selected that as input folder, I get "wait" cursor for several minutes with no indication of what's going on (I lost patience to see how long that would take to complete, and killed it).

With your implementation, import started right away, and was very quick (I think 3x is an under-estimate) - I think parsing step was done in the order of tens of minutes.

However, after parsing was done, indexing turned out problematic. I experienced blocking "wait" cursor while using other applications beyond Slicer. Looking at the status of the Slicer application, it did not seem to be using a lot of CPU, but probably allocating memory very aggressively - I had about 35Gb of disk space, and I had to kill Slicer application because it consumed most of it - I was not able to complete the import because of that.

fedorov:
image

image

lassoan:
Thanks for testing. Indexing results is cached into memory before writing to the database, so memory usage can be significant, but it is surprising that it slows down your computer that much.

I've tested with 30GB data set (first 50 patient of TCIA's TCGA-HNSC) and I did not experience any slowdown on my computer with 32GB RAM because it needed "just" 26GB RAM.

Did you see memory hike during "Update database fields" phase or you run out of memory even before (during the file parsing phase)?

Which dataset you have tested with?

Maybe we need to adjust the maximum tag length that is cached (image data is filtered out but maybe in some data sets there are other large tags that can consume a lot of memory). I could also change the implementation to force database update after a certain batch size is reached. I would need more info (answers to the questions above and some more testing) to see which solution to choose.

fedorov:
> I've tested with 30GB data set (first 50 patient of TCIA's TCGA-HNSC) and I did not experience any slowdown on my computer with 32GB RAM because it needed "just" 26GB RAM. This memory usage of course is not reasonable.

This is the configuration for my system (I inherited it from Ron, and because of this I think it was a top of the line Mac back in 2016!): 

image

> Did you see memory hike during "Update database fields" phase or you run out of memory even before (during the file parsing phase)?

There is a tiny increase towards the end of indexing, but once it starts updating the database, it spikes rather quickly (~45 degree slope in the plot below).

image

> Which dataset you have tested with?

I tested with a non-public dataset, but just now tried with the public [LIDC collection](https://wiki.cancerimagingarchive.net/display/Public/LIDC-IDRI) (which is much larger - 128G! - I happen to have it locally), and I had the same issue.

lassoan:
OK. If you see the hike during database update then it's probably the same hike that I'm seeing, too, so I can address that. Until then, as a workaround, you need to add directories in smaller batches.

fedorov:
Thank you - let me know when the next iteration is ready for testing!

lassoan:
I've pushed a fix, it should not consume significant amount of memory.

It was partly due to a bug that caused reading of pixel data in the memory-cached data set; partly because tens of thousands of data sets were kept in memory (now we force database update after each 5000 files).

Please test it in tomorrow Slicer Preview Release.

fedorov:
Thank you Andras - I'll plan to test it tomorrow, together with the pydicom fix!

fedorov:
This is great Andras, the problem is gone! I was able to index that 60Gb folder in under 15 minutes I think!

Small nitpick - the step "Updating database displayed fields" takes about 2-3 minutes, and during that time the progress bar is not updating (unless I missed it). The application is responsive, and it is nothing more than a small annoyance. Just in case you want to calibrate a bit better progress reporting.

lassoan:
That's great.

Yes, displayed fields update does not provide detailed progress status. Usually it takes only a few seconds to complete, but for large data sets indeed it can take a while.

fedorov:
Andras, if you didn't think about it, I suggest you announce this improvement on the forum. This improvement is very important. I think users will appreciate this update a lot.

lassoan:
Yes, I'm planning to write a forum post about it.

How do you like the data tree display on the left and that you can click on the eye icon to show/hide volumes?
Have you noticed that you can load a patient/study/series by double-click?

lassoan:
Merged.

pieper:
It's been working really well for me 👍

PR #1230, opened 2019-10-14
ENH: Resample curves from GUI and constrain points

smrolfe:
@lassoan This PR adds a menu to markups module to resample open or closed curves with an option to constrain to a model surface. It also adds utilities to resample with surface constraint.

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
@@ -19,25 +19,38 @@
 
 // MRML includes
 #include "vtkCurveGenerator.h"
+#include "vtkMRMLMarkupsNode.h"
 #include "vtkMRMLMarkupsDisplayNode.h"
 #include "vtkMRMLMarkupsFiducialStorageNode.h"
 #include "vtkMRMLScene.h"
+#include "vtkMRMLTransformNode.h"
+
 
 // VTK includes
+#include <vtkBoundingBox.h>
 #include <vtkCutter.h>
 #include <vtkDoubleArray.h>
 #include <vtkFrenetSerretFrame.h>
+#include <vtkGeneralTransform.h>
+#include <vtkGenericCell.h>
+
 #include <vtkLine.h>
 #include <vtkMatrix4x4.h>
 #include <vtkNew.h>
+#include <vtkOBBTree.h>
 #include <vtkObjectFactory.h>
 #include <vtkPlane.h>
 #include <vtkPointData.h>
+#include <vtkPointLocator.h>
 #include <vtkPolyData.h>
+#include <vtkPolyDataNormals.h>
 #include <vtkStringArray.h>
+#include <vtkTransformPolyDataFilter.h>
 
 // STD includes
 #include <sstream>
+
+class vtkMRMLModelNode;
 
 //----------------------------------------------------------------------------
 vtkMRMLNodeNewMacro(vtkMRMLMarkupsCurveNode);
@@ -182,7 +195,232 @@ double vtkMRMLMarkupsCurveNode::GetCurveLengthBetweenStartEndPointsWorld(vtkIdTy
     return this->GetCurveLengthWorld(0, endCurvePointIndex + 1) + this->GetCurveLengthWorld(startCurvePointIndex, -1);
   }
 }
+//---------------------------------------------------------------------------
+bool vtkMRMLMarkupsCurveNode::SetControlPointLabels(vtkStringArray* labels, vtkPoints* points)
+{
+  return this->SetControlPointLabelsWorld(labels, points);
+}
+//---------------------------------------------------------------------------
+bool vtkMRMLMarkupsCurveNode::ResampleCurveSurface(double controlPointDistance, vtkMRMLModelNode* modelNode, double maximumSearchRadiusTolerance)
+{
+	if(!modelNode)
+	  {
+	  vtkErrorMacro("vtkMRMLMarkupsCurveNode::ResampleCurveSurface failed: Constraint surface is not valid");
+	  return false;
+	  }
+	if (maximumSearchRadiusTolerance <= 0 || maximumSearchRadiusTolerance >= 1)
+	  {
+	  vtkErrorMacro("vtkMRMLMarkupsCurveNode::ResampleCurveSurface failed: Invalid search radius");
+	  return false;
+	  }
+	vtkPoints* originalPoints = this->GetCurvePointsWorld();
+	// If there is less than two points there is no segment to resample. Control points are already assumed to be on the surface.
+	if (!originalPoints  || originalPoints->GetNumberOfPoints() < 2)
+      {
+	  vtkErrorMacro("vtkMRMLMarkupsCurveNode::ResampleCurveSurface failed: Invalid number of control points");
+	  return false;
+      }
+	vtkNew<vtkPoints> originalControlPoints;
+	this->GetControlPointPositionsWorld(originalControlPoints);
+	vtkNew<vtkStringArray> originalLabels;
+	this->GetControlPointLabels(originalLabels);
 
+	vtkNew<vtkPoints> interpolatedPoints;
+	vtkMRMLMarkupsCurveNode::ResamplePoints(originalPoints, interpolatedPoints, controlPointDistance, this->CurveClosed);
+
+	vtkPolyData* surfacePolydata = modelNode->GetPolyData();
+	if(!surfacePolydata)
+	  {
+	  vtkErrorMacro("vtkMRMLMarkupsCurveNode::ResampleCurveSurface failed: Constraint surface polydata is not valid");
+	  return false;
+	  }
+
+	vtkMRMLTransformNode* parentTransformNode = modelNode->GetParentTransformNode();
+	if (parentTransformNode)
+	  {
+	  vtkNew<vtkGeneralTransform> modelToWorldTransform;
+	  modelToWorldTransform->Identity();
+	  parentTransformNode->GetTransformToWorld(modelToWorldTransform);
+	  vtkNew<vtkTransformPolyDataFilter> transformPolydataFilter;
+	  transformPolydataFilter->SetInputData(surfacePolydata);
+	  transformPolydataFilter->SetTransform(modelToWorldTransform);
+	  transformPolydataFilter->Update();
+	  surfacePolydata = transformPolydataFilter->GetOutput();
+	  }
+
+	// vtkStaticCellLocator may imporove speed, but FindClosestPoint function is not yet supported
+	vtkNew<vtkPointLocator> pointLocator;
+	pointLocator->SetDataSet(surfacePolydata);
+	pointLocator->BuildLocator();
+
+	vtkDataArray* normalVectorArray = vtkArrayDownCast<vtkDataArray>(surfacePolydata->GetPointData()->GetArray("Normals"));
+	if(!normalVectorArray)
+	  {
+	  vtkNew<vtkPolyDataNormals> normalFilter;
+	  normalFilter->SetInputData(surfacePolydata);
+	  normalFilter->ComputePointNormalsOn();
+	  normalFilter->Update();
+	  vtkPolyData* normalPolydata = normalFilter->GetOutput();
+	  vtkDataArray* normalVectorArray = vtkArrayDownCast<vtkDataArray>(normalPolydata->GetPointData()->GetNormals());
+	  if (!normalVectorArray)
+	    {
+		vtkErrorMacro("vtkMRMLMarkupsCurveNode::ResamplePoints failed: Unable to find or calculate normals");
+		return false;
+		}
+	  }
+
+	double interpolatedPoint[3] = { 0.0 };
+	double projectedPoint[3] = { 0.0 };
+	double segmentStartPoint[3] = { 0.0 };
+	double segmentEndPoint[3] = { 0.0 };
+	double distanceToStart = 0.0;
+	double distanceToEnd = 0.0;
+	vtkNew<vtkPoints> pointNormalArray;
+	// Project each mesh point onto the most external model surface point along a normal vector ray. The normal vector
+	// for each resampled point is estimated from normal vectors at the closest control points on either side.
+	// If there is no intersection between the model and normal vector ray, use the closest model point.
+	for (vtkIdType controlPointIndex = 0; controlPointIndex < interpolatedPoints->GetNumberOfPoints(); controlPointIndex++)
+	  {
+	  interpolatedPoints->GetPoint(controlPointIndex, interpolatedPoint);
+	  vtkIdType segmentStartIndex = vtkMRMLMarkupsNode::GetClosestControlPointIndexToPositionWorld(interpolatedPoint);
+	  originalControlPoints->GetPoint(segmentStartIndex, segmentStartPoint);				// get nearest control point
+	  this->GetClosestPointPositionAlongCurveWorld(interpolatedPoint, projectedPoint);	// get projection of point to curve
+	  if (segmentStartIndex == 0)
+		{
+		originalControlPoints->GetPoint(segmentStartIndex + 1, segmentEndPoint);
+		distanceToStart = 0;
+		distanceToEnd = vtkMath::Distance2BetweenPoints(segmentEndPoint, projectedPoint);
+
+		}
+	  else if (segmentStartIndex == (originalControlPoints->GetNumberOfPoints() - 1))
+	    {
+	    originalControlPoints->GetPoint(segmentStartIndex - 1, segmentEndPoint);
+		distanceToStart = vtkMath::Distance2BetweenPoints(segmentStartPoint, projectedPoint);
+		distanceToEnd = 0;
+	    }
+	  else
+		{
+		double* segmentEndPoint1 = originalControlPoints->GetPoint(segmentStartIndex - 1);
+		double dist1 = vtkMath::Distance2BetweenPoints(segmentEndPoint1, projectedPoint);
+		double* segmentEndPoint2 = originalControlPoints->GetPoint(segmentStartIndex + 1);
+		double dist2 = vtkMath::Distance2BetweenPoints(segmentEndPoint2, projectedPoint);
+		distanceToStart = vtkMath::Distance2BetweenPoints(segmentStartPoint, projectedPoint);
+		if ((dist1 < dist2) && dist1 < vtkMath::Distance2BetweenPoints(segmentEndPoint1, segmentStartPoint))
+		  {
+		  segmentEndPoint[0] = segmentEndPoint1[0];
+		  segmentEndPoint[1] = segmentEndPoint1[1];
+		  segmentEndPoint[2] = segmentEndPoint1[2];
+		  distanceToEnd = dist1;
+		  }
+		else
+		  {
+		  segmentEndPoint[0] = segmentEndPoint2[0];
+		  segmentEndPoint[1] = segmentEndPoint2[1];
+		  segmentEndPoint[2] = segmentEndPoint2[2];
+		  distanceToEnd = dist2;
+		  }
+		}
+	  // estimate normal vector from control points on either side of projected point
+	  vtkIdType pointIdStart = pointLocator->FindClosestPoint(segmentStartPoint);
+	  double startNormal[3] = { 0.0 };
+	  normalVectorArray->GetTuple(pointIdStart, startNormal);
+	  vtkIdType pointIdEnd = pointLocator->FindClosestPoint(segmentEndPoint);
+	  double endNormal[3] = { 0.0 };
+	  normalVectorArray->GetTuple(pointIdEnd, endNormal);
+
+	  double startWeight = distanceToStart / (distanceToStart + distanceToEnd);
+	  double endWeight = distanceToEnd / (distanceToStart + distanceToEnd);
+	  double rayDirection[3] = { 0.0 };
+	  rayDirection[0] = (startWeight*startNormal[0]) + (endWeight*endNormal[0]);
+	  rayDirection[1] = (startWeight*startNormal[1]) + (endWeight*endNormal[1]);
+	  rayDirection[2] = (startWeight*startNormal[2]) + (endWeight*endNormal[2]);
+	  vtkMath::Normalize(rayDirection);
+	  pointNormalArray->InsertNextPoint(rayDirection);
+		
+	  }
+
+	vtkNew<vtkPoints> snappedToSurfaceControlPoints;
+	vtkMRMLMarkupsCurveNode::ConstrainPointsToSurface(interpolatedPoints, pointNormalArray, surfacePolydata,
+	  snappedToSurfaceControlPoints, maximumSearchRadiusTolerance);
+
+	this->SetControlPointPositionsWorld(snappedToSurfaceControlPoints);
+	this->SetControlPointLabelsWorld(originalLabels, originalControlPoints);
+	return true;
+}
+//---------------------------------------------------------------------------
+bool vtkMRMLMarkupsCurveNode::ConstrainPointsToSurface(vtkPoints* originalPoints, vtkPoints* normalVectors, vtkPolyData* surfacePolydata,
+  vtkPoints* surfacePoints, double maximumSearchRadiusTolerance)
+{
+  if (originalPoints->GetNumberOfPoints()!= normalVectors->GetNumberOfPoints())
+	{
+	vtkGenericWarningMacro("vtkMRMLMarkupsCurveNode::ConstrainPointsToSurface failed: invalid inputs");
+	return false;
+	}
+  if (maximumSearchRadiusTolerance <= 0 || maximumSearchRadiusTolerance >= 1)
+    {
+	vtkGenericWarningMacro("vtkMRMLMarkupsCurveNode::ConstrainPointsToSurface failed: Invalid search radius");
+	return false;
+    }
+  vtkNew<vtkOBBTree> surfaceObbTree;
+  surfaceObbTree->SetDataSet(surfacePolydata);
+  surfaceObbTree->BuildLocator();
+  double tolerance = surfaceObbTree->GetTolerance();
+
+  vtkNew<vtkPointLocator> pointLocator;
+  pointLocator->SetDataSet(surfacePolydata);
+  pointLocator->BuildLocator();
+
+  double originalPoint[3] = { 0.0 };
+  double rayDirection[3] = { 0.0 };
+  double exteriorPoint[3] = { 0.0 };
+
+  // Curves are expected to be close to surface. The maximumSearchRadiusTolerance
+  // sets the allowable projection distance as a percentage of the model's
+  // bounding box diagonal in world coordinate system.
+  double polydataBounds[6] = { 0.0 };
+  surfacePolydata->GetBounds(polydataBounds);
+  vtkBoundingBox modelBoundingBox;
+  modelBoundingBox.AddBounds(polydataBounds);
+  double polydataDiagonalLength = modelBoundingBox.GetDiagonalLength();
+  double rayLength = maximumSearchRadiusTolerance*sqrt(polydataDiagonalLength);
+
+  for (vtkIdType controlPointIndex = 0; controlPointIndex < originalPoints->GetNumberOfPoints(); controlPointIndex++)
+	{
+	originalPoints->GetPoint(controlPointIndex, originalPoint);
+	normalVectors->GetPoint(controlPointIndex, rayDirection);
+	//cast ray and find model intersection point
+	double rayEndPoint[3] = { 0.0 };
+	rayEndPoint[0] = originalPoint[0] + rayDirection[0] * rayLength;
+	rayEndPoint[1] = originalPoint[1] + rayDirection[1] * rayLength;
+	rayEndPoint[2] = originalPoint[2] + rayDirection[2] * rayLength;
+
+	double t;
+	double firstIntersectionPoint[3] = { 0.0 };
+	double pcoords[3] = { 0.0 };
+	int subId;
+	vtkIdType cellId;
+	vtkNew <vtkGenericCell> cell;
+	int foundIntersection = surfaceObbTree->IntersectWithLine(rayEndPoint, originalPoint, tolerance, t, exteriorPoint, pcoords, subId, cellId, cell);
+	if(foundIntersection == 0)
+	  {
+	  //if no intersection, reverse direction of normal vector ray
+	  rayEndPoint[0] = originalPoint[0] + rayDirection[0] * -rayLength;
+	  rayEndPoint[1] = originalPoint[1] + rayDirection[1] * -rayLength;
+	  rayEndPoint[2] = originalPoint[2] + rayDirection[2] * -rayLength;
+	  int foundIntersection = surfaceObbTree->IntersectWithLine(originalPoint, rayEndPoint, tolerance, t, exteriorPoint, pcoords, subId, cellId, cell);
+	  if(foundIntersection == 0)
+	    {
+		//if no intersection in either direction, use closest mesh point
+		vtkIdType closestPointId = pointLocator->FindClosestPoint(originalPoint);
+		surfacePolydata->GetPoint(closestPointId, exteriorPoint);
+		vtkGenericWarningMacro("No intersections found");
+	    }
+	  }
+	surfacePoints->InsertNextPoint(exteriorPoint);
+    }
+  return true;
+
+}
 //---------------------------------------------------------------------------
 void vtkMRMLMarkupsCurveNode::ResampleCurveWorld(double controlPointDistance)
 {
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.h
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.h
@@ -20,11 +20,15 @@
 
 // MRML includes
 #include "vtkMRMLDisplayableNode.h"
+#include "vtkMRMLModelNode.h"
 
 // Markups includes
 #include "vtkSlicerMarkupsModuleMRMLExport.h"
 #include "vtkMRMLMarkupsDisplayNode.h"
 #include "vtkMRMLMarkupsNode.h"
+
+// VTK includes
+#include <vtkStringArray.h>
 
 class vtkPlane;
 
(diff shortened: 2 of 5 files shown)


jcfr:
I suggest you squash all the commit together. Let us know if you need help to do so.

smrolfe:
Thanks, I squashed the commits locally and then pushed that to my master branch. Could you let me know if this is correct now in the pr or if I need to do something else?

jcfr:
It looks like something went wrong when you squashed the commit. Indeed, there are still 14 commits associated with the PR.

I suggest you try the following:

```
git rebase --interactive dd85e97a17
```

Commit dd85e97a17 corresponds to the Slicer commit just before your first one.

Then, you a text editor should be open asking to choose what to do with each commit:

image

In the screenshot above, I decided to squash each commit with the previous one.

Last step, is to save and then craft a meaningful commit message.

Let me know if you have any questions

jcfr:
Since the PR was done by integrating smrolfe  master branch into the Slicer one, I suggest you update master.

Looking at your GitHub fork, it looks like you created a new branch named `snapCurvePoints`

To update the PR, I suggest you do the following (assuming `smrolfe` corresponds to your GitHub fork):

```
git checkout snapCurvePoints
git push smrolfe snapCurvePoints:master --force
```

lassoan:
@smrolfe you may also try to use a git client with a GUI. On Windows I use TortoiseGit but there are many others. I find it much easier to do all operations with a GUI, as I can easily see what options are available (don't need to memorize commands and options).

smrolfe:
Thanks @jcfr, it looks like this worked. 

@lassoan thanks, I've installed TortoiseGit and will use it going forward!

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
```diff
@@ -183,6 +189,205 @@ double vtkMRMLMarkupsCurveNode::GetCurveLengthBetweenStartEndPointsWorld(vtkIdTy
...
+	double rayDirection[3] = { 0.0 };
+	double exteriorPoint[3] = { 0.0 };
+
+	// Curve is expected to be close to surface. If projection distance is greater than 10% of image size
+	// it's likely due to a second external surface
+	double bounds[6] = { 0.0 };
+	surfacePolydata->GetBounds(bounds);
+	double point1[3] = { bounds[0], bounds[2], bounds[4]};
+	double point2[3] = { bounds[1], bounds[3], bounds[5] };
+	double polydataSize = vtkMath::Distance2BetweenPoints(point1,point2);
```
lassoan: It would be more elegant to use vtkBoundingBox::GetDiagonalLength to compute polydataSize. Also, size typically refers to integer count, so rename it something like length or diameter.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
```diff
@@ -183,6 +189,205 @@ double vtkMRMLMarkupsCurveNode::GetCurveLengthBetweenStartEndPointsWorld(vtkIdTy
...
+		normalVectors->GetPoint(controlPointIndex, rayDirection);
+		//cast ray and find model intersection point
+		double rayEndPoint[3] = { 0.0 };
+		rayEndPoint[0] = originalPoint[0] + rayDirection[0] * rayLength;
+		rayEndPoint[1] = originalPoint[1] + rayDirection[1] * rayLength;
+		rayEndPoint[2] = originalPoint[2] + rayDirection[2] * rayLength;
+
+		vtkNew<vtkIdList> intersectionIds;
+		vtkNew<vtkPoints> intersectionPoints;
+		obbTree->IntersectWithLine(originalPoint, rayEndPoint, intersectionPoints, intersectionIds);
```
lassoan: Use this method to return the first intersection: https://vtk.org/doc/nightly/html/classvtkOBBTree.html#a268ecf8a6ed5d343d5c98c386918c98b

Since the polydata does not change, we can use a locator that is probably faster: https://vtk.org/doc/nightly/html/classvtkStaticCellLocator.html#details

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.h
```diff
@@ -83,6 +84,15 @@ class  VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkMRMLMarkupsCurveNode : public vt
   /// \return sum of distances between the curve points, returns 0 in case of an error
   double GetCurveLengthBetweenStartEndPointsWorld(vtkIdType startCurvePointIndex, vtkIdType endCurvePointIndex);
 
+  void ResampleCurveSurface(double controlPointDistance, vtkMRMLModelNode* node );
+  // Resample a curve with points constrained to surface
```
lassoan: Put comments above the method.

use /// instad of // (that indicates the document generator to include the comment in the class documentation)

Modules/Loadable/Markups/Resources/UI/qSlicerMarkupsModule.ui
```diff
@@ -714,6 +714,100 @@
...
+              <property name="addEnabled">
+                <bool>false</bool>
+              </property>
+              <property name="removeEnabled">
+                <bool>false</bool>
+              </property>
+            </widget>
+          </item>
+          <item row="2" column="0">
+            <widget class="QLabel" name="ResampleCurveRateLabel">
```
lassoan: rename widget to reflect the actual meaning

Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
```diff
@@ -432,6 +432,33 @@ void qSlicerMarkupsModuleWidgetPrivate::setupUi(qSlicerWidget* widget)
...
+  this->MRMLNodeComboBox_constraintNodeSelector->setEnabled(true);
+
+  QObject::connect(this->applyCurveResamplingPushButton, SIGNAL(clicked()),
+	  q, SLOT(onapplyCurveResamplingPushButtonClicked()));
+
+  //check for active curve and enable 
+  vtkMRMLSelectionNode *selNode = this->selectionNode();
+  if (selNode)
+  {
+	  vtkMRMLNode *mrmlNode = q->mrmlScene()->GetNodeByID(selNode->GetActivePlaceNodeID());
```
lassoan: always use two spaces for alignment (never use tabs)

align braces, too

Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
```diff
@@ -432,6 +432,33 @@ void qSlicerMarkupsModuleWidgetPrivate::setupUi(qSlicerWidget* widget)
...
+	  vtkMRMLMarkupsCurveNode *markupsCurveNode = vtkMRMLMarkupsCurveNode::SafeDownCast(mrmlNode);
```
lassoan: use d->MarkupsNode instead of getting it from selection node etc.

Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
```diff
@@ -815,6 +842,35 @@ void qSlicerMarkupsModuleWidget::updateWidgetFromMRML()
...
+  {
+	  vtkMRMLNode *mrmlNode = this->mrmlScene()->GetNodeByID(selNode->GetActivePlaceNodeID());
+	  vtkMRMLMarkupsCurveNode *markupsCurveNode = vtkMRMLMarkupsCurveNode::SafeDownCast(mrmlNode);
+	  if (markupsCurveNode)
+	  {
+		  d->applyCurveResamplingPushButton->setEnabled(true);
+	  }
+	  else
+	  {
+		vtkMRMLMarkupsClosedCurveNode *markupsClosedCurveNode = vtkMRMLMarkupsClosedCurveNode::SafeDownCast(mrmlNode);
```
lassoan: use d->MarkupsNode

Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
```diff
@@ -815,6 +842,35 @@ void qSlicerMarkupsModuleWidget::updateWidgetFromMRML()
...
+		else
+		{
+			d->applyCurveResamplingPushButton->setEnabled(false);
+		}
+	  }
+	 
+  }
+  else
+  {
+	  d->applyCurveResamplingPushButton->setEnabled(false);
```
lassoan: We should offer resample for all markup nodes, just disable the number of points spinbox. It would allow us to snap any annotation to a model.

Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
```diff
@@ -2403,6 +2460,41 @@ void qSlicerMarkupsModuleWidget::onResetToDefaultDisplayPropertiesPushButtonClic
...
+void qSlicerMarkupsModuleWidget::onapplyCurveResamplingPushButtonClicked()
+{
+  Q_D(qSlicerMarkupsModuleWidget);
+
+  double resampleRate = d->ResampleCurveRateSpinBox->value();
+  vtkMRMLSelectionNode *selNode = d->selectionNode();	
+  if (selNode)
+  {
+	vtkMRMLNode *mrmlNode = this->mrmlScene()->GetNodeByID(selNode->GetActivePlaceNodeID());
+	vtkMRMLMarkupsCurveNode *markupsCurveNode = vtkMRMLMarkupsCurveNode::SafeDownCast(mrmlNode);
```
lassoan: use use d->MarkupsNode

Modules/Loadable/Markups/Resources/UI/qSlicerMarkupsModule.ui
```diff
@@ -714,6 +714,100 @@
...
+          <bool>true</bool>
+        </property>
+        <layout class="QGridLayout" name="gridLayout1">
+          <item row="1" column="0">
+            <widget class="QLabel" name="constraintNodeSelectorLabel">
+              <property name="toolTip">
+                <string>Optionally select a node to constrain placement of resampled curve points. </string>
+              </property>
+              <property name="text">
+                <string>Surface constraint node</string>
```
lassoan: Rename to "Constrain points to surface:".

Modules/Loadable/Markups/Resources/UI/qSlicerMarkupsModule.ui
```diff
@@ -714,6 +714,100 @@
...
+              </property>
+            </widget>
+          </item>
+          <item row="2" column="0">
+            <widget class="QLabel" name="ResampleCurveRateLabel">
+              <property name="toolTip">
+                <string>Select the number of points on the resampled curve. </string>
+              </property>
+              <property name="text">
+                <string>Number of resampled points</string>
```
lassoan: add : to the end

lassoan:
This shows the algorithm error that I mentioned above. Original curve ran slightly above the surface in one region (see red circle in the image). That part of the curve was not snapped to the (very close) surface below but to the (very far) surface above.

image

smrolfe:
Thanks for your feedback @lassoan . The error in the example above is likely due to the points having  a second exterior surface located within the length of the normal vector ray. I set the length of that ray to 10% of the image diameter, which works for most cases I've tried. It might be useful to some users to adjust this property, though it would be difficult to set without understanding the algorithm. Would it make sense to add it under a collapsed "advanced" menu?

smrolfe:
I also agree about making the original curve accessible and will add an output node selector for the results.

We don't have a use case now that requires the dynamic snapping to surface. We could add this later if you think it would be useful.

lassoan:
Have you seen my code review comments?

smrolfe:
Yes, thanks. I am compiling now with these updates.

Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
smrolfe: Could we add resample for other markup node types at a later point? We'd like to get this working for curves as soon as we can.

lassoan: Yes

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
smrolfe: The alternate method for finding the intersection requires a tolerance value to be set. It wasn't clear to me how to set this value from the documentation. I'd prefer to use the other method if it may need to change.

lassoan: Tolerance is needed for almost all geometric operations, otherwise imprecision of floating-point operations would cause lots of problems. In this specific case, the line is considered to cross a cell if it is inside the cell, or it is outside the cell but its distance from the edge of the cell is less than 'tol'.

When you use the "other" method, tolerance is defined by the [Tolerance](https://vtk.org/doc/nightly/html/classvtkLocator.html#af97088c133f28f9f6c4e617615a510e9) member variable, that's why it is not required as an input argument for the method. You can use value of Tolerance member variable.

If you have any doubts about what the two methods do, you can have a look at their source code:
- https://github.com/Kitware/VTK/blob/master/Filters/General/vtkOBBTree.cxx#L663
- https://github.com/Kitware/VTK/blob/master/Filters/General/vtkOBBTree.cxx#L896

As you can see the first method (that you used originally) does many things and we discard most of the computation results. We cannot afford to do this when we have complex meshes or when we'll compute the curve snapping from control points on-the-fly.

smrolfe: Great, thanks for explaining this. I'll make this change.

smrolfe: I ran into an issue with the vtkStaticCellLocator. The code compiles, but then gives the error

`The locator class - vtkStaticCellLocator does not yet support FindClosestPoint`

This only comes up rarely since it's only used when no intersection is found, but I noticed that in this case it was skipping these points rather than using the closest mesh point.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
```diff
@@ -35,10 +37,19 @@
 #include <vtkPointData.h>
 #include <vtkPolyData.h>
 #include <vtkStringArray.h>
+#include <vtkOBBTree.h>
+#include <vtkPointLocator.h>
+#include <vtkPolyDataNormals.h>
+#include <vtkDoubleArray.h>
```
lassoan: Double-check that all includes are required and insert them in alphabetical order (it helps in avoiding including the same include multiple times)

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
```diff
@@ -183,6 +194,182 @@ double vtkMRMLMarkupsCurveNode::GetCurveLengthBetweenStartEndPointsWorld(vtkIdTy
   }
 }
 
+//---------------------------------------------------------------------------
+void vtkMRMLMarkupsCurveNode::ResampleCurveSurface(double controlPointDistance, vtkMRMLModelNode* modelNode, double maximumSearchRadius)
+{
+	if(!modelNode)
+	  {
+	  vtkErrorMacro("vtkMRMLMarkupsCurveNode::ResampleCurveSurface failed: Constraint surface is not valid");
```
lassoan: return

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
```diff
@@ -183,6 +194,182 @@ double vtkMRMLMarkupsCurveNode::GetCurveLengthBetweenStartEndPointsWorld(vtkIdTy
...
+	  }
+	vtkPoints* originalPoints = this->GetCurvePointsWorld();
+	if (!originalPoints  || originalPoints->GetNumberOfPoints() < 2)
+      {
```
lassoan: indentation

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
```diff
@@ -183,6 +194,182 @@ double vtkMRMLMarkupsCurveNode::GetCurveLengthBetweenStartEndPointsWorld(vtkIdTy
...
+	vtkNew<vtkStringArray> originalLabels;
+	this->GetControlPointLabels(originalLabels);
+
+	vtkNew<vtkPoints> interpolatedPoints;
+	vtkMRMLMarkupsCurveNode::ResamplePoints(originalPoints, interpolatedPoints, controlPointDistance, this->CurveClosed);
+
+	vtkPolyData* surfacePolydata = modelNode->GetPolyData();
+	if(!surfacePolydata)
+	  {
+	  vtkErrorMacro("vtkMRMLMarkupsCurveNode::ResampleCurveSurface failed: Constraint surface polydata is not valid");
```
lassoan: return

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
```diff
@@ -183,6 +194,182 @@ double vtkMRMLMarkupsCurveNode::GetCurveLengthBetweenStartEndPointsWorld(vtkIdTy
...
+	  {
+	  vtkNew<vtkPolyDataNormals> normalFilter;
+	  normalFilter->SetInputData(surfacePolydata);
+	  normalFilter->ComputePointNormalsOn();
+	  normalFilter->Update();
+	  vtkPolyData* normalPolydata = normalFilter->GetOutput();
+	  vtkDataArray* normalVectorArray = vtkArrayDownCast<vtkDataArray>(normalPolydata->GetPointData()->GetNormals());
+	  if (!normalVectorArray)
+	    {
+		vtkGenericWarningMacro("vtkMRMLMarkupsCurveNode::ResamplePoints failed: Unable to find or calculate normals");
```
lassoan: Generic warning macro is only intended for static functions, we should use vtkWarningMacro here. But since the operation fails, it would be better to log an error instead.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
```diff
@@ -183,6 +194,182 @@ double vtkMRMLMarkupsCurveNode::GetCurveLengthBetweenStartEndPointsWorld(vtkIdTy
   }
 }
 
+//---------------------------------------------------------------------------
+void vtkMRMLMarkupsCurveNode::ResampleCurveSurface(double controlPointDistance, vtkMRMLModelNode* modelNode, double maximumSearchRadius)
```
lassoan: Since this operation can fail, add a bool return value. Return true on success and false on failure.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
```diff
@@ -183,6 +194,182 @@ double vtkMRMLMarkupsCurveNode::GetCurveLengthBetweenStartEndPointsWorld(vtkIdTy
...
+		return;
+		}
+	  }
+
+	double interpolatedPoint[3] = { 0.0 };
+	double projectedPoint[3] = { 0.0 };
+	double segmentStartPoint[3] = { 0.0 };
+	double segmentEndPoint[3] = { 0.0 };
+	vtkNew<vtkPoints> pointNormalArray;
+	//Project each mesh point onto the most external model surface point along a normal vector ray. The normal vector
```
lassoan: space after //

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
```diff
@@ -183,6 +194,182 @@ double vtkMRMLMarkupsCurveNode::GetCurveLengthBetweenStartEndPointsWorld(vtkIdTy
...
+
+	vtkNew<vtkPoints> interpolatedPoints;
+	vtkMRMLMarkupsCurveNode::ResamplePoints(originalPoints, interpolatedPoints, controlPointDistance, this->CurveClosed);
+
+	vtkPolyData* surfacePolydata = modelNode->GetPolyData();
+	if(!surfacePolydata)
+	  {
+	  vtkErrorMacro("vtkMRMLMarkupsCurveNode::ResampleCurveSurface failed: Constraint surface polydata is not valid");
+	  }
+	vtkNew<vtkPointLocator> pointLocator;
```
lassoan: add a comment why we could not use static locator

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
lassoan: Don't we need to still snap the control point to the surface?

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
```diff
@@ -183,6 +194,182 @@ double vtkMRMLMarkupsCurveNode::GetCurveLengthBetweenStartEndPointsWorld(vtkIdTy
...
+	  double startNormal[3] = { 0.0 };
+	  normalVectorArray->GetTuple(pointIdStart, startNormal);
+	  vtkIdType pointIdEnd = pointLocator->FindClosestPoint(segmentEndPoint);
+	  double endNormal[3] = { 0.0 };
+	  normalVectorArray->GetTuple(pointIdStart, endNormal);
+	  double rayDirection[3] = { 0.0 };
+	  rayDirection[0] = (startNormal[0] + endNormal[0]) / 2.0;
+	  rayDirection[1] = (startNormal[1] + endNormal[1]) / 2.0;
+	  rayDirection[2] = (startNormal[2] + endNormal[2]) / 2.0;
+	  pointNormalArray->InsertNextPoint(rayDirection);
```
lassoan: rayDirection is not a unit vector anymore which causes you problems later. Normalize it carefully, as it may even be null-vector (e.g., for [1,0,0] and [-1,0,0,]).

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
lassoan: if search radius is not an absolute value then indicate it by including "factor" or "tolerance" in the name

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
```diff
@@ -183,6 +194,182 @@ double vtkMRMLMarkupsCurveNode::GetCurveLengthBetweenStartEndPointsWorld(vtkIdTy
...
+	  if (segmentStartIndex == 0)
+		{
+		originalControlPoints->GetPoint(segmentStartIndex + 1, segmentEndPoint);
+		}
+	  else if (segmentStartIndex == (originalControlPoints->GetNumberOfPoints() - 1))
+	    {
+	    originalControlPoints->GetPoint(segmentStartIndex - 1, segmentEndPoint);
+	    }
+	  else
+		{
```
lassoan: indentation

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
```diff
@@ -183,6 +194,182 @@ double vtkMRMLMarkupsCurveNode::GetCurveLengthBetweenStartEndPointsWorld(vtkIdTy
...
+		  segmentEndPoint[2] = segmentEndPoint2[2];
+		  }
+		}
+	  // estimate normal vector from control points on either side of projected point
+	  vtkIdType pointIdStart = pointLocator->FindClosestPoint(segmentStartPoint);
+	  double startNormal[3] = { 0.0 };
+	  normalVectorArray->GetTuple(pointIdStart, startNormal);
+	  vtkIdType pointIdEnd = pointLocator->FindClosestPoint(segmentEndPoint);
+	  double endNormal[3] = { 0.0 };
+	  normalVectorArray->GetTuple(pointIdStart, endNormal);
```
lassoan: shouldn't this be pointIdEnd?

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
```diff
@@ -183,6 +194,182 @@ double vtkMRMLMarkupsCurveNode::GetCurveLengthBetweenStartEndPointsWorld(vtkIdTy
...
+	  double rayDirection[3] = { 0.0 };
+	  rayDirection[0] = (startNormal[0] + endNormal[0]) / 2.0;
```
lassoan: since we compute distance to start and end points anyway, why don't we use those to weight the average (instead of always using 0.5 weight for both)?

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
```diff
@@ -183,6 +194,182 @@ double vtkMRMLMarkupsCurveNode::GetCurveLengthBetweenStartEndPointsWorld(vtkIdTy
...
+	vtkNew<vtkPoints> interpolatedPoints;
+	vtkMRMLMarkupsCurveNode::ResamplePoints(originalPoints, interpolatedPoints, controlPointDistance, this->CurveClosed);
+
+	vtkPolyData* surfacePolydata = modelNode->GetPolyData();
+	if(!surfacePolydata)
+	  {
+	  vtkErrorMacro("vtkMRMLMarkupsCurveNode::ResampleCurveSurface failed: Constraint surface polydata is not valid");
+	  }
+	vtkNew<vtkPointLocator> pointLocator;
+	pointLocator->SetDataSet(surfacePolydata);
```
lassoan: can the vtkOBBTree also find points? If yes then we could avoid building two locators and just use vtkOBBTree in both methods.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
```diff
@@ -183,6 +194,182 @@ double vtkMRMLMarkupsCurveNode::GetCurveLengthBetweenStartEndPointsWorld(vtkIdTy
...
+  double exteriorPoint[3] = { 0.0 };
+
+  // Curve is expected to be close to surface. If projection distance is greater than 10% of image size
+  // it's likely due to a second external surface
+  double polydataBounds[6] = { 0.0 };
+  surfacePolydata->GetBounds(polydataBounds);
+  vtkBoundingBox modelBoundingBox;
+  modelBoundingBox.AddBounds(polydataBounds);
+  double polydataDiagonalLength = modelBoundingBox.GetDiagonalLength();
+  double rayLength = (.01*maximumSearchRadius)*sqrt(polydataDiagonalLength);
```
lassoan: If you mean that maximumSearchRadius is in percentage then include "percent" in the name, but even better, just don't multiply it by 0.01 to follow VTK conventions (in VTK, factors are in 0.0-1.0 range and not in 0.0-100.0). On the GUI it can be displayed as percentage.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
```diff
@@ -183,6 +194,182 @@ double vtkMRMLMarkupsCurveNode::GetCurveLengthBetweenStartEndPointsWorld(vtkIdTy
...
+
+  vtkNew<vtkPointLocator> pointLocator;
+  pointLocator->SetDataSet(surfacePolydata);
+  pointLocator->BuildLocator();
+
+  double originalPoint[3] = { 0.0 };
+  double rayDirection[3] = { 0.0 };
+  double exteriorPoint[3] = { 0.0 };
+
+  // Curve is expected to be close to surface. If projection distance is greater than 10% of image size
```
lassoan: It seems that the 10% is not hardcoded anymore so this comment should be updated.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
```diff
@@ -183,6 +194,182 @@ double vtkMRMLMarkupsCurveNode::GetCurveLengthBetweenStartEndPointsWorld(vtkIdTy
...
+	rayEndPoint[1] = originalPoint[1] + rayDirection[1] * rayLength;
+	rayEndPoint[2] = originalPoint[2] + rayDirection[2] * rayLength;
+
+	double t;
+	double firstIntersectionPoint[3] = { 0.0 };
+	double pcoords[3] = { 0.0 };
+	int subId;
+	vtkIdType cellId;
+	vtkNew <vtkGenericCell> cell;
+	int foundIntersection = surfaceObbTree->IntersectWithLine(rayEndPoint, originalPoint, tolerence, t, exteriorPoint, pcoords, subId, cellId, cell);
```
lassoan: tolerence typo

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.h
```diff
@@ -83,6 +84,15 @@ class  VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkMRMLMarkupsCurveNode : public vt
   /// \return sum of distances between the curve points, returns 0 in case of an error
   double GetCurveLengthBetweenStartEndPointsWorld(vtkIdType startCurvePointIndex, vtkIdType endCurvePointIndex);
 
+  /// Resample a curve with points constrained to surface
+  /// Projection to surface is constrained by maximumSearchRadius, specified as a percentage of the model diagonal distance
```
lassoan: specify a default value for maximumSearchRadius. Clarify valid range.
"Diagonal distance" -> "size of the model's bounding box diagonal in world coordinate system"

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.h
```diff
@@ -83,6 +84,15 @@ class  VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkMRMLMarkupsCurveNode : public vt
...
+  void ResampleCurveSurface(double controlPointDistance, vtkMRMLModelNode* node, double maximumSearchRadius);
+  
+  /// Constrain points to a specified model surface
+  /// Projection to surface is constrained by maximumSearchRadius, specified as a percentage of the model diagonal distance
```
lassoan: comments made on the other method apply here, too

Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
```diff
@@ -2403,6 +2450,61 @@ void qSlicerMarkupsModuleWidget::onResetToDefaultDisplayPropertiesPushButtonClic
   this->markupsLogic()->SetDisplayNodeToDefaults(displayNode);
 }
 
+//-----------------------------------------------------------------------------
+void qSlicerMarkupsModuleWidget::onapplyCurveResamplingPushButtonClicked()
```
lassoan: Apply capital A

Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
```diff
@@ -2403,6 +2450,61 @@ void qSlicerMarkupsModuleWidget::onResetToDefaultDisplayPropertiesPushButtonClic
...
+{
+  Q_D(qSlicerMarkupsModuleWidget);
+
+  double resampleRate = d->SelectNumberOfSamplePointsSpinBox->value();
```
lassoan: it is not really a rate, it is distance

Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
```diff
@@ -2403,6 +2450,61 @@ void qSlicerMarkupsModuleWidget::onResetToDefaultDisplayPropertiesPushButtonClic
...
+
+  vtkMRMLMarkupsCurveNode *markupsCurveNode = vtkMRMLMarkupsCurveNode::SafeDownCast(d->MRMLNodeComboBox_inputNodeSelector->currentNode());
+
+  if (!markupsCurveNode)
+    {
+	vtkMRMLMarkupsClosedCurveNode *markupsCurveNode = vtkMRMLMarkupsClosedCurveNode::SafeDownCast(d->MRMLNodeComboBox_inputNodeSelector->currentNode());
```
lassoan: use the markups curve node member variable instead of getting it again from the GUI

Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
```diff
@@ -2403,6 +2450,61 @@ void qSlicerMarkupsModuleWidget::onResetToDefaultDisplayPropertiesPushButtonClic
...
+	  {
+	  vtkMRMLMarkupsClosedCurveNode *outputNode = vtkMRMLMarkupsClosedCurveNode::SafeDownCast(d->MRMLNodeComboBox_outputNodeSelector->currentNode());
+	  if (!outputNode)
+		{
+		qDebug() << "No valid input curve. ";
+		return;
+		}
+
+	  }
+	if (outputNode->GetClassName() != markupsCurveNode->GetClassName())
```
lassoan: When the input node is selected then you can configure the output node selector to only show nodes of the same class.

Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
```diff
@@ -2403,6 +2450,61 @@ void qSlicerMarkupsModuleWidget::onResetToDefaultDisplayPropertiesPushButtonClic
...
+	  return;
+	  }
+	if (markupsCurveNode == outputNode)
+	  {
+	  outputNode = markupsCurveNode;
+	  }
+	else
+	  {
+	  vtkNew<vtkPoints> originalControlPoints;
+	  markupsCurveNode->GetControlPointPositionsWorld(originalControlPoints);
```
lassoan: There are many other properties to copy (labels, placement status, etc.). It could be better to use Copy() method.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
```diff
@@ -183,6 +194,182 @@ double vtkMRMLMarkupsCurveNode::GetCurveLengthBetweenStartEndPointsWorld(vtkIdTy
...
+      }
+	vtkNew<vtkPoints> originalControlPoints;
+	this->GetControlPointPositionsWorld(originalControlPoints);
+	vtkNew<vtkStringArray> originalLabels;
+	this->GetControlPointLabels(originalLabels);
+
+	vtkNew<vtkPoints> interpolatedPoints;
+	vtkMRMLMarkupsCurveNode::ResamplePoints(originalPoints, interpolatedPoints, controlPointDistance, this->CurveClosed);
+
+	vtkPolyData* surfacePolydata = modelNode->GetPolyData();
```
lassoan: You need to transform the model polydata to the world coordinate system (if parent transform node is set then get TransformToWorld transform and use it in vtkTransformFilter)

Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
smrolfe: The Copy() function was resulting in some unexpected behavior, like the two nodes being tied to the same visibility flag. I also tested this using the Python command line in some simple examples. For now I have copied over the labels and points.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
smrolfe: Since the assumption is that the control points are already snapped to the surface, in the case of a curve with a single point, I think it may make more sense to exit. I added an error message for this case.

Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
lassoan: Yes, Copy may copy more things than you need. Labels and points may be enough for now.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
lassoan: It's a pity that we cannot snap a single point, logging an error message is an acceptable limitation for now.

smrolfe:
Thanks, I have addressed all your comments and it's working well for the data sets I've tested. I'll squash the commits so we can move forward.

lassoan:
This was merged.

PR #1231, opened 2019-10-15
COMP: Fix python-dicom-requirements not installing

jamesobutler:
This fixes an issue with the `python-dicom-requirements` external project not having the specified python packages installed as reported in https://discourse.slicer.org/t/cannot-import-pydicom/8766. 

I don't understand why the current code does not work and why these changes fix the incorrect behavior. I would expect project dependencies to be handled correctly and not require that the dependency packages are included in the main file.  Hopefully someone can provide an explanation.


jamesobutler:
@jcfr @lassoan

lassoan:
Thank you. I'll try to test this and merge as is, if it fixes the problem. @jcfr can review later why the original implementation did not work.

lassoan:
@jamesobutler thank you very much, it works. I've merged it now.

lassoan:
It's all because of the poor name choice and subsequent rename of the pydicom package.

We have pydicom bundled, too. It's just most module developers did not update.

You can replace `import dicom` by `import pydicom as dicom` and things should work as normal (except a few backward-incompatible changes, such as rename of UID to uid - see details [here](https://github.com/pydicom/pydicom/commit/16f5ee721c8872948e629b0026a6864242765d18)).

fedorov:
Thank you for the clarification. I did not realize both versions were packaged.

jcfr:
> I did not realize both versions were packaged.

For reference, see http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=27955

pieper:
We had this on the Slicer5 roadmap already, so I cross-linked to this thread too: https://www.slicer.org/wiki/Documentation/Labs/Slicer5-roadmap#Additional_proposed_changes_to_be_discussed

PR #1232, opened 2019-10-16
ENH: Add insert control point xyz function

sjh26:
Adds functions to insert a control point using a vector (previously only available using ControlPoint structure)

Also has a bugfix for resetting the ID on a control point.  Previously, LastUsedControlPointNumber was no incremented fro each reset, and so if you reset multiple points that would each gave the same ID.

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
@@ -689,6 +689,27 @@ bool vtkMRMLMarkupsNode::InsertControlPoint(ControlPoint *controlPoint, int targ
 }
 
 //-----------------------------------------------------------
+bool vtkMRMLMarkupsNode::InsertControlPointWorld(int n, vtkVector3d pointWorld, std::string label)
+{
+  vtkVector3d point;
+  this->TransformPointFromWorld(pointWorld, point);
+  return this->InsertControlPoint(n, point, label);
+}
+
+
+//-----------------------------------------------------------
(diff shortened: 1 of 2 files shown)


Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
```diff
@@ -1414,7 +1435,9 @@ bool vtkMRMLMarkupsNode::ResetNthControlPointID(int n)
     {
     return false;
     }
+
   this->SetNthControlPointID(n, this->GenerateUniqueControlPointID());
+  this->LastUsedControlPointNumber++;
```
lassoan: This manual incrementing of LastUsedControlPointNumber is quite messy. I think it should be done in GenerateUniqueControlPointID() method.

jcfr: I also thought about this, but `LastUsedControlPointNumber` is not systematically incremented each time GenerateUniqueControlPointID is called ...

sjh26: This case is the only one where it (currently) isn't incremented?
https://github.com/Slicer/Slicer/blob/d3bd8dd0377fd437ad87f0cd6b288c57222571d3/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx#L482

Is this correct or a bug?

lassoan:
Merged (and cleaned up LastUsedControlPointNumber increment in a follow-up commit).

PR #1234, opened 2019-10-22
ENH: Add option to set Segment Editor threshold values from a local ROI

Sunderlandkyl:
The "Local histogram" section within the Segment Editor threshold effect allows users to select a brush and click+drag on the image to create an ROI

Available brushes are:
  * Box
  * Circle
  * Draw
  * Line

Values within the ROI are plotted on a histogram, and the upper and lower threshold values are set based on the maximum, minimum or average value from the histogram.

Clicking and dragging on the histogram will create a manually selected range that will be used to select the threshold values. Right-clicking will clear the manually selected range.

image

Requires https://github.com/commontk/CTK/pull/888

### DIFF
--- a/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorThresholdEffect.py
+++ b/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorThresholdEffect.py
@@ -31,7 +31,27 @@ class SegmentEditorThresholdEffect(AbstractScriptedSegmentEditorEffect):
     self.timer.connect('timeout()', self.preview)
 
     self.previewPipelines = {}
+    self.histogramPipeline = None
     self.setupPreviewDisplay()
+
+    # Histogram stencil setup
+    self.stencil = vtk.vtkPolyDataToImageStencil()
+
+    # Histogram reslice setup
+    self.reslice = vtk.vtkImageReslice()
+    self.reslice.AutoCropOutputOff()
+    self.reslice.SetOptimization(1)
+    self.reslice.SetOutputOrigin(0, 0, 0)
+    self.reslice.SetOutputSpacing(1, 1, 1)
+    self.reslice.SetOutputDimensionality(3)
+    self.reslice.GenerateStencilOutputOn()
+
+    self.imageAccumulate = vtk.vtkImageAccumulate()
+    self.imageAccumulate.SetInputConnection(0, self.reslice.GetOutputPort())
+    self.imageAccumulate.SetInputConnection(1, self.stencil.GetOutputPort())
+
+    self.selectionStartPosition = None
+    self.selectionEndPosition = None
 
   def clone(self):
     import qSlicerSegmentationsEditorEffectsPythonQt as effects
@@ -67,6 +87,7 @@ class SegmentEditorThresholdEffect(AbstractScriptedSegmentEditorEffect):
 
     # Clear preview pipeline and stop timer
     self.clearPreviewDisplay()
+    self.clearHistogramDisplay()
     self.timer.stop()
 
   def setCurrentSegmentTransparent(self):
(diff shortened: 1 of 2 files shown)


Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorThresholdEffect.py
```diff
@@ -444,6 +633,206 @@ def preview(self):
...
+    elif eventId == vtk.vtkCommand.LeftButtonReleaseEvent:
+      self.histogramPipeline.state = HISTOGRAM_STATE_PLACED
+      abortEvent = True
+    elif eventId == vtk.vtkCommand.MouseMoveEvent:
+      if self.histogramPipeline.state == HISTOGRAM_STATE_MOVING:
+        self.histogramPipeline.addPoint(ras)
+        self.updateHistogram()
+        abortEvent = True
+    else:
+      pass
```
lassoan: not needed

Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorThresholdEffect.py
```diff
@@ -444,6 +633,206 @@ def preview(self):
...
+    brushExtent = [0, -1, 0, -1, 0, -1]
+    for i in range(3):
+      brushExtent[2*i] = vtk.vtkMath.Floor(brushBounds[2*i])
+      brushExtent[2*i+1] = vtk.vtkMath.Ceil(brushBounds[2*i+1])
+    if brushExtent[0] > brushExtent[1] or brushExtent[2] > brushExtent[3] or brushExtent[4] > brushExtent[5]:
+      self.histogramFunction.RemoveAllPoints()
+      return
+
+    sliceLogic = self.histogramPipeline.sliceWidget.sliceLogic()
+    backgroundLogic = sliceLogic.GetBackgroundLayer()
```
lassoan: Usually the master volume is the background volume but the user can change that. Maybe we could choose between foreground/background depending on which one is the master volume and if neither then log a warning and use the more visible one.

Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorThresholdEffect.py
```diff
@@ -444,6 +633,206 @@ def preview(self):
...
+    self.averageFunction.RemoveAllPoints()
+    self.averageFunction.AddPoint(average - epsilon, 0.0)
+    self.averageFunction.AddPoint(average, 1.0)
+    self.averageFunction.AddPoint(average + epsilon, 0.0)
+    self.averageFunction.AdjustRange(scalarRange)
+
+    minimumThreshold = lower
+    maximumThreshold = upper
+
+    if self.histogramLowerThresholdMinimumButton.checked:
```
lassoan: it would be a bit nicer to use parameter value in self.scriptedEffect node instead of GUI button state

Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorThresholdEffect.py
```diff
@@ -482,6 +871,251 @@ def __init__(self):
...
+
+    self.rasPoints = vtk.vtkPoints()
+    self.polyData = vtk.vtkPolyData()
+    self.polyData.SetPoints(self.rasPoints)
+
+    lines = vtk.vtkCellArray()
+    self.polyData.SetLines(lines)
+    idArray = lines.GetData()
+    idArray.Reset()
+    idArray.InsertNextTuple1(0)
```
lassoan: why is this needed?

lassoan (commented):
Thank you, this is a great feature! Please implement saving/restoring of user choices (ROI shape, lower/upper value selection) on the GUI to/from parameter node. I've also added a few minor comments.

Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorThresholdEffect.py
Sunderlandkyl: Sure. Currently the background is used for the threshold preview. Should that be changed as well?

lassoan: Yes, it would be nice to use the same logic for both.

lassoan:
Thank you, merged!

PR #1235, opened 2019-10-23
BUG: Fix segments model update when multiple segments are removed

Sunderlandkyl:
When multiple segments are removed within a StartModify/EndModify, not all removed segments were removed from the table.
This was because the onSegmentRemoved iterated over the rows from 0 to rowCount. The index would be invalidated if any rows were removed.
Fixed by iterating through the rows in reverse.


lassoan:
Thank you, merged!

PR #1236, opened 2019-10-24
ENH: Represent all split segments from island effect as one layer

Sunderlandkyl:
Splitting islands to individual segments previously used vtkSlicerSegmentationsModuleLogic::ImportLabelmapToSegmentationNode to create the segments.

This commit:
* Modifies segments split with the island effect using modifySegmentByLabelmap for all segments rather than ImportLabelmapToSegmentationNode.
* Improves two implementations of ImportLabelmapToSegmentationNode to import segments as a single layer from vtkOrientedImageData.


lassoan:
Looks good to me, thank you.

What is the advantage of importing each segment by modifySegmentByLabelmap instead of using one ImportLabelmapToSegmentationNode call?

Sunderlandkyl:
Because in some cases you may end up with 2 layers instead of just one in the case where the segment that you are splitting is already on the same layer as other segments.

ImportLabelmapToSegmentationNode would create a new layer from the input labelmap, which I think is the desired behavior.

This way, the number of layers should never increase.

lassoan:
Thank you, committed r28566.

PR #1237, opened 2019-10-25
BUG: Fix missing labels in closed surface to binary labelmap conversion

Sunderlandkyl:
When converting closed surface segments that have a label value > 255, the segments would not be in the labelmaps representations post-conversion.
This was because the labelmaps generated by vtkClosedSurfaceToBinaryLabelmapConversionRule are of type UNSIGNED_CHAR, and the segment label value was used for the stencil background value.
In the case where the label value exceeded 255, this meant that the labelmap would be missing from the output.

Fixed by setting the label value for binary labelmap created from closed surface to 1.
This is fine, since the PostConvert step will collapse the labelmaps and assign new label values anyway.

### DIFF
--- a/Libs/vtkSegmentationCore/vtkClosedSurfaceToBinaryLabelmapConversionRule.cxx
+++ b/Libs/vtkSegmentationCore/vtkClosedSurfaceToBinaryLabelmapConversionRule.cxx
@@ -46,6 +46,8 @@
 // STD includes
 #include <sstream>
 
+int DEFAULT_LABEL_VALUE = 1;
+
 //----------------------------------------------------------------------------
 vtkSegmentationConverterRuleNewMacro(vtkClosedSurfaceToBinaryLabelmapConversionRule);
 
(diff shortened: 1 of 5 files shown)


Sunderlandkyl:
Added method documentation. This is the same approach/signature that is used in vtkOrientedImageDataResample::MergeImage(). If it would be better, I can change the methods to accept int* for extent.

lassoan:
Is there any difference in Python wrapping of int* (maybe with some wrapping hint to indicate that it expects vector of 6) and int[6]?

Sunderlandkyl:
Probably [[vtk::sizehint(arg, expression)]] would be required.

Sunderlandkyl:
Function signatures modified:
```
  static bool IsLabelInMask(vtkOrientedImageData* binaryLabelmap, vtkOrientedImageData* mask,
    int* extent = nullptr, int maskThreshold=0) VTK_SIZEHINT(extent, 6);
```

Python wrapped successfully.

Sunderlandkyl:
Regarding the possibility of preserving the label value, I believe that it should be possible to implement that system.

I'll look into implementing that and make another pull request.

Sunderlandkyl:
@lassoan 

Signatures now:
```
  static bool IsLabelInMask(vtkOrientedImageData* binaryLabelmap, vtkOrientedImageData* mask, int extent[6]=nullptr, int maskThreshold=0);
```

lassoan:
Thank you! Merged.

PR #1238, opened 2019-10-25
ENH: Support reading segmentations from NIFTI files

Sunderlandkyl:
Previously, segmentation files were read using vtkTeemNrrdReader.
vtkMRMLSegmentationStorage node can now read both NIFTI and NRRD file formats using vtkITKArchetypeImageSeriesVectorReaderFile.


Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
```diff
@@ -138,6 +140,7 @@ void vtkMRMLSegmentationStorageNode::InitializeSupportedReadFileTypes()
   this->SupportedReadFileTypes->InsertNextValue("Segmentation (.seg.vtm)");
   this->SupportedReadFileTypes->InsertNextValue("Segmentation (.nrrd)");
   this->SupportedReadFileTypes->InsertNextValue("Segmentation (.vtm)");
+  this->SupportedReadFileTypes->InsertNextValue("Segmentation (.nii.gz)");
```
lassoan: Add .hdr (nifti format with separate header and pixel data) and .nii (uncompressed single-file nifti).

Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
```diff
@@ -163,6 +166,7 @@ void vtkMRMLSegmentationStorageNode::InitializeSupportedWriteFileTypes()
     {
     this->SupportedWriteFileTypes->InsertNextValue("Segmentation (.seg.nrrd)");
     this->SupportedWriteFileTypes->InsertNextValue("Segmentation (.nrrd)");
+    this->SupportedWriteFileTypes->InsertNextValue("Segmentation (.nii.gz)");
```
lassoan: Can we write segmentation to nifti format? If we can only write labelmap without segmentation metdata then it is better not to list this format among supported write formats (it is more of a lossy export format then).

lassoan (commented):
Thank you, it mostly looks good.

It would be also nice to add a test that reads data from nifti file (e.g., test import from a simple 3D labelmap stored in nifti format) and I've added a couple of comments.

Sunderlandkyl:
Changes made.
Added vtkMRMLSegmentationStorageNodeTest1 to test loading of 3 types of segmentations.
- ITK-SNAP nii.gz
- Slicer pre shared labelmap segmentation
- Slicer post shared labelmap segmentation

lassoan:
Thank you! Merged.

PR #1240, opened 2019-10-29
BUG: Fix segment binary labelmap outline thickness

Sunderlandkyl:
The code for setting the labelmap outline thickness was placed in the polydata section of UpdateSegmentPipelines.
Moved relevant code to the correct location.


lassoan:
Thank you. Committed r28573.

PR #1241, opened 2019-10-30
BUG: Fix bugs in segment editor paint effect

Sunderlandkyl:
* "No input data" message when using paint effect
  When rendering the brush feedback actor with no points in the pipeline, there could be a large number of "No input data" error messages that would pollute the log.
  The pipeline is now cleared before the points are removed to prevent this message from occurring.

* Expansion of segment extent beyond reference geometry
  Normally when editing a segment, the labelmap is padded to contain the modifier extent, and then clipped to the effective extent.
  If vtkOrientedImageDataResample::MergeImage() indicated that the segment was not modified, then the segment would not be reduced to its effective extent, even if it had been padded.
  This is fixed by reducing the segment to its effective extent regardless of whether or not MergeImage indicates that the segment was modified.


lassoan:
Thank you! Merged r28578.

PR #1243, opened 2019-10-31
ENH: Ignore Segment Editor interactor events during touchscreen gesture

Sunderlandkyl:
Pinch/Pan/Rotate can have unwanted effects when used while an editor effect is active.
This commit changes qMRMLSegmentEditorWidget to only react to interaction events while a gesture is not currently active.

### DIFF
--- a/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
+++ b/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
@@ -68,9 +68,13 @@
 #include <vtkSlicerApplicationLogic.h>
 
 // MRML includes
+#include <vtkMRMLCameraDisplayableManager.h>
+#include <vtkMRMLCameraWidget.h>
+#include <vtkMRMLCrosshairDisplayableManager.h>
 #include <vtkMRMLLabelMapVolumeNode.h>
 #include <vtkMRMLSelectionNode.h>
 #include <vtkMRMLScene.h>
+#include <vtkMRMLSliceIntersectionWidget.h>
 #include <vtkMRMLSliceNode.h>
 #include <vtkMRMLTransformNode.h>
 #include <vtkMRMLViewNode.h>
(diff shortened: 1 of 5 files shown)


Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
```diff
@@ -2757,12 +2767,24 @@ void qMRMLSegmentEditorWidget::processEvents(vtkObject* caller,
   vtkMRMLAbstractViewNode* callerViewNode = vtkMRMLAbstractViewNode::SafeDownCast(caller);
   if (callerInteractor)
     {
-    bool abortEvent = activeEffect->processInteractionEvents(callerInteractor, eid, viewWidget);
-    if (abortEvent)
+    if (eid == vtkCommand::StartPanEvent || eid == vtkCommand::StartPinchEvent || eid == vtkCommand::StartRotateEvent)
```
lassoan: Is it possible that two gestures are active at once? If yes then a single Boolean flag may not be enough to robustly determine gesture state.

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
```diff
@@ -2757,12 +2767,24 @@ void qMRMLSegmentEditorWidget::processEvents(vtkObject* caller,
...
+      {
+      self->d_ptr->GestureInProgress = true;
+      }
+    if (eid == vtkCommand::EndPanEvent || eid == vtkCommand::EndPinchEvent || eid == vtkCommand::EndRotateEvent)
+      {
+      self->d_ptr->GestureInProgress = false;
```
lassoan: Would it be possible to get the gesture state from the interactor? It could probably detect it more robustly and would simplify this class.

Sunderlandkyl: Currently no. It would require some changes in VTK.

lassoan: Do you mean that we would need to add a method to vtkRenderWindowInteractor to get current value of CurrentGesture, or we would need something more complicated? If it's a simple change then it could be integrated along your other pending gesture-related VTK enhancements.

Sunderlandkyl: I don't think I have any gesture-related VTK enhancements currently, just some tablet related ones.

The gestures received by QVTKInteractorAdapter are not processed through vtkRenderWindowInteractor, just passed along, so the solution would need to be more complicated.

Ex. Pan Gesture
https://github.com/Kitware/VTK/blob/35ee72189870f414dc72faa08c4610dea666d2ca/GUISupport/Qt/QVTKInteractorAdapter.cxx#L454-L475

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
Sunderlandkyl: Changed the gesture flag to an integer count of gestures in progress.

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
lassoan: Is CurrentGesture member variable usable or that is for other gestures?
https://github.com/Kitware/VTK/blob/master/Rendering/Core/vtkRenderWindowInteractor.cxx#L628

Sunderlandkyl: Same type of gesture, but CurrentGesture is never used by events from Qt, only for ones recognized at the VTK level.

With some changes in VTK it should be possible to change the way that Qt events are handled so that the gesture is stored in CurrentGesture.

Sunderlandkyl:
Using the gesture count, there seems to be mismatch between start and end gesture events. Debugging...

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
lassoan: For this single feature it seems like an overkill to implement significant changes in VTK. However, I'm not sure if we can reliably detect gesture state in a module. We don't always observe events (we add/remove observers as we enter/exit the module), anybody can abort events, etc., so there would be a risk that we miss a single End...Event and from that point gesture state would stuck and all interaction events would be ignored.

If it is hard to implement it in VTK then maybe somewhere in Slicer at higher level classes we could implement it more robustly.

Sunderlandkyl: Does it make sense then to implement a system like the vtkMRMLSliceIntersectionWidget for state management?

lassoan: Do you mean that vtkMRMLSliceIntersectionWidget already has an internal state and so we should get the gesture state from that widget?

Sunderlandkyl: No, I mean that we should implement a separate state in the qMRMLSegmentEditorWidget, but you bring up a good point. Would it make sense to get the gesture state from the vtkMRMLSliceIntersectionWidget?

I seem to recall that we tried to implement a count of the number of gestures in progress, but that for some reason the number of "start gesture" != number of "end gesture" at the Qt level.

lassoan:
Thank you. Merged in r28611.

I'm wondering what happens if the users starts painting and then transitions to a touch gesture. Is the paint stroke properly finished or cancelled?

Sunderlandkyl:
Just tested, The paint stroke is not terminated or cancelled.

PR #1244, opened 2019-11-01
ENH: Center paint effect cylinder brush at mouse location

Sunderlandkyl:
Previously the center of the vtkCylinderSource was offset from the mouse position.
This also caused issues with painting such as not being able to paint a single pixel in certain views, and not being able to paint at all when using a small diameter brush.
Both of these issues were fixed by centering the the brush model at the mouse location.

This commit also implements the BrushPixelMode, although it is not currently accessible to users.


lassoan:
Coudl we do pixel paint automatically when brush size is so small that no voxels would be painted?

Sunderlandkyl:
We could, but it turns out that it is not necessary.
Now that the center of the cylinder is aligned with the mouse, it does not matter how small the brush size is. It will always paint at least one voxel.

lassoan:
Perfect, thank you. I'll test it a bit and merge.

lassoan:
This is awesome, works perfectly. Merged in r28584. Thank you Kyle!

PR #1245, opened 2019-11-02
COMP: new python build needs openssl 1.1.0 and up.

gaoyi:
This commit uses openssl 1.1.1d, which triggers the upgrading curl to 7.66

This finally solved my problem for days.


jcfr:
Thanks again for the contribution.

Any reason this was closed, is there a related discourse discussion ?

pieper:
Here's the discussion:

https://discourse.slicer.org/t/cant-start-slicer-on-linux-on-one-machine/8175/51

I agree with you @jcfr I'm not sure why this was closed.  Looks like the solution @gaoyi said worked for him.

PR #1247, opened 2019-11-03
Remove legacy dicom python package

jamesobutler:
Both the legacy [dicom](https://pypi.org/project/dicom/) and the newer [pydicom](https://pypi.org/project/pydicom/) packages have both been bundled in Slicer since February 8th, 2019 (see 99142bce0076570388373f5746d022c5b7fa33e7).  DICOMPatcher in Slicer core recently began using pydicom instead of dicom in 65b11a04e324f71709cbf04cef4a8e9d2695c0f2.

This PR removes the legacy dicom package as proposed in changes to be included in Slicer 5.
https://www.slicer.org/wiki/Documentation/Labs/Slicer5-roadmap#Additional_proposed_changes_to_be_discussed

### DIFF
--- a/Modules/Scripted/DICOMPatcher/DICOMPatcher.py
+++ b/Modules/Scripted/DICOMPatcher/DICOMPatcher.py
@@ -196,7 +196,7 @@ class ForceSamePatientNameIdInEachDirectory(DICOMPatcherRule):
     self.firstFileInDirectory = True
     self.patientIndex += 1
   def processDataSet(self, ds):
-    import pydicom as dicom
+    import pydicom
     if self.firstFileInDirectory:
       # Get patient name and ID for this folder and save it
       self.firstFileInDirectory = False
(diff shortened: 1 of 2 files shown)


lassoan:
It was not obvious for developers that an obsolete pydicom version was used. We should give at least a couple of months to developers to update their code and then we can remove the old package.

@jamesobutler Could you add a note about this to the migration guide? https://www.slicer.org/wiki/Documentation/Nightly/Developers/Tutorials/MigrationGuide

jamesobutler:
The main motivation that I submitted this change now was based on previous knowledge that it would be removed as part of Slicer 5 which was scheduled to be released in the fall before RSNA.  This was from an old [comment](https://discourse.slicer.org/t/2019-04-23-hangout/6571/2?u=jamesobutler) on the forum so things might've changed since then.

Yes, I can add a note about it to the migration guide, but had to request an account just now.

lassoan:
Yes, that was the plan. I'm not sure if we can make it. There is no strict requirement for RSNA (no large group tutorial this year), but we should have a new stable release at least once a year. We'll discuss this at the next Slicer developer hangout.

jamesobutler:
Note added to the migration page https://www.slicer.org/wiki/Documentation/Nightly/Developers/Tutorials/MigrationGuide/SlicerExtension#Slicer_5.0:_Python_packaging_dicom_to_pydicom

jcfr:
Closing. Changes integrated in [r28603](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28603) and [28604](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28604)

PR #1250, opened 2019-11-07
BUG: Correct calculation of surface normals during curve resampling 

smrolfe:
This commit fixes a bug which causes the application to crash when a curve is resampled and constrained to a surface without an existing normal vector array. 

The current implementation has a bug at line 262  where the vtkDataArray is re-instantiated inside a conditional. This also occurs when the vtkDataArray from 254 is referenced. This fix adds a vtkFloatArray and uses DeepCopy to copy the contents of the vtkDataArray to the vtkFloatArray.


lassoan:
Fix committed in r28612.

PR #1251, opened 2019-11-08
ENH: Improve observer management in vtkSegmentation

Sunderlandkyl:
Previously observers were only added or removed from master representations when AddSegment, RemoveSegment or SetMasterRepresentationModifiedEnabled was called.
This meant that observers were not handled if vtkSegment::AddRepresentation was called, resulting in orphaned or missing observers.

This commit collects all of the AddObserver/RemoveObserver calls for the master representation in a single function UpdateMasterRepresentationObservers.
UpdateMasterRepresentationObservers caches a list of the master representations data objects that are monitored and compares against the list of master representations in the segmentation, and then adds or removes observers as needed.


lassoan:
Thank you. Good catch and fix. Merged in r28610.

PR #1253, opened 2019-11-08
ENH: Enable right-click menu for fiducials

cpinter:
Right-clicking markups in views shows a context menu provided by subject hierarchy. The infrastructure to use the SH in-view context menu for other node types is also established.
- Sink MenuEvent event from markups display node to display node base class
- Display node MenuEvent handled in Subject hierarchy plugin logic. It shows a third type of SH menu called View menu, which offers actions that are applicable in the views
- First plugin to support this is Markups that offers a menu to rename fiducial


Modules/Loadable/Markups/MRML/CMakeLists.txt
```diff
@@ -29,6 +29,8 @@ set(${KIT}_TARGET_LIBRARIES
   ${MRML_LIBRARIES}
   )
 
+set (${KIT}_INCLUDE_DIRS ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_BINARY_DIR} CACHE INTERNAL "" FORCE)
```
lassoan: Why is this needed? These include dirs should be added to cache automatically (you can see in Slicer-build\CMakeCache.txt that there are many ...ModuleLogic_INCLUDE_DIRS variables and they are not set using such code).

Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginLogic.cxx
```diff
@@ -64,11 +75,15 @@ qSlicerSubjectHierarchyPluginLogicPrivate::qSlicerSubjectHierarchyPluginLogicPri
   // Register vtkIdType for use in python for subject hierarchy item IDs
   qRegisterMetaType<vtkIdType>("vtkIdType");
   //qRegisterMetaType<QList<vtkIdType> >("QList<vtkIdType>"); //TODO: Allows returning it but cannot be used (e.g. pluginHandler->currentItems())
+
+  this->ViewMenu = new QMenu(); //TODO: Parent?
```
lassoan: Looks good to me as is (we don't set a parent, so we have to delete it)

Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginLogic.cxx
```diff
@@ -64,11 +75,15 @@ qSlicerSubjectHierarchyPluginLogicPrivate::qSlicerSubjectHierarchyPluginLogicPri
...
 }
 
 //-----------------------------------------------------------------------------
 qSlicerSubjectHierarchyPluginLogicPrivate::~qSlicerSubjectHierarchyPluginLogicPrivate()
-= default;
+{
+  this->ViewMenu->deleteLater(); //TODO: OK?
```
lassoan: Looks good to me. Maybe set this->ViewMenu to nullptr at the end to make the behavior in case of an error (object double-delete, use-after-delete) more predictable.

Modules/Loadable/Markups/MRML/CMakeLists.txt
cpinter: Because the markup node header could not be included in the subject hierarchy plugin cxx without this.

Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginLogic.cxx
cpinter: Thanks!

cpinter:
I like all the ideas about Edit properties and also the special markup types. Since this infrastructure will most probably remain the same, I think we could go ahead and integrate this. Once we have a concrete plan to extend it, I'm happy to help.

Modules/Loadable/Markups/MRML/CMakeLists.txt
lassoan: It is odd that this is needed, because even without this, my Slicer-build/CMakeCache.txt contains `vtkSlicerMarkupsModuleMRML_INCLUDE_DIRS:INTERNAL=C:/D/S4/Modules/Loadable/Markups/MRML;C:/D/S4R/Slicer-build/Modules/Loadable/Markups/MRML`.

It does not hurt to have this, so if you want you can go on and integrate as is, but it would be nice if you could try a clean build without this line and see if it fails.

lassoan:
> Once we have a concrete plan to extend it, I'm happy to help.

I would recommend to have at least these actions:
- Rename point
- Delete point
- Toggle select point
- Edit properties...

Could you add them? It does not have to be in this pull request.

It would be nice to have something like "Show in subject hierarchy tree", which would switch to Data module / Subject Hierarchy tab and select the node.

cpinter:
OK I can add these easily. For Edit properties we can use the current infrastructure right?

Modules/Loadable/Markups/MRML/CMakeLists.txt
cpinter: I started a clean build...

lassoan:
> OK I can add these easily. For Edit properties we can use the current infrastructure right?

yes

Modules/Loadable/Markups/MRML/CMakeLists.txt
cpinter: It builds. I guess then probably there are several lines of the same kind in Slicer. In any case I removed this one.
@lassoan do you have any ideas what I can do to make it build without a clean build in general? Like re-run CMake or clean build just a few projects etc.

lassoan: I don't think anything can replicate the behavior of a clean build.

I only saw this kind of kit include dir caching line in Subject Hierarchy plugin and maybe one more place. They seemed to be special enough that I did not check if they were still needed, but you are right that those may be removed, too.

jcfr: This is probably an oversight. I suggest to remove it and if there are issues later one, I will have a look.

cpinter: For development it not acceptable having to do a complete 3-5 hour clean build when making such changes in CMake. There must be some kind of way to have a successful build without either adding this line or doing a full clean build...

lassoan: There is really no other way if you want to be 100% sure that you test a clean build without actually doing a clean build. If you are sure that your changes only affect Slicer then you may be able to delete just the Slicer binary folder and then VTK, CTK, etc., may not need to be rebuilt.

I have additional test builds set up for things like this, so I can continue my work, while another build is running in the background.

cpinter:
This PR has been integrated in https://github.com/Slicer/Slicer/commit/771bc70d5a2c706d3f37a7ad49f73d12b2167e75

I will push a second PR with the following changes:
* More actions for markpus (and not just for fiducial types): Rename point, Delete point, Toggle select point, Edit properties...
* A whitelist in plugin logic for the actions so that it can be esaily customized

cpinter:
For the record, the follow-up PR: https://github.com/Slicer/Slicer/pull/1262

PR #1256, opened 2019-11-11
COMP: Update to latest Landmark Registration

jamesobutler:
@pieper Here is the corresponding update in the Slicer repo.

This updates Landmark Registration to include a commit that fixes a [failing test](http://slicer.cdash.org/testDetails.php?test=9824573&build=1744757).


lassoan:
Thank you very much. Merged in r28599. It is so great that you were able to take care of these failed tests. It'll make the dashboard clean and make it easy to see any regressions.

PR #1258, opened 2019-11-12
BUG: Fix a set of failing tests due to new SH node as a member variable of MRML scene

jamesobutler:
This PR aims to fix a set of 5 failing tests that began to fail on Sept. 16th due to changes made in aa94228a125cac9e2e128020c08d9a63f2d9bebb. In that commit a subject hierarchy node was added as a member variable of the scene which caused a set of tests to fail all the in the same way of expecting one less node than actual.

The set of tests are:
- vtkMRMLNodeTest1
- vtkMRMLSceneAddSingletonTest
- vtkMRMLSceneImportIDConflictTest
- vtkMRMLSceneImportIDModelHierarchyConflictTest
- vtkMRMLSceneImportIDModelHierarchyParentIDConflictTest

@adamrankin it appears @cpinter alerted you of these failing tests in https://github.com/Slicer/Slicer/pull/1210#issuecomment-532265182, but it looks like this was something that was accidentally forgotten about. Can you take a look and confirm these changes to these tests match the changes you committed?

### DIFF
--- a/Libs/MRML/Core/Testing/vtkMRMLSceneImportIDModelHierarchyConflictTest.cxx
+++ b/Libs/MRML/Core/Testing/vtkMRMLSceneImportIDModelHierarchyConflictTest.cxx
@@ -139,11 +139,12 @@ int ImportIDModelHierarchyConflictTest()
   // and added nodes. New IDs are set by Import to the added nodes.
   // The node ids in the scene after a proper import should be
 
-  scene->Import();
+  scene->Import();  // adds Subject Hierarchy Node
 
   // At this point the scene should be:
   //
   //  Scene
+  //    |---- vtkMRMLSubjectHierarchyNode1
   //    |---- vtkMRMLModelNode1  (valid polydata)
   //    |          |-- ref [displayNodeRef] to vtkMRMLModelDisplayNode1
   //    |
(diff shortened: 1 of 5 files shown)


jcfr:
Thanks for working on this :pray:

Libs/MRML/Core/Testing/vtkMRMLSceneImportIDModelHierarchyConflictTest.cxx
```diff
@@ -166,7 +166,7 @@ int ImportIDModelHierarchyConflictTest()
   //               |-- ref [displayNodeID] to vtkMRMLModelDisplayNode4
   //               |-- ref [associatedNodeRef] to vtkMRMLModelNode2
```
jcfr: Should comment like the one above be updated ?

adamrankin:
Sorry about that, I did overlook the tests.

Libs/MRML/Core/Testing/vtkMRMLSceneImportIDModelHierarchyConflictTest.cxx
jamesobutler: Pushed update that adds comments

cpinter:
Thank you @jamesobutler !

jamesobutler:
@jcfr or another maintainer this PR should be good to go. Hopefully this can be merged before tonight’s nightly build.

lassoan:
Thanks a lot! Merged in r28609.

PR #1260, opened 2019-11-14
COMP: update openssl and curl for python to work correctly

gaoyi:
This change solves my recent building and running issues.


pieper:
```
make[2]: *** No rule to make target `/s/curl-install/lib/libcurl.a', needed by `bin/libRemoteIO.dylib'.  Stop.
```

it seems that for debug builds they renamed the library by adding `-d`

This works around the issue: `cp ./curl-install/lib/libcurl-d.a ./curl-install/lib/libcurl.a`

Not sure how best to fix that.


As an aside, at first, when I tried building this on my mac I got this error below, but this went away after deleting all the `curl*` in my superbuild directory.

```
[ 77%] Performing build step for 'curl'
[  0%] Building C object lib/CMakeFiles/libcurl.dir/warnless.c.o
/s/curl/lib/warnless.c:101:4: error: "SIZEOF_CURL_OFF_T not defined"
#  error "SIZEOF_CURL_OFF_T not defined"
   ^
/s/curl/lib/warnless.c:192:40: error: use of undeclared identifier 'CURL_MASK_SCOFFT'
  return (curl_off_t)(uznum & (size_t) CURL_MASK_SCOFFT);
                                       ^
2 errors generated.
make[5]: *** [lib/CMakeFiles/libcurl.dir/warnless.c.o] Error 1
make[4]: *** [lib/CMakeFiles/libcurl.dir/all] Error 2
make[3]: *** [all] Error 2
make[2]: *** [curl-prefix/src/curl-stamp/curl-build] Error 2
make[1]: *** [CMakeFiles/curl.dir/all] Error 2
make: *** [all] Error 2
```

PR #1261, opened 2019-11-15
BUG: Block MasterRepresentationModified in CollapseBinaryLabelmaps

Sunderlandkyl:
Fixes failing SegmentStatistics test.
If CollapseBinaryLabelmaps is called at the vtkSegmentation level, it can invalidate non master representations, even though the actual content of the segment has not been modified.
This commit prevents MasterRepresentationModified from being called when forceToSingleLayer is false.


Sunderlandkyl:
@lassoan

lassoan:
Looks good to me, I'll merge it now.

PR #1262, opened 2019-11-16
Improved view right-click menu

cpinter:
- More actions for markpus (and not just for fiducial types): Rename point, Delete point, Toggle select point, Edit properties...
- A whitelist in plugin logic for the actions so that it can be easily customized

### DIFF
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginLogic.h
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginLogic.h
@@ -71,8 +71,16 @@ public:
 
   /// Register subject hierarchy core plugins
   /// Note: Registering plugins provided by other modules is the responsibility
-  ///   of the module!
+  ///       of the module!
   void registerCorePlugins();
+
+  /// Get all view menu actions available
+  /// \return List of object names of the view menu actions
+  Q_INVOKABLE QStringList allViewMenuActions();
+  /// Set desired set of view menu actions
+  /// \param actionObjectNames List of view menu actions to consider. Only actions included here by object name
+  ///        are shown in the menu if they are accepted by the owner plugin. The order set here is used.
+  Q_INVOKABLE void setViewMenuActionWhitelist(QStringList actionObjectNames);
 
 protected:
   /// Add observations for node that was added to subject hierarchy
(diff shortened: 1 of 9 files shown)


cpinter:
This is how the whitelist can be used from python:
```
ph = slicer.qSlicerSubjectHierarchyPluginHandler.instance()
pl=ph.pluginLogic()
ma=pl.allViewMenuActions()
ma
('RenamePointAction', 'DeletePointAction', 'ToggleSelectPointAction', 'EditPropertiesAction')
m=['RenamePointAction']
pl.setViewMenuActionWhitelist(m)
```
After running this code snippet, only the rename action of the four available actions will be shown when right-clicking on a markup.

Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginLogic.h
```diff
@@ -71,9 +71,17 @@ class Q_SLICER_MODULE_SUBJECTHIERARCHY_WIDGETS_EXPORT qSlicerSubjectHierarchyPlu
...
+  ///       of the module!
   void registerCorePlugins();
 
+  /// Get all view menu actions available
+  /// \return List of object names of the view menu actions
+  Q_INVOKABLE QStringList allViewMenuActions();
+  /// Set desired set of view menu actions
+  /// \param actionObjectNames List of view menu actions to consider. Only actions included here by object name
+  ///        are shown in the menu if they are accepted by the owner plugin. The order set here is used.
+  Q_INVOKABLE void setViewMenuActionWhitelist(QStringList actionObjectNames);
```
lassoan: I've checked naming conventions for similar methods in Qt and found that "whitelist" is only used once. I think it would be better have these method names instead:
- setVisibleViewMenuActions or setViewMenuActions (rename from setViewMenuActionWhiteList)
- visibleViewMenuActions or viewMenuActions (new)
- allViewMenuActions or allAvailableViewMenuActions (no change needed)

jcfr: `setViewMenuActions()` / `viewMenuActions()`  :+1:

lassoan:
Thank you @cpinter.

I've fixed a crash (wrong number of parameters and wrong method ID was set in qSlicerSubjectHierarchyScriptedPlugin::showViewContextMenuActionsForItem), fixed selecting of node in markups module when "Edit properties" option is used, and tuned method names (availableViewMenuActionNames gets all available action names and setDisplayedViewMenuActionNames sets them).

Merged in http://svn.slicer.org/Slicer4/trunk@28644

PR #1263, opened 2019-11-19
COMP: Update LibArchive to fix build issue with GCC 8.2.0

lassoan:
Update to latest LibArchive release (v3.4.0) to address https://issues.slicer.org/view.php?id=4616


lassoan:
Merged in r28627.

PR #1264, opened 2019-11-20
BUG: Prevent Segment Editor Draw effect from getting stuck

Sunderlandkyl:
When apply was called in the draw effect, the polydata would not be reset unless there was at least one cell in the polydata.
If there were no cells (ie. just a single point). Then reset would not be called, unless more cells were added.
This commit ensures that resetPolyData is called in the apply function, regardless of how many cells there are.


lassoan:
Thank you! Merged in r28643.

PR #1265, opened 2019-11-22
BUG: Fixes normal vector weight bug and update search radius range

smrolfe:
This commit fixes a bug in the resample curves function of the Markups Module. The bug in the estimation of the normal vectors used to project curve points onto a surface resulted in a NAN weight for the final curve point. The weights of the normal points on either side of the projected point needed to be reversed.

The range of allowed the allowed search radius is expanded to include 100% to align with the GUI.


lassoan:
Thank you, merged.

PR #1266, opened 2019-11-24
update openssl and curl. now builds well on both debug and release modes

gaoyi:


pieper:
I did a fresh build and confirmed that this works.  (Note it's not enough to delete the OpenSSL directory at the Superbuild since the download version is stored in CMakeCache, so you can either edit that or do a completely fresh build.

@jcfr what did you think about this change?  Probably the we want to put the curl change back into the Slicer organization fork.

hjmjohnson:
@gaoyi Is this still relevant?  Could you rebase cleanly on the current master branch (not merge) so that the changes can be more easily evaluated?

gaoyi:
Hi Hans,

I did fresh build something like two weeks ago and this is still the case. I’ll re-do that and get back to you.

Best,
yi

PR #1267, opened 2019-11-25
BUG: Fix crash when loading and editing empty segments

Sunderlandkyl:
When loading a segmentation file that contained a segment with an empty segment that was not {0, -1, 0, -1, 0, -1}, there would be a crash when trying to copy or use the image data.
Fixed by setting the segment extent to {0, -1, 0, -1, 0, -1} if it is empty.


lassoan:
Thank you, merged!

PR #1268, opened 2019-11-25
BUG: Fix clearing of segments in ImportLabelmapToSegmentationNode

Sunderlandkyl:
During ImportLabelmapToSegmentationNode, segments should be cleared if they already exist, however the wrong label value was being cleared for the updated segment.
Fixed by clearing the segment before updating the label value (was previously the opposite).
Added a test to SegmentationsModuleTest1 to detect this issue.


lassoan:
Thank you, merged!

PR #1269, opened 2019-11-26
BUG: Fix vtkITKArchetypeImageSeriesReader::CanReadFile()

Sunderlandkyl:
vtkITKArchetypeImageSeriesReader::CanReadFile previously returned true as long as the Archetype was set, and the file exists.
Now, CanReadFile will return true if, in addition to the archetype being set and the file existing, an itk::ImageFileReader can be instantiated for the specified file.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
@@ -341,6 +341,7 @@ int vtkMRMLSegmentationStorageNode::ReadBinaryLabelmapRepresentation4DSpatial(vt
 
   // Get metadata dictionary from image
   itk::MetaDataDictionary metadata = allSegmentLabelmapsImage->GetMetaDataDictionary();
+
   // Read common geometry extent
   std::string commonExtent;
   itk::ExposeMetaData<std::string>(metadata, GetSegmentationMetaDataKey(KEY_SEGMENTATION_EXTENT).c_str(), commonExtent);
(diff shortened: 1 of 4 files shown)


pieper:
Makes sense to me.  Was there a specific issue that requires this change?  (I.e. wouldn't it just have failed later?).

Sunderlandkyl:
It was causing some failing tests in SlicerRT. vtkMRMLSegmentationStorageNode would try to load polydata as an image and report an error. Everything still loaded fine, but the error messages caused the test to fail.

Probably there is some other way to work around the error messages, but I figured that the the better approach would be to determine if vtkITKArchetypeImageSeriesReader can actually read the file before attempting to use it.

pieper:
Yes, makes sense.  Fine to merge from my point of view.

jcfr:
I wonder if there is anyway to refactor the code used in both `RequestInformation` and `CanReadFile` ?

pieper:
Yes, all that code is pretty messy.  Could make sense to take a fresh pass at the whole class.

jcfr:
This could be done later of course and I would have no issue integrated this patch, just wanted to point it out.

@Sunderlandkyl  I will let you assess.

Sunderlandkyl:
I agree, it makes sense to reduce the code duplication.
There's no hurry to integrate, so I'll make the necessary changes.

jcfr:
> so I'll make the changes in this PR.
:pray:

Sunderlandkyl:
I've added a GetImageIO function that is used by both RequestInformation and CanReadFile.

Sunderlandkyl:
@jcfr @pieper @lassoan Let me know if there are any other necessary changes.

lassoan:
The title of this PR is still "do not integrate" - can you confirm that it can be integrated now? Thank you!

Sunderlandkyl:
@lassoan It should be ready to integrate.

cpinter:
Thanks a lot, everyone! I integrated it (https://github.com/Slicer/Slicer/commit/8ea69453690690cee3de21e61e97122622c51e5d) so that we can see in the next nightly if the SlicerRT tests pass, to make progress with this [PR](https://github.com/SlicerRt/SlicerRT/pull/124).

PR #1271, opened 2019-11-27
BUG: Last terminology context is not remembered

lassoan:
Basic terminology items, such as "Liver" is missing from Slicer master list, therefore I would like to use the DICOM master list as default context.
The terminology widget correctly saves the last used context, but since the segment is always initialized with "Segmentation category and type - 3D Slicer General Anatomy list" context, this has no effect.

Remove default terminology to use last context as default context (isntead of force it to always be always Slicer's general anatomy list).

@cpinter Do you know whey we had to add Tissue/Tissue to all segments by default and why not just let it be empty?


fedorov:
> Do you know whey we had to add Tissue/Tissue to all segments by default and why not just let it be empty?

It could be because those attributes cannot be left empty when saving DICOM segmentation, which was one of the driving use cases for terminology.

cpinter:
@lassoan I don't remember the actual reason for this decision, but I do remember that it was not an arbitrary choice :) @fedorov's explanation might be the answer. Maybe it could be tested it with empty terminology information, but I'm sure that there was a good reason to add a valid entry by default.

fedorov:
> it was not an arbitrary choice

Yes, it was not an arbitrary choice. 1) relevant attributes in DICOM need to be populated with **some** codes in order for the saved object to be valid, and 2) "tissue"/"tissue" is a very generic type that has a reasonable chance to be "valid" if the user does not bother to change the default.  The way I understand it, this was the motivation.

lassoan:
Ok, so we'll set a term by default. But then how could we set the default context? Maybe add a default term selector in application settings?

pieper:
I think something like "Unspecified segment" would be a better default than "tissue".

I believe I was the one who made "tissue" the default name for label 1 in the old editor, but it doesn't really fit with all the diverse use cases these days.  Plus "tissue" could actually mean "tissue" or it could mean the user didn't specify.

fedorov:
> I think something like "Unspecified segment" would be a better default than "tissue".

Yes, perhaps. Maybe if/when the user exports the segmentation into DICOM, there could be a required step to specify segment. I guess the idea was to make this process transparent for those users who do not want to spend time specifying what they are doing. Whether that was a good or bad idea - I don't know!

> Ok, so we'll set a term by default. But then how could we set the default context? Maybe add a default term selector in application settings?

Yes, this would be helpful I think.

lassoan:
Maybe there is a better default terminology than tissue/tissue, but my main problem is that the context (that the user can choose items from) is always set to Slicer by default. If we allowed setting of the default item in application settings then it would set both the default context and item details.

pieper:
>  If we allowed setting of the default item in application settings then it would set both the default context and item details.

Yes, that makes sense to me.

cpinter:
+1 on adding a default context selector in Application Settings.

For some reason I thought that it was already there, but apparently not (instead, an entry is added in the ini file about the last context).

lassoan:
Fix is merged in 28694 (default terminology entry for segments can be set in application settings).

PR #1273, opened 2019-11-28
ENH: Allow shift+click pan with threshold editor effect

Sunderlandkyl:
Threshold effect now only starts drawing the histogram area if the shift key is not pressed.
This allows the use of shift+click panning without changing the histogram.


lassoan:
Thank you , merged!

PR #1274, opened 2019-12-03
ENH: Add vtkITK filter for computing label margins

Sunderlandkyl:
vtkITKImageMargin computes image margins using a distance map calculated from itkSignedMaurerDistanceMapImageFilter (http://hdl.handle.net/1926/171).
This commit also modified the hollow and margin segment editor effects to use this filter, instead of vtkImageDilateErode3D.

### DIFF
--- a/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorHollowEffect.py
+++ b/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorHollowEffect.py
@@ -1,6 +1,7 @@
 import os
 import vtk, qt, ctk, slicer
 import logging
+import math
 from SegmentEditorEffects import *
 
 class SegmentEditorHollowEffect(AbstractScriptedSegmentEditorEffect):
@@ -47,13 +48,13 @@ class SegmentEditorHollowEffect(AbstractScriptedSegmentEditorEffect):
     self.shellThicknessMmSpinBox.value = 3.0
     self.shellThicknessMmSpinBox.singleStep = 1.0
 
-    self.kernelSizePixel = qt.QLabel()
-    self.kernelSizePixel.setToolTip("Closest achievable thickness. Constrained by the segmentation's binary labelmap representation spacing.")
+    self.shellThicknessPixel = qt.QLabel()
+    self.shellThicknessPixel.setToolTip("Closest achievable thickness. Constrained by the segmentation's binary labelmap representation spacing.")
 
     shellThicknessFrame = qt.QHBoxLayout()
     shellThicknessFrame.addWidget(self.shellThicknessMmSpinBox)
-    shellThicknessFrame.addWidget(self.kernelSizePixel)
     self.shellThicknessMmLabel = self.scriptedEffect.addLabeledOptionsWidget("Shell thickness:", shellThicknessFrame)
+    self.scriptedEffect.addLabeledOptionsWidget("", self.shellThicknessPixel)
 
     self.applyButton = qt.QPushButton("Apply")
     self.applyButton.objectName = self.__class__.__name__ + 'Apply'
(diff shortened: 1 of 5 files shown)


Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorHollowEffect.py
```diff
@@ -154,49 +134,29 @@ def onApply(self):
...
     if shellMode == MEDIAL_SURFACE:
-      dilateKernelSizePixel = [int((kernelSizePixel[i]+1.0)/4.0)*2+1 for i in range(3)]
-      erodeKernelSizePixel = [int((kernelSizePixel[i]-1.0)/4.0)*2+1 for i in range(3)]
+      margin.SetOuterMarginMm(0.5*shellThicknessMm)
+      margin.SetInnerMarginMm(-0.5*shellThicknessMm)
     elif shellMode == INSIDE_SURFACE:
-      dilateKernelSizePixel = kernelSizePixel
-      erodeKernelSizePixel = [1, 1, 1]
+      margin.SetOuterMarginMm(shellThicknessMm)
+      margin.SetInnerMarginMm(0.0 + 0.000001)
```
lassoan: We don't know what the size of a voxel is in mm (it might be smaller than 0.000001). It would be safer to use 0.001*voxelDiameter.

Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorHollowEffect.py
```diff
@@ -107,17 +95,9 @@ def updateGUIFromMRML(self):
...
-      selectedSegmentLabelmapSpacing = selectedSegmentLabelmap.GetSpacing()
-      kernelSizePixel = self.getKernelSizePixel()
-      if kernelSizePixel[0]<=1 and kernelSizePixel[1]<=1 and kernelSizePixel[2]<=1:
-        self.kernelSizePixel.text = "requested. Not feasible at current resolution."
-        self.applyButton.setEnabled(False)
-      else:
-        thicknessMm = [abs((kernelSizePixel[i]-1.0)/2.0*selectedSegmentLabelmapSpacing[i]) for i in range(3)]
-        self.kernelSizePixel.text = "requested. Actual: {0} x {1} x {2} mm.".format(*thicknessMm)
-        self.applyButton.setEnabled(True)
+      self.marginInfoLabel.text = ""
```
lassoan: Actual shell size is still constrained by the voxel size, so it would be important to keep displaying the request and actual voxel size separately.

Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorMarginEffect.py
```diff
@@ -70,17 +66,6 @@ def createCursor(self, widget):
   def setMRMLDefaults(self):
     self.scriptedEffect.setParameterDefault("MarginSizeMm", 3)
 
-  def getKernelSizePixel(self):
```
lassoan: It would be nice to keep (or add?) display of actual margin size in addition to requested margin size.

lassoan:
@Sunderlandkyl Is this still WIP? Not urgent, it would be just nice to get this improvement.

Sunderlandkyl:
Yes, it's still a WIP.

Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorHollowEffect.py
Sunderlandkyl: Can you elaborate how the method used for stepSize can be applied here?

lassoan: We could round the displayed "actual margin" value to a few decimals, as it is done for stepSize value.

3 floating-point numbers can be quite long, even if rounded, so maybe we could display it in a separate row.

Sunderlandkyl:
Thanks for your patience, I think that the margin/hollow effects should behave as expected with new new filter now.
@lassoan Let me know if I should make any changes.

lassoan:
This works very well and it is very fast now! Thank you. Merged.

PR #1275, opened 2019-12-03
ENH: Adds new icons in Markups module for create new markup actions

smrolfe:
The buttons in the markups module for creating new curve or markups nodes use the same icons as the mouse place mode menu in the upper toolbar. This commit adds icons with a plus symbol in the corner for the add markups node, open curve node, and closed curve node buttons to help users distinguish between the function of these buttons and the mouse place mode menu.


jcfr:
Thanks for the contribution :pray: 

When you have a chance, could you post here a screenshot ?

smrolfe:
screenshotIcon

Thanks, I've added a screenshot with the three new icons highlighted.

lassoan:
Whey there are no '+' signs for line and angle widgets?

smrolfe:
Since the line and angle nodes have a specified number of points, changing the mouse mode creates a new node, which is consistent with the Markups module buttons.

smrolfe:
The '+' signs are intended to show that a new markup node will be created instead of returning to placement mode of the active markup (if existing), which only applies to the curve and fiducial nodes.

jcfr:
I wonder if this would be clearer:
* an "infinity sign" for closed/open curve and fiducial point
* the number of click for the other ones

smrolfe:
Thanks @jcfr. The main issue is that using the same icon in the create and place menu and the Markups module (such as fiducial placement in the screenshot attached) was confusing some users since the function is different. If it's more clear, all the Markup module icons could be modified, even though I don't believe the function is different in the case of angles and lines. 

We liked the '+' on the Markups module icons to indicate that a new markups node is always created, but I think any distinguishing mark would help distinguish between the two sets of icons. Would you suggest adding the infinity symbol/number of clicks to the icons in the create and place menu?

Capture

lassoan:
Line and angle buttons in Markups module perform different action than the buttons on the toolbar ("add a new markup node" vs. "add a new markup point if the active markups node can accept more points and it is the right type otherwise add a new markup point"), so they have to get the "+" sign as well.

smrolfe:
Thanks for the clarification @lassoan , I'll update with those icons.

smrolfe:
I'm adding an updated screenshot with all 5 icons updated.
screenshotIcon5

lassoan:
Thank you, this will work! Just one more thing: we need to make the icons consistent existing "add" icons in Slicer:
- please rename Markups*New.png files to MarkupsAdd*.png.
- use the same '+' symbol as in other "SlicerAdd..." icon images in Base\QTGUI\Resources\Icons. We can make the '+' thicker but then we need to do it consistently for all icons and not just for these new ones.

smrolfe:
Thanks, I have updated the commit to standardize the icon names and '+' symbol. I think the symbols are easily readable and thicker line shouldn't be necessary. Can you let me know what you think? I'm attaching a screenshot.

screenshotIconThin

Modules/Loadable/Markups/Resources/qSlicerMarkupsModule.qrc
```diff
@@ -10,14 +10,19 @@
         <file>Icons/MarkupsListLock.png</file>
         <file>Icons/MarkupsListUnlock.png</file>
         <file>Icons/MarkupsMouseModePlace.png</file>
```
jcfr: Nitpick: Look like the indentation is off

jcfr: I will take care of this while integrating

jcfr:
Closing. Corresponding fixes integrated in [r28666](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28666)

smrolfe:
Thanks @jcfr and @lassoan for your help with this.

PR #1276, opened 2019-12-04
COMP: Fix names of modules to compile with USE_SYSTEM_python

phcerdan:
The name of the modules are sometimes not the same used for the
import command.  For example, for the package pillow, have to use `import PIL`

This made the command `ExternalProject_FindPythonPackage` to fail when
checking version.
Also some modules do not implement `__version__`, but `version()` or
even nothing. Changes have been made accordingly for these cases.

Tested in ArchLinux with command:

```bash
cmake -GNinja -DCMAKE_BUILD_TYPE:STRING=RelWithDebInfo -DSlicer_USE_SYSTEM_curl:BOOL=ON -DSlicer_USE_SYSTEM_python:BOOL=ON -DQt5_DIR:PATH=/usr/lib/cmake/Qt5 ../Slicer-src
```

And installing with `pacman`, from the AUR, or with `pip --user` the required
python packages.

As a side-note, Arch also requires the VTK fix for Python3.8:
https://gitlab.kitware.com/vtk/vtk/merge_requests/5883/commits

### DIFF
--- a/SuperBuild/External_python-dicom-requirements.cmake
+++ b/SuperBuild/External_python-dicom-requirements.cmake
@@ -38,12 +38,21 @@ endif()
 ExternalProject_Include_Dependencies(${proj} PROJECT_VAR proj DEPENDS_VAR ${proj}_DEPENDENCIES)
 
 if(Slicer_USE_SYSTEM_${proj})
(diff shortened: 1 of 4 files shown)


jcfr:
Thanks :pray:  

Closing. Fixes integrated in [r28667](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28667)

PR #1277, opened 2019-12-05
COMP: Make EditorEffects base classes available for extensions

lassoan:
SlicerMacroBuildModuleQtLibrary macro does not add module include directories to CMake variables that are exposed to extensions. However, we need Segmentations/EditorEffects classes to be available for extensions to allow creating editor effects in C++.

See background in https://discourse.slicer.org/t/unable-to-include-qslicersegmentationseditoreffects-when-building-extension/9371/5

### DIFF
--- a/Modules/Loadable/Segmentations/EditorEffects/CMakeLists.txt
+++ b/Modules/Loadable/Segmentations/EditorEffects/CMakeLists.txt
@@ -78,10 +78,9 @@ set(${KIT}_TARGET_LIBRARIES
   )
 
 #-----------------------------------------------------------------------------
-SlicerMacroBuildModuleQtLibrary(
+SlicerMacroBuildModuleWidgets(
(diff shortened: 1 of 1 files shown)


lassoan:
@jcfr is there any drawback of using SlicerMacroBuildModuleWidgets macro for exposing Segment Editor effects C++ base classes to extensions? Do you have any alternative suggestions? We used SlicerMacroBuildModuleQtLibrary before but that did not make these classes available for extensions.

lassoan:
@jcfr can you have a look? Just tell me if it's OK to use the macro like this. Thanks!

jcfr:
I don't foresee any issues. That said, I suggest we introduce a new CMake  function called `SlicerMacroBuildModuleEditorEffect`

hjmjohnson:
@lassoan Can this be merged as is with the new macro added later?

PR #1278, opened 2019-12-05
ENH: Add CropToMinimumExtent to vtkMRMLSegmentationStorageNode

Sunderlandkyl:
If CropToMinimumExtent is false, the segmentation will be saved using the same extent as the reference image.
If CropToMinimumExtent is true, the segmentation will be saved using the effective extent.

The option can be changed using the options widget in the save dialog.
This required moving qSlicerNodeWriterOptionsWidgetPrivate to a private header so that it could be inherited.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLSegmentationStorageNode.h
+++ b/Libs/MRML/Core/vtkMRMLSegmentationStorageNode.h
@@ -162,6 +162,13 @@ public:
   /// Reset supported write file types. Called when master representation is changed
   void ResetSupportedWriteFileTypes();
 
+  /// Save segmentation using the reference geometry.
+  /// If true, the segmentation will be saved using the same extent as the reference image.
+  /// If false, the segmentation will be saved using the effective extent.
+  vtkSetMacro(CropToMinimumExtent, bool);
+  vtkGetMacro(CropToMinimumExtent, bool);
+  vtkBooleanMacro(CropToMinimumExtent, bool);
+
 protected:
   /// Initialize all the supported read file types
   void InitializeSupportedReadFileTypes() override;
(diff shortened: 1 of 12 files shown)


Libs/MRML/Core/vtkMRMLSegmentationStorageNode.h
```diff
@@ -162,6 +162,13 @@ class VTK_MRML_EXPORT vtkMRMLSegmentationStorageNode : public vtkMRMLStorageNode
   /// Reset supported write file types. Called when master representation is changed
   void ResetSupportedWriteFileTypes();
 
+  /// Save segmentation using the reference geometry.
+  /// If true, the segmentation will be saved using the same extent as the reference image.
+  /// If false, the segmentation will be saved using the effective extent.
+  vtkSetMacro(UseReferenceGeometry, bool);
```
lassoan: I think reference geometry may be completely different from the current labelmap geometry (it may be shifted, rotated, etc.). Do we resample to the reference geometry or only use its extent?

Probably we should only use its extent, but then the attribute name should reflect specifically that (e.g., CropToMinimumExtent).

Sunderlandkyl: Ok. Renamed to CropToMinimumExtent.
DetermineCommonLabelmapGeometry call uses either EXTENT_UNION_OF_EFFECTIVE_SEGMENTS or EXTENT_REFERENCE_GEOMETRY depending on CropToMinimumExtent.

https://github.com/Slicer/Slicer/pull/1278/files#diff-0af540967f0f90716cf48277779c9ae8R1133-R1134

lassoan:
Perfect, thank you, I'll merge this now.

lassoan:
Merged!

PR #1279, opened 2019-12-05
BUG: Fix display and derived representations of cleared segments

Sunderlandkyl:
Clearing the last labelmap segment in segmentation would not correctly update the display pipeline, or the derived representations.
This commit ensures that the segments are displayed correctly if cleared using vtkDataObject::Initialize().


Libs/vtkSegmentationCore/vtkBinaryLabelmapToClosedSurfaceConversionRule.cxx
```diff
@@ -258,7 +258,7 @@ bool vtkBinaryLabelmapToClosedSurfaceConversionRule::CreateClosedSurface(vtkOrie
     {
     // empty labelmap
     vtkDebugMacro("Convert: No polygons can be created, input image extent is empty");
-    closedSurfacePolyData->Reset();
+    closedSurfacePolyData->Initialize();
```
lassoan: Just curious, what's the difference between Reset and Initialize?

lassoan:
Merged!

PR #1280, opened 2019-12-06
ENH: Save and load empty nodes without error

cpinter:
When a node had no data in it, for example a transform node had no matrix object set, then WriteDataInternal returned with an error, but also the storage node contained a valid file name and the write state was TransferDone. When loading such a scene, an error was thrown because the file specified did not exist. Thus, this valid behavior could be incorrectly detected as a scene load failure.

This commit introduces the new state CancelledNoData in the storage nodes, allowing the above described situation to be handled in a way that does not indicate error. When loading such a scene, no file is attempted to be loaded, and scene load ends with success.

The supported node types are
- Volume
- Model
- Transform

### DIFF
--- a/Libs/MRML/Core/vtkMRMLModelStorageNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLModelStorageNode.cxx
@@ -122,7 +122,11 @@ bool vtkMRMLModelStorageNode::CanReadInReferenceNode(vtkMRMLNode *refNode)
 //----------------------------------------------------------------------------
 int vtkMRMLModelStorageNode::ReadDataInternal(vtkMRMLNode *refNode)
 {
-  vtkMRMLModelNode *modelNode = dynamic_cast <vtkMRMLModelNode *> (refNode);
+  if (this->GetWriteState() == SkippedNoData)
+    {
+    vtkDebugMacro("ReadDataInternal: empty model file was not saved, ignore loading");
+    return 1;
+    }
 
   std::string fullName = this->GetFullNameFromFileName();
   if (fullName.empty())
@@ -147,6 +151,8 @@ int vtkMRMLModelStorageNode::ReadDataInternal(vtkMRMLNode *refNode)
     }
 
   vtkDebugMacro("ReadDataInternal: extension = " << extension.c_str());
+
+  vtkMRMLModelNode* modelNode = dynamic_cast<vtkMRMLModelNode*>(refNode);
 
   int result = 1;
   try
@@ -327,10 +333,16 @@ int vtkMRMLModelStorageNode::WriteDataInternal(vtkMRMLNode *refNode)
 
   std::string fullName = this->GetFullNameFromFileName();
   if (fullName.empty())
-  {
+    {
     vtkErrorMacro("vtkMRMLModelNode: File name not specified");
     return 0;
-  }
+    }
+
+  if (modelNode->GetMeshConnection() == nullptr || modelNode->GetMesh()->GetNumberOfPoints() == 0)
+    {
+    this->SetWriteStateSkippedNoData();
+    return 1;
+    }
 
   std::string extension = vtkMRMLStorageNode::GetLowercaseExtensionFromFileName(fullName);
 
--- a/Libs/MRML/Core/vtkMRMLStorageNode.h
+++ b/Libs/MRML/Core/vtkMRMLStorageNode.h
@@ -117,6 +117,9 @@ public:
   /// Transferring: data is remote, and the transfer is working to completion
   /// TransferDone: the data is on disk and ready to be read
   /// Cancelled: the user cancelled the remote data transfer
+  /// SkippedNoData: writing or reading was skipped due to data not present.
+  ///                This is a valid state used if and only if the data node
+  ///                is empty and there is nothing to be written or read.
   enum
   {
     Idle,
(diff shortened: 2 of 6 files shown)


pieper:
I guess this makes sense.  Another option would be to have the constructor of these nodes put in a placeholder and enforce that they can't be unset.  Otherwise we end up with this kind of checking logic and workaround everywhere (gui, displayable manager, etc).

But generally I don't mind this fix so LGTM 👍

lassoan:
When we create a new volume, we do not set a valid vtkImageData in it. It is fine everywhere else except for storage, because most file writers refuse to write an invalid/empty file (you cannot write a volume with 0 extent or a polydata with 0 points), so we would eitherbneed to generate some small dummy file (volume of 1 voxel, polydata of 1 point, etc) or flag that there is no file. Adding a dummy data set could cause several complications, so I think adding this flag is the simpler, cleaner solution.

Libs/MRML/Core/vtkMRMLModelStorageNode.cxx
```diff
@@ -332,6 +340,12 @@ int vtkMRMLModelStorageNode::WriteDataInternal(vtkMRMLNode *refNode)
     return 0;
   }
 
+  if (modelNode->GetMeshConnection() == nullptr)
```
lassoan: Can you check if we can write a polygon that has 0 points? If not, then it would be probably better to treat it as CancelledNoData, too.

Libs/MRML/Core/vtkMRMLTransformStorageNode.cxx
```diff
@@ -447,8 +447,16 @@ int vtkMRMLTransformStorageNode::ReadDataInternal(vtkMRMLNode *refNode)
   // check that the file exists
   if (vtksys::SystemTools::FileExists(fullName.c_str()) == false)
     {
-    vtkErrorMacro("ReadDataInternal: transform file '" << fullName.c_str() << "' not found.");
-    return 0;
+    if (this->GetWriteState() == CancelledNoData)
```
lassoan: Should we even try to read if write state was CancelledNoData? Maybe the current behavior is better (e.g., you can prepopulate a scene and add a valid data file later), but a comment to clarify the reasoning would be helpful. It can be described here or at the description of the states.

Libs/MRML/Core/vtkMRMLStorageNode.h
```diff
@@ -117,14 +117,17 @@ class VTK_MRML_EXPORT vtkMRMLStorageNode : public vtkMRMLNode
   /// Transferring: data is remote, and the transfer is working to completion
   /// TransferDone: the data is on disk and ready to be read
   /// Cancelled: the user cancelled the remote data transfer
+  /// CancelledNoData: reading was cancelled due to data not present. Can be
```
lassoan: This is for reading and writing, too, so talk about writing as well (see comment above). I would also make it more clear that it is not just "can be valid" but always valid and is used if and only if the data node is empty and there is nothing to be written or read.

Libs/MRML/Core/vtkMRMLStorageNode.h
```diff
@@ -117,14 +117,17 @@ class VTK_MRML_EXPORT vtkMRMLStorageNode : public vtkMRMLNode
...
   enum
   {
     Idle,
     Pending,
     Scheduled,
     Transferring,
     TransferDone,
-    Cancelled
+    Cancelled,
+    CancelledNoData
```
lassoan: We've discussed this with @cpinter and CancelledNoData seemed like a good name but now I feel that it suggests that something is wrong. I'm wondering if SkippedNoDataAvailable, TransferSkippedNoData, TransferNotNeededEmptyData would be more clear?

@pieper what do you think?

pieper: yes, I prefer SkippedNoData now, that you suggest it.

cpinter: OK I'll change that. Thanks!

cpinter:
@pieper My first implementation to fix this was using a placeholder (identity transform for an empty transform when writing), but it did not exactly correspond to the state of the node in the scene (i.e. it was changed). Also apparently that temporary fix introduced a bug in a custom app, so it seems that it would cause problems with execution too. So after I talked to @lassoan we settled at this new state.

Libs/MRML/Core/vtkMRMLTransformStorageNode.cxx
cpinter: I'm not sure I understand this fully. What do you mean by current behavior? Do you suggest that I move this check out of the if FileExists block?

pieper:
Right @cpinter this is a good fix that is consistent with current design of the nodes.  It's something to keep in mind when creating new node types that they should never be in undefined states that require special cases in all the code that uses them.

Libs/MRML/Core/vtkMRMLTransformStorageNode.cxx
lassoan: If state is "CancelledNoData" but a data file exists by the name that is stored in the storage node. We call ReadData. Possible behaviors:
A. read the file (state has preference over presence of file).
B. do not read the file (presence of file has preference over the state).

I think currently behavior B is implemented consistently, but we should still specify that this is the intended behavior and why. We should probably also always consistently clear the data set if it is read from a "CancelledNoData" storage node.

cpinter:
I updated the commit. If it looks good I'll integrate it soon

cpinter:
Integrated in https://github.com/Slicer/Slicer/commit/63e781dafd5a353cb37ca3f32358bdaa46bd4163

PR #1281, opened 2019-12-06
ENH: Serialize segmentation geometry using vtkNumberToString

Sunderlandkyl:
Previously, segmentation geometry serialization was done using stringstream with a 15 decimal precision.
To ensure that image geometry remains consistent with VTK and ITK, we now use vtkNumberToString, which uses the double_conversion library.


lassoan:
Thank you, merged in r28670.

PR #1283, opened 2019-12-09
BUG: Fix labelmap representation copy between segments

Sunderlandkyl:
When copying labelmap representations between segments, all values in the labelmap were previously copied.
This commit changes vtkSegment::DeepCopy to only copy the label value of the source segment.


Sunderlandkyl:
@lassoan the other alternative would be to modify vtkSegmentation::CopySegmentFromSegmentation to remove other label values from the labelmap after the segment has been copied.

lassoan:
Thank you. Committed r28672.

PR #1286, opened 2019-12-13
BUG: Fix Segment Editor Undo/Redo by reverting r28672

Sunderlandkyl:
Changing the way that lablemap representations were copied broke the Undo/Redo functionality.
Copying segments between segmentations will still need to be fixed, however it is less important than Undo/Redo.

### DIFF
--- a/Libs/vtkSegmentationCore/vtkSegment.cxx
+++ b/Libs/vtkSegmentationCore/vtkSegment.cxx
@@ -21,7 +21,6 @@
 // SegmentationCore includes
 #include "vtkSegment.h"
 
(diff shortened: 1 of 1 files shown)


jcfr:
Good catch. Is there a test that can be updated to capture regression associated with undo/redo ?

Sunderlandkyl:
Probably one of the vtkSegmentationTest could and should be updated to test undo/redo.
I'll start working on it.

jcfr:
> could and should be updated to test undo/redo.
I'll start working on it.

Thanks for working on this :pray:

lassoan:
Thank you, merged.

PR #1287, opened 2019-12-14
Fix failing py_UtilTest and py_MarkupsWidgetsSelfTest

jamesobutler:
The first commit should fix the following failing [py_UtilTest](http://slicer.cdash.org/testDetails.php?test=9849661&build=1773258). The test began to fail when a function was renamed in https://github.com/Slicer/Slicer/commit/a3308548ca4dd5c0ccd44924449abd55fa9d2f01. 

The second commit should fix the following failing [py_MarkupsWidgetsSelfTest ](http://slicer.cdash.org/testDetails.php?test=9849654&build=1773258). The test began to fail when a qMRMLNodeComboBox was renamed in https://github.com/Slicer/Slicer/commit/2a8a518dbb0fce256c3fb094a4e39ae3213ea6ce.


lassoan:
Thanks a lot! Merged.

PR #1289, opened 2019-12-16
BUG: Improve segment deep copy to separate shared labelmaps

Sunderlandkyl:
When copying labelmap representations between segments, all values in the labelmap were previously copied.
This commit changes vtkSegment::DeepCopy to only copy the label value of the source segment.

Reimplements r28672 with some additional changes to ensure that DeepCopy is not used in vtkSegmentationHistory.
This ensures that Undo/Redo behavior is not affected.


lassoan:
Thank you, merged!

PR #1290, opened 2019-12-18
ENH: Update segment center calculation to use largest island

Sunderlandkyl:
~~Previously vtkSegmentationNode::GetSegmentCenter() calculated the center of a segment as the center of the bounding box containing the segment.
This commit changes GetSegmentCenter to calculate and return the the labelmap center of mass.~~

Edit:
Previously vtkSegmentationNode::GetSegmentCenter() calculated the center of a segment as the center of the bounding box containing the entire segment.
This commit changes GetSegmentCenter to calculate and return the the center of the largest island extent.

Also fixes a padding bug in vtkOrientedImageDataResample::ResampleOrientedImageToReferenceOrientedImage that calculated an incorrect extent when the reference extent was empty.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLSegmentationNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLSegmentationNode.cxx
@@ -53,6 +53,9 @@
 #include <vtkMatrix4x4.h>
 #include <vtkTransform.h>
 
+// vtkITK includes
+#include <vtkITKIslandMath.h>
+
 // STD includes
 #include <algorithm>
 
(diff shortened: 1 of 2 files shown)


lassoan:
Thank you, I've tested this and it solves the issue when a large segment is not found because of some speckles far away. However, the computation takes long time for large segments. See for example this scene (jump takes about 10-15sec): https://1drv.ms/u/s!Arm_AFxB9yqHucNuLWLe7SW0g0TsTg?e=V4bHMb

I've also realized that this does not solve the problem of jumping to nowhere if there are two approximately equally large, separate blobs. Computing a "median"-like position would be better than the current "mean". We could keep accurate mean computation as an option somewhere.

What we could do:
- to speed things up: iterate through the effective voxel of the segment labelmap, but only check every N-th column, row, slice (we can start with something like N=8 and if less than 30 points are found then we check the points between them with N=4, N=2, N=1) => 
- to avoid jumping between blobs: performing "keep largest island" before centroid computation would solve this, but it may be too slow (maybe for large N, it would be acceptable); alternatively, we could compute centroid for all the points but then find the closest point to that position and jump to that position
- we could cache the computation result and only recompute if the segment has changed

cpinter:
It would be possible to do keep largest islands on the subsampled (N=8 sounds reasonable) labelmap

Sunderlandkyl:
Segment center is now calculated using the effective extent of the largest island.
Labelmap is resampled to have approximately 64^3 voxels.
Performance should be much better.

lassoan:
This works very well - I've merged it. The only small improvement we should do is that if we don't happen to find any non-zero voxels (can happen when there are just small speckles in the segment) then return the center of the entire effective extent (the old behavior).

PR #1291, opened 2019-12-18
BUG: Remove blockSignals from qMRMLSegmentsModel::updateItemFromSegment

Sunderlandkyl:
Previously signals were being blocked when updateItemDataFromSegment was called.
This meant that the model could not emit dataChanged signals which would trigger an update of the table.

This commit removes the blockSignals call.
Also updates qMRMLSegmentsTableView to remove some unused variables, and add an additional null pointer check.


Sunderlandkyl:
@lassoan this ended up being the issue. I feel somewhat silly for not realizing it earlier.

lassoan:
Thank you, good catch. It's good that nothing "strange" is going on, but we have a clear fix.

lassoan:
Merged in r28696

PR #1294, opened 2019-12-20
BUG: Change vtkSegmentation::DeepCopy to use CopySegment

Sunderlandkyl:
Due to changes in the way vtkSegment::DeepCopy is performed, it is no longer suitable for preserving shared labelmap layers when copying entire segmentations.
This commit fixes the behavior of vtkSegmentation::DeepCopy by utilizing vtkSegmentation::CopySegment (migrated from vtkSegmentationHistory).


Sunderlandkyl:
@lassoan This fixes the SH clone issue.

lassoan:
Thanks a lot for the quick fix. Merged.

PR #1295, opened 2019-12-27
Update ITK 5.1rc01 and SimpleITK

jamesobutler:
This PR updates ITK to 5.1rc01 and updates SimpleITK so that it is compatible with ITK 5.1rc01.  Building of SimpleITK is also now using the upstream repo as I noticed the previous used git hash in the Slicer fork had no additional changes so instead of creating a new branch that is a mirror up to a certain hash I'm just specifying the upstream repo.

I have had a successful build on Windows 10 and with only seemingly already failing tests continuing to fail.
ITK-5 1rc01-update

### DIFF
--- a/SuperBuild.cmake
+++ b/SuperBuild.cmake
@@ -265,8 +265,8 @@ set(BRAINSTools_options
   )
 
 Slicer_Remote_Add(BRAINSTools
-  GIT_REPOSITORY ${EP_GIT_PROTOCOL}://github.com/BRAINSIA/BRAINSTools.git
-  GIT_TAG d0e6c01c712e73f47b7e99c3194514a888a1b4a7 # post ITK 5.0.1
+  GIT_REPOSITORY ${EP_GIT_PROTOCOL}://github.com/Slicer/BRAINSTools.git
+  GIT_TAG f8e5ed125623fea783e3adedcedffb1f114f619a # slicer-2020-01-08-d0e6c01
   LICENSE_FILES "http://www.apache.org/licenses/LICENSE-2.0.txt"
   OPTION_NAME Slicer_BUILD_BRAINSTOOLS
   OPTION_DEPENDS "Slicer_BUILD_CLI_SUPPORT;Slicer_BUILD_CLI"
--- a/SuperBuild/External_ITK.cmake
+++ b/SuperBuild/External_ITK.cmake
@@ -33,7 +33,7 @@ if(NOT DEFINED ITK_DIR AND NOT Slicer_USE_SYSTEM_${proj})
 
   ExternalProject_SetIfNotDefined(
     Slicer_${proj}_GIT_TAG
-    "ff48670261e3bd16ee1c6f5494834f65183a98dd" # pre-v5.1b01 (2019-08-26)
+    "v5.1rc01"
     QUIET
     )
 
--- a/SuperBuild/External_SimpleITK.cmake
+++ b/SuperBuild/External_SimpleITK.cmake
@@ -35,13 +35,13 @@ ExternalProject_Execute(${proj} \"install\" \"${PYTHON_EXECUTABLE}\" Packaging/s
 
   ExternalProject_SetIfNotDefined(
     Slicer_${proj}_GIT_REPOSITORY
-    "${EP_GIT_PROTOCOL}://github.com/Slicer/SimpleITK.git"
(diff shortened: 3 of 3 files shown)


lassoan:
Thanks a lot for this work. It is important to upgrade ITK, as it fixes issues such as [this](https://discourse.slicer.org/t/double-images-when-loading-a-dicom-file/9601/7).

However, BRAINSTools has build errors after this change (in PerformMetricTest). I've submitted a [pull request with a fix](https://github.com/BRAINSia/BRAINSTools/pull/462). We need to wait for it to be integrated and then we need to include BRAINSTools update in this pull request.

jamesobutler:
Since there hasn't been any movement on the associated BRAINS PR I've added a PR that points to @lassoan fork which includes the update needed for it to be ITK 5.1rc01 compatible.

lassoan:
Test error is cause by a change in how ITK loads nifti files (it seems that it is just impossible to get orientation right in nifti files) - see discussion here: https://discourse.itk.org/t/nifti-file-orientation-change-between-2019-08-26-and-itk-v5-1rc01/2594

The effect of this change is probably limited (only affects old nifti files), but it is worth waiting a few days until ITK developers figure out how to proceed.

lassoan:
I've updated this pull request to use Slicer's BRAINSTool fork, which contains the necessary fix to build with ITK 5.1rc01.

lassoan:
Thank you, this is merged now. The two test failures will be fixed by updating file format of the data sets (Analyze format is deprecated see https://discourse.itk.org/t/nifti-file-orientation-change-between-2019-08-26-and-itk-v5-1rc01/2594/24).

lassoan:
After the upgrade, there are 2 build errors on mac: http://slicer.cdash.org/viewBuildError.php?buildid=1795939

@pieper could you have a look at them? It may be some trivial syntax error (there are no problems on Windows and Linux).

pieper:
Yes, I get the same build error on mac locally.  I'll have a look.

hjmjohnson:
I thought that this error should also show up on Linux and windows, but perhaps only under strict compiler flags.

This is addressing a common and sometimes difficult to spot performance problem in VNL that was addressed recently.  The implicit conversions between different types was convenient, but very often performed several implicit conversions.

The fix is to explicitly request the type you want.  In many cases, a better fix is to make a local variable `auto`, or match the type returned by GetVnlMatrix() to avoid the implicit conversions.
```c++
centralMoments.GetVnlMatrix().as_matrix()
```

pieper:
Changing the declarations like this fixes the mac build:

-  vnl_matrix<double> M( 3, 3 );
+  vnl_matrix_fixed<double, 3, 3> M( 3, 3 );

@hjmjohnson is this okay or better to dot he `as_matrix()` approach (if so can you provide the fix?)

hjmjohnson:
> Changing the declarations like this fixes the mac build:
> 
> * vnl_matrix M( 3, 3 );
> 
> * vnl_matrix_fixed<double, 3, 3> M( 3, 3 );
> 
> @hjmjohnson is this okay or better to dot he `as_matrix()` approach (if so can you provide the fix?)

@pieper It is much better to match the types like you did!  That will be a performance gain (perhaps not important here, but in many cases this was occurring inside a "for every pixel" loop.

The ".as_matrix()" will create a temporary vnl_matrix object and copy the contents of the "vnl_matrix_fixed<double, 3, 3> M( 3, 3 )".  Your solution avoids that extra temporary.

pieper:
Thanks Hans!

Pushed as r28728

lassoan:
Thank you @pieper and @hjmjohnson!

jamesobutler:
A user on the forum reported that on macOS SimpleITK reading of images results in Slicer crashing. See https://discourse.slicer.org/t/sitkutils-pullvolumefromslicer-makes-slicer-crash/9781/3.  This is also observed by [this failling test](http://slicer.cdash.org/testDetails.php?test=9875240&build=1797177) on the Mac factory build.  The same test isn't failing on Sam's experimental Windows build.

hjmjohnson:
This is now slightly out of date.  The ITK 5.1rc02 work will be a better stepping stone (which also requires a bit of coordination) for the next version of Slicer.

I recommend closing this issue for now and re-open a new issue for ITK 5.1rc02.

PR #1296, opened 2019-12-28
BUG: Fix setModuleHelpSectionVisible always setting false

jamesobutler:
@lassoan This fixes a minor bug that was included when `setModuleHelpSectionVisible` was originally introduced as part of https://github.com/Slicer/Slicer/commit/21dc780a216f077bdd080fdc64073ccf6b5356e0.


lassoan:
Thank you! Good catch. Merged.

PR #1297, opened 2019-12-30
ENH: Allow using any markups node as CLI module input or output

lassoan:
Requires merging of https://github.com/Slicer/SlicerExecutionModel/pull/122


lassoan:
Merged.

PR #1298, opened 2020-01-02
ENH: update general anatomy segmentation context

fedorov:
Slicer General Anatomy DICOM segmentation context was updated to add
missing terms and missing Slicer integer labels, as discussed in https://github.com/QIICR/dcmqi/issues/388


lassoan:
Thank you, it is merged now (r28707).

PR #1299, opened 2020-01-06
COMP: Fix issue when WebEngine support is disabled

sankhesh:
Fixes issue [4718](https://issues.slicer.org/view.php?id=4718)


lassoan:
Thank you, merged.

PR #1301, opened 2020-01-09
ENH: Add vtkITKLabelShapeStatistics for binary labelmap statistics

Sunderlandkyl:
This commit adds the new vtkITKLabelShapeStatistics class to vtkITK that calculates various statics on binary labelmaps.
Additional statisics provided by the filter have been added to the LabelmapSegmentStatisticsPlugin as optional, non-default statistics.

Current statistics provided by vtkITKLabelShapeStatistics are:
- Centroid coordinates
- Feret diameter
- Perimeter
- Roundness
- Flatness
- Oriented bounding box

### DIFF
--- a/Libs/MRML/Core/vtkMRMLTableStorageNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLTableStorageNode.cxx
@@ -39,6 +39,8 @@
 //------------------------------------------------------------------------------
 vtkMRMLNodeNewMacro(vtkMRMLTableStorageNode);
 
+const char* COMPONENT_SEPERATOR = "_";
+
 //----------------------------------------------------------------------------
 vtkMRMLTableStorageNode::vtkMRMLTableStorageNode()
 {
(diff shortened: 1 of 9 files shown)


Libs/vtkITK/vtkITKLabelShapeStatistics.h
```diff
@@ -0,0 +1,103 @@
...
+#include <vtkMatrix4x4.h>
+#include <vtkSmartPointer.h>
+#include <vtkVector.h>
+
+// std includes
+#include <map>
+
+class vtkPoints;
+
+/// \brief ITK-based utilities for calculating label statistics.
```
lassoan: Please make the description a bit more detailed and add link to corresponding ITK class where meaning of input flags and computed values are defined.

Libs/vtkITK/vtkITKLabelShapeStatistics.h
```diff
@@ -0,0 +1,103 @@
...
+
+  void ClearPerimeter();
+  void AddPerimeter(int labelValue, double perimeter);
+  double GetPerimeter(int labelValue);
+
+  void ClearRoundness();
+  void AddRoundness(int labelValue, double roundness);
+  double GetRoundness(int labelValue);
+
+  void ClearFlatness();
```
lassoan: There are a number of other parameters, such as principal axes directions that would be useful to expose. How have you decided what to include?
It would be nice to expose all possible parameters, but as of now, it would be a lot of copy-paste. Maybe we could add macros or allow enabling/disabling computation and getting values via string keys. vtkVariant could be used as return value type.

Modules/Scripted/SegmentStatistics/SegmentStatisticsPlugins/LabelmapSegmentStatisticsPlugin.py
```diff
@@ -94,4 +181,36 @@ def getMeasurementInfo(self, key):
...
+                                   unitsDicomCode=self.createCodedEntry("mm2", "UCUM", "squared millimeters", True))
+
+    info["roundness"] = \
+      self.createMeasurementInfo(name="Roundness", description="Roundness", units="")
+
+    info["flatness"] = \
+      self.createMeasurementInfo(name="Flatness", description="Flatness", units="")
+
+    info["obb_origin_ras"] = \
+      self.createMeasurementInfo(name="OBB Origin (RAS)", description="OBB origin", units="")
```
lassoan: we can be verbose in the description - spell out oriented bounding box

lassoan:
Add "componentNames" property to table schema. Construct csv file columns names as columnName + "_" + components[i].

Schema example:


columnName | componentNames
-- | --
Centroid | r\|a\|s
Displacement | radial\|tangential
Vector | 0\|1\|2

Sunderlandkyl:
Pull request updated.
- tsv/csv can now read/write multi component columns with the addition of the schema columns listed above.
- vtkITKLabelShapeStatistics now returns a vtkTable as output that contains all calculated statistics.

Libs/MRML/Core/vtkMRMLTableStorageNode.cxx
```diff
@@ -340,106 +340,273 @@ bool vtkMRMLTableStorageNode::ReadTable(std::string filename, vtkMRMLTableNode*
     return 0;
     }
 
-  vtkSmartPointer<vtkTable> table = vtkSmartPointer<vtkTable>::New();
-  for (int col = 0; col < rawTable->GetNumberOfColumns(); ++col)
+  // Struct for managing columns
+  typedef struct
+  {
+    std::string ColumnName = "";
```
lassoan: string is an object, has default constructor, so no need for `= "";`

Libs/MRML/Core/vtkMRMLTableStorageNode.cxx
```diff
@@ -340,106 +340,273 @@ bool vtkMRMLTableStorageNode::ReadTable(std::string filename, vtkMRMLTableNode*
...
+    std::vector<vtkAbstractArray*> RawComponentArrays;
+    int DataType = -1;
```
lassoan: Maybe ScalarType or ValueType could be more clear (for me, "data type" definition includes everything - that it is a scalar/vector, with N components)

Libs/MRML/Core/vtkMRMLTableStorageNode.cxx
```diff
@@ -340,106 +340,273 @@ bool vtkMRMLTableStorageNode::ReadTable(std::string filename, vtkMRMLTableNode*
...
-      continue;
+      ColumnInfo columnInfo;
+      columnInfo.ColumnName = schemaColumnNameArray->GetValue(schemaRowIndex);
+      columnInfo.DataType = tableNode->GetColumnValueTypeFromSchema(columnInfo.ColumnName);
+
+      std::vector<vtkAbstractArray*> componentArrays;
+      std::string componentNamesStr = schemaComponentNamesArray->GetValue(schemaRowIndex);
+      if (componentNamesStr.empty())
+        {
+        vtkAbstractArray* rawColumn = rawTable->GetColumnByName(columnInfo.ColumnName.c_str());
```
lassoan: add check if rawColumn is not nullptr

Libs/MRML/Core/vtkMRMLTableStorageNode.cxx
```diff
@@ -340,106 +340,273 @@ bool vtkMRMLTableStorageNode::ReadTable(std::string filename, vtkMRMLTableNode*
...
+        componentArrays.push_back(rawColumn);
+        }
+      else
+        {
+        std::stringstream ss(componentNamesStr);
+        std::string componentName;
+        while (std::getline(ss, componentName, '|'))
+          {
+          std::string componentColumnName = columnInfo.ColumnName + "_" + componentName;
+          vtkAbstractArray* rawColumn = rawTable->GetColumnByName(componentColumnName.c_str());
```
lassoan: add rawColumn nullptr check

Libs/MRML/Core/vtkMRMLTableStorageNode.cxx
```diff
@@ -340,106 +340,273 @@ bool vtkMRMLTableStorageNode::ReadTable(std::string filename, vtkMRMLTableNode*
...
+        continue;
+        }
+      columnInfo.ColumnName = column->GetName();
+      columnInfo.DataType = tableNode->GetColumnValueTypeFromSchema(columnInfo.ColumnName);
+      columnInfo.RawComponentArrays.push_back(column);
+      columnDetails.push_back(columnInfo);
       }
+    }
+
+  vtkSmartPointer<vtkTable> table = vtkSmartPointer<vtkTable>::New();
```
lassoan: Add 1-2 summary sentences about what is going to happen here. Since the method is a bit too long, you may even consider to split it into a few methods. This is a potential split point position.

Libs/MRML/Core/vtkMRMLTableStorageNode.cxx
```diff
@@ -340,106 +340,273 @@ bool vtkMRMLTableStorageNode::ReadTable(std::string filename, vtkMRMLTableNode*
...
+         numberOfTuples = std::max(numberOfTuples, componentArray->GetNumberOfTuples());
+         }
+       typedColumn->SetNumberOfTuples(numberOfTuples);
+
+       int componentIndex = 0;
+       for (vtkAbstractArray* componentColumn : components)
+         {
+         if (componentColumn == nullptr)
+           {
+           continue;
```
lassoan: we should log a warning and maybe even insert a dummy column (with all empty values)

Libs/MRML/Core/vtkMRMLTableStorageNode.cxx
```diff
@@ -340,106 +340,273 @@ bool vtkMRMLTableStorageNode::ReadTable(std::string filename, vtkMRMLTableNode*
...
+           }
+
+         vtkStringArray* rawColumn = vtkStringArray::SafeDownCast(componentColumn);
+         if (!rawColumn)
+           {
+           continue;
```
lassoan: we should log a warning and maybe even insert a dummy column (with all empty values)

Libs/MRML/Core/vtkMRMLTableStorageNode.cxx
```diff
@@ -340,106 +340,273 @@ bool vtkMRMLTableStorageNode::ReadTable(std::string filename, vtkMRMLTableNode*
...
+      {
+      std::string columnName;
+      if (oldColumn->GetName())
+        {
+        columnName = oldColumn->GetName();
+        }
+      for (int componentIndex = 0; componentIndex < numberOfComponents; ++componentIndex)
+        {
+        std::stringstream newColumnNameSS;
+        newColumnNameSS << columnName << "_";
```
lassoan: add a const for "_" (no need for class member just file static)

Libs/MRML/Core/vtkMRMLTableStorageNode.cxx
```diff
@@ -530,6 +697,63 @@ bool vtkMRMLTableStorageNode::WriteSchema(std::string filename, vtkMRMLTableNode
       }
     }
 
+  vtkStringArray* componentNamesArray = vtkStringArray::SafeDownCast(schemaTable->GetColumnByName("componentNames"));
+  if (componentNamesArray == nullptr)
+    {
+    vtkNew<vtkStringArray> newArray;
+    newArray->SetName("componentNames");
+    newArray->SetNumberOfValues(schemaTable->GetNumberOfRows());
+    schemaTable->AddColumn(newArray.GetPointer());
```
lassoan: do we still need `GetPointer()`?

Libs/vtkITK/vtkITKLabelShapeStatistics.h
```diff
@@ -0,0 +1,76 @@
...
+
+  enum ShapeStatistic
+  {
+    Centroid,
+    OrientedBoundingBox,
+    FeretDiameter,
+    Perimeter,
+    Roundness,
+    Flatness,
+    LastStatistc,
```
lassoan: typo
also, use convention (_Last) to distinguish the last special element a bit more

Sunderlandkyl:
Made the changes. Running ctest now.

Sunderlandkyl:
Tests completed.
Several tests failed. but I think it's due to the fact that I haven't updated my superbuild in a while, and not due to this change: http://slicer.cdash.org/viewTest.php?onlyfailed&buildid=1800275

lassoan:
I've tested it and it works great! I'll merge as is.

Could you please add add in a follow-up pull request:
- principal moment and axes computation (similar to OBB but more robust)
- description of each metric (e.g., roundness: characterizes how much the object shape is similar to a sphere (0 = flat object, 1 = for spherical object), computed as ratio of ... and ...)
- display number of components in tables module column properties GUI (read-only for now)

The new vtkITK filter will be also usable in the future in Islands (effect for filtering not only based on size but also on shape, orientation, etc.). We may also extend the filter's interface in the future to allow computing grayscale metrics.

lassoan:
Thank you, merged:
https://github.com/Slicer/Slicer/commit/4e48c12277a872a03ac021ffbc8257d75d047745

PR #1302, opened 2020-01-12
BUG: Fix stdout fetching in export Slicer Data Bundle

cpinter:
The problem (on Windows) when exporting the scene as a Secondary Capture DICOM using Slicer Data Bundle was that the dump file after the dcmdump had problems:
- The line endings were replaced by unescaped "\r\n" strings
- The whole dump was wrapped in a b"..."
So after adding the candygram string that identified the file for the importer as a Slicer Data Bundle, the first tag could not be processed by the dump2dcm call due to the double quote at the end (and there were error messages about too long lines due to the unescaped line endings), and was stripped. Because of this, the examine step failed and the file could not be loaded as a Slicer scene.
Solution was used from https://forum.qt.io/topic/85064/qbytearray-to-string


cpinter:
@pieper I found the problem with MRB SC export/import, see above. Not sure if this is only a Windows issue, since you said a round-trip test was successful on Mac recently. In any case, it fixes the problem with export for me.

cpinter:
Another smaller issue is that the Scalar Volume plugin also tells that it can load the series when in reality it cannot ("Could not load: Slicer Data Bundle as a Scalar Volume"). The problem is that both loadables are selected after the examine step. Maybe they return with the same confidence. I think this is due to using most tags from the "reference file", and it looks both a valid scalar volume and a SC. I have not verified these conjectures yet though, but I thought I'd share it now in case you have a better idea.

image

lassoan:
The mrb import plugin can be very confident in loading the bundle file, so it should return a very high confidence value (0.9?). If one confidence value stands out then then only that loadable should be selected (if still multiple loadables were selected then it is a bug).

pieper:
Thanks for the fixes @cpinter !

I agree with @lassoan about the confidence value.  If you can add that fix and commit I can test it on mac/linux.

cpinter:
I will, thanks!

cpinter:
I made the change, and confirmed that now there is no error on import. I updated the PR. @pieper can you please try this change on Mac? I don't have access to Mac anymore... Thanks!

pieper:
@cpinter tested the branch on mac and it works 👍 

A few lingering things just to mention though (not critical, but since we are looking at this):

* it gets saved into a custom temporary directory, which is printed in the dialog but no place else, so the only way to use it currently is to check the 'import exported data' option so it shows up back in the database.  Maybe we should allow people to save to a custom location (means changing Modules/Scripted/DICOMLib/Widgets/qSlicerDICOMExportDialog.cxx).

* The only way to get to the scene export that I see is by selecting a node in the data tree and picking export from the context menu.  Maybe we should expose it at a higher level but I'm not sure where that would be in the Subject Hierarchy interface (right click on the background maybe?)

* We can't currently drag-and-drop these files back on Slicer, but we could add a hook for that (like discussed [here](https://discourse.slicer.org/t/proposition-name-volumes-imported-via-data-by-series-description/9779/13))

Modules/Scripted/DICOMPlugins/DICOMSlicerDataBundlePlugin.py
```diff
@@ -65,6 +65,7 @@ def examineFiles(self,files):
         loadable.name = name + ' - as Slicer Scene'
         loadable.selected = True
         loadable.tooltip = 'Appears to contain a slicer scene'
```
lassoan: If we are confident that it is a Slicer scene (I think we should be fairly certain) then the tooltip text should be a bit more confident, too

cpinter:
@pieper thanks for testing!

* I agree completely, we should allow saving to custom location. I was planning to do this a bit later, once more of the function I'm working on is ready.

* Yes, I noticed this, and I couldn't think of an obvious solution. Right-clicking the background is a good idea! But then it brings up the issue that I've had with right-clicking the background :) Many times there is very little space that is background (when there are enough items to fill the allocated height). It is always an option to show the scene item, but it would introduce an extra not-so-useful line right on the top, and consume some horizontal place too, due to the tree structure, so I didn't want to go that way. Maybe I should add more padding if possible? Any thoughts on this?

* It would be nice to allow drag&drop. Let me come back to this when working on the custom location.

lassoan:
If I had to choose, I would prefer a root element over increased whitespace, because users might not realize that right-clicking on whitespace has a special meaning.

cpinter:
Hi @pieper and @lassoan,

If there is no objection I'd like to integrate this. One last thing that I'd like to sort out in this PR is what I described above about the temporary files:

> some temporary files are in the folder after exporting. In particular what bothers me is the template.dcm file, which I'm not sure the DICOM importer tries to import or not when importing the folder containing the bundle. So I propose removing these files (template.dcm, dump.dcm, and scene folder - maybe keep the scene.zip?) after a successful export.

What do you think about this?

The rest of the issues (about usability) I'd address at a later time.

Thanks!

pieper:
Cleaning up any unneeded files makes perfect sense.

Overall I'm very happy to see this integrated.  Usability issues can be further addressed in the future, maybe when we have better ideas about use cases.

Thanks 👍 !

PR #1303, opened 2020-01-13
ENH: Enable ResampleOrientedImageToReferenceGeometry for all scalars

Sunderlandkyl:
Previously, only VTK_UNSIGNED_CHAR, VTK_UNSIGNED_SHORT, and VTK_SHORT were supported in ResampleOrientedImageToReferenceGeometry.
This commit removes the restriction and allows all scalar types to be used.

### DIFF
--- a/Libs/vtkSegmentationCore/vtkOrientedImageDataResample.cxx
+++ b/Libs/vtkSegmentationCore/vtkOrientedImageDataResample.cxx
@@ -487,16 +487,6 @@ bool vtkOrientedImageDataResample::ResampleOrientedImageToReferenceGeometry(vtkO
     return false;
     }
 
-  // Only support the following scalar types
-  int inputImageScalarType = inputImage->GetScalarType();
-  if ( inputImageScalarType != VTK_UNSIGNED_CHAR
-    && inputImageScalarType != VTK_UNSIGNED_SHORT
-    && inputImageScalarType != VTK_SHORT )
-    {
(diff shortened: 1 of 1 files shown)


jcfr:
Thanks :+1: 

Where you able to find more details in the comment introducing the original restriction ?

Sunderlandkyl:
I couldn't find the reason for the original restriction. Perhaps @cpinter or @lassoan know.

jcfr:
cc: @pieper

pieper:
I do not know - I don't see any reason to restrict the types.

cpinter:
I don't remember the actual reason. My best guess is that we originally created the class for resampling labelmaps.

Sunderlandkyl:
Right, that makes sense.

The issue that I'm running into is that apparently Freesurfer sometimes saves labelmaps using float. The labelmaps import fine into segmentations, but fail suddenly when exporting/resampling.

Probably we could also perform some intelligent casting on the imported labelmap to convert it to an integer scalar type.

lassoan:
> My best guess is that we originally created the class for resampling labelmaps.

Yes, we originally developed this method for resampling segmentation (labelmap representation). At some point we updated the underlying infrastructure to be completely generic, so this check can be removed.

lassoan:
Thank you, merged.

https://github.com/Slicer/Slicer/commit/4800add31e82f16c585b439aa31683b1b2b6ef22

PR #1304, opened 2020-01-17
BUG: Fix handling of legacy node reference attributes in scene files

Sunderlandkyl:
Previously, if both legacy and current node reference attributes existed with the XML tag for a MRMLNode, both reference IDs would be added to the node.
This commit removes the writing of the legacy node reference attributes to the XML tag, and ensures that for scenes that contain both types of attributes, the legacy format is ignored if the same node reference is found in the "references" attribute.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLNode.cxx
@@ -367,53 +367,63 @@ void vtkMRMLNode::ReadXMLAttributes(const char** atts)
   vtkMRMLReadXMLBooleanMacro(undoEnabled, UndoEnabled)
   vtkMRMLReadXMLEndMacro();
 
-  std::map<std::string, std::string> references;
-
+  std::set<std::string> references;
+  const char** xmlReadAtts = atts;
   const char* attName;
   const char* attValue;
-  while (*atts != nullptr)
+  while (*xmlReadAtts != nullptr)
     {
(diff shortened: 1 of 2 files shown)


cpinter:
Great! Thanks, Kyle :)

Libs/MRML/Core/vtkMRMLNode.cxx
```diff
@@ -368,52 +368,60 @@ void vtkMRMLNode::ReadXMLAttributes(const char** atts)
...
-       {
-       std::stringstream ss(attValue);
-       while (!ss.eof())
-         {
-         std::string id;
-         ss >> id;
-         if (!id.empty())
-           {
-           if (references.find(id) == references.end() ||
-             references.find(id)->second != referenceRole)
```
lassoan: The same node may be referenced with different roles (e.g., in SlicerRT DVH plot, the same nodes are referenced using DoseVolumeHistogram.dvhPlotSeries and plotSeries role). In such cases the logic is still not correct because in "references" variable we only store the last encountered role name.

We should just keep a list of roles (can be a std::set<std::string>) and if we find that a reference role has been encountered before then we don't add any nodes with it again.

Sunderlandkyl:
@lassoan I made the change. Anything else that needs to be done?

lassoan:
Thank you, I've merged it now (r28748).

PR #1305, opened 2020-01-17
ENH: Use TBB for threading in ITK

lassoan:
TBB offers better load balancing for unevenly complex work distribution and plays better with non-ITK load on the CPU.

As a side effect, it also improves granularity of filter progress reporting (https://discourse.slicer.org/t/bilateral-filter-not-working/9797/13).

Let's NOT INTEGRATE this until a Windows factory machine is down. Once a new nightly build is available then we can integrate this and have the same Slicer version with/without TBB for clean performance comparison.


lassoan:
Integrated in r28749.

PR #1306, opened 2020-01-17
COMP: Fix compile error in vtkITKLabelShapeStatistics

Sunderlandkyl:
vtkSetMacro and vtkGetMacro do not work with std::vector.
This commit fixes vtkITKLabelShapeStatistics by using vtkSetStdVectorMacro from vtkAddon instead.


Sunderlandkyl:
Not sure why this didn't get caught before, but I encountered the error when compiling Slicer on a different computer.

lassoan:
Merged, thank you!

PR #1309, opened 2020-01-22
Model and markups save in LPS by default

lassoan:
Slicer has been using LPS coordinate system when storing images, transforms, and markups. All medical image computing software (maybe except a few very old ones) uses LPS coordinate system, too. Slicer started writing coordinate system name in mesh files (STL, VTK, VTP, OBJ, PLY) but still used RAS by default for saving models. This caused inconsistencies when Slicer-exported data was used in third-party software. See https://issues.slicer.org/view.php?id=4445 for details.

This commit changes default coordinate system in saved models to LPS. When loading a model that has coordinate system information embedded in the file then that coordinate system is used. If the model does not contain coordinate system information then the CoordinateSystem property of model storage node is used. CoordinateSystem is set to LPS by default but the user can change it in "Add data" dialog (in option column)

Backward compatibility:
- all old Slicer scenes are loaded correctly (if model coordinate system is not specified in the mrml file then it is assumed to be RAS)
- all standalone model files that Slicer created since 2017-09-27 (Slicer-4.8.0) are read correctly (except obj files created by Slicer-4.6 and Slicer-4.8)
- all files created by external software that use LPS coordinate system are read correctly
- files that require manual selection of RAS coordinate system when loading without a scene: model files created by Slicer-4.6 (2017-09-27) and earlier; and obj files created by Slicer-4.6 and Slicer-4.8 (between 2016-10-11 and 2018-03-26).

Also cleaned up markups node coordinate system writing (use LPS/RAS string instead of magic numbers, similar style as model nodes, added tests).

### DIFF
--- a/Libs/MRML/Core/Testing/vtkMRMLModelStorageNodeTest1.cxx
+++ b/Libs/MRML/Core/Testing/vtkMRMLModelStorageNodeTest1.cxx
@@ -1,12 +1,12 @@
 /*=auto=========================================================================
 
-  Portions (c) Copyright 2005 Brigham and Women's Hospital (BWH)
-  All Rights Reserved.
+Portions (c) Copyright 2005 Brigham and Women's Hospital (BWH)
+All Rights Reserved.
 
-  See COPYRIGHT.txt
-  or http://www.slicer.org/copyright/copyright.txt for details.
+See COPYRIGHT.txt
+or http://www.slicer.org/copyright/copyright.txt for details.
 
-  Program:   3D Slicer
+Program:   3D Slicer
 
 =========================================================================auto=*/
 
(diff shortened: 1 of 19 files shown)


lassoan:
This is ready for integration now. @pieper @jcfr @cpinter please have a look. I would like to merge this within a few days.

pieper:
This sounds like a reasonable plan, and doing the break at Slicer5 release is the best time (easiest to document and explain).

One idea would be that when reading an ambiguous file (one where the coordinate reference system is not declared) we could generate a warning dialog to call attention to the new behavior (the dialog can have a "don't warn me again" setting).  It would also be good to always generate a log warning when reading an ambiguous file.

lassoan:
There is an info log each time when reading a model file that does not have coordinate system info embedded. I made it first a warning but it generated too many false alarms (all the old scene files and any file imported from third-party software generated a warning).

Adding a popup would not be very easy and it is quite likely that it would be turned off very soon and would not be very helpful. I would also expect that most files will be interpreted correctly by default and if not then the flip will be obvious. We can also add stronger warnings if it turns out to be a bigger issue than we expected.

pieper:
We discussed this a bit on the slicer hangout today and I'm still of the opinion that an log message is two subtle a warning for this breaking change.  I would rather that people got a clear and hard to miss warning.  People could easily fall into a trap of loading an old file and not realizing it was flipped.  I can think of several situations where this might be non-obvious to the user.

As a start, in [this block](https://github.com/Slicer/Slicer/pull/1309/files#diff-5ba57bc567ee43c855189a3b4f93aefeR371-R373) we should keep a flag on the storage node that indicates that the coordinate system from the original file was ambiguous.  That way we would be able to check the scene for ambiguous nodes after the Add Data dialog (or other operations) took place.

We could expose that value as a warning icon in the Data module tree so people would know which data was ambiguous.

It would also be good to expose the functionality of `vtkMRMLModelStorageNode::ConvertBetweenRASAndLPS`, perhaps as a subject hierarchy plugin for the context menu, so that it could be fixed easily.

lassoan:
Thanks for the review and comments. I'll get back to this later when I have more time.

lassoan:
Discussed at the Slicer developer meeting yesterday and agreed that it can be merged as is.
Merged r28794.

PR #1310, opened 2020-01-23
ENH: Add loadFromDICOMWeb method to DICOMUtils

Sunderlandkyl:
@pieper

### DIFF
--- a/Modules/Scripted/DICOMLib/DICOMUtils.py
+++ b/Modules/Scripted/DICOMLib/DICOMUtils.py
@@ -676,3 +676,61 @@ def loadLoadables(loadablesByPlugin, messages=None, progressCallback=None):
   slicer.mrmlScene.RemoveObserver(sceneObserverTag)
 
   return loadedNodeIDs
+
+def importFromDICOMWeb(dicomWebEndpoint, studyUID, seriesUID=None, accessToken=None):
+  """
+  Downloads and imports DICOM series from a DICOMweb instance.
+  Example usage:
+    from DICOMLib import DICOMUtils
+    
+    loadedUIDs = DICOMUtils.importFromDICOMWeb(dicomWebEndpoint="https://yourdicomweburl/dicomWebEndpoint",
+                                             studyUID="2.16.840.1.113669.632.20.1211.10000509338")
+                                             accessToken="YOUR_ACCESS_TOKEN")
+  :param dicomWebEndpoint: Endpoint URL for retrieving the study/series from DICOMweb
+  :param studyUID: UID for the study to be downloaded
+  :param seriesUID: UID for the series to be downloaded. If not specified, all series will be downloaded from the study
+  :param accessToken: Optional access token for the query
+  :return: List of imported study UIDs
+  """
+
+  from dicomweb_client.api import DICOMwebClient
+  import random
+
+  if accessToken is None:
+    client = DICOMwebClient(url = dicomWebEndpoint)
+  else:
+    client = DICOMwebClient(
+              url = dicomWebEndpoint,
+              headers = { "Authorization": "Bearer {}".format(accessToken) },
+              )
+    
+
+  seriesList = client.search_for_series(study_instance_uid=studyUID)
+  seriesUIDs = []
+  if not seriesUID is None:
+    seriesUIDs = [seriesUID]
+  else:
+    for series in seriesList:
+      currentSeriesUID = series['0020000E']['Value'][0]
+      seriesUIDs.append(currentSeriesUID)
+
+  fileNumber = 0
+  for currentSeriesUID in seriesUIDs:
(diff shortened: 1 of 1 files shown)


Modules/Scripted/DICOMLib/DICOMUtils.py
```diff
@@ -668,3 +668,35 @@ def onNodeAdded(caller, event, calldata):
   slicer.mrmlScene.RemoveObserver(sceneObserverTag)
 
   return loadedNodeIDs
+
+def loadFromDICOMWeb(dicomWebEndpoint, studyUID, seriesUID="", accessToken=""):
```
lassoan: Unless there is a strong reason to do otherwise, use None as default parameter.

lassoan: Add method documentation. Preferably also include an example of how it can be used.

Modules/Scripted/DICOMLib/DICOMUtils.py
```diff
@@ -668,3 +668,35 @@ def onNodeAdded(caller, event, calldata):
...
+      seriesUIDs.append(currentSeriesUID)
+
+  fileNumber = 0
+  for currentSeriesUID in seriesUIDs:
+    instances = client.retrieve_series(
+      study_instance_uid=studyUID,
+      series_instance_uid=currentSeriesUID)
+
+    outputDirectory = (slicer.dicomDatabase.databaseDirectory +
+      "/" + studyUID + "/" + currentSeriesUID + "/")
```
lassoan: IDs may be very long and so when they are used for creating folders then they are often longer than maximum path length on Windows. Instead, you can create unique folder using date&time and/or hashes.

Modules/Scripted/DICOMLib/DICOMUtils.py
```diff
@@ -668,3 +668,35 @@ def onNodeAdded(caller, event, calldata):
...
+    if not os.access(outputDirectory, os.F_OK):
+      os.makedirs(outputDirectory)
+    for instance in instances:
+      filename = outputDirectory + str(fileNumber) + ".dcm"
+      instance.save_as(filename)
+      fileNumber += 1
+    importDicom(outputDirectory)
```
lassoan: it could be nice to return loaded node IDs

pieper:
Looks great to me, thanks.

hjmjohnson:
@Sunderlandkyl Can you please rebase, and address the seemingly minor improvements to documentation and default values?

Modules/Scripted/DICOMLib/DICOMUtils.py
pieper: I think the correct letter case for the method should be `loadFromDICOMweb` for consistency with the [official trademark style](https://www.dicomstandard.org/dicomweb/).  Then the variable name would be `dicomWEB`.  (I know, it looks awkward - perhaps we make a style exception here.)

Sunderlandkyl:
@hjmjohnson Sure, I'll update the pull request.

Modules/Scripted/DICOMLib/DICOMUtils.py
lassoan: I don't think that trademark style would apply to code (only to displayed text) but if we change the method name to "loadFromDICOMweb" then add a comment there that this is intentional (otherwise someone may "fix" it when comes across the coding style violation).

DICOMWeb and DICOMweb are about equally bad (hard to read/inconsistent with other) so I don't have a strong preference there. But dicomWeb looks more friendly and consistent to me than dicomWEB (it almost shouts at me).

pieper: Agreed!  I've started using 'dicom' more often than DICOM in regular text to avoid that extra screaming element.  

Anyway yeah, let's leave it as is.

Modules/Scripted/DICOMLib/DICOMUtils.py
Sunderlandkyl: This function doesn't load nodes into slicer, just downloads and imports into the database.
I've renamed the function to importFromDICOMWeb to reflect this, and return a list of loaded series UIDs.

lassoan:
Thank you, merged. I have also added registerSlicerURLHandler (https://github.com/Slicer/Slicer/commit/49df340fc046d0479be8ada53523d92643cbac6d) helper function, which registers Slicer as handler of slicer:// URLs on Windows.

Currently, if I type the example URL then no data set is loaded. Has the handling of DICOMweb URLs integrated into Slicer core?

Sunderlandkyl:
No, I have the code here for handling urls. I'll make the pull request shortly.

lassoan:
Thank you, it's not urgent at all, I just wanted to test this to see if it works.

PR #1313, opened 2020-02-03
BUG: Fix replace for shared labelmaps with SetBinaryLabelmapToSegment

Sunderlandkyl:
Calling vtkSlicerSegmentationsModuleLogic::SetBinaryLabelmapToSegment with replace mode previously copied the contents of the modifier labelmap to the segment labelmap without checking to see if it is a shared labelmap.
Fixed by checking if the segment labelmap is shared. If the segment is shared the label value cleared for the segment and masking operations are used instead.


lassoan:
Thank you, merged in r28752.

PR #1314, opened 2020-02-03
ENH: Add more segment label statistics and display table component info

Sunderlandkyl:
New label shape statics:
- Elongation
- Principal moments
- Principal axes

Adds component information (component count and component names) within qSlicerTableColumnPropertiesWidget. Additional information is only shown if there is more than one component.

image

Also:
- Fixes a bug in qSlicerTableCOlumnPropertiesWidget that caused many connections to be added to the QLineEdit textEdited() signal.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLTableNode.h
+++ b/Libs/MRML/Core/vtkMRMLTableNode.h
@@ -147,7 +147,7 @@ public:
   /// Remove all columns from the table (including associated properties).
   /// Returns with true on success.
   bool RemoveAllColumns();
-  
+
   ///
   /// Add an empty row at the end of the table
   /// Returns the index of the inserted row or -1 on failure.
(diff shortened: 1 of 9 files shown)


lassoan:
Thank you!

I'm not sure if circleci has re-run the checks after your force-push. Could you please check?

Libs/MRML/Core/vtkMRMLTableNode.h
```diff
@@ -291,6 +291,19 @@ class VTK_MRML_EXPORT vtkMRMLTableNode : public vtkMRMLStorableNode
   /// Get value type id from string (uses vtkImageScalarTypeNameMacro)
   static std::string GetValueTypeAsString(int valueType);
 
+  /// TODO
```
lassoan: Could you complete these?

Sunderlandkyl: Done

lassoan:
I've tested this a bit more and it works really well. I've just found on major issue that would be important to fix before merging (and a few tiny ones that would be nice to address).

1. Table load fails: I've computed full segment statistics, saved to .tsv file, and when loaded the .tsv file using Add data, a table with a single empty row showed up (there were many errors and warnings in the log).

Saved table+schema:
https://1drv.ms/u/s!Arm_AFxB9yqHuqUzR4aUiJGm9_gGkQ?e=oFa0Nz

2. Minor issues

- when creating a new table node in Tables module, these warnings are logged a couple of times: `[WARNING][Qt] 05.02.2020 09:57:59 [] (unknown:0) - class QString __cdecl qSlicerTableColumnPropertiesWidget::columnProperty(class QString) const  failed: table node is not selected`
- "long" column type is not displayed in table module GUI (when selecting "Number of voxels" column in segment statistics results table,  Column properties / Data type combobox does not display any type)
- in the schema file, we have "0" in componentNames column for all columns that have only a single component. If the user hasn't specified any component name for a single-components column then the corresponding component name entry should be left empty in the schema file.

Sunderlandkyl:
Updated the vtkMRMLTableStorageNodeTest1 to check reading/writing multi-component tables with a schema.

lassoan:
Awesome, thanks a lot, I've merged this now.

This will be a very useful feature for many projects. Please write an announcement on discourse (tagged with "feature"). Maybe also try to find those couple of topics on discourse where people asked about "diameter" or bounding box size of segments and add a link to the announcement post.

To create a nice illustration for oriented bounding box, you can use the script referred to in this post: https://discourse.slicer.org/t/reslice-reorient-slice-view/10093/12?u=lassoan

Sunderlandkyl:
The updated vtkMRMLTableStorageNodeTest1 is failing for some reason:
http://slicer.cdash.org/testDetails.php?test=9907802&build=1817244

I'm looking into it.

PR #1316, opened 2020-02-06
ENH: Load drag-and-dropped files as DICOM while in DICOM module

lassoan:
To simplify DICOM import (minimize number of clicks), this PR changes drag-and-drop behavior so that while the DICOM module is active, files and folders are always interpreted as DICOM data (without asking the user).


pieper:
Looks good to me 👍  thanks!

We might want to make a settings panel option to have this behavior all the time.  That way people who do a lot of work with dicom files would have more predictable and easier behavior.

lassoan:
Merged.

> We might want to make a settings panel option to have this behavior all the time. That way people who do a lot of work with dicom files would have more predictable and easier behavior.

Good idea, we should keep simplifying DICOM import/loading.

PR #1317, opened 2020-02-06
BUG: Fix table read/write error

Sunderlandkyl:
This commit ensures that single component vectors that have no component name do not return a component name of "0".


Sunderlandkyl:
@lassoan this fix was meant to be in the previous pull request, however it somehow got removed. I probably missed staging this change.

lassoan:
I've forced-push a stylistic change. Before merging, it would be nice to ensure that generated component names are unique.

Sunderlandkyl:
Component names should now be unique when saving.

lassoan:
Thanks a lot, merged (r28755).

PR #1318, opened 2020-02-10
ENH: Improve performance of segment island effect

Sunderlandkyl:
Adds a NodeModify context to the split segments function to reduce the number of modified events.
Also wraps the color generated from vtkMRMLSegmentationDisplayNode::GenerateSegmentColor() around if it goes out of bounds of the generic anatomy color table node. This prevents a large number of errors from being printed to the console if the number of segments exceeds the number of colors in the table.


lassoan:
Thank you, merged!

PR #1319, opened 2020-02-10
ENH: Use UTF-8 as string encoding

lassoan:
In general, all std::string or char buffers in Slicer are now expected to contain UTF-8 encoded string. This is what was the default on Mac OSX and Linux and it has been made available as an option on Windows as well. This is where VTK and ITK (and most other software libraries) are moving towards, too.

This will allow loading and saving files that has special characters in their names.

Requires yet unmerged changes in CTK: https://github.com/commontk/CTK/pull/902


Libs/MRML/Core/Testing/cube-utf8.mrml
```diff
@@ -1,4 +1,43 @@
-<MRML>
-<Camera
- id="vtkMRMLCameraNode1" name="camera(攝像頭)"  hideFromEditors="true"  position="1.26247 0.628757 -0.668912" focalPoint="0 0 0" viewUp="-0.566328 0.729902 -0.382773" parallelProjection="false" parallelScale="0" active="true"></Camera>
-</MRML>
+<?xml version="1.0" encoding="utf-8"?>
+<MRML>
+
+    <Camera
+        id="vtkMRMLCameraNode1"
+        name="camera(ae=æ AE=Æ a:=ä A:=Ä o|=ø O|=Ø o:=ö O:=Ö ao=å Ao=Å)"
```
pieper: Is this what was intended?

Libs/MRML/Core/Testing/cube-utf8.mrml
```diff
@@ -1,4 +1,43 @@
...
+    </Camera>
+
+    <VolumeArchetypeStorage
+        id="vtkMRMLVolumeArchetypeStorageNode1" name=""  hideFromEditors="true"
+        fileName="TestData/fixed.nrrd" >
+    </VolumeArchetypeStorage>
+
+    <VolumeDisplay
+        id="vtkMRMLScalarVolumeDisplayNode1"
+        name="árvíztűrő tükörfúrógép"
```
pieper: Is this what was intended?

Libs/MRML/Core/Testing/cube-utf8.mrml
lassoan: Yes. Does it look like an accident? It is a couple of Nordic characters with how it should approximately look like (using ASCII characters) and the real Unicode character.

Libs/MRML/Core/Testing/cube-utf8.mrml
lassoan: Yes, this is a well-known phrase that contain all the accented characters of Hungarian language. (to be precise: well-known to Hungarians)

jcfr:
Thanks for the work on this :pray: 

I didn't have a chance to look at this PR or the one in CTK. Could we add this to the agenda for tomorrow dev meeting ?

Libs/MRML/Core/Testing/cube-utf8.mrml
pieper: Right - I wasn't sure if this was on purpose.  It looked like something that might have come from some autoconvert, but now that you explain it sounds like a good test.

Libs/MRML/Core/Testing/cube-utf8.mrml
pieper: 😄

lassoan:
If there is no further feedback then I'll merge this tomorrow.

jcfr:
> If there is no further feedback then I'll merge this tomorrow.

Shouldn't we finalize this one first https://github.com/Slicer/SlicerExecutionModel/pull/124

I am assuming the warnings et all discussed there will still apply.

lassoan:
SEM is quite independent: if the UTF-8 branch gets merged then non-ascii strings can be transferred correctly between Slicer and the CLI. Currently, we don't support non-ascii strings anyway, so it would not be a regression.

I've fixed the test failure in SEM. I don't think we should fix the warning, it is just a warning that the mt.exe tool included in that Windows SDK does not know this option yet (it was introduced after that Windows SDK was released). Despite the warning, mt.exe can still embed the manifest and the built exexutable works well on recent Windows 10 systems.

lassoan:
@jcfr Should I wait for merge of https://github.com/Slicer/SlicerExecutionModel/pull/124 or can proceed with merging this without SEM update for now (and then update it when you have a chance to review that)?

This merge should happen before the git switch, as many files are modified.

jcfr:
> I don't think we should fix the warning, it is just a warning that the mt.exe tool included in that Windows SDK does not know this option yet (it was introduced after that Windows SDK was released)

Makes sense. 

Could you add an exception to https://github.com/Slicer/Slicer/blob/master/CMake/CTestCustom.cmake.in#L45 

> Should I wait for merge of Slicer/SlicerExecutionModel#124

It has just been merged.

Once the CTest exception has been added. I have no problem with integrating this PR. :tada: :pray:

lassoan:
Thanks for these. If you don't mind, I would add the warning suppression in a few days, after we saw which systems report the manifest authoring warning and to confirm that the change is necessary and indeed fixes the warning.

lassoan:
Thanks for the reviews, this is merged now in r28783. I'll add the CMake warning exception within a few days.

PR #1320, opened 2020-02-10
ENH: Display name rather than description in Segment Statistics options

Sunderlandkyl:
Previously the description was used in the Segment Statistics options widget, which could be long.
This changes the displayed text to be the name instead. The description (and other relevant info) are still displayed in the tooltip.


lassoan:
Works very well, thank you! Merged.

PR #1321, opened 2020-02-10
BUG: Remove broken method of gdb usage.

JEHoctor:
Users should attach to a running Slicer using the method described here:
https://www.slicer.org/wiki/Documentation/Nightly/Developers/Tutorials/Debug_Instructions#GDB_debug_by_attaching_to_running_process_.5BRECOMMENDED.5D

I assume that ddd usage is affected in the same was as gdb usage, as ddd
is a graphical wrapper around debuggers including gdb. To that end, this
form of ddd usage is removed as well.


JEHoctor:
Thanks @jcfr for encouraging me to submit this PR.

This removes the `--gdb` command line flag on Linux. Using this flag doesn't currently work, and attaching GDB after starting the application is the currently recommended method.

Concerns:
1. This patch is founded on the assumption that the Slicer developers do not want to fix `--gdb` in the near term. If there is a fix to this functionality, adopting it may be preferable to removing the flag.
2. I also remove the `--ddd` flag because I _assume_ it has the same issues. I have not tested this.

pieper:
LGTM - thanks for the PR and the description 👍

jcfr:
Thanks @JEHoctor 

Outstanding first PR 🎉 🙏 

I will integrate tomorrow.

jcfr:
Closing. Fix integrated in [r28772](http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=28772)

PR #1323, opened 2020-02-11
BUG: Fix Segment Statistics principal axes calculation if OBB is off

Sunderlandkyl:
Principal axes are now transformed to RAS at the centroid position (was previously the OBB origin).
Added some additional error handling to prevent a Traceback in the event that a statistic could not be calculated by vtkITKLabelShapeStatistics.


lassoan:
Thank you, merged!

PR #1325, opened 2020-02-15
BUG: Allow reading of FreeSurfer models saved with vtk extension

Sunderlandkyl:
When imported into Slicer, FreeSurfer models are read using a vtkMRMLFreeSurferModelStorageNode.
When saving, the storage node has no write methods, resulting in it falling back on the parent implementation and saving as a vtk (or other vtkMRMLModelNode format).
Trying to read the saved vtk files will cause an error, since vtkMRMLFreeSurferModelStorageNode does not handle non-FreeSurfer file types.

This fixes the issue by adding a workaround to revert to the parent implementation if the file extension is vtk.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLFreeSurferModelStorageNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLFreeSurferModelStorageNode.cxx
@@ -120,6 +120,13 @@ int vtkMRMLFreeSurferModelStorageNode::ReadDataInternal(vtkMRMLNode *refNode)
   std::string extension = vtkMRMLStorageNode::GetLowercaseExtensionFromFileName(fullName);
   vtkDebugMacro("ReadDataInternal: extension = " << extension.c_str());
 
+  // Node was saved with vtk extension.
+  // Should be handled by vtkMRMLModelStorageNode.
+  if (extension == ".vtk")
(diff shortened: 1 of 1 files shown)


mikebind:
Thanks for taking a look at this and posting a fix, Kyl.  Will I be able to tell on this page if/when it has been incorporated into a nightly Slicer build for Windows?

sjh26:
When it is merged, a maintainer will make a comment  here with the details, and it will be available in the nightly build the next day.

lassoan:
The PR looks good to me.

@sjh26 could you merge it? (it's holiday today in Canada and I'm not sure when I will get to my computer)

sjh26:
Done!  Merged in https://github.com/Slicer/Slicer/commit/797bfed3c34a7735dfceb15b0f53f34c37f8ff6a (r28771)

Sunderlandkyl:
Thanks @sjh26!

lassoan:
@Sunderlandkyl [This commit](https://github.com/Slicer/Slicer/commit/a73fdbaf641dc7162953ade6f59d8e46865d609c) was supposed to fix FreeSurfer model saving. Could you check what caused the regression? (or maybe it solved a different issue?)

PR #1326, opened 2020-02-17
COMP: removing unused include from vtkImageGrowCutSegment.cxx

dzenanz:
This makes reusing this file outside of Slicer easier.


lassoan:
Merged in r28779.

PR #1327, opened 2020-02-18
ENH: Add surface based curve interpolation to markups

Sunderlandkyl:
Adds a Dijkstra's algorithm based pathfinding to generate an interpolated curve for open and closed curve markup types.
By default, the only metric used is the distance between each vertex, however scalar values can also optionally be used (see below).

- Add a combobox to select curve interpolation type to the markups module
- Convert vtkCurveGenerator to a vtkPolyDataAlgorithm
- Scalar weighted pathfinding:
  - The pathfinding edge cost can be weighted using the point scalars of the mesh (currently weight = distance / scalar^2)
  - The formula used for generating the scalar values can be specified with vtkMRMLMarkupsCurveNode::SetSurfaceScalarWeightFunction() using vtkFunctionParser syntax
  - All point scalars can be used in the function (ex. "scalar1 + scalar2").

image

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkCurveGenerator.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkCurveGenerator.cxx
@@ -17,17 +17,26 @@
 
 ==============================================================================*/
 
+// Markups MRML includes
 #include "vtkCurveGenerator.h"
 #include "vtkLinearSpline.h"
 
+// VTK includes
 #include <vtkCardinalSpline.h>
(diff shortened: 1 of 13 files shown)


lassoan:
@Sunderlandkyl I've accidentally committed this along with another commit (I'm so looking forward to getting rid of git-svn soon!), so I'm closing this pull request now. We'll start advertising the feature to users after the code review comments are addressed (especially the scene save/load problem).

PR #1328, opened 2020-02-19
BUG: Fix crash when running island effects on empty segment

Sunderlandkyl:
If the segment labelmap had a dimension of {0, 0, 0}, the vtkSlicerSegmentationsModuleLogic::GetAllLabelValues() would cause a crash by trying to run the empty image through vtkImageAccumulate.
Fixed by returning from the function if the dimensions of the image are 0.


lassoan:
Great, thank you, merged!

PR #1329, opened 2020-02-19
COMP: Fix setting EditorEffect cursor icons

chribaue:
Cursor icons for EditorEffects were not visualized in the actual slice view, but only the widget's top part (the red/green/yellow area)

### DIFF
--- a/Modules/Scripted/EditorLib/EditBox.py
+++ b/Modules/Scripted/EditorLib/EditBox.py
@@ -347,7 +347,7 @@ class EditBox(VTKObservationMixin):
       self.currentOption.__del__()
       self.currentOption = None
       for tool in self.currentTools:
-        tool.sliceWidget.unsetCursor()
+        tool.sliceWidget.sliceView().unsetViewCursor()
         tool.cleanup()
       self.currentTools = []
 
@@ -406,7 +406,7 @@ class EditBox(VTKObservationMixin):
       # make an appropriate cursor for the tool
       cursor = self.cursorForEffect(effectName)
       for tool in self.currentTools:
-        tool.sliceWidget.setCursor(cursor)
+        tool.sliceWidget.sliceView().setViewCursor(cursor)
 
   def cursorForEffect(self,effectName):
     """Return an instance of QCursor customized for the given effectName.


lassoan:
Merged. Note that Editor module has been deprecated for more than 3 years now. It will be moved out to an extension within 1-2 months, and it will not be maintained anymore. It is time to update any of your modules that still rely on the legacy Editor module to use Segment Editor module instead. We would be happy to help you if you have any questions.

chribaue:
I am aware that the legacy editor is marked deprecated. Our interactive segmentation tool already supports the new Segment Editor as well, but some people prefer to use the legacy Editor. And they think our Editor effect is broken because the icon does not appear, as shown in our tutorial videos.

I'd like to make the same change to the latest stable release of Slicer, since the issue also exists there. This way the tool will work as expected for the stable releases of Slicer4.

jcfr:
> I'd like to make the same change to the latest stable release of Slicer, since the issue also exists there. This way the tool will work as expected for the stable releases of Slicer4.

We do not plan to release `4.10.3`, this means the change will not be available within any of the `4.x` stable releases.

That said, since your changes is integrated to master, it will be part of tomorrow preview release.

jcfr:
>  but some people prefer to use the legacy Editor

It would be great to gather some specifics here and see how (and if) we could improve the current segment editor.

chribaue:
> It would be great to gather some specifics here and see how (and if) we could improve the current segment editor.

The new segmentation infrastructure has certainly many advantages over labelmaps that I would not want to miss. 

The Segment Editor's user-interface offers more options and information compared to the legacy Editor and requires an additional "Add" click before a user can start with the actual segmentation. For novice users and users that I re-trained on our Editor effect from using the legacy Editor to using the Segmentation Editor, these extras seemed to be more distracting and overwhelming, compared to using the legacy Editor. For their use case, there was no benefit of the more complex interface and user's who were familiar with the legacy Editor rather preferred the  simpler interface.

This is just a personal observation and after training the users were able to operate the Segment Editor just as well. So, I guess, it is only a small issue.

lassoan:
These are all very easy to address.

> Editor and requires an additional "Add" click before a user can start with the actual segmentation

You can pre-populate as many segments as you need, with colors, names, etc.

> extras seemed to be more distracting and overwhelming

You can show/hide various sections of the widget, hide or reorder effects, etc. as needed. See [qMRMLSegmentEditorWidget API](http://apidocs.slicer.org/master/classqMRMLSegmentEditorWidget.html). If you are not sure how to hide any complexity from the users then let us know.

fedorov:
@chribaue I don't know if this helps, but @che85 did a lot of similar UI simplification for [QuantitativeReporting](https://github.com/QIICR/QuantitativeReporting). Some of the components he abstracted into [SlicerDevelopmentToolbox](https://github.com/QIICR/SlicerDevelopmentToolbox). With various updates to Slicer and python version, I think there are likely regressions (Christian moved on to a different lab, and it doesn't look like he's been using it much), but maybe it can be reused in your module. At least that was the intent.

PR #1332, opened 2020-02-21
ENH: making the mask image optional (to match documentation)

dzenanz:
The input is required, otherwise the class produces all-zero output.

Side note: this mask has the opposite semantics from usual (a separate issue).

### DIFF
--- a/Modules/Loadable/Segmentations/Logic/vtkImageGrowCutSegment.cxx
+++ b/Modules/Loadable/Segmentations/Logic/vtkImageGrowCutSegment.cxx
@@ -141,7 +141,7 @@ bool vtkImageGrowCutSegment::vtkInternal::InitializationAHP(
     seedLabelVolumePtr = static_cast<LabelPixelType*>(seedLabelVolume->GetScalarPointer());
     }
   MaskPixelType* maskLabelVolumePtr = nullptr;
-  if (seedLabelVolume != nullptr)
+  if (maskLabelVolume != nullptr)
     {
     maskLabelVolumePtr = static_cast<MaskPixelType*>(maskLabelVolume->GetScalarPointer());
     }
--- a/Modules/Loadable/Segmentations/Logic/vtkImageGrowCutSegment.cxx
+++ b/Modules/Loadable/Segmentations/Logic/vtkImageGrowCutSegment.cxx
@@ -191,7 +191,7 @@ bool vtkImageGrowCutSegment::vtkInternal::InitializationAHP(
     // The neighborhood size is everywhere the same (size of m_NeighborIndexOffsets)
     // except at the edges of the volume, where the neighborhood size is 0.
     m_NumberOfNeighbors.resize(dimXYZ);
-    const unsigned char numberOfNeighbors = m_NeighborIndexOffsets.size();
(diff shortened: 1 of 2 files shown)


Modules/Loadable/Segmentations/Logic/vtkImageGrowCutSegment.h
```diff
@@ -19,8 +19,8 @@ class VTK_SLICER_SEGMENTATIONS_LOGIC_EXPORT vtkImageGrowCutSegment : public vtkI
   /// Set input seed label volume (input 1)
   void SetSeedLabelVolume(vtkImageData* labelImage) { this->SetInputData(1, labelImage); }
 
-  /// Set mask volume (input 2). Optional.
-  /// If this volume is specified then only those regions outside the mask (where mask has zero value)
+  /// Set mask volume (input 2).
```
lassoan: Have you intentionally removed "optional" word? It is important that this mask is not required.

dzenanz: That is the point of this PR. That input is not optional. If it is not set, the output is an all zeros image.

lassoan: That's a bug. Can you check if changing this line:

https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/Segmentations/Logic/vtkImageGrowCutSegment.cxx#L144

to 

    if (maskLabelVolumePtr != nullptr)

fixes the issue?

dzenanz: No, it doesn't.

Possible correction: the output might be just a default-created `vtkImageData`, not an all-zeros image when mask is not set.

lassoan: Anyway, it's just a small bug that I would rather see fixed then having it as a documented limitation. If you use this class in some project anyway then it may worth spending 5-10 minutes with debugging why the mask is not completely ignored when it is not set.

dzenanz: Now that I know it is supposed to work, I might do that.

dzenanz:
I was planning to make a new PR, but I forgot this PR was coming from my `master` branch. The force push turns this into a small feature update.

Modules/Loadable/Segmentations/Logic/vtkImageGrowCutSegment.cxx
```diff
@@ -191,7 +191,7 @@ bool vtkImageGrowCutSegment::vtkInternal::InitializationAHP(
     // The neighborhood size is everywhere the same (size of m_NeighborIndexOffsets)
     // except at the edges of the volume, where the neighborhood size is 0.
     m_NumberOfNeighbors.resize(dimXYZ);
-    const unsigned char numberOfNeighbors = m_NeighborIndexOffsets.size();
+    const unsigned char numberOfNeighbors = static_cast<unsigned char>(m_NeighborIndexOffsets.size());
```
hjmjohnson: +1

lassoan:
Merged.

PR #1333, opened 2020-02-21
ENH: Update surface tracing for markup curves

Sunderlandkyl:
- Miscellaneous name/style changes to surface curve related functions
- Fix calculation of metrics when curve type is changed
- Add vtkSlicerDijkstraGraphGeodesicPath to expose additional customization options for how the scalars are weighted
- Many other changes addressing feedback

**Curve settings widget:**
image

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.h
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.h
@@ -174,25 +174,28 @@ public:
   void SetCurveTypeToCardinalSpline();
   void SetCurveTypeToKochanekSpline();
   void SetCurveTypeToPolynomial();
-  void SetCurveTypeToShortestSurfaceDistance(vtkMRMLModelNode* modelNode=nullptr);
+  void SetCurveTypeToShortestDistanceOnSurface(vtkMRMLModelNode* modelNode=nullptr);
 
   /// Node reference role for the surface that is used in the shortest surface distance curve type
-  const char* GetSurfaceMeshNodeReferenceRole() { return "surfaceMesh"; };
-  const char* GetSurfaceMeshNodeReferenceMRMLAttributeName() { return "surfaceMeshRef"; };
+  const char* GetShortestDistanceSurfaceNodeReferenceRole() { return "shortestDistanceSurface"; };
+  const char* GetShortestDistanceSurfaceNodeReferenceMRMLAttributeName() { return "shortestDistanceSurfaceRef"; };
 
-  /// The model node that is used for calculating the shortest surface distance curve type
-  void SetAndObserveModelNode(vtkMRMLModelNode* modelNode);
-  vtkMRMLModelNode* GetModelNode();
+  /// The model node that is used as the surface mesh for finding the shortest distance path on the surface mesh.
+  /// Used by the ShortestDistanceOnSurface curve type.
+  void SetAndObserveShortestDistanceSurfaceNode(vtkMRMLModelNode* modelNode);
+  vtkMRMLModelNode* GetShortestDistanceSurfaceNode();
 
-  /// Whether to scale the distance of the points by some scalar weight value.
-  /// Only applies to shortest surface distance curve type.
-  bool GetUseSurfaceScalarWeights();
-  void SetUseSurfaceScalarWeights(bool useSurfaceScalarWeights);
+  /// The method that should be used to combine the distance with the scalar value.
+  /// Uses the COST_FUNCTION_X enums from vtkSlicerDijkstraGraphGeodesicPath.
+  int GetSurfaceCostFunctionType();
+  void SetSurfaceCostFunctionType(int surfaceCostFunctionType);
+  static const char* GetSurfaceCostFunctionTypeAsString(int surfaceCostFunctionType);
+  static int GetSurfaceCostFunctionTypeFromString(const char* name);
 
   /// The scalar weight function that is used for modifying the weight on each vertex.
   /// The the currently active point scalar array is availiable as the "activeScalar" variable.
-  const char* GetSurfaceScalarWeightFunction();
-  void SetSurfaceScalarWeightFunction(const char* function);
+  const char* GetSurfaceDistanceWeightingFunction();
+  void SetSurfaceDistanceWeightingFunction(const char* function);
 
   //@{
   /// Get/set how many curve points are inserted between control points.
@@ -202,9 +205,9 @@ public:
   //@}
 
 protected:
-  vtkSmartPointer<vtkTransformPolyDataFilter> PolyDataToWorldTransformer;
-  vtkSmartPointer<vtkArrayCalculator> ScalarCalculator;
-  vtkSmartPointer<vtkPassArrays> PassArray;
+  vtkSmartPointer<vtkTransformPolyDataFilter> SurfaceToWorldTransformer;
+  vtkSmartPointer<vtkArrayCalculator> SurfaceScalarCalculator;
+  vtkSmartPointer<vtkPassArrays> SurfacePassArray;
   const char* ActiveScalar;
 
 protected:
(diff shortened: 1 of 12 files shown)


Sunderlandkyl:
@lassoan let me know if I missed any changes.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.h
```diff
@@ -174,25 +174,26 @@ class  VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkMRMLMarkupsCurveNode : public vt
...
+  vtkMRMLModelNode* GetShortestDistanceSurfaceNode();
 
-  /// Whether to scale the distance of the points by some scalar weight value.
-  /// Only applies to shortest surface distance curve type.
-  bool GetUseSurfaceScalarWeights();
-  void SetUseSurfaceScalarWeights(bool useSurfaceScalarWeights);
+  /// The method that should be used to combine the distance with the scalar value.
+  /// Uses the COST_FUNCTION_X enums from vtkSlicerDijkstraGraphGeodesicPath.
+  int GetSurfaceCostFunction();
+  void SetSurfaceCostFunction(int surfaceCostFunction);
```
lassoan: The current name suggests that it is a function (similar to SurfaceScalarWeightFunction, which is a string). Adding "Type" in the name (SurfaceCostFunctionType) would be better.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
```diff
@@ -1522,56 +1522,56 @@ void vtkMRMLMarkupsNode::ConvertOrientationWXYZToMatrix(double orientationWXYZ[4
...
+orientationMatrix[8] = zz * 2 + s;    // (2,2)
 }
 
 //----------------------------------------------------------------------
 vtkPoints* vtkMRMLMarkupsNode::GetCurvePoints()
 {
   this->CurveGenerator->Update();
   if (!this->CurveGenerator->GetOutput())
-    {
+  {
```
lassoan: we should still use old-style brace indentation (but hopefully we can get rid of it soon)

Modules/Loadable/Markups/MRML/vtkSlicerDijkstraGraphGeodesicPath.cxx
```diff
@@ -0,0 +1,188 @@
...
+      }
+    default:
+      {
+      return "";
+      }
+    }
+}
+
+//------------------------------------------------------------------------------
+const char* vtkSlicerDijkstraGraphGeodesicPath::GetCostFunctionAsHumanReadableString(int costFunction)
```
lassoan: It should be GetCostFunctionFromString and do just enum<->code string conversion. Human-readable string <-> enum conversion should happen in the GUI layer (same as for the other similar function).

Modules/Loadable/Markups/MRML/vtkSlicerDijkstraGraphGeodesicPath.cxx
```diff
@@ -0,0 +1,188 @@
...
+    }
+}
+
+//----------------------------------------------------------------------------
+int vtkSlicerDijkstraGraphGeodesicPath::RequestData(
+  vtkInformation* vtkNotUsed(request),
+  vtkInformationVector** inputVector,
+  vtkInformationVector* outputVector)
+{
+  vtkInformation* inInfo = inputVector[0]->GetInformationObject(0);
```
lassoan: add a comment how it is related to the parent class implementation (what's the difference, why)

Modules/Loadable/Markups/MRML/vtkSlicerDijkstraGraphGeodesicPath.cxx
```diff
@@ -0,0 +1,188 @@
...
+
+  this->ShortestPath(input, this->StartVertex, this->EndVertex);
+  this->TraceShortestPath(input, output, this->StartVertex, this->EndVertex);
+  return 1;
+}
+
+//------------------------------------------------------------------------------
+void vtkSlicerDijkstraGraphGeodesicPath::BuildAdjacency(vtkDataSet* inData)
+{
+  Superclass::BuildAdjacency(inData);
```
lassoan: add a comment how it is related to the parent class implementation (what's the difference, why)

Modules/Loadable/Markups/MRML/vtkSlicerDijkstraGraphGeodesicPath.h
```diff
@@ -0,0 +1,87 @@
...
+// VTK includes
+#include <vtkDijkstraGraphGeodesicPath.h>
+
+// export
+#include "vtkSlicerMarkupsModuleMRMLExport.h"
+
+//
+// Set built-in type.  Creates member Set"name"() (e.g., SetVisibility());
+//
+#define vtkSetAdjacencyMacro(name,type) \
```
lassoan: Why do we need this, why here, and with a macro?

Modules/Loadable/Markups/MRML/vtkSlicerDijkstraGraphGeodesicPath.h
```diff
@@ -0,0 +1,87 @@
...
+{
+public:
+  vtkTypeMacro(vtkSlicerDijkstraGraphGeodesicPath, vtkDijkstraGraphGeodesicPath);
+  static vtkSlicerDijkstraGraphGeodesicPath* New();
+  void PrintSelf(ostream& os, vtkIndent indent) override;
+
+  enum
+  {
+    COST_FUNCTION_DISTANCE,
+    COST_FUNCTION_ADDITIVE,
```
lassoan: COST_FUNCTION_SCALING (or COST_FUNCTION_SCALE, or COST_FUNCTION_MULTIPLICATIVE) would be a very basic type, too, it would be nice to be able to experiment with it

Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
```diff
@@ -868,29 +875,58 @@ void qSlicerMarkupsModuleWidget::updateWidgetFromMRML()
...
+    QString prefixString;
+    QString suffixString;
+    switch (costFunction)
+      {
+      case vtkSlicerDijkstraGraphGeodesicPath::COST_FUNCTION_ADDITIVE:
+        prefixString = "distance + ";
+        break;
+      case vtkSlicerDijkstraGraphGeodesicPath::COST_FUNCTION_INVERSE_SQUARED:
+        prefixString = "distance / (";
+        suffixString = "^2";
```
lassoan:
```suggestion
        suffixString = ") ^ 2";
```

Modules/Loadable/Markups/MRML/vtkSlicerDijkstraGraphGeodesicPath.h
Sunderlandkyl: In the base class, the adjacency is only recalculated if the input is modified. I added the RecalculateAdjacency varaible that functions could set to indicate that edge calculation parameters have changed, and that the Adjacency should be recomputed.

I added this macro (and created overridden methods for BuildAdjacency and RequestData) to set/use this variable, rather than copy-pasting the SetX macro for each function that needed the adjacency to be rebuilt.

lassoan: OK, I understand now. Set macros are somewhat opaque but they are so simple that they are acceptable. However, in this case, it makes things a bit too complicated. Simpler solutions could be to just repeating those few lines in simple methods; or do things like storing of the CostFunction and UseScalarWeights values that were used when computing adjacency last time, or store last time adjacency related input values changed in a timestamp - and do not recompute adjacency if not needed.

Sunderlandkyl:
@lassoan Changes made.

Regarding the prefix/suffixed QLineEdit, I didn't find a simple way to do it (other than placing QLabels on either side). I think that if I was going to implement it, I would add a sub-classed QLineEdit with that functionality to CTK. If that's something that you think is worthwhile, then I can look into it.

lassoan:
Thank you!

Prefix /suffix display is more than good enough for now. We can finalize/polish the GUI in the coming weeks as wet gather more experience and feedback from users.

lassoan:
Thank you, it works great, merged.

lassoan:
It would be great if you could create a 1-minute youtube video and discourse post about it.

PR #1335, opened 2020-02-23
ENH: Add back vtkCurveGenerator SetInputPoints and GetOutputPoints

Sunderlandkyl:
Restore removed methods from vtkCurveGenerator. Allows setting of input and retrieval of output as viewpoints.


lassoan:
thank you, merged!

PR #1336, opened 2020-02-24
BUG: Fix markup curve surface interpolation with incompatible mesh

Sunderlandkyl:
Some mesh required a cleanup and/or conversion to a triangular mesh before vtkSlicerDijkstraGraphGeodesicPath could generate a path.
Input mesh are now run through vtkCleanPolyData and vtkTriangleFilter before being used for pathfinding.
Also reduce errors for mesh with no active scalars by routing the input through a vtkPassThroughFilter to avoid scalar calculation when it is not needed.


lassoan:
thank you, merged!

PR #1337, opened 2020-03-01
BUG: Disable default interaction of ffmpeg with stdin

fepegar:
If Slicer is run in the background process group, it hangs when calling ffmpeg with subprocess.Popen().
This commit disables interaction with stdin to prevent this from happenning on Linux and macOS.

See discussion on Slicer forum: https://discourse.slicer.org/t/screen-capture-module-makes-slicer-hang/10456.
Solution found on SO: https://stackoverflow.com/q/47115191/3956024.


fepegar:
I guess my commit should have started with "BUG:" as well.

lassoan:
Thank you, nice work! Merged.

PR #1338, opened 2020-03-02
ENH: Add patient birth date and sex to exported DICOM scalar volume

cpinter:
Patient birth date and sex were not exported in the series by the scalar volume plugin (and CreateDICOMSeries module), even if the fields were filled. These two tags are now also supported.


pieper:
LGTM - thank you!

cpinter:
Thanks for the review!
Integrated in https://github.com/Slicer/Slicer/commit/02d21534c37a577c3dcdf1cf72e9c8fe5d7165c9

PR #1339, opened 2020-03-03
ENH: Add new plane markup type

Sunderlandkyl:
This commit adds a new plane type of markup. Planes are defined by 3 control points. The first control point defines the plane origin. An offset can be specified which transforms the origin and vectors of the plane using a 4x4 matrix.

Shift+Left click and drag translates the offset of the plane in 3D and slice views.
The vtkMRMLMarkupsPlaneNode class also includes some convenience functions for generating a plane without placing the markup points (SetOrigin(), SetNormal(), and SetPlaneAxes()) that can be specified in either local or world coordinates.

Future work will include more advanced plane widgets/interactions (movement/rotation along specific axis, adjusting size, etc).

image


jcfr:
Trying to notify @commontk-bot

Sunderlandkyl:
Moved to https://github.com/Slicer/Slicer/pull/4746

PR #1340, opened 2020-03-05
DOC: Update copyright assignment to NumFOCUS

hjmjohnson:
As part of the process of the Insight Software Consortium joining the
NumFOCUS non-profit corporation as a Fiscally Sponsored Project, all
assets, including intellectual property assets, are being transferred
to enable dissolution of the Insight Software Consortium non-profit
corporate legal entity.


hjmjohnson:
@pieper I replaced old boiler plate with new boiler plate.  I am not considering if the original boiler plate was correct.

jcfr:
Thanks @pieper

PR #1341, opened 2020-03-05
COMP: Large 64 bit integers loose precision in double

hjmjohnson:
warning: implicit conversion from 'unsigned long' to 'double'
changes value from 18446744073709551615 to 18446744073709551616 [-Wimplicit-int-float-conversion]


hjmjohnson:
There are many of these warnings that are not yet addressed.  Before moving forward, I'd like to confirm that there is an agreement on making similar changes to address the warnings.

lassoan:
Thank you, merged.

PR #1342, opened 2020-03-05
STYLE: Remove redundant void argument lists

hjmjohnson:
Find and remove redundant void argument lists.


lassoan:
Merged. Thank you very much!

PR #1345, opened 2020-03-11
ENH: Add command line url handling and DICOM import from DICOMweb url

Sunderlandkyl:
Adds the urlReceived signal to qSlicerCoreApplication that is emitted when a url beginning with the "slicer://" protocol is processed from the command line.
This also adds handling for "slicer://viewer/..." urls, which will import the specified dataset using DICOMweb.


Sunderlandkyl:
@lassoan Here is the url handling code.

Sunderlandkyl:
Migrated to new repo: https://github.com/Slicer/Slicer/pull/4727

